Issue #798, opened 2020-12-17
(deployment): Deployment cancellation does not go through if deployment action has not yet been registered in UpdateSystemPolicyService pendingActions queue


shaguptashaikh:
**Describe the bug**
When a deployment gets cancelled, it can be in progress and in any of the possible stages, i.e 
    Stage 1. Preparation - dependency resolution/artifact download/config generation 
    Stage 2. Done with prep but waiting until components stop deferring deployment via IPC
    Stage 3. Executing the change i.e. merging config, lifecycle changes, monitoring, (rollback if something fails) etc

In Stage 3, we don't allow cancellation because it can leave the system broken/ nondeterministic. 
In Stage 2 and 1, cancellation should go through because it is the only way to get out of situations like long retry loop/components forever deferring deployments via IPC. 

Right now, deployment cancellation does not go through because the [check to determine if cancellation is allowed](https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/29292a7eee6d88a8d9cfac5778e06df34fcca9d0/src/main/java/com/aws/greengrass/lifecyclemanager/UpdateSystemPolicyService.java#L119) based on the stage is incorrect

**To Reproduce**
We discovered this accidentally, when there was some misconfiguration in region and endpoint configs, the deployment is not able to reach cloud component service during dependency resolution, but the MQTT connection works fine so it can poll deployments and cancellation signals. If you try the same thing and try to cancel the deployment, this issue can be reproduced.

**Expected behavior**
When deployment is in Stage 1 and it gets cancelled, the cancellation and further deployments should go through

**Actual behavior**
When deployment was stuck in Stage 1 and it got cancelled, the cancellation or any further deployments did not go through and the original deployment stayed stuck in Stage 1

**Environment**
 - OS: [e.g. Ubuntu 20.04]
 - JDK version:
 - Nucleus version: 2.0.3

**Additional context**

E.g. what is the impact of the bug? - Deployment cancellation does not take effect when deployment is still in preparation mode(download recipes/resolve components/download artifacts etc)

shaguptashaikh:
Addressing the issue in this PR https://github.com/aws-greengrass/aws-greengrass-nucleus/pull/799

shaguptashaikh:
Fix has been merged

PR #799, opened 2020-12-17
fix(deployment): Handle deployment cancellation correctly when deployment is not yet registered with UpdateSystemPolicyService

shaguptashaikh:
fix(e2e): Add 1 TPS limit to GCS createComponent calls for test component creation
fix(benchmark): Remove benchmark workflow

**Issue #, if available:**
https://github.com/aws-greengrass/aws-greengrass-nucleus/issues/798
[See issue for details]

**Description of changes:**
Correctly handle the stage deployment is in when evaluating if it can be cancelled. There was a regression introduced in device deployment handling which causes the evaluation to not work in cases when deployment is not yet registered in the UpdateSystemPolicyService's pendingActions queue, i.e. while deployment is still in dependency resolution/download artifacts stage.

Necessary to ensure deployment cancellation can be handled correctly

Added a new E2E test `GIVEN_deployment_in_dependency_resolution_stage_WHEN_cancel_event_received_THEN_deployment_should_be_cancelled`, tested that it passes
```
mvn surefire:test@integration-tests -Dgroups="E2E" -DexcludedGroups="" -Dsurefire.argLine="" -Dtest=DeploymentE2ETest#GIVEN_deployment_in_dependency_resolution_stage_WHEN_cancel_event_received_THEN_deployment_should_be_cancelled
```

**Checklist:**


src/main/java/com/aws/greengrass/lifecyclemanager/UpdateSystemPolicyService.java
```diff
@@ -116,7 +117,8 @@ public boolean discardPendingUpdateAction(String tag) {
         }
         // Signal components that they can resume their work since the update is not going to happen
         lifecycleIPCAgent.sendPostComponentUpdateEvent(new PostComponentUpdateEvent());
-        return pendingActions.remove(tag) != null;
+        pendingActions.remove(tag);
+        return true;
     }
```
hui-yang, line 122: What happens is the action is in progress, like mentioned in the javadoc?

shaguptashaikh: It will return false on line 116

MikeDombo (dismissed):

src/main/java/com/aws/greengrass/lifecyclemanager/UpdateSystemPolicyService.java
```diff
@@ -47,7 +48,7 @@
     // String identifies the action, the pair consist of timeout and an action. The timeout
     // represents the value in seconds the kernel will wait for components to respond to
     // an precomponent update event
-    private final Map<String, UpdateAction> pendingActions = new LinkedHashMap<>();
+    private final Map<String, UpdateAction> pendingActions = Collections.synchronizedMap(new LinkedHashMap<>());
```
fengwa-aws, line 51: Why not just use ConcurrentHashMap?

MikeDombo: We need to keep ordering I believe

shaguptashaikh: It needs to be a LinkedHashMap so all pending actions are executed in order, right now we only ever have 1 action because we use this only for deployments and we process 1 deployment at a time, but this service was written to accommodate more actions and execute them in order of registration, and we've always left it like that

MikeDombo (approved):
Thanks!

fengwa-aws (approved):

fahadmohammed01 (approved):

PR #801, opened 2020-12-18
fix: set permissions on paths that we create to 755 instead of 700

MikeDombo:
**Issue #, if available:**

**Description of changes:**
Paths created by Greengrass have been created with 700 permissions which is causing issues because if multiple levels of directories are created, then not all users have permissions to traverse the tree. Specifically, `ggc_user` wouldn't have permission to execute a python script because not all of the directories on the path are readable/executable by ggc_user.

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/util/Utils.java
+++ b/src/main/java/com/aws/greengrass/util/Utils.java
@@ -610,7 +610,7 @@ public final class Utils {
                 // This only supports POSIX compliant file permission right now. We will need to
                 // change this when trying to support Greengrass in Non-POSIX OS.
                 Files.createDirectories(p,
-                        PosixFilePermissions.asFileAttribute(PosixFilePermissions.fromString("rwx------")));
+                        PosixFilePermissions.asFileAttribute(PosixFilePermissions.fromString("rwxr-xr-x")));
             }
         }
     }


src/main/java/com/aws/greengrass/util/Utils.java
```diff
@@ -610,7 +610,7 @@ public static void createPaths(Path... paths) throws IOException {
                 // This only supports POSIX compliant file permission right now. We will need to
                 // change this when trying to support Greengrass in Non-POSIX OS.
                 Files.createDirectories(p,
-                        PosixFilePermissions.asFileAttribute(PosixFilePermissions.fromString("rwx------")));
+                        PosixFilePermissions.asFileAttribute(PosixFilePermissions.fromString("rwxr-xr-x")));
```
fengwa-aws, line 613: Is this the default permission and can it be overridden? I'm thinking about the cases when a directory needs to be read protected from all.

MikeDombo: We always set the permissions appropriately in `NucleusPaths`, but that class doesn't set the permissions on the parent directories like this one does.

fengwa-aws (approved):

philcali (approved):

Issue #802, opened 2020-12-18
(easysetup): iam:CreatePolicy required, even though policy exists


fiLLLip:
**Describe the bug**
When adopting a new Greengrass Core, we need temp credentials to do a lot of things. I don't want those credentials to be able to create new IAM policies and roles, so I created them myself, since they are reused across all adoptions. But, I cannot do the adoption since I don't have iam:CreatePolicy rights.
https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/8a9bd022c7a1da82012eec4a006839d7b7e36f70/src/main/java/com/aws/greengrass/easysetup/DeviceProvisioningHelper.java#L399
Here you run createPolicy directly, and it will never answer that it already exists since we don't have permission to create.
I would propose running the getPolicy first for checking if it exists instead of relying on the error message.

**To Reproduce**
Run easysetup with minimal permissions

**Expected behavior**
It should say it exist and move on.

**Actual behavior**
```
Attaching TES role policy to IoT thing...
Error while trying to setup Greengrass Nucleus
software.amazon.awssdk.services.iam.model.IamException: User: arn:aws:sts::xxx:assumed-role/GreengrassCommissioningRole/tempcore-30293-setup is not authorized to perform: iam:CreatePolicy on resource: policy GreengrassV2TokenExchangeRoleAccess
```

**Environment**
 - OS: Raspbian Stretch
 - JDK version: OpenJDK8
 - Nucleus version: latest?

**Additional context**
I need to add very permissive permissions to a role that will be running in the first-run scripts of my embedded device to add it as a temp core device in greengrass.

fiLLLip:
Workaround:

Add iam:CreatePolicy permission with the specific policy as resource :+1: 

```
              - Effect: Allow
                Action:
                  - "iam:CreatePolicy"
                Resource: !Ref GreengrassV2TokenExchangeRoleAccess
```

shaguptashaikh:
Hi, thanks for reporting this. We are looking into it and will post an update here once done.

[...]

PR #805, opened 2020-12-24
improvement(easysetup): Do not attempt creating role policy if it exists

shaguptashaikh:
**Issue #, if available:**
https://github.com/aws-greengrass/aws-greengrass-nucleus/issues/802

**Description of changes:**
When `--provision` option is set during setup, it currently creates the token exchange IAM role policy directly. Users may create the role/role alias/policy resources outside the setup in order to scope down the permissions needed on the device for setup even further. In that case, checking for the resource through the GetPolicy API first and creating only if the resource does not exist would be desirable.

Manually tested provisioning and starting nucleus

**Checklist:**

### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/e2e/BaseE2ETestCase.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/e2e/BaseE2ETestCase.java
@@ -39,9 +39,11 @@ import software.amazon.awssdk.services.greengrassv2.model.ComponentDeploymentSpe
 import software.amazon.awssdk.services.greengrassv2.model.CreateComponentVersionResponse;
 import software.amazon.awssdk.services.greengrassv2.model.CreateDeploymentRequest;
 import software.amazon.awssdk.services.greengrassv2.model.CreateDeploymentResponse;
+import software.amazon.awssdk.services.greengrassv2.model.DeleteCoreDeviceRequest;
 import software.amazon.awssdk.services.greengrassv2.model.DeploymentComponentUpdatePolicy;
 import software.amazon.awssdk.services.greengrassv2.model.DeploymentConfigurationValidationPolicy;
 import software.amazon.awssdk.services.greengrassv2.model.DeploymentPolicies;
+import software.amazon.awssdk.services.greengrassv2.model.ResourceNotFoundException;
 import software.amazon.awssdk.services.iam.IamClient;
 import software.amazon.awssdk.services.iot.IotClient;
 import software.amazon.awssdk.services.iot.model.CreateThingGroupResponse;
@@ -400,6 +402,13 @@ public class BaseE2ETestCase implements AutoCloseable {
         createdDeployments.forEach(greengrassClient::cancelDeployment);
         createdDeployments.clear();
 
+        try {
+            greengrassClient.deleteCoreDevice(
+                    DeleteCoreDeviceRequest.builder().coreDeviceThingName(thingInfo.getThingName()).build());
+        } catch (ResourceNotFoundException e) {
+            logger.atDebug().kv("coreDevice", thingInfo.getThingName()).log("No core device to delete");
+        }
+
         deviceProvisioningHelper.cleanThing(iotClient, thingInfo, false);
         createdThingGroups.forEach(thingGroup -> IotJobsUtils.cleanThingGroup(iotClient, thingGroupName));
         createdThingGroups.clear();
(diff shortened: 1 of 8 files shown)


src/main/java/com/aws/greengrass/easysetup/DeviceProvisioningHelper.java
```diff
@@ -390,31 +399,55 @@ public void setupIoTRoleForTes(String roleName, String roleAliasName, String cer
...
-            outStream.printf("IAM policy named \"%s\" already exists. Please attach it to the IAM role if not "
-                    + "already%n", rolePolicyName);
-            return Optional.empty();
         }
     }
 
+    private Optional<String> getPolicyArn(String policyName, Region awsRegion) {
+        String partition = awsRegion.metadata().partition().id();
+        try {
+            String policyArn = String.format(MANAGED_IAM_POLICY_ARN_FORMAT, partition, policyName);
```
MikeDombo, line 431: is it possible to use the `Arn` datatype instead of building it ourselves using string formatting?

src/integrationtests/java/com/aws/greengrass/integrationtests/e2e/BaseE2ETestCase.java
```diff
@@ -419,7 +428,8 @@ protected void setupTesRoleAndAlias() throws InterruptedException {
         deviceProvisioningHelper.setupIoTRoleForTes(TES_ROLE_NAME, TES_ROLE_ALIAS_NAME, thingInfo.getCertificateArn());
         if (tesRolePolicyArn == null || !tesRolePolicyArn.isPresent()) {
             tesRolePolicyArn = deviceProvisioningHelper
-                    .createAndAttachRolePolicy(TES_ROLE_NAME, TES_ROLE_POLICY_NAME, TES_ROLE_POLICY_DOCUMENT);
```
hui-yang, line 422: Just realized this, should we change e2e tests to target prod?
Maybe add a separate task for this.

src/main/java/com/aws/greengrass/easysetup/DeviceProvisioningHelper.java
```diff
@@ -142,6 +147,7 @@ public DeviceProvisioningHelper(String awsRegion, String environmentStage, Print
                 : EnvironmentStage.fromString(environmentStage);
         this.iotClient = IotSdkClientFactory.getIotClient(awsRegion, envStage);
         this.iamClient = IamSdkClientFactory.getIamClient(awsRegion);
+        this.stsClient = StsSdkClientFactory.getStsClient(awsRegion);
```
hui-yang, line 150: Update the required permission for e2e tests in readme https://github.com/aws-greengrass/aws-greengrass-nucleus/blame/main/DEVELOPING.md#L51-L57

src/main/java/com/aws/greengrass/easysetup/DeviceProvisioningHelper.java
```diff
@@ -108,6 +110,8 @@
     private static final String IOT_ROLE_POLICY_NAME_PREFIX = "GreengrassTESCertificatePolicy";
     private static final String GREENGRASS_CLI_COMPONENT_NAME = "aws.greengrass.Cli";
     private static final String INITIAL_DEPLOYMENT_NAME_FORMAT = "Deployment for %s";
+    private static final String IAM_POLICY_ARN_FORMAT = "arn:%s:iam::%s:policy/%s";
+    private static final String MANAGED_IAM_POLICY_ARN_FORMAT = "arn:%s:iam::aws:policy/%s";
```
hui-yang, line 114: Do we have a managed policy?

src/integrationtests/java/com/aws/greengrass/integrationtests/e2e/BaseE2ETestCase.java
shaguptashaikh: Not sure if we want to, should discuss

src/main/java/com/aws/greengrass/easysetup/DeviceProvisioningHelper.java
shaguptashaikh: We don't but since the arn is used to make a get request, it should look for it in both categories, this bit changed a bit now though after switching to the Arn builder

src/main/java/com/aws/greengrass/easysetup/DeviceProvisioningHelper.java
```diff
@@ -390,31 +398,54 @@ public void setupIoTRoleForTes(String roleName, String roleAliasName, String cer
...
 
+    private Optional<String> getPolicyArn(String policyName, Region awsRegion) {
+        String partition = awsRegion.metadata().partition().id();
+        Arn.Builder policyArnBuilder = Arn.builder().partition(partition).service("iam").resource(policyName);
+        try {
+            return Optional.of(iamClient.getPolicy(software.amazon.awssdk.services.iam.model.GetPolicyRequest.builder()
+                    .policyArn(policyArnBuilder.build().toString()).build()).policy().arn());
+        } catch (ResourceNotFoundException | NoSuchEntityException mnf) {
+            try {
+                return Optional.of(iamClient.getPolicy(
```
MikeDombo, line 435: Add a comment for what these 2 cases are trying to do.

src/main/java/com/aws/greengrass/easysetup/DeviceProvisioningHelper.java
shaguptashaikh: I made this change in previous revision but had to revert it because it turns out that arns built like that(e.g. `arn:aws:iam:null:null:policy/GreengrassV2TokenExchangeRoleAccess`) do not work well with global services like IAM and when the ARNs are different based on managed vs customer created resources.

MikeDombo (approved):

fahadmohammed01 (approved):

PR #806, opened 2020-12-30
fix(provision): load Nucleus recipe from build file to component store and config store

hui-yang:
**Issue #, if available:**

**Description of changes:**
Problems:
1. GIVEN nucleus and CLI provisioned on the device, WHEN use CLI to create local deployment to change Nucleus config while device is offline, THEN deployment fails because Nucleus recipe is missing in component store
    1. If Nucleus recipe does exist and references an artifact, deployment still fails because artifact is also missing 
1. GIVEN nucleus provisioned on the device, WHEN create a first deployment (local or cloud) to change any Nucleus config, THEN deployment succeeds but Nucleus always restarts, because Nucleus bootstrap lifecycle step is not initialized and any changes in bootstrap step leads to the restart workflow. 

The solution to both problems is to properly initialize Nucleus service config and component store during the initial provision.
1. Initialize JVM options from current setting
1. Copy Nucleus recipe from `{release_zip_unpack_dir}/conf/recipe.yaml` to local component store
   1. Copy unpacked artifacts to local component store (as it's referenced in lifecyle.bootstrap in the recipe)
1. Load dependencies and lifecycle definitions from the recipe to config store

To enable Nucleus deployments when device is offline.
To prevent unnecessary restarts in Nucleus config deployments.

End-to-end scenarios:
1. GIVEN nucleus and CLI provisioned on the device, WHEN device goes offline, AND use CLI to create a deployment to change Nucleus config `logging.level=DEBUG`, THEN deployment succeeds (and Nucleus doesn't restart)
1. GIVEN nucleus and CLI provisioned on the device, WHEN device goes offline, AND use CLI to create a deployment to change Nucleus config `jvmOptions=-Xmx64m`, THEN Nucleus restarts and deployment succeeds
1. GIVEN nucleus v2.0.0 and CLI provisioned on the device, WHEN create a cloud deployment of nucleus v2.0.1, THEN Nucleus restarts into the new version and deployment succeeds. WHEN device goes offline, AND use CLI to create a deployment of Nucleus v2.0.0, THEN Nucleus restarts into the old version (using the local recipe) and deployment succeeds

When device is online, deployments will always download latest recipe from Cloud and overwrite the local copy (if the content doesn't match).
In this PR, the changes focus on common scenarios of local deployments when offline. As a result, the local recipe and artifacts don't have to be exact matches as in Cloud component service.
- Local recipe doesn't have artifacts section
- Unpacked artifacts are copied to component store instead of the original artifact zip file 

**Checklist:**


tilo-chen:
Looking good. Maybe test with `BootstrapManager#isBootstrapRequired` to make sure it won't have unexpected restart.

src/main/java/com/aws/greengrass/lifecyclemanager/Kernel.java
```diff
@@ -603,15 +612,64 @@ public Kernel parseArgs(String... args) {
...
+    private void initializeNucleusServiceConfigIfNotExist(KernelAlternatives kernelAlternatives,
+                                                          DeviceConfiguration deviceConfiguration) {
+        Topics nucleusLifecycle = getConfig().lookupTopics(DEFAULT_VALUE_TIMESTAMP, SERVICES_NAMESPACE_TOPIC,
+                deviceConfiguration.getNucleusComponentName(), SERVICE_LIFECYCLE_NAMESPACE_TOPIC);
+        if (nucleusLifecycle.children.size() != 0) {
+            logger.atDebug().log("Nucleus lifecycle has already been initialized");
+            return;
+        }
+
+        // Persist initial Nucleus launch parameters
```
tilo-chen, line 638: So we're moving this line in nucleus bootstrap script: `echo "{configuration:/jvmOptions}" > "$KERNEL_ROOT/alts/current/launch.params"`
to here. Nice!

src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
```diff
@@ -349,6 +354,17 @@ public Topic getRootCAFilePath() {
...
     }
 
+    /**
+     * Get JVM options. Default to current JVM options input.
+     *
+     * @return JVM options config topic
+     */
+    public Topic getJvmOptions() {
+        String jvmOptions = ManagementFactory.getRuntimeMXBean().getInputArguments().stream().sorted()
+                .filter(s -> !s.startsWith(JVM_OPTION_ROOT_PATH)).collect(Collectors.joining(" "));
```
tilo-chen, line 364: Hmm why exclude `-Droot`?

src/main/java/com/aws/greengrass/lifecyclemanager/Kernel.java
hui-yang: Yeah. The bootstrap script only gets to run in deployments. Here in provisioning I need to set it up explicitly

src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
hui-yang: The loader script will pass in `-Droot` https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/main/scripts/loader#L27 
The Nucleus root path won't be changed even if customers configure it in jvmOptions via deployments

src/main/java/com/aws/greengrass/lifecyclemanager/Kernel.java
```diff
@@ -603,15 +612,64 @@ public Kernel parseArgs(String... args) {
...
         setupProxy();
+
         return this;
     }
 
     private void setupProxy() {
         ProxyUtils.setProxyProperties(context.get(DeviceConfiguration.class));
     }
 
+    private void initializeNucleusServiceConfigIfNotExist(KernelAlternatives kernelAlternatives,
```
shaguptashaikh, line 629: After adding getting the recipe from the build file, will we still keep the properties file as well for version? or should the version now come from the recipe?

src/main/java/com/aws/greengrass/lifecyclemanager/Kernel.java
```diff
@@ -603,15 +612,64 @@ public Kernel parseArgs(String... args) {
         }
 
         // Update device configuration from commandline arguments after loading config files
-        kernelCommandLine.updateDeviceConfiguration(getContext().get(DeviceConfiguration.class));
+        DeviceConfiguration deviceConfiguration = getContext().get(DeviceConfiguration.class);
+        kernelCommandLine.updateDeviceConfiguration(deviceConfiguration);
+        // After configuration is fully loaded, initialize Nucleus service config
+        initializeNucleusServiceConfigIfNotExist(kernelAlts, deviceConfiguration);
```
shaguptashaikh, line 618: There's some initialization logic in `DeviceConfiguration.initializeNucleusComponentConfig()` for initializing version and a few more things for the Nucleus component config, is there a way we can consolidate that with this here?

src/main/java/com/aws/greengrass/lifecyclemanager/Kernel.java
hui-yang: Ideally properties file can be removed. Let me see if I can get the order of the events correct

src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
```diff
@@ -355,6 +360,17 @@ public Topic getRootCAFilePath() {
...
+        return getTopic(DEVICE_PARAM_JVM_OPTIONS).dflt(jvmOptions);
```
shaguptashaikh, line 371: Should we remove the default for the `jvmOptions` config from the Nucleus recipe now that the default it evaluated here?

src/main/java/com/aws/greengrass/lifecyclemanager/Kernel.java
hui-yang: I put it here because the interpolation should happen after all configs are loaded. Let me see if this can be optimized.

src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
hui-yang: Do you mean this should be removed instead of default to empty string? https://code.amazon.com/packages/GreengrassV2ComponentsReleaseCDK/blobs/47911fb22576071ee8ec09c7bf0eef3ef421ae35/--/lib/components/greengrass-nucleus.ts#L68

src/test/java/com/aws/greengrass/lifecyclemanager/LogManagerHelperTest.java
```diff
@@ -168,7 +167,7 @@ void GIVEN_null_logger_config_WHEN_subscribe_THEN_correctly_reconfigures_all_log
         NucleusPaths nucleusPaths = mock(NucleusPaths.class);
         Topics rootConfigTopics = mock(Topics.class);
         when(rootConfigTopics.findOrDefault(any(), anyString(), anyString(), anyString())).thenReturn(new ArrayList<>());
-        when(configuration.lookup(anyString(), anyString())).thenReturn(mock(Topic.class));
+        //lenient().when(configuration.lookup(anyString(), anyString())).thenReturn(mock(Topic.class));
```
MikeDombo, line 170: remove.

src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
```diff
@@ -506,22 +604,19 @@ public String getNucleusVersion() {
...
-                            .resolve(NUCLEUS_BUILD_METADATA_FILENAME))) {
-                NUCLEUS_BUILD_PROPERTIES.load(is);
-            }
-
-            String version = NUCLEUS_BUILD_PROPERTIES.getProperty(NUCLEUS_VERSION_BUILD_METADATA_KEY);
-            if (version != null) {
-                return version;
+            com.amazon.aws.iot.greengrass.component.common.ComponentRecipe recipe = getRecipeSerializer()
+                    .readValue(locateCurrentKernelUnpackDir().resolve(NUCLEUS_BUILD_METADATA_DIRECTORY)
+                                    .resolve(NUCLEUS_RECIPE_FILENAME).toFile(),
```
MikeDombo, line 611: We need to modify the pipeline to write the version number into the recipe now instead of properties?

conf/recipe.yaml
```diff
@@ -1,3 +1,8 @@
+#
```
MikeDombo, line 1: double check that this recipe matches exactly what is in the CDK.

src/main/java/com/aws/greengrass/deployment/DeploymentService.java
```diff
@@ -510,57 +511,72 @@ private void copyRecipesAndArtifacts(Deployment deployment) throws InvalidReques
         }
     }
 
-    @SuppressWarnings("PMD.ExceptionAsFlowControl")
     private void copyRecipesToComponentStore(Path from) throws IOException {
         try (Stream<Path> files = Files.walk(from)) {
             for (Path r : files.collect(Collectors.toList())) {
```
MikeDombo, line 516: Can you add a skip for directories? We currently get a false error that some directory is not a recipe, which we can safely ignore.

src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
```diff
@@ -226,6 +235,95 @@ private void initializeNucleusComponentConfig() {
...
+            }
+            ComponentRecipe componentRecipe = resolvedRecipe.get();
+            nucleusComponentName = componentRecipe.getComponentName();
+
+            // Add Nucleus version
+            String componentVersion = componentRecipe.getVersion().toString();
+            kernel.getConfig().lookup(SERVICES_NAMESPACE_TOPIC, nucleusComponentName,
+                    VERSION_CONFIG_KEY).dflt(componentVersion);
+
+            kernel.getConfig().lookup(SETENV_CONFIG_NAMESPACE, GGC_VERSION_ENV).withValue(componentVersion);
```
MikeDombo, line 298: you can't put this here since in most cases this code will never execute. Put this somewhere that always executes so that we always pass the version in.

src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
hui-yang: Yes. Two options
1. Update version in recipe
2. Keep build.properties and here we interpolate the recipe with correct version

I'm currently going with #1 but it can be changed

conf/recipe.yaml
hui-yang: Yes, I've verified in UAT but I will be updating CDK to update the recipe

src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
hui-yang: Ok. I can move it back to where it was

src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
hui-yang: I'm keep it in mind when updating the cdk

src/main/java/com/aws/greengrass/lifecyclemanager/Kernel.java
hui-yang: I did some experiments. `initializeNucleusComponentConfig` was called in DeviceConfiguration constructor. Since I introduced ComponentStore RecipeLoader etc here in the init, changes in `initializeNucleusComponentConfig` will cause circular dependency issues. 

I still kept the logic here for now

src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
```diff
@@ -226,6 +246,97 @@ private void initializeNucleusComponentConfig() {
...
+            kernel.getConfig().lookup(DEFAULT_VALUE_TIMESTAMP, SERVICES_NAMESPACE_TOPIC,
+                    nucleusComponentName, SERVICE_DEPENDENCIES_NAMESPACE_TOPIC)
+                    .dflt(kernelConfigResolver.generateServiceDependencies(nucleusDependencies));
+
+            // Add Nucleus lifecycle (after config interpolation)
+            Object interpolatedLifecycle = kernelConfigResolver.interpolate(componentRecipe.getLifecycle(),
+                    new ComponentIdentifier(componentRecipe.getComponentName(), componentRecipe.getVersion()),
+                    nucleusDependencies.keySet(),
+                    kernel.getConfig().lookupTopics(SERVICES_NAMESPACE_TOPIC).toPOJO());
+            nucleusLifecycle.replaceAndWait((Map<String, Object>) interpolatedLifecycle);
```
shaguptashaikh, line 332: Should this be `replaceAndWait()` or `dflt()` ? What happens when Nucleus restarts and already has lifecycle config from the tlog?

src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
```diff
@@ -226,6 +246,97 @@ private void initializeNucleusComponentConfig() {
...
+                throw new PackageLoadingException("Failed to resolve Nucleus recipe");
+            }
+            ComponentRecipe componentRecipe = resolvedRecipe.get();
+            nucleusComponentName = componentRecipe.getComponentName();
+
+            // Add Nucleus version
+            Semver componentVersion = componentRecipe.getVersion();
+            kernel.getConfig().lookup(SERVICES_NAMESPACE_TOPIC, nucleusComponentName,
+                    VERSION_CONFIG_KEY).withValue(componentVersion.toString());
+            kernel.getConfig().lookup(SETENV_CONFIG_NAMESPACE, GGC_VERSION_ENV).withValue(componentVersion.toString());
```
shaguptashaikh, line 306: This is being done in `initializeNucleusComponentConfig` as well, is this duplicate?

hui-yang: `initializeNucleusComponentConfig` only knows about the fallback version 2.0.0.
You have a good point though. I will move `SETENV_CONFIG_NAMESPACE.GGC_VERSION_ENV` setting to a common code path after version is initialized.

src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
hui-yang: It's prevented with this check https://github.com/aws-greengrass/aws-greengrass-nucleus/pull/806/files/77b497ecff63d8fdcc1771c88d586700ab512160#diff-3de627667101ba5b23e01542a659fd03c9a728449b99d59d0c309ccdbe5badc5R276-R281

Topics only has `updateFromMap` and `replaceAndWait`, but no `dflt`

shaguptashaikh (dismissed):

assembly.xml
```diff
@@ -27,6 +27,7 @@
             <outputDirectory>/conf</outputDirectory>
             <includes>
                 <include>nucleus-build.properties</include>
```
MikeDombo, line 29: remove the properties file?

src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
```diff
@@ -226,6 +245,96 @@ private void initializeNucleusComponentConfig() {
...
+    /**
+     * Load Nucleus component information from build recipe.
+     *
+     * @param kernelAlts KernelAlternatives instance
+     */
+    public void initializedNucleusFromRecipe(KernelAlternatives kernelAlts) {
+        String nucleusComponentName = getNucleusComponentName();
+        Topics nucleusLifecycle = kernel.getConfig().lookupTopics(DEFAULT_VALUE_TIMESTAMP, SERVICES_NAMESPACE_TOPIC,
+                nucleusComponentName, SERVICE_LIFECYCLE_NAMESPACE_TOPIC);
+        if (nucleusLifecycle.children.size() != 0) {
```
MikeDombo, line 277: use isEmpty?

src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
```diff
@@ -226,6 +245,96 @@ private void initializeNucleusComponentConfig() {
...
+            logger.atDebug().log("Nucleus lifecycle has already been initialized");
+            return;
+        }
+
+        persistInitialLaunchParams(kernelAlts);
+        try {
+            Path unpackDir = locateCurrentKernelUnpackDir();
+            Path recipePath = unpackDir.resolve(NUCLEUS_BUILD_METADATA_DIRECTORY)
```
MikeDombo, line 285: validate that the path actually exists before attempting to copy it?

src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
```diff
@@ -226,6 +245,96 @@ private void initializeNucleusComponentConfig() {
...
+            nucleusComponentName = componentRecipe.getComponentName();
+
+            // Add Nucleus version
+            Semver componentVersion = componentRecipe.getVersion();
+            kernel.getConfig().lookup(SERVICES_NAMESPACE_TOPIC, nucleusComponentName,
+                    VERSION_CONFIG_KEY).withValue(componentVersion.toString());
+
+            // Copy unpacked artifacts to component store
+            Path unpackArtifactPath = kernel.getContext().get(NucleusPaths.class).unarchiveArtifactPath(
+                    new ComponentIdentifier(nucleusComponentName, componentVersion), "aws.greengrass.nucleus");
```
MikeDombo, line 308: Is this hardcoded name correct/the only way to do it?

src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
```diff
@@ -226,6 +245,96 @@ private void initializeNucleusComponentConfig() {
...
+            Utils.copyFolderRecursively(unpackDir, unpackArtifactPath, REPLACE_EXISTING, NOFOLLOW_LINKS,
```
MikeDombo, line 309: this might be copying too much, if I don't unpack into a directory which contains only the zip stuff, like if I unzip into my home directory.

We may want to restrict the copying more. Just something to consider anyway.

MikeDombo (commented):
Can you describe in the description what this change allows us to do that we couldn't do before, as well as the testing you've performed to validate the change (any manual testing).

src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
hui-yang: If it's ok not to copy over the license files, I can explicitly copy the bin/ conf/ lib/ directories only

MikeDombo: Seems ok to me.
@jbutler @philcali  any issues with this? We still distribute the licenses, we'd just not be copying them into the unzip path.

MikeDombo (dismissed):

shaguptashaikh (approved):

MikeDombo (commented):
Did you account for cases where there is _some_ existing configuration for the nucleus already in the config such as when the customer has manually created the config.yaml file for manual provisioning?

hui-yang:
> Did you account for cases where there is _some_ existing configuration for the nucleus already in the config such as when the customer has manually created the config.yaml file for manual provisioning?

Yes. The main logic is still in `deviceConfiguration.initializeNucleusFromRecipe`. I've updated it so that all the setup steps should always be reached.

src/main/java/software/amazon/awssdk/http/apache/internal/conn/SdkTlsSocketFactory.java
```diff
@@ -52,7 +52,7 @@ protected final void prepareSocket(final SSLSocket socket) {
         // BEGIN GG MODIFICATIONS
         try {
             SSLParameters params = new SSLParameters();
-            params.setApplicationProtocols(new String[]{"x-amzn-http-ca", "http/1.1"});
+            //params.setApplicationProtocols(new String[]{"x-amzn-http-ca", "http/1.1"});
```
MikeDombo, line 55: Why did you do this?

hui-yang: It's a mistake. I'll fix it

MikeDombo (approved):

shaguptashaikh (approved):

PR #809, opened 2021-01-07
fix: Catch null topic in publish

MikeDombo:
**Issue #, if available:**

**Description of changes:**
Catching the case where the topic is null in a publish message so that we can return an appropriate error.

**Checklist:**


rbattle (dismissed):

src/main/java/com/aws/greengrass/builtin/services/pubsub/PubSubIPCEventStreamAgent.java
```diff
@@ -107,9 +108,13 @@ public PublishToTopicResponse publish(String topic, byte[] binaryMessage, String
         return handlePublishToTopicRequest(topic, serviceName, Optional.empty(), Optional.of(binaryMessage));
     }
 
+    @SuppressWarnings("PMD.PreserveStackTrace")
     private PublishToTopicResponse handlePublishToTopicRequest(String topic, String serviceName,
                                                                Optional<Map<String, Object>> jsonMessage,
                                                                Optional<byte[]> binaryMessage) {
+        if (topic == null) {
```
saranyailla, line 115: Do we also want to check if the topic is empty?

MikeDombo: No, empty is weird, but technically it would work.

saranyailla (approved):

rbattle (approved):

PR #810, opened 2021-01-07
improvement(various): No cloud calls if configured to run offline

tilo-chen:
**Issue #, if available:**

**Description of changes:**
Check if device is configured to run offline, then check and skip cloud calls in:
- IoTJobsHelper
- ShadowDeploymentListener
- MqttClient
- FleetStatusService
- TelemetryAgent
- ComponentManager

Avoid unnecessary cloud communication if device is configured to run offline.

While offline, install Nucleus and CLI, do a local deployment, then shutdown.
Tested with a Hello World recipe with an artifact declared without checksum.
Deployment succeeds. Also, if artifact doesn't exist, fails with expected error.

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/componentmanager/ComponentManager.java
+++ b/src/main/java/com/aws/greengrass/componentmanager/ComponentManager.java
@@ -151,6 +151,12 @@ public class ComponentManager implements InjectionActions {
             // otherwise try to negotiate with cloud
             logger.atInfo().setEventType("negotiate-version-with-cloud-start").log("Negotiating version with cloud");
 
+            if (!deviceConfiguration.isDeviceConfiguredToTalkToCloud()) {
+                throw new NoAvailableComponentVersionException(String.format(
+                        "Device is configured to run offline and no local applicable version found for component '%s' "
+                                + "satisfying requirement '%s'.", componentName, versionRequirements));
+            }
+
             resolvedComponentId =
                     negotiateVersionWithCloud(componentName, versionRequirements, localCandidateOptional.orElse(null));
 
--- a/src/main/java/com/aws/greengrass/componentmanager/ComponentManager.java
+++ b/src/main/java/com/aws/greengrass/componentmanager/ComponentManager.java
@@ -148,11 +148,12 @@ public class ComponentManager implements InjectionActions {
             resolvedComponentId = localCandidateOptional.get();
         } else {
             // Otherwise try to negotiate with cloud
-            logger.atInfo().setEventType("negotiate-version-with-cloud-start").log("Negotiating version with cloud");
-
             if (deviceConfiguration.isDeviceConfiguredToTalkToCloud()) {
+                logger.atInfo().setEventType("negotiate-version-with-cloud-start")
+                        .log("Negotiating version with cloud");
                 resolvedComponentId = negotiateVersionWithCloud(componentName, versionRequirements,
                         localCandidateOptional.orElse(null));
+                logger.atInfo().setEventType("negotiate-version-with-cloud-end").log("Negotiated version with cloud");
             } else {
                 // Device running offline. Use the local candidate if present, otherwise fails
                 if (localCandidateOptional.isPresent()) {
@@ -165,8 +166,6 @@ public class ComponentManager implements InjectionActions {
                                     + " '%s' satisfying requirement '%s'.", componentName, versionRequirements));
                 }
             }
-
-            logger.atInfo().setEventType("negotiate-version-with-cloud-end").log("Negotiated version with cloud");
         }
 
         logger.atInfo().setEventType("resolve-component-version-end").kv("ResolvedComponent", resolvedComponentId)
@@ -370,9 +369,9 @@ public class ComponentManager implements InjectionActions {
             ArtifactDownloader downloader = artifactDownloaderFactory
                     .getArtifactDownloader(componentIdentifier, artifact, packageArtifactDirectory);
             if (downloader.downloadRequired()) {
-                if (!deviceConfiguration.isDeviceConfiguredToTalkToCloud()) {
-                    throw new PackageDownloadException("Artifact download required but device is configured to run "
-                            + "offline");
+                if (!downloader.downloadReady()) {
+                    throw new PackageDownloadException(String.format(
+                            "Download required for artifact %s but device configs are invalid for download", artifact));
                 }
                 // Check disk size limits before download
                 // TODO: [P41215447]: Check artifact size for all artifacts to download early to fail early
--- a/src/main/java/com/aws/greengrass/componentmanager/ComponentManager.java
+++ b/src/main/java/com/aws/greengrass/componentmanager/ComponentManager.java
@@ -369,9 +369,11 @@ public class ComponentManager implements InjectionActions {
             ArtifactDownloader downloader = artifactDownloaderFactory
                     .getArtifactDownloader(componentIdentifier, artifact, packageArtifactDirectory);
             if (downloader.downloadRequired()) {
-                if (!downloader.downloadReady()) {
+                Optional<String> errorMsg = downloader.checkDownloadable();
+                if (errorMsg.isPresent()) {
                     throw new PackageDownloadException(String.format(
-                            "Download required for artifact %s but device configs are invalid for download", artifact));
+                            "Download required for artifact %s but device configs are invalid:%n%s",
+                            artifact.getArtifactUri(), errorMsg.get()));
                 }
                 // Check disk size limits before download
                 // TODO: [P41215447]: Check artifact size for all artifacts to download early to fail early
--- a/src/main/java/com/aws/greengrass/componentmanager/ComponentManager.java
+++ b/src/main/java/com/aws/greengrass/componentmanager/ComponentManager.java
@@ -367,18 +367,21 @@ public class ComponentManager implements InjectionActions {
                 .addKeyValue(PACKAGE_IDENTIFIER, componentIdentifier).log();
 
         for (ComponentArtifact artifact : artifacts) {
-            // check disk space before download
-            // TODO: [P41215447]: Check artifact size for all artifacts to download early to fail early
-            long usableSpaceBytes = componentStore.getUsableSpace();
-            if (usableSpaceBytes < DEFAULT_MIN_DISK_AVAIL_BYTES) {
-                throw new SizeLimitException(
-                        String.format("Disk space critical: %d bytes usable, %d bytes minimum allowed",
-                                usableSpaceBytes, DEFAULT_MIN_DISK_AVAIL_BYTES));
-            }
             ArtifactDownloader downloader = artifactDownloaderFactory
                     .getArtifactDownloader(componentIdentifier, artifact, packageArtifactDirectory);
-
             if (downloader.downloadRequired()) {
+                if (!deviceConfiguration.isDeviceConfiguredToTalkToCloud()) {
+                    throw new PackageDownloadException("Artifact download required but device is configured to run "
+                            + "offline");
+                }
+                // Check disk size limits before download
+                // TODO: [P41215447]: Check artifact size for all artifacts to download early to fail early
+                long usableSpaceBytes = componentStore.getUsableSpace();
+                if (usableSpaceBytes < DEFAULT_MIN_DISK_AVAIL_BYTES) {
+                    throw new SizeLimitException(
+                            String.format("Disk space critical: %d bytes usable, %d bytes minimum allowed",
+                                    usableSpaceBytes, DEFAULT_MIN_DISK_AVAIL_BYTES));
+                }
                 long downloadSize = downloader.getDownloadSize();
                 long storeContentSize = componentStore.getContentSize();
                 if (storeContentSize + downloadSize > getConfiguredMaxSize()) {
--- a/src/main/java/com/aws/greengrass/componentmanager/ComponentManager.java
+++ b/src/main/java/com/aws/greengrass/componentmanager/ComponentManager.java
@@ -136,7 +136,6 @@ public class ComponentManager implements InjectionActions {
                     .log("Found the best local candidate that satisfies the requirement.");
         } else {
             logger.atInfo().log("Can't find a local candidate that satisfies the requirement.");
-
         }
         ComponentIdentifier resolvedComponentId;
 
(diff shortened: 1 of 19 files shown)


src/main/java/com/aws/greengrass/easysetup/GreengrassSetup.java
```diff
@@ -287,10 +287,13 @@ void performSetup() throws IOException, DeviceConfigurationException, URISyntaxE
             throw new RuntimeException(e);
         }
 
-        //initialize the device provisioning helper
-        this.deviceProvisioningHelper = new DeviceProvisioningHelper(awsRegion, environmentStage, this.outStream);
-
         if (needProvisioning) {
+            if (!deviceConfiguration.isDeviceConfiguredToTalkToCloud()) {
```
MikeDombo, line 291: this isn't right. The provisioning will perform the device configuration so that it can talk to the cloud.

src/main/java/com/aws/greengrass/status/FleetStatusService.java
```diff
@@ -138,10 +138,16 @@ public FleetStatusService(Topics topics, MqttClient mqttClient, DeploymentStatus
...
         super(topics);
-
         this.mqttClient = mqttClient;
         this.deploymentStatusKeeper = deploymentStatusKeeper;
         this.kernel = kernel;
+        this.deviceConfiguration = deviceConfiguration;
+
+        if (!deviceConfiguration.isDeviceConfiguredToTalkToCloud()) {
+            logger.atWarn().log("Device not configured to talk to AWS Iot cloud. FSS is offline");
+            return;
```
MikeDombo, line 148: Same here as with the MQTT client. Don't partially configure the class, construct the whole thing, but modify its behavior after construction.

src/main/java/com/aws/greengrass/telemetry/TelemetryAgent.java
```diff
@@ -122,12 +122,18 @@ public TelemetryAgent(Topics topics, MqttClient mqttClient, DeviceConfiguration
...
-        this.publisher = new MqttChunkedPayloadPublisher<>(this.mqttClient);
-        this.publisher.setMaxPayloadLengthBytes(MAX_PAYLOAD_LENGTH_BYTES);
         this.ses = ses;
         this.executorService = executorService;
         this.metricsAggregator = ma;
         this.deviceConfiguration = deviceConfiguration;
+
+        if (!deviceConfiguration.isDeviceConfiguredToTalkToCloud()) {
+            logger.atWarn().log("Device not configured to talk to AWS Iot cloud. Telemetry is offline");
+            return;
```
MikeDombo, line 132: Same

MikeDombo (changes requested):

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -136,6 +136,11 @@ public MqttClient(DeviceConfiguration deviceConfiguration,
                       ScheduledExecutorService ses) {
         this(deviceConfiguration, null, ses);
 
+        if (!deviceConfiguration.isDeviceConfiguredToTalkToCloud()) {
```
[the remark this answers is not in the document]
tilo-chen: Thanks! Makes sense. I was thinking that this might save some resource.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -294,6 +299,10 @@ protected MqttClient(DeviceConfiguration deviceConfiguration, Spool spool, Sched
     @SuppressWarnings("PMD.CloseResource")
     public synchronized void subscribe(SubscribeRequest request)
             throws ExecutionException, InterruptedException, TimeoutException {
+        if (!deviceConfiguration.isDeviceConfiguredToTalkToCloud()) {
+            throw new ExecutionException(new Exception("Cannot subscribe because device is configured to run offline"));
```
MikeDombo, line 303: I'm not so sure that throwing an exception is the right thing to do at this point. Open to discussion, but I think it would be better to just log and return here instead of throwing.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -424,6 +433,11 @@ private synchronized void spoolMessage()  {
      * Iterate the spooler queue to publish all the spooled message.
      */
     protected void spoolTask() {
+        if (!deviceConfiguration.isDeviceConfiguredToTalkToCloud()) {
```
MikeDombo, line 436: I wouldn't have this here. Just put it in the `publish` method instead.

MikeDombo (changes requested):
How about all the cloud calls during a deployment?

fengwa-aws:
What happens if a customer later does a local deployment to configure the `DeviceConfiguration` so that `isDeviceConfiguredToTalkToCloud` is changed to `True`? Do all the cloud related services restart and check the flag again? Have you tested that scenario?

tilo-chen:
> What happens if a customer later does a local deployment to configure the `DeviceConfiguration` so that `isDeviceConfiguredToTalkToCloud` is changed to `True`? Do all the cloud related services restart and check the flag again? Have you tested that scenario?

No. I didn't consider that case. I think those services can rely on `MqttClient` connection status callback to get back online. Shouldn't need to restart the services.
On the other hand, it's much easier for the customer to just restart the nucleus with `--provision` options right? Because manually setting up the cert, keys, policy, etc. can be nontrivial.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
tilo-chen: Changed to not throw. Consistent with how subscription failure is handled.

tilo-chen:
> How about all the cloud calls during a deployment?

If the device is configured to run offline, I think it's safe to assume only local deployment will get submitted. That workflow doesn't interact with cloud.

MikeDombo:
> > How about all the cloud calls during a deployment?
> 
> If the device is configured to run offline, I think it's safe to assume only local deployment will get submitted. That workflow doesn't interact with cloud.

No, that's the point, a local deployment still calls cloud APIs. Your assertion is not correct. We call the cloud for resolving component versions.

src/main/java/com/aws/greengrass/componentmanager/ComponentManager.java
```diff
@@ -148,11 +147,24 @@ ComponentMetadata resolveComponentVersion(String componentName, Map<String, Requ
                     + " candidate as the resolved version without negotiating with cloud.");
             resolvedComponentId = localCandidateOptional.get();
         } else {
-            // otherwise try to negotiate with cloud
+            // Otherwise try to negotiate with cloud
             logger.atInfo().setEventType("negotiate-version-with-cloud-start").log("Negotiating version with cloud");
 
-            resolvedComponentId =
-                    negotiateVersionWithCloud(componentName, versionRequirements, localCandidateOptional.orElse(null));
+            if (deviceConfiguration.isDeviceConfiguredToTalkToCloud()) {
```
shaguptashaikh, line 153: Do we not need similar changes for artifacts download? Is that already covered?

tilo-chen: So I saw [here](https://docs.aws.amazon.com/greengrass/v2/developerguide/getting-started.html#create-first-component) the recipe for local deployment doesn't declare artifacts (just directly uses it). Is this always the case? Then we don't need change because artifacts will be empty.

MikeDombo: No, if the recipe lists artifacts then we might try to download them (if they don't exist already).

shaguptashaikh (approved):

src/main/java/com/aws/greengrass/componentmanager/ComponentManager.java
MikeDombo: Move this check above the log line, otherwise the log doesn't make sense since we're skipping.

src/main/java/com/aws/greengrass/componentmanager/ComponentManager.java
```diff
@@ -355,18 +367,21 @@ void prepareArtifacts(ComponentIdentifier componentIdentifier, List<ComponentArt
...
-            if (usableSpaceBytes < DEFAULT_MIN_DISK_AVAIL_BYTES) {
-                throw new SizeLimitException(
-                        String.format("Disk space critical: %d bytes usable, %d bytes minimum allowed",
-                                usableSpaceBytes, DEFAULT_MIN_DISK_AVAIL_BYTES));
-            }
             ArtifactDownloader downloader = artifactDownloaderFactory
                     .getArtifactDownloader(componentIdentifier, artifact, packageArtifactDirectory);
-
             if (downloader.downloadRequired()) {
+                if (!deviceConfiguration.isDeviceConfiguredToTalkToCloud()) {
```
MikeDombo, line 373: I'd probably move this check inside of the artifact downloader. We might have downloaders which don't require this configuration and thus having the limitation wouldn't make sense.

src/main/java/com/aws/greengrass/status/FleetStatusService.java
```diff
@@ -184,6 +184,10 @@ public FleetStatusService(Topics topics, MqttClient mqttClient, DeploymentStatus
         this.mqttClient.addToCallbackEvents(callbacks);
 
         TestFeatureParameters.registerHandlerCallback(this.getName(), this::handleTestFeatureParametersHandlerChange);
+
+        if (!deviceConfiguration.isDeviceConfiguredToTalkToCloud()) {
+            this.isConnected.set(false);
```
MikeDombo, line 189: How does this affect FSS? Make sure you understand the ramifications of it. For example, do we cache data in memory while waiting for a connection, could we blow up our memory usage because of this?

src/main/java/com/aws/greengrass/telemetry/TelemetryAgent.java
```diff
@@ -147,6 +147,10 @@ public TelemetryAgent(Topics topics, MqttClient mqttClient, DeviceConfiguration
         // Subscribe to thing name changes.
         deviceConfiguration.getThingName()
                 .subscribe((why, node) -> updateThingNameAndPublishTopic(Coerce.toString(node)));
+
+        if (!deviceConfiguration.isDeviceConfiguredToTalkToCloud()) {
+            this.isConnected.set(false);
```
MikeDombo, line 152: Same question as in FSS above.

src/main/java/com/aws/greengrass/componentmanager/ComponentManager.java
tilo-chen: Good point!

src/main/java/com/aws/greengrass/status/FleetStatusService.java
tilo-chen: So FSS data is computed ad-hoc and not cached. It just skips the computation if it's not connected.
Also it looks like we can skip the scheduled task and listeners setup here because right now connection can't be brought online without a restart.

src/main/java/com/aws/greengrass/telemetry/TelemetryAgent.java
tilo-chen: Telemetry data is emitted as log files and not cached in memory. We can probably also skip scheduled upload for the same reason as FSS.

src/main/java/com/aws/greengrass/componentmanager/ComponentManager.java
```diff
@@ -355,18 +366,21 @@ void prepareArtifacts(ComponentIdentifier componentIdentifier, List<ComponentArt
...
-            if (usableSpaceBytes < DEFAULT_MIN_DISK_AVAIL_BYTES) {
-                throw new SizeLimitException(
-                        String.format("Disk space critical: %d bytes usable, %d bytes minimum allowed",
-                                usableSpaceBytes, DEFAULT_MIN_DISK_AVAIL_BYTES));
-            }
             ArtifactDownloader downloader = artifactDownloaderFactory
                     .getArtifactDownloader(componentIdentifier, artifact, packageArtifactDirectory);
-
             if (downloader.downloadRequired()) {
+                if (!downloader.downloadReady()) {
```
MikeDombo, line 372: maybe `canDownload`?

MikeDombo: Have that return an Optional string so that we can tell the user _why_ it is unable to download, not only that it can't do it.

src/main/java/com/aws/greengrass/telemetry/TelemetryAgent.java
```diff
@@ -142,11 +142,18 @@ public TelemetryAgent(Topics topics, MqttClient mqttClient, DeviceConfiguration
         getPeriodicAggregateTimeTopic();
         getPeriodicPublishTimeTopic();
         schedulePeriodicAggregateMetrics(false);
```
MikeDombo, line 144: why not move L144 also? We don't need to schedule right?

tilo-chen: It can be helpful to keep the periodic aggregation. Even though it can't be uploaded, may still be useful offline?

MikeDombo: OK.

src/main/java/com/aws/greengrass/componentmanager/GreengrassComponentServiceClientFactory.java
```diff
@@ -53,8 +53,8 @@
 public class GreengrassComponentServiceClientFactory {
 
     private static final Logger logger = LogManager.getLogger(GreengrassComponentServiceClientFactory.class);
-
     private GreengrassV2Client cmsClient;
+    private final boolean configValid;
```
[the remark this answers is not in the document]
MikeDombo: Nevermind.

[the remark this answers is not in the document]
tilo-chen: This is accessed via `@getter` on the whole class. Is it an anti-pattern?

src/main/java/com/aws/greengrass/componentmanager/ComponentManager.java
```diff
@@ -355,18 +366,23 @@ void prepareArtifacts(ComponentIdentifier componentIdentifier, List<ComponentArt
...
-                                usableSpaceBytes, DEFAULT_MIN_DISK_AVAIL_BYTES));
-            }
             ArtifactDownloader downloader = artifactDownloaderFactory
                     .getArtifactDownloader(componentIdentifier, artifact, packageArtifactDirectory);
-
             if (downloader.downloadRequired()) {
+                Optional<String> errorMsg = downloader.checkDownloadable();
+                if (errorMsg.isPresent()) {
+                    throw new PackageDownloadException(String.format(
+                            "Download required for artifact %s but device configs are invalid:%n%s",
```
MikeDombo, line 375: "device configuration is invalid: %s". I would NOT put a newline inside it.

shaguptashaikh (approved):

MikeDombo (approved):

shaguptashaikh (approved):

PR #811, opened 2021-01-12
improvement: Make Greengrass-ATS endpoint configurable

MikeDombo:
**Issue #, if available:**

**Description of changes:**
Makes the Greengrass-ATS endpoint port configurable from 8443 (default) to 443 (optional).
Adds an option in the Nucleus configuration called "greengrassDataPlanePort" which can be configured through a deployment. Needed to copy `SdkTlsSocketFactory` from the SDK and override `prepareSocket` so that we can add ALPN to the socket before it is used by the SDK since the SDK has no such support for this right now.

Note to reviewers: do not review `SdkTlsSocketFactory` except for what is between the "modification markers". The rest of the code is directly from the SDK and should not be changed.

Some customers require that all traffic uses port 443, this change enables that functionality. Default is still to use port 8443.

Manually verified that when set to port 443 the nucleus is still able to perform a deployment (which requires that the ats endpoint is working correctly). Verified that without the changes to the SDK socket factory to add ALPN, we are rejected by the cloud.

**Checklist:**

### DIFF
--- a/src/main/java/software/amazon/awssdk/http/apache/internal/conn/SdkTlsSocketFactory.java
+++ b/src/main/java/software/amazon/awssdk/http/apache/internal/conn/SdkTlsSocketFactory.java
@@ -1,0 +1,122 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package software.amazon.awssdk.http.apache.internal.conn;
+
+import org.apache.http.HttpHost;
+import org.apache.http.conn.ssl.SSLConnectionSocketFactory;
+import org.apache.http.protocol.HttpContext;
+import software.amazon.awssdk.annotations.SdkInternalApi;
+import software.amazon.awssdk.http.apache.internal.net.SdkSocket;
+import software.amazon.awssdk.http.apache.internal.net.SdkSslSocket;
+import software.amazon.awssdk.utils.Logger;
+
+import java.io.IOException;
+import java.net.InetSocketAddress;
+import java.net.Socket;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import javax.net.ssl.HostnameVerifier;
+import javax.net.ssl.SSLContext;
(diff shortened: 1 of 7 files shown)


src/main/java/software/amazon/awssdk/http/apache/internal/conn/SdkTlsSocketFactory.java
```diff
@@ -0,0 +1,122 @@
...
+
+    /**
+     * {@inheritDoc} Used to enforce the preferred TLS protocol during SSL handshake.
+     */
+    @Override
+    protected final void prepareSocket(final SSLSocket socket) {
+        // BEGIN GG MODIFICATIONS
+        SSLParameters params = new SSLParameters();
+        params.setApplicationProtocols(new String[]{"x-amzn-http-ca", "http/1.1"});
+        socket.setSSLParameters(params);
```
fengwa-aws, line 55: So are you saying these modifications are needed to connect to port 443? How do existing customers connect to port 443 using the AWS SDK then?

MikeDombo: These are only needed because of IoT Core which requires that HTTPS connections use ALPN when connecting to port 443 instead of 8443.

ShirleyZheng92 (approved):

shaguptashaikh (commented):
How do we expect this option to be used? By providing the config.yaml in setup?

MikeDombo:
> How do we expect this option to be used? By providing the config.yaml in setup?

Yes, like all things that's an option. It _can_ also be updated in a deployment since it is a property of the nucleus now.

shaguptashaikh:
> Yes, like all things that's an option. It _can_ also be updated in a deployment since it is a property of the nucleus now.

Right, but unlike other nucleus properties, for changing this port through deployment, you have to make that deployment that calls the Greengrass dataplane endpoint and that deployment will use the default port 8443.

shaguptashaikh (approved):

fengwa-aws (approved):

PR #815, opened 2021-01-19
fix: Catch NoSuchMethodError in ALPN implementation for old JDK8 versions

MikeDombo:
**Issue #, if available:**

**Description of changes:**
Catches `NoSuchMethodError` in the ALPN implementation since ALPN did not launch with Java 8, but was only backported from Java 9 in April 2020 with JDK8u252. By catching the exception we will enable users to still use port 8443 without issue, it is just that port 443 won't work for the greengrass-ats endpoint.

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/lifecyclemanager/PluginService.java
+++ b/src/main/java/com/aws/greengrass/lifecyclemanager/PluginService.java
@@ -14,12 +14,10 @@ import java.net.URL;
 import java.net.URLClassLoader;
 import java.util.Map;
 import javax.inject.Inject;
-import javax.inject.Singleton;
 
 import static com.aws.greengrass.componentmanager.KernelConfigResolver.VERSION_CONFIG_KEY;
 import static com.aws.greengrass.deployment.bootstrap.BootstrapSuccessCode.REQUEST_RESTART;
 
-@Singleton
 public class PluginService extends GreengrassService {
     @Inject
     private EZPlugins ezPlugins;
--- a/src/main/java/com/aws/greengrass/lifecyclemanager/PluginService.java
+++ b/src/main/java/com/aws/greengrass/lifecyclemanager/PluginService.java
@@ -14,10 +14,12 @@ import java.net.URL;
 import java.net.URLClassLoader;
 import java.util.Map;
 import javax.inject.Inject;
+import javax.inject.Singleton;
 
 import static com.aws.greengrass.componentmanager.KernelConfigResolver.VERSION_CONFIG_KEY;
 import static com.aws.greengrass.deployment.bootstrap.BootstrapSuccessCode.REQUEST_RESTART;
 
+@Singleton
 public class PluginService extends GreengrassService {
     @Inject
     private EZPlugins ezPlugins;
--- a/src/main/java/software/amazon/awssdk/http/apache/internal/conn/SdkTlsSocketFactory.java
+++ b/src/main/java/software/amazon/awssdk/http/apache/internal/conn/SdkTlsSocketFactory.java
@@ -50,9 +50,16 @@ public class SdkTlsSocketFactory extends SSLConnectionSocketFactory {
     @Override
     protected final void prepareSocket(final SSLSocket socket) {
         // BEGIN GG MODIFICATIONS
-        SSLParameters params = new SSLParameters();
-        params.setApplicationProtocols(new String[]{"x-amzn-http-ca", "http/1.1"});
-        socket.setSSLParameters(params);
+        try {
+            SSLParameters params = new SSLParameters();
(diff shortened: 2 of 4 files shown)


src/main/java/software/amazon/awssdk/http/apache/internal/conn/SdkTlsSocketFactory.java
```diff
@@ -50,9 +50,16 @@ public SdkTlsSocketFactory(final SSLContext sslContext, final HostnameVerifier h
...
         // BEGIN GG MODIFICATIONS
-        SSLParameters params = new SSLParameters();
-        params.setApplicationProtocols(new String[]{"x-amzn-http-ca", "http/1.1"});
-        socket.setSSLParameters(params);
+        try {
+            SSLParameters params = new SSLParameters();
+            params.setApplicationProtocols(new String[]{"x-amzn-http-ca", "http/1.1"});
+            socket.setSSLParameters(params);
+        } catch (NoSuchMethodError e) {
+            log.debug(() -> "Unable to configure socket for ALPN. Ports other than 443 may not work.");
```
nikkhilmuthye, line 58: nit: Why debug?

nikkhilmuthye (approved):

src/main/java/software/amazon/awssdk/http/apache/internal/conn/SdkTlsSocketFactory.java
MikeDombo: This method is called for every single SDK operation that needs a socket, so this would massively pollute the logs if it were any higher.

fengwa-aws: `Ports other than 443 may not work.`
-- Do you mean `may work`?

fengwa-aws (approved):

src/main/java/software/amazon/awssdk/http/apache/internal/conn/SdkTlsSocketFactory.java
```diff
@@ -50,9 +50,16 @@ public SdkTlsSocketFactory(final SSLContext sslContext, final HostnameVerifier h
...
         // BEGIN GG MODIFICATIONS
-        SSLParameters params = new SSLParameters();
-        params.setApplicationProtocols(new String[]{"x-amzn-http-ca", "http/1.1"});
-        socket.setSSLParameters(params);
+        try {
+            SSLParameters params = new SSLParameters();
+            params.setApplicationProtocols(new String[]{"x-amzn-http-ca", "http/1.1"});
+            socket.setSSLParameters(params);
+        } catch (NoSuchMethodError e) {
+            log.debug(() -> "Unable to configure socket for ALPN. Ports other than 443 may still work.");
```
rbattle, line 58: Do we know here if they are using port 443 - and if so log at a higher level like warn?

src/main/java/com/aws/greengrass/lifecyclemanager/PluginService.java
```diff
@@ -14,10 +14,12 @@
 import java.net.URLClassLoader;
 import java.util.Map;
 import javax.inject.Inject;
+import javax.inject.Singleton;
 
 import static com.aws.greengrass.componentmanager.KernelConfigResolver.VERSION_CONFIG_KEY;
 import static com.aws.greengrass.deployment.bootstrap.BootstrapSuccessCode.REQUEST_RESTART;
 
+@Singleton
```
rbattle, line 22: does `PluginService` itself ever get instantiated? Or do classes just extend it?

src/main/java/software/amazon/awssdk/http/apache/internal/conn/SdkTlsSocketFactory.java
MikeDombo: No, we can't tell because this is so low level. We're just making a socket which is later used to connect to some port. We don't have enough info to do anything smarter.

src/main/java/com/aws/greengrass/lifecyclemanager/PluginService.java
MikeDombo: Only extending classes.

MikeDombo: But the annotation should apply to extending classes, right?

rbattle: The singleton annotation is not inherited by subclasses - annotations can have an `@Inherited` meta-annotation on them - but singleton does not

MikeDombo: OK thanks, I reverted it. Will look into it later.

fengwa-aws (approved):

rbattle (approved):

PR #827, opened 2021-01-22
fix(shadow): prevent duplicated shadow deployment at startup

tilo-chen:
**Issue #, if available:**

**Description of changes:**
Add a check for the first shadow doc received since startup. Ignore the deployment if `reported` section exists and the ARN matches desired ARN. Because the deployment had already been processed and reported.

Fix duplicated shadow deployment when nucleus restarts.

Manual tests:
On EC2:
- From cloud console, deploy a component to the device that will run successfully
- Verify shadow status reported as success
- restart
- Verify log has `Deployment result already reported. Ignoring shadow update at startup.` for the previous deployment config ARN. No duplicate deployment queued
- Deploy a component to the device that will take a long time to install
- Verify shadow status reported as in progress
- restart
- Verify last device deployment is picked up again and can finish successfully.
- Deploy a component to the device that will fail to install
- Verify shadow status reported as failure
- restart
- Verify log has `Deployment result already reported. Ignoring shadow update at startup.` for the previous deployment config ARN. No duplicate deployment queued
- Deploy a component that takes long to install
- Cancel deployment before it finishes
- Verify desired shadow status is `canceled` and reported status is `in_progress`
- restart
- Verify log has `Deployment was canceled. Ignoring shadow update at startup` for the previous deployment config ARN. No duplicate deployment queued

Adding UAT scenario for the above manual test plan.

`@Smoke` run on ubuntu DUT all passed.

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/deployment/DeploymentService.java
+++ b/src/main/java/com/aws/greengrass/deployment/DeploymentService.java
@@ -81,7 +81,6 @@ public class DeploymentService extends GreengrassService {
     public static final String DEPLOYMENT_SERVICE_TOPICS = "DeploymentService";
     public static final String GROUP_TO_ROOT_COMPONENTS_TOPICS = "GroupToRootComponents";
     public static final String COMPONENTS_TO_GROUPS_TOPICS = "ComponentToGroups";
-    public static final String LAST_SUCCESSFUL_SHADOW_DEPLOYMENT_ID_TOPIC = "LastSuccessfulShadowDeploymentId";
     public static final String GROUP_TO_ROOT_COMPONENTS_VERSION_KEY = "version";
     public static final String GROUP_TO_ROOT_COMPONENTS_GROUP_CONFIG_ARN = "groupConfigArn";
     public static final String GROUP_TO_ROOT_COMPONENTS_GROUP_NAME = "groupConfigName";
@@ -209,17 +208,9 @@ public class DeploymentService extends GreengrassService {
                         logger.atInfo().kv(DEPLOYMENT_ID_LOG_KEY_NAME, currentDeploymentTaskMetadata.getDeploymentId())
                                 .log("The current deployment cannot be cancelled");
                     }
-                } else if (deployment.getDeploymentType().equals(DeploymentType.SHADOW)) {
+                } else if (DeploymentType.SHADOW.equals(deployment.getDeploymentType())) {
                     // The deployment type is shadow
-                    if (deployment.getId()
-                            .equals(Coerce.toString(config.lookup(LAST_SUCCESSFUL_SHADOW_DEPLOYMENT_ID_TOPIC)))) {
-                        // On device start up, Shadow listener will fetch the shadow and schedule a shadow deployment
-                        // Discard the deployment if Kernel starts up from a tlog file
-                        // and has already processed deployment
-                        logger.atInfo().kv(DEPLOYMENT_ID_LOG_KEY_NAME, deployment.getId())
-                                .log("Skip the already processed shadow deployment");
-                        deploymentQueue.remove();
-                    } else if (currentDeploymentTaskMetadata != null && DeploymentType.SHADOW
+                    if (currentDeploymentTaskMetadata != null && DeploymentType.SHADOW
                             .equals(currentDeploymentTaskMetadata.getDeploymentType())) {
                         // A new device deployment invalidates the previous deployment, cancel the ongoing device
                         //deployment and wait till the new device deployment can be picked up.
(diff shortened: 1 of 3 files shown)


src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
```diff
@@ -245,6 +247,14 @@ protected void shadowUpdated(Map<String, Object> desired, Integer version) {
         boolean cancelDeployment = DESIRED_STATUS_CANCELED.equals(desired.get(DESIRED_STATUS_KEY));
 
         synchronized (ShadowDeploymentListener.class) {
```
MikeDombo, line 249: the code in the method isn't very nullsafe, can you go over all of it and add null checks when reading the fields?

MikeDombo (dismissed):

src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
```diff
@@ -242,18 +244,31 @@ protected void shadowUpdated(Map<String, Object> desired, Integer version) {
...
         String configurationArn = configuration.getConfigurationArn();
+        if (configurationArn == null) {
+            logger.atError().log("Desired state has null configuration ARN. Ignoring shadow update");
+            return;
+        }
         boolean cancelDeployment = DESIRED_STATUS_CANCELED.equals(desired.get(DESIRED_STATUS_KEY));
 
         synchronized (ShadowDeploymentListener.class) {
+            if (lastConfigurationArn == null && reported != null && configurationArn
+                    .equals(reported.get(ARN_FOR_STATUS_KEY))) {
```
hui-yang, line 255: Need clarification on a few more cases
1. If `cancelDeployment` and `lastConfigurationArn == null`, skip
1. If `!cancelDeployment` and `reported.get(STATUS_KEY) == in_progress`, should we still process the deployment, so the behavior is the same as jobs deployment? (If I understand correctly)
1. If `!cancelDeployment` and `reported.get(STATUS_KEY)` is a final state, skip

tilo-chen: Cleaned up the cases a bit. Case 2 I think we should still process it.

fahadmohammed01 (approved):

src/main/java/com/aws/greengrass/deployment/DeploymentService.java
```diff
@@ -211,15 +210,7 @@ protected void startup() throws InterruptedException {
                     }
                 } else if (deployment.getDeploymentType().equals(DeploymentType.SHADOW)) {
```
MikeDombo, line 211: can you switch the order of the equals here to prevent NPE?

src/main/java/com/aws/greengrass/deployment/DeploymentService.java
```diff
@@ -209,17 +208,9 @@ protected void startup() throws InterruptedException {
...
+                } else if (DeploymentType.SHADOW.equals(deployment.getDeploymentType())) {
                     // The deployment type is shadow
-                    if (deployment.getId()
-                            .equals(Coerce.toString(config.lookup(LAST_SUCCESSFUL_SHADOW_DEPLOYMENT_ID_TOPIC)))) {
-                        // On device start up, Shadow listener will fetch the shadow and schedule a shadow deployment
-                        // Discard the deployment if Kernel starts up from a tlog file
-                        // and has already processed deployment
-                        logger.atInfo().kv(DEPLOYMENT_ID_LOG_KEY_NAME, deployment.getId())
-                                .log("Skip the already processed shadow deployment");
-                        deploymentQueue.remove();
```
MikeDombo, line 221: why aren't we removing from the queue anymore?

tilo-chen: This check is basically moved here and we won't queue the duplicated deployment any more: https://github.com/aws-greengrass/aws-greengrass-nucleus/pull/827/files#diff-59c7fe23a5126f4c98b9b2baf83faa49534d72dbff905244da12031b7c523392R272-R276
This check was not covering cases other than last deployment succeeded. The new logic will cover those as well.

MikeDombo (commented):
Have you manually built and verified the fix since making these larger changes?

tilo-chen:
> Have you manually built and verified the fix since making these larger changes?

Haven't yet. Running through some scenarios right now.

MikeDombo (approved):

hui-yang (approved):

MikeDombo (approved):

hui-yang (approved):

src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
```diff
@@ -128,8 +129,9 @@ private void subscribeToShadowTopics() {
                 updateNamedShadowSubscriptionRequest.shadowName = DEPLOYMENT_SHADOW_NAME;
                 updateNamedShadowSubscriptionRequest.thingName = thingName;
                 iotShadowClient.SubscribeToUpdateNamedShadowAccepted(updateNamedShadowSubscriptionRequest,
```
aimiebell, line 131: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Method names should start with a lower case letter (NM_METHOD_NAMING_CONVENTION)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-method-names-should-start-with-a-lower-case-letter-nm-method-naming-convention).
Methods should be verbs, in mixed case with the first letter lowercase, with the first letter of each internal word capitalized.

MikeDombo: This method is from the IoT Device SDK v2: https://github.com/aws/aws-iot-device-sdk-java-v2/blob/main/sdk/src/main/java/software/amazon/awssdk/iot/iotshadow/IotShadowClient.java#L719

We don't get to choose the name of the method as it is part of the library.

PR #828, opened 2021-01-22
fix: Dynamic reconfigure GG client when port changes

MikeDombo:
**Issue #, if available:**

**Description of changes:**
Update the GG client when the port changes.

**Checklist:**


shaguptashaikh (commented):
Why can't we reconfigure the data plane client instead of restarting nucleus? The former is what happens when region changes, so it should work for the port as well right?

MikeDombo:
> Why can't we reconfigure the data plane client instead of restarting nucleus? The former is what happens when region changes, so it should work for the port as well right?

Thanks, really good call. Updated.

shaguptashaikh (dismissed):

shaguptashaikh (approved):

hui-yang (approved):

fengwa-aws (approved):

PR #829, opened 2021-01-25
fix(spooler): Improving spooler latency

awszztt:
**Description of changes:**
In the MqttClient, instead of using the  ScheduledExecutorService, executorService is used to control the spooling process. While the mqttClient is online, the spooler always working and waiting for the next available message if the the queue is empty. 

Original, the spooler would sleep for 5 seconds if the queue is empty. The sleeping time would result in the latency of the message publishing. 

test with "mvn verify"

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
+++ b/src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
@@ -15,6 +15,10 @@ import com.aws.greengrass.mqttclient.spool.SpoolerStoreException;
 import com.aws.greengrass.util.Coerce;
 import com.aws.greengrass.util.LockScope;
 import com.aws.greengrass.util.ProxyUtils;
+import com.aws.greengrass.util.Utils;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import lombok.AccessLevel;
+import lombok.Getter;
 import software.amazon.awssdk.crt.auth.credentials.X509CredentialsProvider;
 import software.amazon.awssdk.crt.http.HttpProxyOptions;
 import software.amazon.awssdk.crt.io.ClientBootstrap;
@@ -39,7 +43,9 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.CopyOnWriteArraySet;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
+import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
@@ -101,10 +107,11 @@ public class MqttClient implements Closeable {
     private final ClientBootstrap clientBootstrap;
     private final CallbackEventManager callbackEventManager = new CallbackEventManager();
     private final Spool spool;
-    private final ScheduledExecutorService ses;
+    private final ExecutorService executorService;
     private final AtomicReference<Future<?>> spoolingFuture = new AtomicReference<>();
     private final int maxInFlightPublishes = DEFAULT_MAX_IN_FLIGHT_PUBLISHES;
 
+    @Getter(AccessLevel.PROTECTED)
     private final MqttClientConnectionEvents callbacks = new MqttClientConnectionEvents() {
         @Override
         public void onConnectionInterrupted(int errorCode) {
@@ -117,6 +124,7 @@ public class MqttClient implements Closeable {
         @Override
         public void onConnectionResumed(boolean sessionPresent) {
             setMqttOnline(true);
+            triggerSpooler();
         }
     };
 
@@ -128,13 +136,15 @@ public class MqttClient implements Closeable {
 
     /**
      * Constructor for injection.
+     *
      * @param deviceConfiguration device configuration
      * @param ses                 scheduled executor service
+     * @param executorService     executor service
      */
     @Inject
-    public MqttClient(DeviceConfiguration deviceConfiguration,
-                      ScheduledExecutorService ses) {
-        this(deviceConfiguration, null, ses);
+    public MqttClient(DeviceConfiguration deviceConfiguration, ScheduledExecutorService ses,
+                      ExecutorService executorService) {
+        this(deviceConfiguration, null, ses, executorService);
 
         HttpProxyOptions httpProxyOptions = ProxyUtils.getHttpProxyOptions(deviceConfiguration);
 
@@ -149,42 +159,36 @@ public class MqttClient implements Closeable {
                             mqttTopics.findOrDefault(DEFAULT_MQTT_KEEP_ALIVE_TIMEOUT, MQTT_KEEP_ALIVE_TIMEOUT_KEY)))
                     .withPingTimeoutMs(
                             Coerce.toInt(mqttTopics.findOrDefault(DEFAULT_MQTT_PING_TIMEOUT, MQTT_PING_TIMEOUT_KEY)))
-                    .withSocketOptions(new SocketOptions()).withTimeoutMs(
-                            Coerce.toInt(mqttTopics.findOrDefault(DEFAULT_MQTT_SOCKET_TIMEOUT,
-                                    MQTT_SOCKET_TIMEOUT_KEY)));
+                    .withSocketOptions(new SocketOptions()).withTimeoutMs(Coerce.toInt(
+                            mqttTopics.findOrDefault(DEFAULT_MQTT_SOCKET_TIMEOUT, MQTT_SOCKET_TIMEOUT_KEY)));
         } else {
             String tesRoleAlias = Coerce.toString(deviceConfiguration.getIotRoleAlias());
 
-            try (TlsContextOptions x509TlsOptions = TlsContextOptions.createWithMtlsFromPath(
-                    Coerce.toString(deviceConfiguration.getCertificateFilePath()),
-                    Coerce.toString(deviceConfiguration.getPrivateKeyFilePath()))) {
+            try (TlsContextOptions x509TlsOptions = TlsContextOptions
+                    .createWithMtlsFromPath(Coerce.toString(deviceConfiguration.getCertificateFilePath()),
+                            Coerce.toString(deviceConfiguration.getPrivateKeyFilePath()))) {
 
                 x509TlsOptions.withCertificateAuthorityFromPath(null,
                         Coerce.toString(deviceConfiguration.getRootCAFilePath()));
 
                 try (ClientTlsContext x509TlsContext = new ClientTlsContext(x509TlsOptions)) {
-                    this.credentialsProvider =
-                            new X509CredentialsProvider.X509CredentialsProviderBuilder()
-                                    .withClientBootstrap(clientBootstrap).withTlsContext(x509TlsContext)
-                                    .withEndpoint(Coerce.toString(deviceConfiguration.getIotCredentialEndpoint()))
-                                    .withRoleAlias(tesRoleAlias)
-                                    .withThingName(Coerce.toString(deviceConfiguration.getThingName()))
-                                    .withProxyOptions(httpProxyOptions).build();
+                    this.credentialsProvider = new X509CredentialsProvider.X509CredentialsProviderBuilder()
+                            .withClientBootstrap(clientBootstrap).withTlsContext(x509TlsContext)
+                            .withEndpoint(Coerce.toString(deviceConfiguration.getIotCredentialEndpoint()))
+                            .withRoleAlias(tesRoleAlias)
+                            .withThingName(Coerce.toString(deviceConfiguration.getThingName()))
+                            .withProxyOptions(httpProxyOptions).build();
 
                     this.builderProvider =
                             (clientBootstrap) -> AwsIotMqttConnectionBuilder.newMtlsBuilderFromPath(null, null)
                                     .withEndpoint(Coerce.toString(deviceConfiguration.getIotDataEndpoint()))
-                                    .withCleanSession(false)
-                                    .withBootstrap(clientBootstrap)
-                                    .withKeepAliveMs(Coerce.toInt(mqttTopics
-                                            .findOrDefault(DEFAULT_MQTT_KEEP_ALIVE_TIMEOUT,
-                                                    MQTT_KEEP_ALIVE_TIMEOUT_KEY)))
-                                    .withPingTimeoutMs(Coerce.toInt(
+                                    .withCleanSession(false).withBootstrap(clientBootstrap).withKeepAliveMs(
+                                            Coerce.toInt(mqttTopics.findOrDefault(DEFAULT_MQTT_KEEP_ALIVE_TIMEOUT,
+                                                    MQTT_KEEP_ALIVE_TIMEOUT_KEY))).withPingTimeoutMs(Coerce.toInt(
                                             mqttTopics.findOrDefault(DEFAULT_MQTT_PING_TIMEOUT, MQTT_PING_TIMEOUT_KEY)))
                                     .withSocketOptions(new SocketOptions()).withTimeoutMs(Coerce.toInt(mqttTopics
-                                    .findOrDefault(DEFAULT_MQTT_SOCKET_TIMEOUT, MQTT_SOCKET_TIMEOUT_KEY)))
-                                    .withWebsockets(true)
-                                    .withWebsocketCredentialsProvider(credentialsProvider)
+                                            .findOrDefault(DEFAULT_MQTT_SOCKET_TIMEOUT, MQTT_SOCKET_TIMEOUT_KEY)))
+                                    .withWebsockets(true).withWebsocketCredentialsProvider(credentialsProvider)
                                     .withWebsocketSigningRegion(Coerce.toString(deviceConfiguration.getAWSRegion()))
                                     .withWebsocketProxyOptions(httpProxyOptions);
                 }
@@ -194,9 +198,9 @@ public class MqttClient implements Closeable {
 
     protected MqttClient(DeviceConfiguration deviceConfiguration,
                          Function<ClientBootstrap, AwsIotMqttConnectionBuilder> builderProvider,
-                         ScheduledExecutorService ses) {
+                         ScheduledExecutorService ses, ExecutorService executorService) {
         this.deviceConfiguration = deviceConfiguration;
-        this.ses = ses;
+        this.executorService = executorService;
 
         mqttTopics = this.deviceConfiguration.getMQTTNamespace();
         this.builderProvider = builderProvider;
@@ -241,8 +245,7 @@ public class MqttClient implements Closeable {
                     return;
                 }
 
-                logger.atInfo().kv("modifiedNode", node.getFullName())
-                        .kv("changeType", what)
+                logger.atInfo().kv("modifiedNode", node.getFullName()).kv("changeType", what)
                         .log("Reconfiguring MQTT clients");
 
                 // Reconnect in separate thread to not block publish thread
@@ -276,16 +279,18 @@ public class MqttClient implements Closeable {
     }
 
     // constructor specific for unit test with spooler
-    protected MqttClient(DeviceConfiguration deviceConfiguration, Spool spool, ScheduledExecutorService ses,
-                         boolean mqttOnline) {
+    protected MqttClient(DeviceConfiguration deviceConfiguration, Spool spool, boolean mqttOnline,
+                         Function<ClientBootstrap, AwsIotMqttConnectionBuilder> builderProvider,
+                         ExecutorService executorService) {
         this.deviceConfiguration = deviceConfiguration;
         mqttTopics = this.deviceConfiguration.getMQTTNamespace();
         eventLoopGroup = new EventLoopGroup(Coerce.toInt(mqttTopics.findOrDefault(1, MQTT_THREAD_POOL_SIZE_KEY)));
         hostResolver = new HostResolver(eventLoopGroup);
         clientBootstrap = new ClientBootstrap(eventLoopGroup, hostResolver);
         this.spool = spool;
-        this.ses = ses;
         this.mqttOnline.set(mqttOnline);
+        this.builderProvider = builderProvider;
+        this.executorService = executorService;
     }
 
     /**
@@ -341,6 +346,21 @@ public class MqttClient implements Closeable {
     private Optional<Map.Entry<MqttTopic, AwsIotMqttClient>> findExistingSubscriberForTopic(String topic) {
         return subscriptionTopics.entrySet().stream().filter(s -> s.getKey().isSupersetOf(new MqttTopic(topic)))
                 .findAny();
+    }
+
+    @SuppressFBWarnings("JLM_JSR166_UTILCONCURRENT_MONITORENTER")
+    private void triggerSpooler() {
+        // Do not synchronize on MqttClient because that causes a dead lock
+        synchronized (spoolingFuture) {
+            if (spoolingFuture.get() == null || spoolingFuture.get().isDone()
+                    && !spoolingFuture.get().isCancelled()) {
+                try {
+                    spoolingFuture.set(executorService.submit(this::runSpooler));
+                } catch (RejectedExecutionException e) {
+                    logger.atWarn().log("Failed to run MQTT spooler", e);
+                }
+            }
+        }
     }
 
     /**
@@ -412,8 +432,8 @@ public class MqttClient implements Closeable {
             return future;
         }
 
-        boolean willDropTheRequest = !mqttOnline.get() && request.getQos().getValue() == 0
-                && !spool.getSpoolConfig().isKeepQos0WhenOffline();
+        boolean willDropTheRequest = !mqttOnline.get() && request.getQos().getValue() == 0 && !spool.getSpoolConfig()
+                .isKeepQos0WhenOffline();
 
         if (willDropTheRequest) {
             SpoolerStoreException e = new SpoolerStoreException("Device is offline. Dropping QoS 0 message.");
@@ -424,7 +444,7 @@ public class MqttClient implements Closeable {
 
         try {
             spool.addMessage(request);
-            spoolMessage();
+            triggerSpooler();
         } catch (InterruptedException | SpoolerStoreException e) {
             logger.atDebug().log("Fail to add publish request to spooler queue", e);
             future.completeExceptionally(e);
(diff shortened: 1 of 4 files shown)


src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -433,39 +444,40 @@ public synchronized void unsubscribe(UnsubscribeRequest request)
         return CompletableFuture.completedFuture(0);
     }
 
-    private synchronized void spoolMessage()  {
-        if (spoolingFuture.get() == null || spoolingFuture.get().isCancelled()) {
-            spoolingFuture.set(ses.scheduleWithFixedDelay(this::spoolTask, 0, 5, TimeUnit.SECONDS));
+    protected CompletableFuture<Integer> spoolSingleMessage() throws InterruptedException {
+        final long id = spool.popId();
```
MikeDombo, line 448: does this call block until an ID is available, or what happens if the spooler is empty?

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -559,8 +572,8 @@ public boolean connected() {
     @Override
     public synchronized void close() {
         // Shut down spooler and then no more message will be published
-        if (spoolingFuture.get() != null) {
-            spoolingFuture.get().cancel(true);
+        if (isSpooling.get()) {
+            executorService.shutdownNow();
```
MikeDombo, line 576: Absolutely do not do this. You don't own the executor service, you may not shut it down.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -117,6 +121,9 @@ public void onConnectionInterrupted(int errorCode) {
         @Override
         public void onConnectionResumed(boolean sessionPresent) {
             setMqttOnline(true);
+            if (isSpooling.compareAndSet(false, true)) {
+                executorService.execute(() -> spoolMessages());
```
MikeDombo, line 125: this needs to become more robust now that it isn't scheduled. If `spoolMessages` throws an unchecked exception then it won't start executing again until this callback is triggered.

MikeDombo: I also don't think that this is sufficient. This callback is only called when the connection is first offline, then online again, so if the device never went offline the spooler would never run.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
MikeDombo: Revert this back to how it was, that was correct.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
awszztt: If the spooler is empty, it will pause here and wait until there is new message inserted into the queue.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
awszztt: The "onConnectionResumed" will be also run when the connection is firstly connected. https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/main/src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java#L175

awszztt: > this needs to become more robust now that it isn't scheduled. If `spoolMessages` throws an unchecked exception then it won't start executing again until this callback is triggered.

This is a good callout, it is necessary to make sure the spooler is up and running as expected.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -105,10 +106,15 @@
     private final AtomicReference<Future<?>> spoolingFuture = new AtomicReference<>();
     private final int maxInFlightPublishes = DEFAULT_MAX_IN_FLIGHT_PUBLISHES;
 
+    @Getter
     private final MqttClientConnectionEvents callbacks = new MqttClientConnectionEvents() {
         @Override
         public void onConnectionInterrupted(int errorCode) {
             setMqttOnline(false);
+            if (spoolingFuture.get() != null && !spoolingFuture.get().isCancelled()) {
+                spoolingFuture.get().cancel(true);
```
jbutler, line 115: Careful about how we handle this case - interrupting the publish loop now means that we need to make sure we don't lose a message. A different approach is to put the publish loop to sleep while the device is offline, rather than starting and stopping the task.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -433,39 +443,40 @@ public synchronized void unsubscribe(UnsubscribeRequest request)
...
+                        logger.atError().log("Failed to publish the message via Spooler", throwable);
+                    }
+                });
     }
 
     /**
      * Iterate the spooler queue to publish all the spooled message.
      */
-    protected void spoolTask() {
+    protected void spoolMessages() {
```
jbutler, line 470: Like we talked about, rename this since it isn't spooling messages, this is taking messages off of the spool. Maybe something like `publishTask`

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -587,5 +607,10 @@ public void addToCallbackEvents(MqttClientConnectionEvents callbacks) {
 
     protected void setMqttOnline(boolean networkStatus) {
         mqttOnline.set(networkStatus);
+        if (networkStatus) {
+            waitMqttOnline.complete(true);
```
MikeDombo, line 611: you can't just set this to true here. The network here could be disabled.

MikeDombo: Sorry, nevermind.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -95,16 +97,19 @@
...
     private final AtomicBoolean mqttOnline = new AtomicBoolean(false);
+    CompletableFuture<Boolean> waitMqttOnline = new CompletableFuture<>();
 
     private final EventLoopGroup eventLoopGroup;
     private final HostResolver hostResolver;
     private final ClientBootstrap clientBootstrap;
     private final CallbackEventManager callbackEventManager = new CallbackEventManager();
     private final Spool spool;
-    private final ScheduledExecutorService ses;
+    protected final ScheduledExecutorService ses;
```
MikeDombo, line 107: you aren't using the ses.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -95,16 +97,19 @@
...
+    private final ExecutorService executorService;
     private final AtomicReference<Future<?>> spoolingFuture = new AtomicReference<>();
     private final int maxInFlightPublishes = DEFAULT_MAX_IN_FLIGHT_PUBLISHES;
 
+    @Getter
```
MikeDombo, line 112: make this protected so that it is only available to tests.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -433,50 +445,58 @@ public synchronized void unsubscribe(UnsubscribeRequest request)
...
+
+    /**
+     * Iterate the spooler queue to publish all the spooled message.
+     */
+    protected void publishAllTasks() {
+        while (!Thread.currentThread().isInterrupted()) {
+            try {
+                getConnection(false).connect().get();
+                List<CompletableFuture<?>> publishRequests = new ArrayList<>();
+                while (waitMqttOnline.get()) {
```
MikeDombo, line 477: instead of waiting on a future, you can just quit the thread and save resources. Then when we're back online this will start executing again.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -433,50 +445,58 @@ public synchronized void unsubscribe(UnsubscribeRequest request)
...
+                                publishRequests.toArray(new CompletableFuture[0])).get();
+                        publishRequests = publishRequests.stream().filter(f ->
+                                !f.isDone()).collect(Collectors.toList());
+                    }
                 }
+            } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+                if (spoolingFuture.get() != null) {
+                    spoolingFuture.get().cancel(true);
+                }
```
MikeDombo, line 495: Don't do this. The only way the thread is interrupted is if the task was cancelled already. This thread should not be managing itself, that's what the outside class is already doing.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -433,50 +445,58 @@ public synchronized void unsubscribe(UnsubscribeRequest request)
...
+                logger.atDebug().log("Shutting down spooler task");
+            } catch (ExecutionException e) {
```
MikeDombo, line 497: What about any other type of exception?

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -433,50 +448,63 @@ public synchronized void unsubscribe(UnsubscribeRequest request)
...
-                    publishRequests = publishRequests.stream().filter(f -> !f.isDone()).collect(Collectors.toList());
+    /**
+     * Iterate the spooler queue to publish all the spooled message.
+     */
+    @SuppressWarnings("PMD.AvoidCatchingThrowable")
+    protected void publishAllTasks() {
+        List<CompletableFuture<?>> publishRequests = new ArrayList<>();
+        while (!Thread.currentThread().isInterrupted()) {
+            try {
+                getConnection(false).connect().get();
```
MikeDombo, line 480: Why are you manually calling connect? You shouldn't be doing this, the publish already handles it.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
awszztt: this is used to reconnect the connection, I am considering whether to change the executor to the executorService as well?

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -433,50 +448,63 @@ public synchronized void unsubscribe(UnsubscribeRequest request)
...
+                        CompletableFuture.anyOf(
+                                publishRequests.toArray(new CompletableFuture[0])).get();
+                        publishRequests = publishRequests.stream().filter(f ->
+                                !f.isDone()).collect(Collectors.toList());
+                    }
+                }
+                break;
+            } catch (ExecutionException e) {
+                logger.atError().log("Error when publishing from spooler", e);
+                if (Objects.equals(Utils.getUltimateCause(e).getClass(), InterruptedException.class)) {
```
MikeDombo, line 498: just use `instanceof`

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -433,50 +448,63 @@ public synchronized void unsubscribe(UnsubscribeRequest request)
...
+                    Thread.currentThread().interrupt();
+                    logger.atDebug().log("Shutting down spooler task");
                 }
+            } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+                logger.atDebug().log("Shutting down spooler task");
+            } catch (Throwable ex) {
+                logger.atError().log("Unchecked Error when publishing from spooler", ex);
```
MikeDombo, line 506: lowercase E

src/test/java/com/aws/greengrass/mqttclient/MqttClientTest.java
```diff
@@ -101,6 +103,7 @@ void beforeEach() {
         lenient().when(mockConnection.unsubscribe(any())).thenReturn(CompletableFuture.completedFuture(0));
         lenient().when(mockConnection.publish(any(), any(), anyBoolean()))
                 .thenReturn(CompletableFuture.completedFuture(0));
+        Thread.interrupted();
```
MikeDombo, line 106: why are you interrupting the main thread?

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
MikeDombo: That's fine to keep then, but why did you change the visibility of it.

src/test/java/com/aws/greengrass/mqttclient/MqttClientTest.java
```diff
@@ -405,74 +409,197 @@ void GIVEN_add_message_to_spooler_throw_interrupted_exception_WHEN_publish_THEN_
...
+
+        verify(client).publishAllTasks();
+        verify(awsIotMqttClient).publish(any(), any(), anyBoolean());
+        verify(spool).getMessageById(anyLong());
+        verify(spool).removeMessageById(anyLong());
+        // The 3rd call is to trigger Interrupted Exception and exit the loop
+        verify(spool, times(2)).popId();
+        verify(client, times(2)).publishSingleTask();
+
+        Thread.interrupted();
```
MikeDombo, line 509: why are you interrupting the main thread everywhere?

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
MikeDombo: Take this out of the loop. There's no reason for it to be inside the loop.

src/test/java/com/aws/greengrass/mqttclient/MqttClientTest.java
awszztt: This is for the testing issue. In order to make it method exit the infinite loop, I introduced the InterruptedException and called Thread.currentThread().interrupt();  when exiting the loop. This is to clear the interrupting flag.  https://github.com/junit-team/junit4/issues/1365

awszztt: But this line should be removed, it should be only applied to a specific test.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
awszztt: the format check would complain if I keep it as private

src/test/java/com/aws/greengrass/mqttclient/MqttClientTest.java
MikeDombo: You're running `publishAllTasks` on the main thread? Why is the thread getting interrupted then, who is doing the interrupting?

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
MikeDombo: What check? It was private before.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -433,50 +448,63 @@ public synchronized void unsubscribe(UnsubscribeRequest request)
...
+                }
+                break;
+            } catch (ExecutionException e) {
+                logger.atError().log("Error when publishing from spooler", e);
+                if (Objects.equals(Utils.getUltimateCause(e).getClass(), InterruptedException.class)) {
+                    Thread.currentThread().interrupt();
+                    logger.atDebug().log("Shutting down spooler task");
                 }
+            } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
```
MikeDombo, line 503: You don't need to interrupt here. You just need to exit the loop which will make the thread stop since it has nothing to do.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
awszztt: In the edition of mainline, we called "getConnection(false).connect().get();" to make sure the OnInitConnect callback would be called and set the mqttClientOnline as True. So that the first run could start. 

The reason I put it in the while loop because the connect.get() may throw ExecutionExecption and the while loop may not start.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
awszztt: Exception for the InterrupedException, it would not stop. IF the queue is empty, it will wait until there is new message.

MikeDombo: There's more than one way to exit a loop.

MikeDombo: Instead of interrupting the thread, just break out of the loop.

MikeDombo (dismissed):

MikeDombo (approved):

fengwa-aws:
How did we test this change? Can we put the latency data comparison in the description?

awszztt:
> How did we test this change? Can we put the latency data comparison in the description?

I am not officially testing it yet. I am testing it by modifying the existing UAT test and read from the logs to make sure it has been sent out immediately. But I can summarize some data based on that.

src/test/java/com/aws/greengrass/mqttclient/MqttClientTest.java
```diff
@@ -82,6 +83,7 @@
     Spool spool;
 
     ScheduledExecutorService ses = new ScheduledThreadPoolExecutor(1);
+    ExecutorService executorService = TestUtils.synchronousExecutorService();
```
MikeDombo, line 86: Make sure we shut these down, we don't want stray threads.

MikeDombo (approved):

MikeDombo (approved):

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -433,50 +452,70 @@ public synchronized void unsubscribe(UnsubscribeRequest request)
...
-        if (spoolingFuture.get() == null || spoolingFuture.get().isCancelled()) {
-            spoolingFuture.set(ses.scheduleWithFixedDelay(this::spoolTask, 0, 5, TimeUnit.SECONDS));
+    @SuppressWarnings("PMD.AvoidCatchingThrowable")
+    protected CompletableFuture<Integer> publishSingleSpoolerMessage() throws InterruptedException {
+        final long id = spool.popId();
+        PublishRequest request = spool.getMessageById(id);
+        if (request == null) {
+            return null;
+        }
+        MqttMessage m = new MqttMessage(request.getTopic(), request.getPayload());
```
jbutler, line 462: It's not difficult to imagine spooler implementations where `getMessageById` is interruptible. Can we move this inside the try block? If we don't, I expect we'll have a bug in the future where messages get dropped.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -433,50 +452,70 @@ public synchronized void unsubscribe(UnsubscribeRequest request)
...
+                }
+                break;
+            } catch (ExecutionException e) {
+                logger.atError().log("Error when publishing from spooler", e);
+                if (Utils.getUltimateCause(e) instanceof InterruptedException) {
+                    logger.atDebug().log("Shutting down spooler task");
+                    break;
                 }
+            } catch (InterruptedException e) {
+                logger.atDebug().log("Shutting down spooler task");
```
fengwa-aws, line 514: I would prefer to log `InterruptedException` at Info or Warn. Customers would want to know if the spooler is shut down. Same for line511.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -433,50 +452,70 @@ public synchronized void unsubscribe(UnsubscribeRequest request)
...
-    private synchronized void spoolMessage()  {
-        if (spoolingFuture.get() == null || spoolingFuture.get().isCancelled()) {
-            spoolingFuture.set(ses.scheduleWithFixedDelay(this::spoolTask, 0, 5, TimeUnit.SECONDS));
+    @SuppressWarnings("PMD.AvoidCatchingThrowable")
+    protected CompletableFuture<Integer> publishSingleSpoolerMessage() throws InterruptedException {
+        final long id = spool.popId();
+        PublishRequest request = spool.getMessageById(id);
+        if (request == null) {
+            return null;
+        }
```
fengwa-aws, line 461: Can this actually happen in any normal condition? When would it happen? Regardless a log message is warranted.

fengwa-aws: Also I would prefer we don't return null value unless null is a legit value in certain use case, which doesn't sound like it here.

MikeDombo: It is backed by a Map, so it returns null if the ID can't be found in the map.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -433,50 +452,70 @@ public synchronized void unsubscribe(UnsubscribeRequest request)
...
+            return getConnection(false).publish(m, request.getQos(), request.isRetain())
+                    .whenComplete((packetId, throwable) -> {
+                        // packetId is the SDK assigned ID. Ignore this and instead use the spooler ID
+                        if (throwable == null) {
+                            spool.removeMessageById(id);
+                            logger.atDebug().kv("id", id).kv("topic", request.getTopic())
+                                    .log("Successfully published message");
+                        } else {
+                            spool.addId(id);
+                            logger.atError().log("Failed to publish the message via Spooler", throwable);
```
fengwa-aws, line 473: Since the message is added back, the log can be:
```
Failed to publish the message via Spooler and will retry
```

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
fengwa-aws: but why would the id not be found in the map if it's popped off from the queue? Does it get removed/updated? Is there any operation not synchronized? Sounds like only a bug can cause the situation, right?

MikeDombo: Seems that's the case, yes. @awszztt?

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -433,50 +452,70 @@ public synchronized void unsubscribe(UnsubscribeRequest request)
...
+        try {
+            return getConnection(false).publish(m, request.getQos(), request.isRetain())
+                    .whenComplete((packetId, throwable) -> {
+                        // packetId is the SDK assigned ID. Ignore this and instead use the spooler ID
+                        if (throwable == null) {
+                            spool.removeMessageById(id);
+                            logger.atDebug().kv("id", id).kv("topic", request.getTopic())
+                                    .log("Successfully published message");
+                        } else {
+                            spool.addId(id);
```
fengwa-aws, line 472: Here and line477. Have you thought about how to handle a "poison pill" message? For example, if the message size is larger than 128KB and rejected by IoT, do we end up retrying infinitely and blocking all the other messages?

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -343,6 +348,20 @@ public synchronized void subscribe(SubscribeRequest request)
                 .findAny();
     }
 
+    @SuppressFBWarnings("JLM_JSR166_UTILCONCURRENT_MONITORENTER")
+    private void triggerSpooler() {
+        // Do not synchronize on MqttClient because that causes a dead lock
+        synchronized (spoolingFuture) {
+            if (spoolingFuture.get() == null || spoolingFuture.get().isDone() || spoolingFuture.get().isCancelled()) {
```
fengwa-aws, line 355: Why check `spoolingFuture.get().isCancelled()` and `spoolingFuture.get().isDone()`? `isCancelled()` and `isDone()` only returns true is `spoolingFuture` is cancelled or done, which can only happen when spooler is shutdown (the runSpooler task thread is interrupted). The two condition checks seem unnecessary. This also introduces race condition between `triggerSpooler()` method and `close()` method.

MikeDombo: If the spooler dies or quits because the network is out (the thread quits when mqtt is offline)

fengwa-aws: Alright. I missed the first break statement. Then we need to think about how to address the race condition.

MikeDombo: @awszztt just remove the `isCancelled` check and that will resolve the race.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
jbutler: When the spool fills up, QoS 0 messages can get dropped from the backing store, but the ID isn't removed from the queue. Really the spool should hide that

awszztt: @jbutler Yes, that is the case.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
awszztt: yes.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
fengwa-aws: Why can't the id be removed as well? This doesn't sound ideal.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
fengwa-aws: If a future is cancelled, isDone() is also true. So I'm not sure if that is sufficient.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
awszztt: The id is stored in LinkedBlockingDeque and id of the message with QoS could be at anywhere in the queue. Wondering it is worth to do the extra operation to remove this id from the queue.

fengwa-aws: Okay. Then why do you even to return null to the caller? Why not just pop and discard those ids until you reach one that exists in the map.  I would also add a debug logline as well.

awszztt: yes, I will make some change to achieve this.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
awszztt: Right now, we always retry until it succeed. For the message length, it would be better to reject it when it is trying to insert the queue if that is too long. But for other non-retryable message, it may block the queue. Do we need to add some retry logic here?

awszztt: I think it is better to create new class called SpoolMessage which has three field, (1) id, (2)publish request, (3) retriedTime. The decision on whether to add the message back could be made in this block of code based on the 'retriedTimes'. In this way, we could prevent other kind of non-retryable exception

Or add it as TODO and make changes in the next release?

awszztt:
The preliminary data is shown as https://quip-amazon.com/9Sv6Aw5OgX35/Spooler-latency-fix

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
fengwa-aws: Ideally we should be able to distinguish whether the exception is retryable or not. If it's not retryable, then the message can be discarded and log an error.

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -172,9 +178,12 @@ public void popOutMessagesWithQosZero() {
...
         while (messageIdIterator.hasNext() && addJudgementWithCurrentSpoolerSize(needToCheckCurSpoolerSize)) {
-            long idToBeRemoved = messageIdIterator.next();
-            if (getMessageById(idToBeRemoved).getQos().getValue() == 0) {
-                removeMessageById(idToBeRemoved);
+            long id = messageIdIterator.next();
+            PublishRequest request = getMessageById(id);
+            if (request.getQos().getValue() == 0) {
+                removeMessageById(id);
+                logger.atDebug().kv("id", id).kv("topic", request.getTopic())
+                        .log("The message which Qos is 0 will be dropped when offline.");
```
MikeDombo, line 186: this message is a bit confusing because it isn't dropping the message in the future, it is dropping the message right now.

"The message has QoS 0 and the spooler is configured to drop QoS 0 when offline. Dropping message now."

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -452,29 +453,38 @@ public synchronized void unsubscribe(UnsubscribeRequest request)
...
         } catch (Throwable t) {
-            spool.addId(id);
+            // valid id is starting from 0
+            if (id >= 0) {
+                spool.addId(id);
+            }
+
+            if (Utils.getUltimateCause(t) instanceof InterruptedException) {
+                throw new InterruptedException("Interrupted while publishing from spooler");
+            }
```
MikeDombo, line 486: this isn't necessary, we already check the cause in the caller of this method by looking at the cause of the execution exception. Did you have any additional reasoning that I'm not getting?

awszztt: In the future, the maxInFlightPublishes may set up bigger than one. Thus, the exception would not be necessary to be thrown out and catch by the 'runSpooler' https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/main/src/main/java/com/aws/greengrass/mqttclient/MqttClient.java#L472

MikeDombo: Even in that case though I believe it would work since we call `.get()` on the combined future which would throw up the exception since that future would be completed already. I suppose this code can't hurt, just think it isn't really necessary.

awszztt: I mean if that interruptedException happen, we can get it immediately. Just a little bitter concern whether it is possible that the message is still publishing, even thought the Mqttclient already requested to be shut down.

MikeDombo (approved):

MikeDombo (dismissed):

fengwa-aws (approved):

fengwa-aws (approved):

MikeDombo (approved):

PR #831, opened 2021-01-26
improvement(s3): don't fallthrough if GetBucketLocation fails

MikeDombo:
**Issue #, if available:**

**Description of changes:**
Currently if GetBucketLocation fails with any exception, we assume that the bucket is in us-east-1 which is a bad assumption. This change will throw the exception up if GetBucketLocation fails with an unknown exception. 

This change also improves tests which were leaving stray threads running.
This change also fixes the stray processes which we were seeing in GitHub which was due to us not being able to kill the `sudo` processes due to a lack of permissions.

**Checklist:**

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -31,7 +31,7 @@ jobs:
       - name: Build with Maven
         env:
           AWS_REGION: us-west-2
-        run: mvn -ntp -U clean verify
+        run: mvn -ntp -U verify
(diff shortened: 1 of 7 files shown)


rbattle:
Since it has nothing to do with the first commit, the commit message for the second commit should also have a conventional commit message like: `fix: process killing with privilege` or something to that effect

rbattle (approved):

nikkhilmuthye (approved):

PR #832, opened 2021-01-26
feat(fleetstatusservice): make periodic update interval for Fleet Status Service configurable

nikkhilmuthye:
**Issue #, if available:**

**Description of changes:**
If the configured value is less than the default (84600 seconds), then it is set to the default.
If the configured value is more than the default, then the periodic update interval for FSS is set to that configured value.

To ensure that the customers have the ability to increase the cadence at which the status of the device is sent periodically.

Added unit tests.

**Checklist:**
 - [X] Updated the README if applicable

### DIFF
--- a/src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
+++ b/src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
@@ -105,6 +105,7 @@ public class DeviceConfiguration {
     public static final String RUN_WITH_DEFAULT_WINDOWS_USER = "windowsUser";
     public static final String RUN_WITH_DEFAULT_POSIX_SHELL = "posixShell";
     public static final String RUN_WITH_DEFAULT_POSIX_SHELL_VALUE = "sh";
+    public static final String FLEET_STATUS_CONFIG_TOPICS = "fleetStatus";
 
     public static final String IOT_ROLE_ALIAS_TOPIC = "iotRoleAlias";
     public static final String COMPONENT_STORE_MAX_SIZE_BYTES = "componentStoreMaxSizeBytes";
@@ -220,6 +221,15 @@ public class DeviceConfiguration {
      */
     public Topics getTelemetryConfigurationTopics() {
         return getTopics(TELEMETRY_CONFIG_LOGGING_TOPICS);
+    }
+
+    /**
+     * Get the fleet status configuration.
+     *
+     * @return Configuration for fleet status service.
+     */
+    public Topics getStatusConfigurationTopics() {
+        return getTopics(FLEET_STATUS_CONFIG_TOPICS);
     }
 
     /**
--- a/src/main/java/com/aws/greengrass/status/FleetStatusService.java
+++ b/src/main/java/com/aws/greengrass/status/FleetStatusService.java
@@ -44,7 +44,6 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 import javax.inject.Inject;
 
-import static com.aws.greengrass.componentmanager.KernelConfigResolver.CONFIGURATION_CONFIG_KEY;
 import static com.aws.greengrass.deployment.DeploymentService.COMPONENTS_TO_GROUPS_TOPICS;
 import static com.aws.greengrass.deployment.DeploymentService.DEPLOYMENT_DETAILED_STATUS_KEY;
 import static com.aws.greengrass.deployment.DeploymentService.DEPLOYMENT_FAILURE_CAUSE_KEY;
@@ -62,8 +61,8 @@ public class FleetStatusService extends GreengrassService {
     public static final String DEFAULT_FLEET_STATUS_SERVICE_PUBLISH_TOPIC =
             "$aws/things/{thingName}/greengrassv2/health/json";
     public static final String FLEET_STATUS_TEST_PERIODIC_UPDATE_INTERVAL_SEC = "fssPeriodicUpdateIntervalSec";
-    public static final int DEFAULT_PERIODIC_UPDATE_INTERVAL_SEC = 86_400;
-    static final String FLEET_STATUS_PERIODIC_UPDATE_INTERVAL_SEC = "periodicUpdateIntervalSec";
+    public static final int DEFAULT_PERIODIC_PUBLISH_INTERVAL_SEC = 86_400;
+    static final String FLEET_STATUS_PERIODIC_PUBLISH_INTERVAL_SEC = "periodicStatusPublishIntervalSeconds";
     static final String FLEET_STATUS_SEQUENCE_NUMBER_TOPIC = "sequenceNumber";
     static final String FLEET_STATUS_LAST_PERIODIC_UPDATE_TIME_TOPIC = "lastPeriodicUpdateTime";
     private static final int MAX_PAYLOAD_LENGTH_BYTES = 128_000;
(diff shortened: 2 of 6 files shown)


src/main/java/com/aws/greengrass/status/FleetStatusService.java
```diff
@@ -165,15 +163,18 @@ public FleetStatusService(Topics topics, MqttClient mqttClient, DeploymentStatus
...
+        deviceConfiguration.getPeriodicStatusPublishIntervalTopic()
+                .dflt(DEFAULT_PERIODIC_PUBLISH_INTERVAL_SEC)
                 .subscribe((why, newv) -> {
                     int newPeriodicUpdateIntervalSec = Coerce.toInt(newv);
                     // Do not update the scheduled interval if it is less than the default.
-                    if (newPeriodicUpdateIntervalSec < DEFAULT_PERIODIC_UPDATE_INTERVAL_SEC) {
+                    if (newPeriodicUpdateIntervalSec < DEFAULT_PERIODIC_PUBLISH_INTERVAL_SEC) {
+                        logger.atWarn().log("Not updating periodic publish interval for Fleet Status Service to "
+                                        + "{} because it is less than the minimum default: {}",
+                                newPeriodicUpdateIntervalSec, DEFAULT_FLEET_STATUS_SERVICE_PUBLISH_TOPIC);
```
hui-yang, line 174: Do we need to reset to DEFAULT_PERIODIC_PUBLISH_INTERVAL_SEC, if the current periodicPublishIntervalSec is longer than DEFAULT_PERIODIC_PUBLISH_INTERVAL_SEC?

src/main/java/com/aws/greengrass/status/FleetStatusService.java
```diff
@@ -145,8 +143,8 @@ public FleetStatusService(Topics topics, MqttClient mqttClient, DeploymentStatus
         this.publisher = new MqttChunkedPayloadPublisher<>(this.mqttClient);
         this.architecture = platformResolver.getCurrentPlatform()
                 .getOrDefault(PlatformResolver.ARCHITECTURE_KEY, PlatformResolver.UNKNOWN_KEYWORD);
-        this.periodicUpdateIntervalSec = TestFeatureParameters.retrieveWithDefault(Double.class,
-                FLEET_STATUS_TEST_PERIODIC_UPDATE_INTERVAL_SEC, periodicUpdateIntervalSec).intValue();
+        this.periodicPublishIntervalSec = TestFeatureParameters.retrieveWithDefault(Double.class,
+                FLEET_STATUS_TEST_PERIODIC_UPDATE_INTERVAL_SEC, periodicPublishIntervalSec).intValue();
```
hui-yang, line 147: If the previous deployment changes the config to a smaller value, it seems the smaller value will be kept in config but not affecting the FSS publish. Is it correct?
In this case, if nucleus restarts, will FSS pick up the smaller value here and actually use it for publish?

src/main/java/com/aws/greengrass/status/FleetStatusService.java
nikkhilmuthye: No. We should keep the same interval as before. This is the same logic we have for telemetry. If the customer wants to set it to default, they can.

hui-yang: Then we need to persist the actual interval somewhere? especially in the restart case

nikkhilmuthye: Discussed offline. This scenario is handled and that we will use the value in the device configuration on start of fresh nucleus or on a restart.

hui-yang (dismissed):

MikeDombo (approved):

src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
```diff
@@ -100,6 +100,7 @@
     public static final String DEVICE_PARAM_AWS_REGION = "awsRegion";
     public static final String DEVICE_MQTT_NAMESPACE = "mqtt";
     public static final String DEVICE_SPOOLER_NAMESPACE = "spooler";
+    public static final String FLEET_STATUS_CONFIG_TOPICS = "status";
```
MikeDombo, line 103: status seems too generic IMO. Maybe just "fleetStatus" or "coreStatus"? Other suggestions would be great.

src/main/java/com/aws/greengrass/status/FleetStatusService.java
```diff
@@ -62,8 +61,8 @@
     public static final String DEFAULT_FLEET_STATUS_SERVICE_PUBLISH_TOPIC =
             "$aws/things/{thingName}/greengrassv2/health/json";
     public static final String FLEET_STATUS_TEST_PERIODIC_UPDATE_INTERVAL_SEC = "fssPeriodicUpdateIntervalSec";
-    public static final int DEFAULT_PERIODIC_UPDATE_INTERVAL_SEC = 86_400;
-    static final String FLEET_STATUS_PERIODIC_UPDATE_INTERVAL_SEC = "periodicUpdateIntervalSec";
+    public static final int DEFAULT_PERIODIC_PUBLISH_INTERVAL_SEC = 86_400;
+    public static final String FLEET_STATUS_PERIODIC_PUBLISH_INTERVAL_SEC = "periodicStatusPublishIntervalSeconds";
```
MikeDombo, line 65: You're changing a config key's name, how will the affect current users? The settings won't apply because it will be looking at the wrong name, right?

src/test/java/com/aws/greengrass/status/FleetStatusServiceTest.java
```diff
@@ -433,7 +434,38 @@ void GIVEN_component_status_change_WHEN_periodic_update_triggered_THEN_MQTT_Sent
...
+
+        // Set up all the mocks
+        when(mockDeploymentStatusKeeper.registerDeploymentStatusConsumer(any(), consumerArgumentCaptor.capture(), anyString())).thenReturn(true);
+        doNothing().when(context).addGlobalStateChangeListener(addGlobalStateChangeListenerArgumentCaptor.capture());
+        when(mockDeviceConfiguration.getStatusConfigurationTopics()).thenReturn(statusConfigTopics);
+        when(context.get(ScheduledExecutorService.class)).thenReturn(ses);
+
+        // Create the fleet status service instance
+        fleetStatusService = createFSS();
+        fleetStatusService.startup();
```
MikeDombo, line 458: Do we shut it down anywhere?

src/main/java/com/aws/greengrass/status/FleetStatusService.java
nikkhilmuthye: Right. We were not looking at this key. So it won't affect existing customers.

MikeDombo: How can you say that? It is read on L168 on the left side.

src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
nikkhilmuthye: I think "fleetStatus" seems reasonable.

src/test/java/com/aws/greengrass/status/FleetStatusServiceTest.java
nikkhilmuthye: We do shut it down in the cleanup function.

src/main/java/com/aws/greengrass/status/FleetStatusService.java
nikkhilmuthye: We never documented this configuration anywhere. Plus it was looking at the config space of FSS component which I believe will not work. I forget the reason but I believe to have configuration set for components in the Nucleus, we added the system configurations.

MikeDombo (approved):

hui-yang (approved):

PR #833, opened 2021-01-27
fix(mqtt): Terminate previous mqtt session when creating a new connection

fahadmohammed01:
**Issue #, if available:**

Greengrass v2 currently tracks the number of subscriptions that each MQTT client made, since IoT Core limits subscriptions to 50 per connection. This has a bit of a conflict when Greengrass restarts because Greengrass no longer knows how many subscriptions the client has; it would assume that the client has 0 subscriptions, even though it connected with a persistent session which may have resubscribe to up to 50 topics.

The behavior of the client which has over-subscribed is that the initial 50 subscriptions work correctly, but subsequent subscriptions, despite not erroring in anyway, simply do not work (no messages come in for them).

**Description of changes:**

When a new mqtt connection is made, the client connects with cleansession = true, then disconnects and connects with cleansession = false. This invalidates the session with IoT Core. This is done the first time a connection is established, for subsequent connects the client uses cleansession = false.

Unit and E2E tests

**Checklist:**
 - [X ] Updated or added new unit tests
 - [ X] Updated or added new end-to-end tests

### DIFF
--- a/src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
+++ b/src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
@@ -47,6 +47,7 @@ class AwsIotMqttClient implements Closeable {
     @Getter
     private final String clientId;
     private MqttClientConnection connection;
+    private CompletableFuture<Boolean> connectionFuture = null;
     private final AtomicBoolean currentlyConnected = new AtomicBoolean();
     private final CallbackEventManager callbackEventManager;
     private final AtomicBoolean initalConnect = new AtomicBoolean(true);
@@ -150,7 +151,13 @@ class AwsIotMqttClient implements Closeable {
     }
 
     protected synchronized CompletableFuture<Boolean> connect() {
-        if (connection != null) {
+        // future not done indicates an ongoing connect attempt, caller should wait on that future
+        // instead of starting another connect attempt.
+        if (connectionFuture != null && !connectionFuture.isDone()) {
+            return connectionFuture;
+        }
+        // We're already connected, there's nothing to do
+        if (currentlyConnected.get()) {
             return CompletableFuture.completedFuture(true);
         }
         // For the initial connect, client connects with cleanSession=true and disconnects.
@@ -164,19 +171,23 @@ class AwsIotMqttClient implements Closeable {
             });
         }
 
-        return voidCompletableFuture.thenCompose((b) -> establishConnection(false)).thenApply((sessionPresent) -> {
-            currentlyConnected.set(true);
-            logger.atInfo().kv("sessionPresent", sessionPresent).log("Successfully connected to AWS IoT Core");
-            if (!sessionPresent) {
-                resubscribe();
-            }
-            callbackEventManager.runOnInitialConnect(sessionPresent);
-            return sessionPresent;
-        });
+        connectionFuture = voidCompletableFuture.thenCompose((b) -> establishConnection(false))
+                .thenApply((sessionPresent) -> {
+                    currentlyConnected.set(true);
+                    logger.atInfo().kv("sessionPresent", sessionPresent)
+                            .log("Successfully connected to AWS IoT Core");
+                    if (!sessionPresent) {
+                        resubscribe();
+                    }
+                    callbackEventManager.runOnInitialConnect(sessionPresent);
+                    return sessionPresent;
+                });
+
+        return connectionFuture;
     }
 
     @SuppressWarnings("PMD.NullAssignment")
-    CompletableFuture<Boolean> establishConnection(boolean overrideCleanSession) {
+    private CompletableFuture<Boolean> establishConnection(boolean overrideCleanSession) {
         // Always use the builder provider here so that the builder is updated with whatever
         // the latest device config is
         try (AwsIotMqttConnectionBuilder builder = builderProvider.get()) {
--- a/src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
+++ b/src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
@@ -49,7 +49,7 @@ class AwsIotMqttClient implements Closeable {
     private MqttClientConnection connection;
     private final AtomicBoolean currentlyConnected = new AtomicBoolean();
     private final CallbackEventManager callbackEventManager;
-    private final AtomicBoolean initalConnect = new  AtomicBoolean(true);
+    private final AtomicBoolean initalConnect = new AtomicBoolean(true);
 
     @Getter(AccessLevel.PACKAGE)
     private final MqttClientConnectionEvents connectionEventCallback = new MqttClientConnectionEvents() {
@@ -144,11 +144,10 @@ class AwsIotMqttClient implements Closeable {
         // Synchronize here instead of method signature to make mockito work without deadlocking
         synchronized (this) {
             logger.atInfo().log("Reconnecting MQTT client most likely due to device configuration change");
-            disconnect();
+            disconnect().get(getTimeout(), TimeUnit.MILLISECONDS);
             connect().get(getTimeout(), TimeUnit.MILLISECONDS);
         }
     }
-
 
     protected synchronized CompletableFuture<Boolean> connect() {
         if (connection != null) {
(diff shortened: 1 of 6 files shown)


src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
```diff
@@ -49,6 +49,7 @@
     private MqttClientConnection connection;
     private final AtomicBoolean currentlyConnected = new AtomicBoolean();
     private final CallbackEventManager callbackEventManager;
+    private final AtomicBoolean initalConnect = new  AtomicBoolean(true);
```
MikeDombo, line 52: spacing seems off here.

src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
```diff
@@ -222,7 +243,7 @@ synchronized boolean connected() {
     }
 
     @SuppressWarnings("PMD.NullAssignment")
-    private synchronized void disconnect() {
+    synchronized void disconnect() {
```
MikeDombo, line 246: make it protected if you're wanting to use it in tests.

src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
```diff
@@ -148,33 +149,53 @@ void reconnect() throws TimeoutException, ExecutionException, InterruptedExcepti
...
         if (connection != null) {
             return CompletableFuture.completedFuture(true);
         }
+        // For the initial connect, client connects with cleanSession=true and disconnects.
+        // This deletes any previous session information maintained by IoT Core.
+        // For subsequent connects, the client connects with cleanSession=false
+        if (initalConnect.get()) {
+            try {
+                establishConnection(true).get(getTimeout(), TimeUnit.MILLISECONDS);
+                disconnect();
```
MikeDombo, line 163: disconnect can sometimes take a very long time, or even timeout. You may need to do some manual testing or talk to the SDK team to figure that one out.

src/test/java/com/aws/greengrass/mqttclient/AwsIotMqttClientTest.java
```diff
@@ -29,11 +29,7 @@
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.spy;
-import static org.mockito.Mockito.times;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.*;
```
MikeDombo, line 32: no * imports.

src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
```diff
@@ -148,33 +149,53 @@ void reconnect() throws TimeoutException, ExecutionException, InterruptedExcepti
...
     protected synchronized CompletableFuture<Boolean> connect() {
         if (connection != null) {
             return CompletableFuture.completedFuture(true);
         }
+        // For the initial connect, client connects with cleanSession=true and disconnects.
+        // This deletes any previous session information maintained by IoT Core.
+        // For subsequent connects, the client connects with cleanSession=false
+        if (initalConnect.get()) {
+            try {
+                establishConnection(true).get(getTimeout(), TimeUnit.MILLISECONDS);
```
MikeDombo, line 162: this is now a blocking call, so you might want to consider making `connect` more asynchronous so that we don't block threads for a long time.

src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
fahadmohammed01: Addressed

src/test/java/com/aws/greengrass/mqttclient/AwsIotMqttClientTest.java
fahadmohammed01: updated likewise

src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
fahadmohammed01: updated likewise

src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
fahadmohammed01: refactored to make connect asynchronous.

src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
```diff
@@ -143,38 +144,54 @@ void reconnect() throws TimeoutException, ExecutionException, InterruptedExcepti
...
-    @SuppressWarnings("PMD.NullAssignment")
     protected synchronized CompletableFuture<Boolean> connect() {
         if (connection != null) {
             return CompletableFuture.completedFuture(true);
         }
+        // For the initial connect, client connects with cleanSession=true and disconnects.
+        // This deletes any previous session information maintained by IoT Core.
+        // For subsequent connects, the client connects with cleanSession=false
+        CompletableFuture<Void> voidCompletableFuture =  CompletableFuture.completedFuture(null);
+        if (initalConnect.get()) {
```
MikeDombo, line 160: While the things inside these callbacks are running you need to ensure that no other thread can come in and call connect() and execute these things at the same time. This is possible to do, but difficult to get right. I implemented it in an earlier closed PR by saving one future and returning it to anyone who called connect() so that all callers were waiting on the same connect() future.

fahadmohammed01: Changed to ensure there that only one thread can request a connect and all subsequent calls to connect will wait for an ongoing connect operation.

src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
```diff
@@ -143,38 +145,64 @@ void reconnect() throws TimeoutException, ExecutionException, InterruptedExcepti
...
+        }
+
+        connectionFuture = voidCompletableFuture.thenCompose((b) -> establishConnection(false))
+                .thenApply((sessionPresent) -> {
+                    currentlyConnected.set(true);
+                    logger.atInfo().kv("sessionPresent", sessionPresent)
+                            .log("Successfully connected to AWS IoT Core");
+                    if (!sessionPresent) {
+                        resubscribe();
+                    }
```
MikeDombo, line 181: we now need to resubscribe no matter what because the session which might be present won't have the subscriptions that we want.

src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
```diff
@@ -222,26 +250,27 @@ synchronized boolean connected() {
...
-                } finally {
+    protected synchronized CompletableFuture<Void> disconnect() {
+        currentlyConnected.set(false);
+        if (connection != null) {
+            logger.atDebug().log("Disconnecting from AWS IoT Core");
+            return connection.disconnect().whenComplete((future, error) -> {
+                logger.atDebug().log("Successfully disconnected from AWS IoT Core");
+                synchronized (this) {
                     connection.close();
                     connection = null;
```
MikeDombo, line 261: might want to reset the connection future here too.

src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
fahadmohammed01: Manually tested by performing connect -> publish/subscribe -> disconnect actions in a loop. Did not see disconnect timing out. Connect operation did fail a couple of times (twice in 100 attempts) but increasing the timeout did not fix that issue. Even when the connect failed, when it was retried for the next publish operation it went through.

src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
fahadmohammed01: I don't think the logic changed to get to the state "if session present won't have the subscriptions that we want".

MikeDombo: Perhaps you're right because of the initialConnection bool

src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
fahadmohammed01: we restrict the connect() operation such that there is only one ongoing connect at any particular time. If connect is ongoing and disconnect is called, resetting the connectionFuture here will allow another connect operation to go through when the previous one is still in progress. Ideally we should not do that unless I am missing something

fahadmohammed01: Also connect and disconnect are protected so we have control on how they are called.

MikeDombo (dismissed):

shaguptashaikh (dismissed):

shaguptashaikh (dismissed):

MikeDombo (approved):

shaguptashaikh (approved):

PR #838, opened 2021-01-28
fix(platform resolver): determine if on macos using os.name instead of uname -a

### LINKED ISSUE #837: (platform resolver): Incorrectly identifies MacOS system if hostname contains "darwin" or "macos"
**Describe the bug**
https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/8a9bd022c7a1da82012eec4a006839d7b7e36f70/src/main/java/com/aws/greengrass/config/PlatformResolver.java#L103

Uses `uname -a` which includes the hostname. If the hostname includes "darwin" the PlatformResolver will incorrectly identify the OS as Mac OSX.

**To Reproduce**
Change hostname to 'darwin' and launch Greengrass

**Expected behavior**
PlatformResolver shouldn't be resolving Linux machines with a hostname containing "darwin" to MacOSX

**Actual behavior**
Tell us what actually happened.

**Environment**
 - OS: Jetson Nano, running Ubuntu.
 - JDK version:
 - Nucleus version:

**Additional context**

E.g. what is the impact of the bug?

MikeDombo:
**Issue #, if available:**
Closes #837 

**Description of changes:**
Instead of using `uname -a` to determine the platform, switch to using `os.name` from the Java system properties.

`uname -a` contains the hostname in the output, so if the hostname was "darwin" or "macos" we were incorrectly identifying the machine as a mac.

Validate that on macos the os.name property is "Mac OS X"

Verified that we detect macos and linux correctly, and even after setting the hostname to "darwin" it still detects correctly.

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/config/PlatformResolver.java
+++ b/src/main/java/com/aws/greengrass/config/PlatformResolver.java
@@ -94,22 +94,22 @@ public class PlatformResolver {
         return platform;
     }
 
+    /**
+     * Get the OS type for this system.
+     *
+     * @return one of our supported identified OSes
+     */
     @SuppressFBWarnings("DMI_HARDCODED_ABSOLUTE_FILENAME")
-    private static String getOSInfo() {
+    public static String getOSInfo() {
         if (isWindows) {
             return OS_WINDOWS;
         }
-        try {
-            String sysver = Exec.sh("uname -a").toLowerCase();
-
-            if (sysver.contains("darwin")) {
-                return OS_DARWIN;
-            }
-            if (Files.exists(Paths.get("/proc"))) {
-                return OS_LINUX;
-            }
-        } catch (IOException | InterruptedException e) {
-            logger.error("Error trying to determine OS", e);
+        String osName = System.getProperty("os.name").toLowerCase();
+        if (osName.contains("mac os")) {
+            return OS_DARWIN;
+        }
+        if (Files.exists(Paths.get("/proc"))) {
+            return OS_LINUX;
         }
         return UNKNOWN_KEYWORD;
     }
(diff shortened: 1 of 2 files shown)


shaguptashaikh (commented):
Don't we need to change `Platform.getInstance()` as well? https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/main/src/main/java/com/aws/greengrass/util/platforms/Platform.java#L48

MikeDombo:
> Don't we need to change `Platform.getInstance()` as well? https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/main/src/main/java/com/aws/greengrass/util/platforms/Platform.java#L48

Yes, just updated that as well.

src/main/java/com/aws/greengrass/config/PlatformResolver.java
```diff
@@ -94,22 +94,22 @@ public PlatformResolver(DeviceConfiguration deviceConfiguration) {
...
-
-            if (sysver.contains("darwin")) {
-                return OS_DARWIN;
-            }
-            if (Files.exists(Paths.get("/proc"))) {
-                return OS_LINUX;
-            }
-        } catch (IOException | InterruptedException e) {
-            logger.error("Error trying to determine OS", e);
+        String osName = System.getProperty("os.name").toLowerCase();
```
fengwa-aws, line 107: Do you know how the `os.name` property is set?

MikeDombo: It is builtin to the JVM.

shaguptashaikh (approved):

fengwa-aws (dismissed):

src/main/java/com/aws/greengrass/config/PlatformResolver.java
MikeDombo: https://github.com/openjdk/jdk/blob/9604ee82690f89320614b37bfef4178abc869777/src/java.base/macosx/native/libjava/java_props_macosx.c#L234-L235

https://github.com/openjdk/jdk/blob/9604ee82690f89320614b37bfef4178abc869777/src/java.base/unix/native/libjava/java_props_md.c#L403

fengwa-aws (approved):

shaguptashaikh (approved):

philcali (approved):

K-AWS (approved):

PR #841, opened 2021-02-03
fix(startup): ensure nucleus version config is always available

shaguptashaikh:
**Issue #, if available:**
If a config file is manually provided at startup with certain nucleus component config but no component version config(or other required keys for nucleus component), inferred version from build file or fallback version if even that's not present(which shouldn't happen in real usage) should be used instead of forcing user to include version in manually created config or finding out too late that it's needed
Also adds a null check in dependency resolution which depends on the version config and and fails without proper message if it's missing

**Description of changes:**

See description

Added unit tests

**Checklist:**

### DIFF
--- a/src/test/java/com/aws/greengrass/deployment/DeviceConfigurationTest.java
+++ b/src/test/java/com/aws/greengrass/deployment/DeviceConfigurationTest.java
@@ -5,6 +5,7 @@
 
 package com.aws.greengrass.deployment;
 
+import com.amazon.aws.iot.greengrass.component.common.ComponentType;
 import com.amazon.aws.iot.greengrass.component.common.RecipeFormatVersion;
 import com.aws.greengrass.componentmanager.ComponentStore;
 import com.aws.greengrass.componentmanager.KernelConfigResolver;
@@ -18,6 +19,7 @@ import com.aws.greengrass.deployment.exceptions.ComponentConfigurationValidation
 import com.aws.greengrass.lifecyclemanager.Kernel;
 import com.aws.greengrass.lifecyclemanager.KernelAlternatives;
 import com.aws.greengrass.testcommons.testutilities.GGExtension;
+import com.aws.greengrass.util.Coerce;
 import com.aws.greengrass.util.NucleusPaths;
 import com.aws.greengrass.util.Utils;
 import com.vdurmont.semver4j.Semver;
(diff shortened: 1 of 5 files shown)


src/test/java/com/aws/greengrass/deployment/DeviceConfigurationTest.java
```diff
@@ -73,9 +82,81 @@ void GIVEN_bad_cred_endpoint_config_WHEN_validate_THEN_fails() {
...
+            assertEquals("0.0.0", deviceConfiguration.getNucleusVersion());
+        } finally {
+            context.close();
+        }
+    }
+
+    @Test
+    void GIVEN_no_existing_config_WHEN_init_device_config_THEN_use_default_nucleus_config() throws Exception {
+        Context context = new Context();
+        try {
```
MikeDombo, line 149: use try with resources here and elsewhere

MikeDombo:
To fix the commitlint, make the commit have a lowercase `Ensure`

MikeDombo (dismissed):

hui-yang (approved):

MikeDombo (commented):
Need to add logging of the deployment errors as well so that we can have proper logs for the next time this happens.

shaguptashaikh:
> Need to add logging of the deployment errors as well so that we can have proper logs for the next time this happens.

I'm not sure which error logs you're referring to, I see that if errors like PackageDownloadException in this case are set as failure cause, the status detail already picks up the message. And DeploymentService also logs any unknown exceptions so I think we're covered. I'll send out a separate PR if something is still missing

MikeDombo:
> > Need to add logging of the deployment errors as well so that we can have proper logs for the next time this happens.
> 
> I'm not sure which error logs you're referring to, I see that if errors like PackageDownloadException in this case are set as failure cause, the status detail already picks up the message. And DeploymentService also logs any unknown exceptions so I think we're covered. I'll send out a separate PR if something is still missing

We don't properly log here: https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/4b87b714b44ab837cedfb024b8207b6402716a16/src/main/java/com/aws/greengrass/deployment/DefaultDeploymentTask.java#L126

If we had logged, we could have identified the problem within seconds.

MikeDombo:
If we merge Hui's change, do we still need this, or does her's take care of it?

shaguptashaikh:
> If we merge Hui's change, do we still need this, or does her's take care of it?

I can rebase on that change but I want to get the tests from this PR in at least

hui-yang (approved):

MikeDombo (approved):

PR #850, opened 2021-02-08
fix(spooler): poison pill protection

awszztt:
**Issue #, if available:**
Currently, the MqttClient.publish would complete the future immediately if the message has been added to the spooler and does not provide the information whether the messages have been really published out. 

**Description of changes:**
1. Check whether the publish request is valid 
2. Set the maximum retried limit to the message published and make this setting configurable

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
+++ b/src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
@@ -11,6 +11,7 @@ import com.aws.greengrass.deployment.DeviceConfiguration;
 import com.aws.greengrass.logging.api.Logger;
 import com.aws.greengrass.logging.impl.LogManager;
 import com.aws.greengrass.mqttclient.spool.Spool;
+import com.aws.greengrass.mqttclient.spool.SpoolMessage;
 import com.aws.greengrass.mqttclient.spool.SpoolerStoreException;
 import com.aws.greengrass.util.Coerce;
 import com.aws.greengrass.util.LockScope;
@@ -32,6 +33,7 @@ import software.amazon.awssdk.crt.mqtt.MqttMessage;
 import software.amazon.awssdk.iot.AwsIotMqttConnectionBuilder;
 
 import java.io.Closeable;
+import java.nio.charset.StandardCharsets;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.List;
@@ -56,6 +58,7 @@ import java.util.concurrent.locks.ReadWriteLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.function.Consumer;
 import java.util.function.Function;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import javax.inject.Inject;
 
@@ -68,6 +71,7 @@ import static com.aws.greengrass.deployment.DeviceConfiguration.DEVICE_PARAM_ROO
 import static com.aws.greengrass.deployment.DeviceConfiguration.DEVICE_PARAM_THING_NAME;
 import static com.aws.greengrass.mqttclient.AwsIotMqttClient.TOPIC_KEY;
 
+@SuppressWarnings({"PMD.AvoidDuplicateLiterals"})
 public class MqttClient implements Closeable {
     private static final Logger logger = LogManager.getLogger(MqttClient.class);
     private static final String MQTT_KEEP_ALIVE_TIMEOUT_KEY = "keepAliveTimeoutMs";
@@ -83,10 +87,20 @@ public class MqttClient implements Closeable {
     static final String MQTT_OPERATION_TIMEOUT_KEY = "operationTimeoutMs";
     static final int DEFAULT_MQTT_OPERATION_TIMEOUT = (int) Duration.ofSeconds(30).toMillis();
     static final String MQTT_MAX_IN_FLIGHT_PUBLISHES_KEY = "maxInFlightPublishes";
-    static final int DEFAULT_MAX_IN_FLIGHT_PUBLISHES = 1;
+    static final int DEFAULT_MAX_IN_FLIGHT_PUBLISHES = 5;
     public static final int MAX_SUBSCRIPTIONS_PER_CONNECTION = 50;
     public static final String CLIENT_ID_KEY = "clientId";
     public static final int EVENTLOOP_SHUTDOWN_TIMEOUT_SECONDS = 2;
+    static final String MQTT_MAX_OF_MESSAGE_SIZE_IN_BYTES_KEY = "maxMessageSizeInBytes";
+    static final String MQTT_MAX_OF_PUBLISH_RETRY_COUNT_KEY = "maxPublishRetry";
+    static final int DEFAULT_MQTT_MAX_OF_PUBLISH_RETRY_COUNT = 100;
+    // http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718023
+    static final int MQTT_MAX_LIMIT_OF_MESSAGE_SIZE_IN_BYTES = 256 * 1024 * 1024; // 256 MB
+    // https://docs.aws.amazon.com/general/latest/gr/iot-core.html#limits_iot
+    static final int DEFAULT_MQTT_MAX_OF_MESSAGE_SIZE_IN_BYTES = 128 * 1024; // 128 kB
+    static final int MAX_NUMBER_OF_FORWARD_SLASHES = 7;
+    static final int MAX_LENGTH_OF_TOPIC = 256;
+    static final int IOT_MAX_LIMIT_IN_FLIGHT_OF_QOS1_PUBLISHES = 100;
 
     // Use read lock for MQTT operations and write lock when changing the MQTT connection
     private final ReadWriteLock connectionLock = new ReentrantReadWriteLock(true);
@@ -109,7 +123,9 @@ public class MqttClient implements Closeable {
     private final Spool spool;
     private final ExecutorService executorService;
     private final AtomicReference<Future<?>> spoolingFuture = new AtomicReference<>();
-    private final int maxInFlightPublishes = DEFAULT_MAX_IN_FLIGHT_PUBLISHES;
+    private int maxInFlightPublishes;
+    private int maxPublishRetryCount;
+    private int maxPublishMessageSize;
 
     @Getter(AccessLevel.PROTECTED)
     private final MqttClientConnectionEvents callbacks = new MqttClientConnectionEvents() {
@@ -204,6 +220,7 @@ public class MqttClient implements Closeable {
 
         mqttTopics = this.deviceConfiguration.getMQTTNamespace();
         this.builderProvider = builderProvider;
+        validateAndSetMqttPublishConfiguration();
 
         eventLoopGroup = new EventLoopGroup(Coerce.toInt(mqttTopics.findOrDefault(1, MQTT_THREAD_POOL_SIZE_KEY)));
         hostResolver = new HostResolver(eventLoopGroup);
@@ -237,6 +254,10 @@ public class MqttClient implements Closeable {
                         .childOf(DEVICE_PARAM_CERTIFICATE_FILE_PATH) || node.childOf(DEVICE_PARAM_ROOT_CA_PATH) || node
                         .childOf(DEVICE_PARAM_AWS_REGION))) {
                     return;
+                }
+
+                if (node.childOf(DEVICE_MQTT_NAMESPACE)) {
+                    validateAndSetMqttPublishConfiguration();
                 }
 
                 // Only reconnect when the region changed if the proxy exists
@@ -293,6 +314,39 @@ public class MqttClient implements Closeable {
         this.mqttOnline.set(mqttOnline);
         this.builderProvider = builderProvider;
         this.executorService = executorService;
+        validateAndSetMqttPublishConfiguration();
+    }
+
+    private void validateAndSetMqttPublishConfiguration() {
+        maxInFlightPublishes = Coerce.toInt(mqttTopics
+                .findOrDefault(DEFAULT_MAX_IN_FLIGHT_PUBLISHES,
+                        MQTT_MAX_IN_FLIGHT_PUBLISHES_KEY));
+        if (maxInFlightPublishes > IOT_MAX_LIMIT_IN_FLIGHT_OF_QOS1_PUBLISHES) {
+            logger.atWarn()
+                    .kv(MQTT_MAX_IN_FLIGHT_PUBLISHES_KEY, maxInFlightPublishes)
+                    .kv("Max acceptable configuration", IOT_MAX_LIMIT_IN_FLIGHT_OF_QOS1_PUBLISHES)
+                    .log(String.format("The configuration of %s may hit the AWS IoT Core restricting number of "
+                                    + "unacknowledged QoS=1 publish requests per client. "
+                                    + "Will change to the maximum allowed setting: %d",
+                            MQTT_MAX_IN_FLIGHT_PUBLISHES_KEY, IOT_MAX_LIMIT_IN_FLIGHT_OF_QOS1_PUBLISHES));
+
+            maxInFlightPublishes = IOT_MAX_LIMIT_IN_FLIGHT_OF_QOS1_PUBLISHES;
+        }
+
+        maxPublishMessageSize = Coerce.toInt(mqttTopics.findOrDefault(DEFAULT_MQTT_MAX_OF_MESSAGE_SIZE_IN_BYTES,
+                MQTT_MAX_OF_MESSAGE_SIZE_IN_BYTES_KEY));
+        if (maxPublishMessageSize > MQTT_MAX_LIMIT_OF_MESSAGE_SIZE_IN_BYTES) {
+            logger.atWarn()
+                    .kv(MQTT_MAX_OF_MESSAGE_SIZE_IN_BYTES_KEY, maxPublishMessageSize)
+                    .kv("Max acceptable configuration", MQTT_MAX_LIMIT_OF_MESSAGE_SIZE_IN_BYTES)
+                    .log(String.format("The configuration of %s exceeds the max limit and will change to "
+                                    + "the maximum allowed setting: %d bytes", MQTT_MAX_OF_MESSAGE_SIZE_IN_BYTES_KEY,
+                            MQTT_MAX_LIMIT_OF_MESSAGE_SIZE_IN_BYTES));
+            maxPublishMessageSize = MQTT_MAX_LIMIT_OF_MESSAGE_SIZE_IN_BYTES;
+        }
+
+        maxPublishRetryCount =  Coerce.toInt(mqttTopics.findOrDefault(DEFAULT_MQTT_MAX_OF_PUBLISH_RETRY_COUNT,
+                MQTT_MAX_OF_PUBLISH_RETRY_COUNT_KEY));
     }
 
     /**
@@ -434,6 +488,14 @@ public class MqttClient implements Closeable {
             return future;
         }
 
+        try {
+            isValidPublishRequest(request);
+        } catch (SpoolerStoreException e) {
+            logger.atError().kv("topic", request.getTopic()).log("Invalid publish request: {}", e.getMessage());
+            future.completeExceptionally(e);
+            return future;
+        }
+
         boolean willDropTheRequest = !mqttOnline.get() && request.getQos().getValue() == 0 && !spool.getSpoolConfig()
                 .isKeepQos0WhenOffline();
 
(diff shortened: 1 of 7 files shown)


src/integrationtests/java/com/aws/greengrass/integrationtests/e2e/mqttclient/MqttTest.java
```diff
@@ -59,12 +61,16 @@ void GIVEN_mqttclient_WHEN_subscribe_and_publish_THEN_receives_all_messages()
...
+            assertEquals(id, ++lastMessageId);
         }).build());
 
         for (int i = 0; i < NUM_MESSAGES; i++) {
-            client.publish(PublishRequest.builder().topic("A/B/C").payload("What's up".getBytes(StandardCharsets.UTF_8))
+            String id = String.valueOf(i);
+            Integer res = client.publish(PublishRequest.builder().topic("A/B/C").payload(id.getBytes(StandardCharsets.UTF_8))
                     .build()).get(5, TimeUnit.SECONDS);
             logger.atInfo().kv("total", i + 1).log("Added 1 message to spooler.");
+            assertEquals(0, res);
```
MikeDombo, line 73: I wouldn't assert on this as we do not control its value.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -83,10 +85,15 @@
...
     static final String MQTT_MAX_IN_FLIGHT_PUBLISHES_KEY = "maxInFlightPublishes";
-    static final int DEFAULT_MAX_IN_FLIGHT_PUBLISHES = 1;
+    static final int DEFAULT_MAX_IN_FLIGHT_PUBLISHES = 5;
     public static final int MAX_SUBSCRIPTIONS_PER_CONNECTION = 50;
     public static final String CLIENT_ID_KEY = "clientId";
     public static final int EVENTLOOP_SHUTDOWN_TIMEOUT_SECONDS = 2;
+    static final int DEFAULT_PUBLISH_RETRIED_COUNTER = 100;
+    // http://www.steves-internet-guide.com/mqtt-broker-message-restrictions/#:~:text=The%20MQTT%20protocol%20allows%20messages,especially%20public%20brokers%2C%20will%20restrict.
+    static final int MAXIMUM_SIZE_OF_MQTT_MESSAGE_SIZE = 268_435_455;
+    // https://docs.aws.amazon.com/general/latest/gr/greengrass.html
```
MikeDombo, line 95: why link to greengrass? This is a IoT Core restriction right?

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -83,10 +85,15 @@
     static final String MQTT_OPERATION_TIMEOUT_KEY = "operationTimeoutMs";
     static final int DEFAULT_MQTT_OPERATION_TIMEOUT = (int) Duration.ofSeconds(30).toMillis();
     static final String MQTT_MAX_IN_FLIGHT_PUBLISHES_KEY = "maxInFlightPublishes";
-    static final int DEFAULT_MAX_IN_FLIGHT_PUBLISHES = 1;
+    static final int DEFAULT_MAX_IN_FLIGHT_PUBLISHES = 5;
     public static final int MAX_SUBSCRIPTIONS_PER_CONNECTION = 50;
     public static final String CLIENT_ID_KEY = "clientId";
     public static final int EVENTLOOP_SHUTDOWN_TIMEOUT_SECONDS = 2;
+    static final int DEFAULT_PUBLISH_RETRIED_COUNTER = 100;
+    // http://www.steves-internet-guide.com/mqtt-broker-message-restrictions/#:~:text=The%20MQTT%20protocol%20allows%20messages,especially%20public%20brokers%2C%20will%20restrict.
```
MikeDombo, line 93: Link to the actual specification. http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718023

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -239,6 +249,11 @@ protected MqttClient(DeviceConfiguration deviceConfiguration,
                     return;
                 }
 
+                if (node.childOf(DEVICE_MQTT_NAMESPACE)) {
+                    maxInFlightPublishes = Coerce.toInt(mqttTopics.lookup(MQTT_MAX_IN_FLIGHT_PUBLISHES_KEY));
+                    logger.atInfo().kv("value", maxInFlightPublishes).log("updating maxInFlightPublishes");
```
MikeDombo, line 254: this will unnecessarily log because you aren't checking that anything changed. I would not bother logging this.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -239,6 +249,11 @@ protected MqttClient(DeviceConfiguration deviceConfiguration,
                     return;
                 }
 
+                if (node.childOf(DEVICE_MQTT_NAMESPACE)) {
+                    maxInFlightPublishes = Coerce.toInt(mqttTopics.lookup(MQTT_MAX_IN_FLIGHT_PUBLISHES_KEY));
```
MikeDombo, line 253: Use findOrDefault again, this might be null.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -432,6 +447,15 @@ public synchronized void unsubscribe(UnsubscribeRequest request)
             return future;
         }
 
+        try {
+            isValidPublishRequest(request);
+        } catch (SpoolerStoreException e) {
+            logger.atDebug().kv("topic", request.getTopic())
```
MikeDombo, line 453: provide a clear log message such as `invalid publish request: {}` and then interpolate the error message there.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -442,37 +466,72 @@ public synchronized void unsubscribe(UnsubscribeRequest request)
...
+        }
+
+        // Topic should not contain wildcard characters
+        String topic = request.getTopic();
+        if (topic.contains("#") || topic.contains("+")) {
+            throw new SpoolerStoreException("The topic of publish request should not contain wildcard characters.");
+        }
+
+        // Topic should not have no more than maximum number of forward slashes (/)
+        if (topic.chars().filter(num -> num == '/').count() > MAXIMUM_NUMBER_OF_FORWARD_SLASHES) {
```
MikeDombo, line 496:
```
A topic in a publish or subscribe request can have no more than 7 forward slashes (/). This excludes the first 3 slashes in the mandatory segments for Basic Ingest topics ($AWS/rules/rule-name/).
```

We probably want to set our maximum to 10 in that case, or check for the reserved topic.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -442,37 +466,72 @@ public synchronized void unsubscribe(UnsubscribeRequest request)
...
+            throw new SpoolerStoreException(String.format("The topic of publish request should not contain "
```
MikeDombo, line 497: "The topic of publish requests must not have more than %d forward slashes."

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -442,37 +466,72 @@ public synchronized void unsubscribe(UnsubscribeRequest request)
...
+            SpoolMessage spooledMessage = spool.getMessageById(id);
+            PublishRequest request = spooledMessage.getRequest();
             MqttMessage m = new MqttMessage(request.getTopic(), request.getPayload());
 
             long finalId = id;
             return getConnection(false).publish(m, request.getQos(), request.isRetain())
                     .whenComplete((packetId, throwable) -> {
                         // packetId is the SDK assigned ID. Ignore this and instead use the spooler ID
                         if (throwable == null) {
+                            spooledMessage.getFuture().complete(0);
```
MikeDombo, line 517: complete it with the correct value, why not. `.complete(packetId)`

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -442,37 +466,72 @@ public synchronized void unsubscribe(UnsubscribeRequest request)
...
                             spool.removeMessageById(finalId);
                             logger.atDebug().kv("id", finalId).kv("topic", request.getTopic())
                                     .log("Successfully published message");
                         } else {
-                            spool.addId(finalId);
-                            logger.atError().log("Failed to publish the message via Spooler and will retry",
-                                    throwable);
+                            if (!mqttOnline.get() || spooledMessage.getRetried().get()
```
MikeDombo, line 522: the online check seems to not be useful since IoT Core drops us for most of the bad packet types. And in cases where sending a message doesn't cause us to drop, then we don't want to hold that against the message.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
fengwa-aws: Probably should log this as an error.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -442,37 +470,90 @@ public synchronized void unsubscribe(UnsubscribeRequest request)
...
+        int messageSize = request.getPayload().length;
+        if (messageSize > MAXIMUM_SIZE_OF_MQTT_MESSAGE_SIZE) {
+            throw new SpoolerStoreException(String.format("The message size exceeds the maximum size of %d ",
+                    MAXIMUM_SIZE_OF_MQTT_MESSAGE_SIZE));
+        }
+
+        // Topic should not contain wildcard characters
+        String topic = request.getTopic();
+        if (topic.contains("#") || topic.contains("+")) {
+            throw new SpoolerStoreException("The topic of publish request should not contain wildcard characters.");
```
fengwa-aws, line 496: What wildcard character? The exception message should include the wrong input.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -442,37 +470,90 @@ public synchronized void unsubscribe(UnsubscribeRequest request)
...
+
+        if (topic.getBytes(StandardCharsets.UTF_8).length > DEFAULT_MAXIMUM_LENGTH_OF_TOPIC) {
+            logger.atInfo().kv("topic", topic).log("new topic - for size");
+            String errMsg = String.format("The topic size of publish request must be no "
+                            + "larger than {} bytes of UTF-8 encoded characters. This excludes the first "
+                            + "3 mandatory segments for Basic Ingest topics ($AWS/rules/rule-name/).",
+                    DEFAULT_MAXIMUM_LENGTH_OF_TOPIC);
+            throw new SpoolerStoreException(errMsg);
+        }
+        return true;
```
fengwa-aws, line 522: This value is never used. You can update the method to return "void".

src/integrationtests/java/com/aws/greengrass/integrationtests/e2e/mqttclient/MqttTest.java
```diff
@@ -59,10 +61,13 @@ void GIVEN_mqttclient_WHEN_subscribe_and_publish_THEN_receives_all_messages()
         client.subscribe(SubscribeRequest.builder().topic("A/B/C").callback((m) -> {
             cdl.countDown();
             logger.atInfo().kv("remaining", cdl.getCount()).log("Received 1 message from cloud.");
+            Integer id = Integer.parseInt(new String(m.getPayload(), StandardCharsets.UTF_8));
+            assertEquals(id, ++lastMessageId);
```
MikeDombo, line 65: lastMessageId should be an AtomicInteger, otherwise this method can have multi threading problems.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -442,37 +470,92 @@ public synchronized void unsubscribe(UnsubscribeRequest request)
...
+
+        // Topic should not contain wildcard characters
+        String topic = request.getTopic();
+        if (topic.contains("#") || topic.contains("+")) {
+            throw new SpoolerStoreException("The topic of publish request should not contain wildcard "
+                    + "characters of '#' or '+'");
+        }
+
+        String reservedTopicTemplate = "^\\$AWS/rules/\\S+/\\S+";
+        String prefixOfReservedTopic = "^\\$AWS/rules/\\S+?/";
```
MikeDombo, line 501: I'm pretty sure that `$aws` is case insensitive

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -442,37 +470,92 @@ public synchronized void unsubscribe(UnsubscribeRequest request)
...
+        if (Pattern.matches(reservedTopicTemplate, topic)) {
+            // remove the prefix of "$AWS/rules/rule-name/"
+            topic = topic.split(prefixOfReservedTopic, 2)[1];
+        }
+
+        // Topic should not have no more than maximum number of forward slashes (/)
+        if (topic.chars().filter(num -> num == '/').count() > DEFAULT_MAXIMUM_NUMBER_OF_FORWARD_SLASHES) {
+            logger.atInfo().kv("topic", topic).log("new topic - for /");
```
MikeDombo, line 509: this log seems really really weird. What are you trying to tell the user with this?

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -442,37 +470,92 @@ public synchronized void unsubscribe(UnsubscribeRequest request)
...
+            String errMsg = String.format("The topic of publish request must have no"
+                    + "more than %d forward slashes (/). This excludes the first 3 slashes in the mandatory segments "
+                    + "for Basic Ingest topics ($AWS/rules/rule-name/).", DEFAULT_MAXIMUM_NUMBER_OF_FORWARD_SLASHES);
+            throw new SpoolerStoreException(errMsg);
+        }
+
+        // Check the topic size
+        if (topic.getBytes(StandardCharsets.UTF_8).length > DEFAULT_MAXIMUM_LENGTH_OF_TOPIC) {
+            logger.atInfo().kv("topic", topic).log("new topic - for size");
```
MikeDombo, line 518: same here

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -442,37 +470,92 @@ public synchronized void unsubscribe(UnsubscribeRequest request)
...
                         if (throwable == null) {
+                            spooledMessage.getFuture().complete(packetId);
                             spool.removeMessageById(finalId);
                             logger.atDebug().kv("id", finalId).kv("topic", request.getTopic())
                                     .log("Successfully published message");
                         } else {
-                            spool.addId(finalId);
-                            logger.atError().log("Failed to publish the message via Spooler and will retry",
-                                    throwable);
+                            if (spooledMessage.getRetried().get()
```
MikeDombo, line 546: Just use `getAndIncrement()` right here so that it is automatically incremented and you don't need to do it on L551

fengwa-aws (commented):
Given that we change the behavior of publish, we should verify whether we need to change the client code that calls the publish method.

src/integrationtests/java/com/aws/greengrass/integrationtests/e2e/mqttclient/MqttTest.java
awszztt: I am planing the second message is sent out after the first QOS 1 message has been successfully published and should not have any race here. If it exist, that would be the problem of the future from publish method.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
awszztt: Yes, I will update this.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
awszztt: Sorry, it is for testing. I will remove it here.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
awszztt: Same one as here.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -239,6 +252,11 @@ protected MqttClient(DeviceConfiguration deviceConfiguration,
                     return;
                 }
 
+                if (node.childOf(DEVICE_MQTT_NAMESPACE)) {
+                    maxInFlightPublishes = Coerce.toInt(mqttTopics.findOrDefault(DEFAULT_MAX_IN_FLIGHT_PUBLISHES,
+                                    MQTT_MAX_IN_FLIGHT_PUBLISHES_KEY));
```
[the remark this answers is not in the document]
awszztt: Have update the validation for this configuration.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -442,37 +485,90 @@ public synchronized void unsubscribe(UnsubscribeRequest request)
...
     }
 
     @SuppressWarnings({"PMD.AvoidCatchingThrowable", "PMD.PreserveStackTrace"})
     protected CompletableFuture<Integer> publishSingleSpoolerMessage() throws InterruptedException {
         long id = -1L;
         try {
             id = spool.popId();
-            PublishRequest request = spool.getMessageById(id);
+            SpoolMessage spooledMessage = spool.getMessageById(id);
+            PublishRequest request = spooledMessage.getRequest();
```
MikeDombo, line 546: Let's check that the future isn't done yet at this point. This way our client can cancel the future and then we won't send it (if it hasn't been sent already of course).

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -83,10 +87,17 @@
...
     static final int DEFAULT_MQTT_OPERATION_TIMEOUT = (int) Duration.ofSeconds(30).toMillis();
     static final String MQTT_MAX_IN_FLIGHT_PUBLISHES_KEY = "maxInFlightPublishes";
-    static final int DEFAULT_MAX_IN_FLIGHT_PUBLISHES = 1;
+    static final int DEFAULT_MAX_IN_FLIGHT_PUBLISHES = 5;
     public static final int MAX_SUBSCRIPTIONS_PER_CONNECTION = 50;
     public static final String CLIENT_ID_KEY = "clientId";
     public static final int EVENTLOOP_SHUTDOWN_TIMEOUT_SECONDS = 2;
+    static final int DEFAULT_PUBLISH_RETRIED_COUNTER = 100;
+    // http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718023
+    static final int MAXIMUM_SIZE_OF_MQTT_MESSAGE_SIZE = 268_435_455;
```
jbutler, line 96: https://docs.aws.amazon.com/general/latest/gr/iot-core.html#message-broker-limits

IoT max message size is 128kb

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -442,37 +485,89 @@ public synchronized void unsubscribe(UnsubscribeRequest request)
...
+                            spooledMessage.getFuture().complete(packetId);
                             spool.removeMessageById(finalId);
                             logger.atDebug().kv("id", finalId).kv("topic", request.getTopic())
                                     .log("Successfully published message");
                         } else {
-                            spool.addId(finalId);
-                            logger.atError().log("Failed to publish the message via Spooler and will retry",
-                                    throwable);
+                            if (spooledMessage.getRetried().getAndIncrement()
+                                    < DEFAULT_PUBLISH_RETRIED_COUNTER) {
```
jbutler, line 560: I'm not sure how best to handle this. I see the desire to have a way to recover if our message validation isn't good enough, but this also opens up the door to drop QoS 1 messages.

IMO we cannot drop QoS 1 messages. The customer is ultimately in control of the data that they send, and if they have a poison pill, they're doing it to themselves. I don't think we should drop data here, personally.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
awszztt: I think the max message size limit is about to increase. @MikeDombo Or it is better to set this to the current limit 128 kB for now?

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
awszztt: If the Spooler queue is full, we would reject the insertion and leave the Error the message for the customer. Just wondering may be it is fine to leave some Error message and leave the space for the customer to handle this.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
MikeDombo: I'm reticent to put in that limit given that IoT Core may change it. Need to talk offline.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
fengwa-aws: The challenge is, there is only one queue in the spooler and all the components share the queue. If the queue is blocked, it impacts the whole Greengrass core.

PR #851, opened 2021-02-08
feat(AwsIotMqttClient): resubscribe robust retry

tilo-chen:
**Issue #, if available:**

**Description of changes:**
`AwsIotMqttClient`: Implement MQTT continuous subscription retry upon connection recovery.

Before: when MQTT connection resumed, retry all subscribed topics once only if session persisted.
After: when MQTT connection resumed, retry failed subscriptions. Keep retrying until succeeded or connection interrupted again, with backoff delay.

- Add unit tests for resubscribe logic.
- Add unit tests for `WrapperMqttClientConnection` and `MqttClient` to increase test coverage.
- Fix flaky `TelemetryAgentTest`:
    - [Link to failed run](https://github.com/aws-greengrass/aws-greengrass-nucleus/runs/1891190361). Null pointer exception because `TestFeatureParameters.internalEnableTestingFeatureParameters` was called before mock was fully setup. Fixed by reordering.
    - [Link to failed run](https://github.com/aws-greengrass/aws-greengrass-nucleus/runs/1906728743). `RejectedExecutionException` because `internalDisableTestingFeatureParameters` was called after executor was shutdown. Because TA [registers the callback async-ly](https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/3b2350aec4dcacca2e88c06401e63b0078dd051b/src/main/java/com/aws/greengrass/telemetry/TelemetryAgent.java#L293), the callback could be registered after TA itself is shutdown if the test is really short. Fixed by moving `internalDisableTestingFeatureParameters` from `AfterEach` to individual test cases that changed the default handler. Also removed unnecessary code from one test case.
- [Link to failed run](https://github.com/aws-greengrass/aws-greengrass-nucleus/runs/1891370493). Fix flaky `PeriodicFleetStatusServiceTest`: was not waiting until all services started. FSS status could miss some services. Fixed by waiting for `main service` to finish.

Making MQTT network recovery more robust.
Fix flaky tests.

- Unit test
- Smoke UATs
- UAT `Deployment-3-T6`: deploy to a group with intermittent connectivity

**Checklist:**

### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/status/PeriodicFleetStatusServiceTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/status/PeriodicFleetStatusServiceTest.java
@@ -12,6 +12,7 @@ import com.aws.greengrass.integrationtests.BaseITCase;
 import com.aws.greengrass.integrationtests.util.ConfigPlatformResolver;
 import com.aws.greengrass.lifecyclemanager.GreengrassService;
 import com.aws.greengrass.lifecyclemanager.Kernel;
+import com.aws.greengrass.lifecyclemanager.KernelCommandLine;
 import com.aws.greengrass.mqttclient.MqttClient;
 import com.aws.greengrass.mqttclient.PublishRequest;
 import com.aws.greengrass.status.ComponentStatusDetails;
@@ -35,6 +36,7 @@ import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
 
@@ -61,6 +63,7 @@ class PeriodicFleetStatusServiceTest extends BaseITCase {
         ignoreExceptionOfType(context, TLSAuthException.class);
         CountDownLatch fssRunning = new CountDownLatch(1);
         CountDownLatch deploymentServiceRunning = new CountDownLatch(1);
+        AtomicBoolean mainServiceFinished = new AtomicBoolean();
         allComponentsInFssUpdate = new CountDownLatch(1);
         fleetStatusDetails = new AtomicReference<>();
         CompletableFuture cf = new CompletableFuture();
@@ -76,7 +79,8 @@ class PeriodicFleetStatusServiceTest extends BaseITCase {
             try {
                 fleetStatusDetails.set(OBJECT_MAPPER.readValue(publishRequest.getPayload(),
                         FleetStatusDetails.class));
-                if (kernel.orderedDependencies().size() == fleetStatusDetails.get().getComponentStatusDetails().size()) {
+                if (mainServiceFinished.get() && kernel.orderedDependencies().size() == fleetStatusDetails.get()
+                        .getComponentStatusDetails().size()) {
                     allComponentsInFssUpdate.countDown();
                 }
             } catch (JsonMappingException ignored) { }
@@ -89,12 +93,15 @@ class PeriodicFleetStatusServiceTest extends BaseITCase {
                     fssRunning.countDown();
                 }
                 FleetStatusService fleetStatusService = (FleetStatusService) service;
-                fleetStatusService.setPeriodicUpdateIntervalSec(10);
+                fleetStatusService.setPeriodicUpdateIntervalSec(5);
                 fleetStatusService.schedulePeriodicFleetStatusDataUpdate(false);
             }
             if (service.getName().equals(DeploymentService.DEPLOYMENT_SERVICE_TOPICS)
                     && newState.equals(State.RUNNING)) {
                 deploymentServiceRunning.countDown();
+            }
+            if (service.getName().equals(KernelCommandLine.MAIN_SERVICE_NAME) && newState.equals(State.FINISHED)) {
+                mainServiceFinished.set(true);
             }
         });
         // set required instances from context
--- a/src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
+++ b/src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
@@ -291,7 +291,7 @@ class AwsIotMqttClient implements Closeable {
                             .log("Subscription interrupted. Cancelling subscriptions");
                     allSubFutures.cancel(true);
                 } catch (ExecutionException e) {
-                    // Do nothing. Errors already handled in the above whenComplete stage
+                    // Do nothing. Errors already handled in individual subscription future's whenComplete stage
                 }
             }, delayMillis, TimeUnit.MILLISECONDS);
 
--- a/src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
+++ b/src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
@@ -259,7 +259,6 @@ class AwsIotMqttClient implements Closeable {
                 droppedSubscriptionTopics.putAll(subscriptionTopics);
             }
             if (!droppedSubscriptionTopics.isEmpty() && (resubscribeFuture == null || resubscribeFuture.isDone())) {
-                logger.info("ignoring");
                 resubscribeFuture = executorService.submit(this::resubscribeDroppedTopicsTask);
             }
         }
--- a/src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
+++ b/src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
@@ -259,6 +259,7 @@ class AwsIotMqttClient implements Closeable {
                 droppedSubscriptionTopics.putAll(subscriptionTopics);
             }
             if (!droppedSubscriptionTopics.isEmpty() && (resubscribeFuture == null || resubscribeFuture.isDone())) {
+                logger.info("ignoring");
                 resubscribeFuture = executorService.submit(this::resubscribeDroppedTopicsTask);
             }
         }
--- a/src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
+++ b/src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
@@ -12,6 +12,7 @@ import com.aws.greengrass.util.Coerce;
 import com.aws.greengrass.util.ProxyUtils;
 import lombok.AccessLevel;
 import lombok.Getter;
+import lombok.Setter;
 import software.amazon.awssdk.crt.CRT;
 import software.amazon.awssdk.crt.mqtt.MqttClientConnection;
 import software.amazon.awssdk.crt.mqtt.MqttClientConnectionEvents;
@@ -21,10 +22,15 @@ import software.amazon.awssdk.crt.mqtt.QualityOfService;
 import software.amazon.awssdk.iot.AwsIotMqttConnectionBuilder;
 
 import java.io.Closeable;
+import java.time.Duration;
+import java.util.Iterator;
 import java.util.Map;
+import java.util.Random;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
@@ -40,9 +46,11 @@ import javax.inject.Provider;
 class AwsIotMqttClient implements Closeable {
     static final String TOPIC_KEY = "topic";
     private static final String QOS_KEY = "qos";
+    private static final Random RANDOM = new Random();
     private final Logger logger = LogManager.getLogger(AwsIotMqttClient.class).createChild()
             .dfltKv(MqttClient.CLIENT_ID_KEY, (Supplier<String>) this::getClientId);
 
+    private final ExecutorService executorService;
     private final Provider<AwsIotMqttConnectionBuilder> builderProvider;
     @Getter
     private final String clientId;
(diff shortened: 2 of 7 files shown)


src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
```diff
@@ -199,14 +210,54 @@ int getTimeout() {
                 MqttClient.DEFAULT_MQTT_OPERATION_TIMEOUT, MqttClient.MQTT_OPERATION_TIMEOUT_KEY));
     }
 
-    private void resubscribe() {
-        subscriptionTopics.forEach((key, value) -> {
-            subscribe(key, value).whenComplete((i, t) -> {
-                if (t != null) {
-                    logger.atError().kv(TOPIC_KEY, key).kv(QOS_KEY, value.name()).log("Unable to resubscribe to topic");
+    private void resubscribe(boolean sessionPresent) {
```
MikeDombo, line 213: synchronize this to ensure that we don't get multiple competing futures.

src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
```diff
@@ -199,14 +210,54 @@ int getTimeout() {
...
+                droppedSubscriptionTopics.putAll(subscriptionTopics);
+            }
+            if (resubscribeFuture == null || resubscribeFuture.isDone()) {
+                resubscribeFuture = executorService.submit(this::resubscribeDroppedTopicsTask);
+            }
+        }
+    }
+
+    private void resubscribeDroppedTopicsTask() {
+        logger.atInfo().log("Re-subscribing to topics");
```
MikeDombo, line 227: debug

src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
```diff
@@ -199,14 +210,54 @@ int getTimeout() {
...
+        while (true) {
+            logger.atInfo().kv("failed topics", droppedSubscriptionTopics.keySet())
```
MikeDombo, line 229: debug.

Don't print the subscription topics, only the dropped ones.

src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
```diff
@@ -199,14 +210,54 @@ int getTimeout() {
...
+            }
+            if (resubscribeFuture == null || resubscribeFuture.isDone()) {
+                resubscribeFuture = executorService.submit(this::resubscribeDroppedTopicsTask);
+            }
+        }
+    }
+
+    private void resubscribeDroppedTopicsTask() {
+        logger.atInfo().log("Re-subscribing to topics");
+        while (true) {
```
MikeDombo, line 228: why not `while(!droppedSubscriptionTopics.isEmpty())`?

src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
```diff
@@ -199,14 +210,54 @@ int getTimeout() {
...
+                                .log("Failed to re-subscribe to topic, will retry later");
+                    }
+                });
+            }
+            // If there's still failed subscriptions, wait for some time before trying again
+            if (droppedSubscriptionTopics.isEmpty()) {
+                break;
+            } else {
+                try {
+                    Thread.sleep(waitTimeToSubscribeAgainMillis + RANDOM.nextInt(waitTimeJitterRangeMillis));
```
MikeDombo, line 254: instead of sleeping for 2 minutes, we might want to instead use a scheduled executor and schedule another run in 2 minutes.

src/test/java/com/aws/greengrass/mqttclient/AwsIotMqttClientTest.java
```diff
@@ -56,6 +62,7 @@
 
     CallbackEventManager callbackEventManager;
     Topics mockTopic;
+    ExecutorService executorService = Executors.newCachedThreadPool();
```
MikeDombo, line 65: make sure you shut this down in the `afterEach` using `shutdownNow()`

src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
```diff
@@ -199,14 +210,54 @@ int getTimeout() {
...
+                        logger.atError().kv(TOPIC_KEY, subscriptionEntry.getKey())
+                                .kv(QOS_KEY, subscriptionEntry.getValue().name())
+                                .log("Failed to re-subscribe to topic, will retry later");
+                    }
+                });
+            }
+            // If there's still failed subscriptions, wait for some time before trying again
+            if (droppedSubscriptionTopics.isEmpty()) {
+                break;
+            } else {
```
MikeDombo, line 252: This isn't correct since you didn't wait for the futures to complete. The subscription requests will all be sent out, but it is unlikely that you'd have the responses yet.

MikeDombo: Additionally, since you don't wait for the future to complete, you're going to loop around and then it will send out requests again, even though the first requests could still be outstanding and not actually failed.

src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
```diff
@@ -40,15 +46,27 @@
...
     private final AtomicBoolean currentlyConnected = new AtomicBoolean();
     private final CallbackEventManager callbackEventManager;
+    private final Consumer<MqttMessage> messageHandler;
+    private final Topics mqttTopics;
+    @Getter(AccessLevel.PACKAGE)
+    private final Map<String, QualityOfService> subscriptionTopics = new ConcurrentHashMap<>();
+    private final Map<String, QualityOfService> droppedSubscriptionTopics = new ConcurrentHashMap<>();
+    private Future<?> resubscribeFuture;
+    @Setter
+    private static long waitTimeToSubscribeAgainMillis = Duration.ofMinutes(2).toMillis();
```
fengwa-aws, line 67: NIT: make this a constant. Also why 2 minutes? Probably want to add a comment to explain that.

src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
```diff
@@ -40,15 +46,27 @@
...
+    @Setter
+    private static int waitTimeJitterRangeMillis = 10_000;
```
fengwa-aws, line 69: Same as above

src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
```diff
@@ -199,14 +210,54 @@ int getTimeout() {
...
+
+                Map.Entry<String, QualityOfService> subscriptionEntry = iterator.next();
+                subscribe(subscriptionEntry.getKey(), subscriptionEntry.getValue()).whenComplete((i, t) -> {
+                    if (t == null) {
+                        iterator.remove();
+                    } else {
+                        logger.atError().kv(TOPIC_KEY, subscriptionEntry.getKey())
+                                .kv(QOS_KEY, subscriptionEntry.getValue().name())
+                                .log("Failed to re-subscribe to topic, will retry later");
+                    }
```
fengwa-aws, line 246: How can I debug it as a customer? Include the exception message in the error.

src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
tilo-chen: Needed to set this to a tiny value for testing purposes so didn't make it `final`.

Open to discussion what's the appropriate delay. I'm using 2 minutes because I just copied `IotJobHelper`'s subscription retry delay.

src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
```diff
@@ -227,14 +242,58 @@ int getTimeout() {
...
+    private void resubscribeDroppedTopicsTask() {
+        long delayMillis = 0;  // don't delay the first run
+        while (!droppedSubscriptionTopics.isEmpty()) {
+            logger.atDebug().kv("droppedTopics", droppedSubscriptionTopics.keySet()).kv("delayMillis", delayMillis)
+                    .log("Subscribing to dropped topics");
+            ScheduledFuture<?> scheduledFuture = ses.schedule(() -> {
+                for (Map.Entry<String, QualityOfService> entry : droppedSubscriptionTopics.entrySet()) {
+                    CompletableFuture<Integer> subFuture = subscribe(entry.getKey(), entry.getValue());
+                    try {
+                        subFuture.get();
```
MikeDombo, line 273: We can send out all the subscription requests at once. Then use `future.whenComplete` to handle the outcomes from the future. You can use CompleteableFuture.allOf to get you a future that represents all of the other futures completing.

src/test/java/com/aws/greengrass/mqttclient/AwsIotMqttClientTest.java
```diff
@@ -40,6 +53,7 @@
 @ExtendWith({GGExtension.class, MockitoExtension.class})
 class AwsIotMqttClientTest {
 
+    public static final int VERIFY_TIMEOUT_MILLIS = 1000;
```
MikeDombo, line 56: this shouldn't need to be public in a test.

src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
tilo-chen: Hmm I actually tried that approach and had 2 concerns:
- If some of the future fails,`allOf` will fail, then we seem to have no way to know which ones succeeded and remove those in the next run.
- We probably had to handle interruption [like this](https://github.com/aws-greengrass/aws-greengrass-nucleus/pull/851/commits/165470c714ced3b327efceec1ca08a181baa6f9b#diff-bfd141b30d873984288154d5640bb36b2fec76b936defe54e86977c49312e77eR285-R292), making this a little too complex?

MikeDombo: You don't need to know which one failed in the allOf, instead you can do something like

```
for topics:
  fut = subscribe(...);
  list.add(fut.whenComplete((i, t) -> {
    // Handle completion logic to remove topic from retry list if it worked.
  }));
allOf(list).get(); // errors can be ignored here since they're handled in the whenComplete, except for InterruptedException
```

src/integrationtests/java/com/aws/greengrass/integrationtests/status/PeriodicFleetStatusServiceTest.java
```diff
@@ -89,13 +93,16 @@ void setupKernel(ExtensionContext context) throws Exception {
...
-                fleetStatusService.setPeriodicUpdateIntervalSec(10);
+                fleetStatusService.setPeriodicUpdateIntervalSec(5);
                 fleetStatusService.schedulePeriodicFleetStatusDataUpdate(false);
             }
             if (service.getName().equals(DeploymentService.DEPLOYMENT_SERVICE_TOPICS)
                     && newState.equals(State.RUNNING)) {
                 deploymentServiceRunning.countDown();
             }
+            if (service.getName().equals(KernelCommandLine.MAIN_SERVICE_NAME) && newState.equals(State.FINISHED)) {
+                mainServiceFinished.set(true);
```
MikeDombo, line 104: this boolean doesn't seem to be read anywhere?

src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
```diff
@@ -227,14 +245,68 @@ int getTimeout() {
...
+     */
+    private synchronized void resubscribe(boolean sessionPresent) {
+        // Don't bother spin up a thread if subscription is empty
+        if (!subscriptionTopics.isEmpty()) {
+            // If connected without a session, all subscriptions are dropped and need to be resubscribed
+            if (!sessionPresent) {
+                droppedSubscriptionTopics.putAll(subscriptionTopics);
+            }
+            if (!droppedSubscriptionTopics.isEmpty() && (resubscribeFuture == null || resubscribeFuture.isDone())) {
+                logger.info("ignoring");
```
MikeDombo, line 262: remove?

src/test/java/com/aws/greengrass/telemetry/TelemetryAgentTest.java
```diff
@@ -129,9 +129,10 @@ void cleanUp() throws IOException, InterruptedException {
         TelemetryConfig.getInstance().closeContext();
         telemetryAgent.shutdown();
         ses.shutdownNow();
+        executorService.shutdownNow();
         context.close();
-        ses.awaitTermination(2, TimeUnit.SECONDS);
-        TestFeatureParameters.internalDisableTestingFeatureParameters();
```
MikeDombo, line 134: keep this so that even if a test fails we don't have the test feature thing messing up future tests.

MikeDombo (dismissed):

src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
```diff
@@ -227,14 +245,67 @@ int getTimeout() {
...
+            }
+        }
+    }
+
+    private void resubscribeDroppedTopicsTask() {
+        long delayMillis = 0;  // don't delay the first run
+        while (!droppedSubscriptionTopics.isEmpty()) {
+            logger.atDebug().event(RESUB_LOG_EVENT).kv("droppedTopics", droppedSubscriptionTopics.keySet())
+                    .kv("delayMillis", delayMillis).log("Subscribing to dropped topics");
+            ScheduledFuture<?> scheduledFuture = ses.schedule(() -> {
```
fengwa-aws, line 272: What's the reason of using `ses`? Would we end up scheduling multiple resubscribe tasks?

tilo-chen: We won't schedule multiple resub tasks because later we call `future.get` to wait for it. So this is actually equivalent to `while { ... Thread.sleep }`. I thought waiting with `Future.get` instead of `Thread.sleep` has better performance (I haven't done experiment though). Downside is that the code is much complicated.

fengwa-aws (dismissed):

MikeDombo (approved):

fengwa-aws (approved):

PR #855, opened 2021-02-11
fix(startup): don't crash nucleus on service load exceptions

hui-yang:
**Issue #, if available:**

**Description of changes:**
Handle service load exception on Nucleus launch
1. Refactor `public GreengrassService locate(String name)` to `locate` which surface exceptions and `locateIgnoreError` which proceeds exceptions using surrogates.
    1. `locateIgnoreError` are used during Nucleus launch, loading new services in deployments, and service setting up dependencies.
1. Deployments are required to fix unloadable services. 
    1. Fixes to plugin services will trigger Nucleus restart. 
    1. Non-plugin services will not restart Nucleus, but old instance will be closed and new one created with new config. When closing unloadable service, its depending services will not block the shutdown.

Minor improvements
1. Track group to root components information if deployment is configured to not rollback on failure
2. Fix error details for Nucleus restart workflow, which should be the persisted error messages during Nucleus activate stage. If missing, there must be errors launching Nucleus in activate stage. 

To prevent such error cases bricking devices.

End-to-end test scenarios:
1. Nucleus can launch successfully under different ServiceLoadExceptions.
2. Unloadable service can be fixed with deployments.

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
+++ b/src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
@@ -18,6 +18,7 @@ import com.aws.greengrass.deployment.model.DeploymentDocument;
 import com.aws.greengrass.deployment.model.DeploymentResult;
 import com.aws.greengrass.lifecyclemanager.GreengrassService;
 import com.aws.greengrass.lifecyclemanager.Kernel;
+import com.aws.greengrass.lifecyclemanager.UnloadableService;
 import com.aws.greengrass.lifecyclemanager.UpdateAction;
 import com.aws.greengrass.lifecyclemanager.UpdateSystemPolicyService;
 import com.aws.greengrass.lifecyclemanager.exceptions.ServiceLoadException;
@@ -37,6 +38,8 @@ import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.stream.Collectors;
 import javax.inject.Inject;
 
@@ -238,6 +241,7 @@ public class DeploymentConfigMerger {
         private Set<String> servicesToUpdate;
         private Set<String> servicesToRemove;
         private Set<String> alreadyBrokenServices;
+        private Set<String> alreadyUnloadableServices;
 
         /**
          * Constructs an object based on the current Kernel state and the config to be merged.
@@ -271,6 +275,13 @@ public class DeploymentConfigMerger {
                     return false;
                 }
             }).collect(Collectors.toSet());
+            this.alreadyUnloadableServices = runningDeployableServices.stream().filter(name -> {
+                try {
+                    return kernel.locate(name) instanceof UnloadableService;
+                } catch (ServiceLoadException e) {
+                    return false;
+                }
+            }).collect(Collectors.toSet());
         }
 
         /**
@@ -282,7 +293,7 @@ public class DeploymentConfigMerger {
             // For rollback, services the deployment originally intended to add should be removed
             // and services it intended to remove should be added back
             return new AggregateServicesChangeManager(kernel, servicesToRemove, servicesToUpdate, servicesToAdd,
-                    alreadyBrokenServices);
+                    alreadyBrokenServices, alreadyUnloadableServices);
         }
 
         /**
@@ -290,9 +301,9 @@ public class DeploymentConfigMerger {
          *
          * @throws ServiceLoadException when any service to be started could not be located
          */
-        public void startNewServices() throws ServiceLoadException {
+        public void startNewServices() {
             for (String serviceName : servicesToAdd) {
-                GreengrassService service = kernel.locate(serviceName);
+                GreengrassService service = kernel.locateIgnoreError(serviceName);
                 if (service.shouldAutoStart()) {
                     service.requestStart();
                 }
@@ -309,6 +320,24 @@ public class DeploymentConfigMerger {
                 GreengrassService service = kernel.locate(serviceName);
                 if (service.currentOrReportedStateIs(State.BROKEN)) {
                     service.requestReinstall();
+                }
+            }
+        }
+
+        /**
+         * Close all unloadable service instances, initialize each service with updated config and add to context.
+         *
+         * @throws ServiceLoadException when any service to be replaced could not be located
+         */
+        public void replaceUnloadableService() throws ServiceLoadException {
+            for (String serviceName : alreadyUnloadableServices) {
+                try {
+                    kernel.locate(serviceName).close().get(30, TimeUnit.SECONDS);
+                    kernel.getContext().remove(serviceName);
+                    kernel.locateIgnoreError(serviceName).requestReinstall();
+                } catch (InterruptedException | ExecutionException | TimeoutException e) {
+                    logger.atError().kv("serviceName", serviceName)
+                            .log("Failed to close unloadable service", e);
                 }
             }
         }
--- a/src/main/java/com/aws/greengrass/deployment/DeploymentService.java
+++ b/src/main/java/com/aws/greengrass/deployment/DeploymentService.java
@@ -75,6 +75,7 @@ import static com.aws.greengrass.deployment.DeploymentConfigMerger.DEPLOYMENT_ID
 import static com.aws.greengrass.deployment.converter.DeploymentDocumentConverter.LOCAL_DEPLOYMENT_GROUP_NAME;
 import static com.aws.greengrass.deployment.model.Deployment.DeploymentStage.DEFAULT;
 import static com.aws.greengrass.deployment.model.Deployment.DeploymentType;
+import static com.aws.greengrass.deployment.model.DeploymentResult.DeploymentStatus.FAILED_ROLLBACK_NOT_REQUESTED;
 
 @ImplementsService(name = DeploymentService.DEPLOYMENT_SERVICE_TOPICS, autostart = true)
 public class DeploymentService extends GreengrassService {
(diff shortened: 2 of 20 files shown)


src/main/java/com/aws/greengrass/lifecyclemanager/Kernel.java
```diff
@@ -354,117 +356,140 @@ public Topics findServiceTopic(String serviceName) {
...
+     */
+    public GreengrassService locateIgnoreError(String name) {
+        return context.getValue(GreengrassService.class, name).computeObjectIfEmpty(v -> {
+            try {
+                return locateGreengrassService(v, name, this::locateIgnoreError);
+            } catch (ServiceLoadException e) {
+                logger.atError().log("Cannot load service", e);
+                return new BrokenService(config.lookupTopics(DEFAULT_VALUE_TIMESTAMP, SERVICES_NAMESPACE_TOPIC, name),
+                        e);
+            }
```
philcali, line 378: How does this `BrokenService` manifest in the tlog and in other places like the debug console?

hui-yang: BrokenService is still a GreengrassService. Tlog should be handled by the underlying config Topics, where each BrokenService has its own entries identified by serviceName.
Same for debug console, I'm expecting customers to see their service with serviceName in BROKEN state. This requires some testing for sure.

philcali: Makes sense. :+1:

src/main/java/com/aws/greengrass/lifecyclemanager/BrokenService.java
```diff
@@ -0,0 +1,38 @@
...
+        this.serviceLoadException = e;
+    }
+
+    @Override
+    protected void install() {
+        serviceErrored(serviceLoadException);
+    }
+
+    @Override
+    public boolean isBootstrapRequired(Map<String, Object> newServiceConfig) {
```
MikeDombo, line 29: why do we want to bootstrap? How does that help us recover? Won't it just be unnecessarily restarting?

src/main/java/com/aws/greengrass/lifecyclemanager/Kernel.java
```diff
@@ -354,117 +356,140 @@ public Topics findServiceTopic(String serviceName) {
...
+     * @return found service
+     */
+    public GreengrassService locateIgnoreError(String name) {
+        return context.getValue(GreengrassService.class, name).computeObjectIfEmpty(v -> {
+            try {
+                return locateGreengrassService(v, name, this::locateIgnoreError);
+            } catch (ServiceLoadException e) {
+                logger.atError().log("Cannot load service", e);
+                return new BrokenService(config.lookupTopics(DEFAULT_VALUE_TIMESTAMP, SERVICES_NAMESPACE_TOPIC, name),
+                        e);
```
MikeDombo, line 377: Do we only want to do this when the service actually exists, does this method check for that?

What I mean is that if we lookup a service which doesn't exist, then I still want the exception and not a "BrokenService".

shaguptashaikh (commented):
1. How will customers know that there's a broken service? 
2. How does a BrokenService instance get fixed? Looks like it'll only get fixed if customers make a deployment that causes nucleus restart? Is that ideal?

src/main/java/com/aws/greengrass/lifecyclemanager/BrokenService.java
hui-yang: The intention here is to restart if a deployment tries to fix the broken service, which fails to load previously but has a placeholder instance.

src/main/java/com/aws/greengrass/lifecyclemanager/Kernel.java
hui-yang: By `service which doesn't exist`, do you mean the config node is missing in config files? In this case, BrokenService is also created. Otherwise it's still a failure case which can brick device if nucleus restarts.

MikeDombo: Right, when the node doesn't exist, that doesn't mean that it is broken, it means that it doesn't exist. Those are different cases.

hui-yang:
>     1. How will customers know that there's a broken service?
The service should show as in broken state in FSS and deployment status reports.
> 
>     2. How does a BrokenService instance get fixed? Looks like it'll only get fixed if customers make a fix deployment that causes nucleus restart? Is that ideal?

Yes, it requires a deployment with nucleus restart. Here BrokenService is only created if a service cannot be loaded, e.g. missing config, bad artifacts, and I intend to replace this instance with the actual service after a nucleus restart deployment.

src/main/java/com/aws/greengrass/lifecyclemanager/Kernel.java
hui-yang: Is it better to rename BrokenService to UnloadableService?

src/main/java/com/aws/greengrass/lifecyclemanager/UnloadableService.java
```diff
@@ -0,0 +1,53 @@
...
+        Topic versionTopic = getConfig().find(VERSION_CONFIG_KEY);
+        if (versionTopic == null) {
+            logger.atInfo().log("Bootstrap is required: current service version unknown");
+            return true;
+        }
+        if (!versionTopic.getOnce().equals(newServiceConfig.get(VERSION_CONFIG_KEY))) {
+            logger.atInfo().log("Bootstrap is required: service version changed");
+            return true;
+        }
+        if (!getConfig().toPOJO().equals(newServiceConfig)) {
```
shaguptashaikh, line 41: Will the service config topics always be equal to the newServiceConfig generated for deployment when customer hasn't changed anything about the unloadable service? Do we create any child config for the service at load time or runtime that might cause differences?

src/main/java/com/aws/greengrass/lifecyclemanager/UnloadableService.java
```diff
@@ -0,0 +1,53 @@
...
+        this.serviceLoadException = e;
+    }
+
+    @Override
+    protected void install() {
+        serviceErrored(serviceLoadException);
+    }
+
+    @Override
+    public boolean isBootstrapRequired(Map<String, Object> newServiceConfig) {
```
shaguptashaikh, line 31: I'm still a little concerned about requiring nucleus restart for fixing unloadbale services, technically load doesn't require restart, this fix is just supposed to avoid breaking the nucleus completely if a service, generic or internal, happens to fail to load during restart. But for subsequent fixes to them, the deployment may not otherwise warrant a restart. Is it possible to avoid it, maybe by handling this during the service start/restart/remove handling and monitoring in deployment activator?

hui-yang: Good point. Let me try that out

src/main/java/com/aws/greengrass/deployment/DeploymentService.java
```diff
@@ -323,9 +307,11 @@ private void finishCurrentDeployment() throws InterruptedException {
                     if (result.getFailureCause() != null) {
                         statusDetails.put(DEPLOYMENT_FAILURE_CAUSE_KEY, result.getFailureCause().getMessage());
                     }
-                    // TODO: [P41179126] Update the groupToRootPackages mapping in config for the case where there
```
shaguptashaikh, line 326: Is this TODO resolved now?

shaguptashaikh: nvm

src/main/java/com/aws/greengrass/lifecyclemanager/UnloadableService.java
```diff
@@ -0,0 +1,103 @@
...
+        if (Utils.stringHasChanged(Coerce.toString(getConfig().find(VERSION_CONFIG_KEY)), newVersion)) {
+            logger.atInfo().log("Bootstrap is required: plugin version changed");
+            return true;
+        }
+        try {
+            Path pluginJar = config.getContext().get(NucleusPaths.class).artifactPath(new ComponentIdentifier(getName(),
+                    new Semver(newVersion))).resolve(getName() + JAR_FILE_EXTENSION);
+
+            if (!pluginJar.toFile().exists() || !pluginJar.toFile().isFile()) {
+                logger.atInfo().log("Bootstrap is not required: plugin JAR not found " + pluginJar);
```
shaguptashaikh, line 77: nit - avoid concat with a `+`

src/main/java/com/aws/greengrass/lifecyclemanager/UnloadableService.java
```diff
@@ -0,0 +1,103 @@
...
+                return false;
+            }
+
+            if (Files.getLastModifiedTime(pluginJar).toMillis() > exceptionTimestamp) {
```
shaguptashaikh, line 81: Do we need to check if the jar is modified or is this covered by version check? If there's a different version of the plugin component, downloader would download the jar in the new version's artifact location even if the jar wasn't modified between versions. It will be very rare that the jar already existed on the device because the version had previously been in use and didn't get wiped out during component store cleanup.

shaguptashaikh (approved):

src/main/java/com/aws/greengrass/lifecyclemanager/UnloadableService.java
hui-yang: New version of plugin should be downloaded to a version-specific artifact location, and should be covered by version check.
This is for cases where there's no change in plugin version, but jar was corrupted previously somehow and to be fixed now in deployments.
The service config input here is the fully resolved config, which can be picked up from existing components as well as from the deployment document. This condition is to check if the deployment document contains fix to the broken plugin.

src/main/java/com/aws/greengrass/deployment/DeploymentService.java
hui-yang: Yes. I marked it as resolved.

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
```diff
@@ -313,6 +324,24 @@ public void reinstallBrokenServices() throws ServiceLoadException {
...
 
+        /**
+         * Close all unloadable service instances, initialize each service with updated config and add to context.
+         *
+         * @throws ServiceLoadException when any service to be replaced could not be located
+         */
+        public void replaceUnloadableService() throws ServiceLoadException {
+            for (String serviceName : alreadyUnloadableServices) {
+                try {
+                    kernel.locate(serviceName).close().get(30, TimeUnit.SECONDS);
```
fahadmohammed01, line 335: When calling close() here, it waits for depernder services to exit. How is the depender services notified that it has to exit?

shaguptashaikh (approved):

fahadmohammed01 (approved):

PR #865, opened 2021-02-18
feat(docker): initial implementation for docker image artifact do…

shaguptashaikh:
…wnloader

**Issue #, if available:**

**Description of changes:**
Implement an artifact downloader which performs docker registry login if applicable and downloads images during the deployment download step. 

*Immediate follow up items outside this PR*
- Add more robustness to intermittent network, close out TODOs etc

*Further fast follow changes planned in this area* 
- This is intended to be a plugin, the code will be moved out of nucleus as soon as time permits, and checks around the plugin dependency being required have been added to make it possible to move this code out as fast follow.
-  Further support for registries

New tests

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/componentmanager/ComponentManager.java
+++ b/src/main/java/com/aws/greengrass/componentmanager/ComponentManager.java
@@ -7,8 +7,11 @@ package com.aws.greengrass.componentmanager;
 
 import com.amazon.aws.iot.greengrass.component.common.ComponentType;
 import com.amazon.aws.iot.greengrass.component.common.Unarchive;
+import com.aws.greengrass.componentmanager.builtins.ArtifactDownloader;
+import com.aws.greengrass.componentmanager.builtins.ArtifactDownloaderFactory;
 import com.aws.greengrass.componentmanager.converter.RecipeLoader;
 import com.aws.greengrass.componentmanager.exceptions.InvalidArtifactUriException;
+import com.aws.greengrass.componentmanager.exceptions.MissingRequiredComponentsException;
 import com.aws.greengrass.componentmanager.exceptions.NoAvailableComponentVersionException;
 import com.aws.greengrass.componentmanager.exceptions.PackageDownloadException;
 import com.aws.greengrass.componentmanager.exceptions.PackageLoadingException;
@@ -19,8 +22,7 @@ import com.aws.greengrass.componentmanager.models.ComponentIdentifier;
 import com.aws.greengrass.componentmanager.models.ComponentMetadata;
 import com.aws.greengrass.componentmanager.models.ComponentRecipe;
 import com.aws.greengrass.componentmanager.models.RecipeMetadata;
-import com.aws.greengrass.componentmanager.plugins.ArtifactDownloader;
-import com.aws.greengrass.componentmanager.plugins.ArtifactDownloaderFactory;
+import com.aws.greengrass.componentmanager.plugins.Image;
 import com.aws.greengrass.config.Topic;
 import com.aws.greengrass.dependency.InjectionActions;
 import com.aws.greengrass.deployment.DeviceConfiguration;
@@ -64,7 +66,9 @@ import javax.inject.Inject;
 
 import static com.aws.greengrass.componentmanager.KernelConfigResolver.PREV_VERSION_CONFIG_KEY;
 import static com.aws.greengrass.componentmanager.KernelConfigResolver.VERSION_CONFIG_KEY;
+import static com.aws.greengrass.componentmanager.plugins.DockerManagerService.DOCKER_MANAGER_PLUGIN_SERVICE_NAME;
 import static com.aws.greengrass.deployment.converter.DeploymentDocumentConverter.ANY_VERSION;
+import static com.aws.greengrass.tes.TokenExchangeService.TOKEN_EXCHANGE_SERVICE_TOPICS;
 import static org.apache.commons.io.FileUtils.ONE_MB;
 
 public class ComponentManager implements InjectionActions {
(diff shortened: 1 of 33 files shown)


src/main/java/com/aws/greengrass/componentmanager/plugins/DefaultDockerClient.java
```diff
@@ -0,0 +1,117 @@
...
+        // TODO : [Blocker] Ensure credentials in command to be run do not get logged
+        CliResponse response = runDockerCmd(String.format("docker login %s -u %s -p %s", registry.getRegistryId(),
+                registry.getCredentials().getUsername(), registry.getCredentials().getPassword()));
+        if (response.exit.isPresent()) {
+            if (response.exit.get() != 0) {
+                throw new ImagePullServiceException(
+                        String.format("Error logging into the registry using credentials - %s", response.out));
+            }
+        }
+        return true;
```
hui-yang, line 56: Do we need both the boolean return value and the exceptions?
If `response.exit` is not present, is login still successful?

src/main/java/com/aws/greengrass/componentmanager/plugins/DefaultDockerClient.java
```diff
@@ -0,0 +1,117 @@
...
+                    throw new ImagePullRetryableException(
+                            String.format("Error logging into the registry using credentials - %s", response.out));
+                }
+                if (response.getOut().contains("repository does not exist or may require 'docker login'")) {
+                    throw new InvalidImageOrAccessDeniedException(
+                            String.format("Invalid image or login - %s", response.out));
+                }
+            }
+        }
+        return true;
```
hui-yang, line 84: Same questions as above.

Add more logging for other cases?

src/main/java/com/aws/greengrass/componentmanager/plugins/DockerImageDownloader.java
```diff
@@ -0,0 +1,155 @@
...
+        // TODO : Maybe worth checking if device is configured such that it can get TES credentials for ECR.
+        // N/A for images from other registries
+        return Optional.empty();
+    }
+
+    @Override
+    public Long getDownloadSize() throws PackageDownloadException, InterruptedException {
+        // Not supported for docker images
+        return null;
+    }
```
hui-yang, line 130: Do it mean customers are required to clean up docker images manually?

src/main/java/com/aws/greengrass/componentmanager/ComponentManager.java
```diff
@@ -335,6 +338,41 @@ private ComponentMetadata getComponentMetadata(ComponentIdentifier componentIden
...
+            Optional<ComponentRecipe> recipeOption = componentStore.findPackageRecipe(pkgId);
+            if (!recipeOption.isPresent()) {
+                throw new PackageLoadingException(
+                        "Unexpected error - cannot find recipe for a component to be " + "prepared");
+            }
+            for (ComponentArtifact artifact : recipeOption.get().getArtifacts()) {
+                // TODO : Make this more generic & use dedicated component type
+                if (artifact.getArtifactUri().getSchemeSpecificPart().equals(ArtifactDownloaderFactory.DOCKER_SCHEME)) {
+                    // TODO : Find better name for component
+                    if (!pkgIds.contains("aws.greengrass.DockerPlugin")) {
```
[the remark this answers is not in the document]
shaguptashaikh: 1) Fixed it.

2) The plugin will be required at download time, hence checking the dependency closure since even if the component is not a dependency of other components, it can still be present in the deployment independently

3) The plugin code is currently in the nucleus like in this PR, but ultimately, it needs to be moved out of the nucleus like other plugins. And for that, we need a check for it always having to be included in deployment for backwards compatibility when we move it out

src/main/java/com/aws/greengrass/componentmanager/ComponentManager.java
```diff
@@ -335,6 +338,41 @@ private ComponentMetadata getComponentMetadata(ComponentIdentifier componentIden
...
+                        throw new PackageLoadingException("Special components with docker artifacts must include the "
+                                + "aws.greengrass.DockerPlugin download plugin in the deployment");
+                    }
+                    Image image = Image.fromArtifactUri(artifact.getArtifactUri());
+                    if (image.getRegistry().isEcrRegistry() && image.getRegistry().isPrivateRegistry() && !pkgIds
+                            .contains(TOKEN_EXCHANGE_SERVICE_TOPICS)) {
+                        throw new PackageLoadingException("Components with private ECR docker artifacts must include "
+                                + "the aws.greengrass.TokenExchangeService plugin in the deployment");
+                    }
```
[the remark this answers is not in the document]
shaguptashaikh: Same as above, it's a download time dependency, and the downloader code will be made its own plugin soon but we need these checks for backwards compatibility when we get to that stage

src/main/java/com/aws/greengrass/componentmanager/plugins/DefaultDockerClient.java
shaguptashaikh: We don't need it, it's just that RetryUtils is written in way it requires a return value. I will spend some time to fix it after other issues are resolved

src/main/java/com/aws/greengrass/componentmanager/plugins/DockerImageDownloader.java
shaguptashaikh: For now, that's the decision. We discussed certain ways to address automatic cleanup, but it'll be a follow up item

src/main/java/com/aws/greengrass/componentmanager/ComponentManager.java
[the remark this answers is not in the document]
hui-yang: Does it mean customer need at least two deployments to use this feature? first deployment to add docker plugin, and second to deploy a docker artifact component.
If that's the case, instead of finding `aws.greengrass.DockerPlugin` in the resolved component list, we should find in `kernel.locate` in order to make sure it's already loaded and running

src/main/java/com/aws/greengrass/componentmanager/plugins/DockerManagerService.java
```diff
@@ -0,0 +1,33 @@
...
+
+import javax.inject.Inject;
+import javax.inject.Singleton;
+
+/**
+ * Loads docker image artifact downloader. Currently just a placeholder since the downloader is part of the
+ * nucleus but when the downloader needs to be moved out into its own plugin, this service will be
+ * instantiated for the plugin.
+ */
+@ImplementsService(name = "aws.greengrass.DockerContainerManager")
```
hui-yang, line 20: nit: DockerImageManager?

src/main/java/com/aws/greengrass/componentmanager/plugins/Image.java
```diff
@@ -0,0 +1,126 @@
...
+     * Split a string, docker registry+image spec within this class, by given separator that identifies unique parts of
+     * the specified string.
+     *
+     * @param toSplit                       String to split
+     * @param separator                     Separator to split by
+     * @param setAllLeftWhenSeparatorAbsent When separator is not present, if true, all of source string will be set as
+     *                                      the first return value. If false, all of source string will be set as the
+     *                                      second return value.
+     * @param defaultWhenSeparatorAbsent    When separator is not present, use return this value for the other return
+     *                                      value than the one assigned with all of source string
```
hui-yang, line 93: nit: the comment is a little confusing

hui-yang (approved):

src/main/java/com/aws/greengrass/componentmanager/plugins/DefaultDockerClient.java
```diff
@@ -0,0 +1,150 @@
...
+     * @param registry Registry to log into, with credentials encapsulated
+     * @return if docker login was successful
+     * @throws ImagePullServiceException           error in authenticating with the registry
+     * @throws UserNotAuthorizedForDockerException when current user is not authorized to use docker
+     * @throws IOException                         unexpected error
+     */
+    public boolean login(Registry registry)
+            throws ImagePullServiceException, UserNotAuthorizedForDockerException, IOException {
+        // TODO : [Blocker] Ensure credentials in command to be run do not get logged
+        CliResponse response = runDockerCmd(String.format("docker login %s -u %s -p %s", registry.getEndpoint(),
```
tilo-chen, line 52: If you're looking for programmatic replacement for CLI, unfortunately there's [no official Java SDK](https://docs.docker.com/engine/api/sdk/#unofficial-libraries). We're using https://github.com/docker-java/docker-java for in UAT framework which is a good candidate.

src/main/java/com/aws/greengrass/componentmanager/builtins/ArtifactDownloader.java
```diff
@@ -250,5 +251,35 @@ protected String getErrorString(String reason) {
...
                 identifier.getName(), identifier.getVersion().toString()) + reason;
     }
+
+    /**
+     * Check if an instance of implemented class supports checking component store size depending on
+     * if the artifact is located in greengrass artifact store or third party.
+     *
+     * @return evaluation result
+     */
+    public boolean checkComponentStoreSize() {
```
tilo-chen, line 261: "**can**CheckComponentStoreSize"?

src/main/java/com/aws/greengrass/componentmanager/plugins/DefaultDockerClient.java
shaguptashaikh: Discussed offline about this

src/main/java/com/aws/greengrass/componentmanager/builtins/ArtifactDownloader.java
shaguptashaikh: I meant for it to convey if if there's a point in checking component store size, we of course can check the component store size always but there's no use doing that if artifacts are located in a third party store like docker images

src/main/java/com/aws/greengrass/componentmanager/plugins/DockerManagerService.java
hui-yang: Also use a constant for the service name?

shaguptashaikh: Changed to constant, not using Image in the name since I don't want to tie the component to images, so keeping it generic

hui-yang (approved):

tilo-chen (approved):

PR #867, opened 2021-02-22
fix(shadowdeployment): handle out of order shadow updates

fahadmohammed01:
**Issue #, if available:**

We update the shadow when a deployment completes which will queue the MQTT message to be delivered eventually. If multiple changes happen, then the messages may be processed out of order, thus causing the shadow to be in an inconsistent state.

**Description of changes:**
Update the shadow using optimistic locking, which adds a version field to the update. The shadow will only be updated in the case that the current shadow version matches the version provided in the update. If a shadow update is rejected, device will fetch the latest shadow and retry the update with the updated shadow version.  This will ensure that the shadow is always in a consistent state and will eventually show the current state.

E2E tests.

Tests:
When a shadow update is rejected, device fetches the latest shadow document.
When a shadow document with outdated deployment status is received, device updates the reported section with the latest deployment status.

**Checklist:**
 - [ X] Updated or added new end-to-end tests

### DIFF
--- a/src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
+++ b/src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
@@ -39,6 +39,7 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 import javax.inject.Inject;
 
@@ -95,7 +96,7 @@ public class ShadowDeploymentListener implements InjectionActions {
         }
     };
     private String lastConfigurationArn;
-    private Integer lastVersion;
+    private AtomicInteger lastVersion = new AtomicInteger();
     private final AtomicReference<Map<String, Object>> lastDeploymentStatus = new AtomicReference();
     protected static final Random JITTER = new Random();
 
@@ -216,7 +217,7 @@ public class ShadowDeploymentListener implements InjectionActions {
             updateNamedShadowRequest.shadowName = DEPLOYMENT_SHADOW_NAME;
             updateNamedShadowRequest.thingName = thingName;
             updateNamedShadowRequest.state = shadowState;
-            updateNamedShadowRequest.version = lastVersion;
+            updateNamedShadowRequest.version = lastVersion.get();
             iotShadowClient.PublishUpdateNamedShadow(updateNamedShadowRequest, QualityOfService.AT_LEAST_ONCE)
                     .get(TIMEOUT_FOR_PUBLISHING_TO_TOPICS_SECONDS, TimeUnit.SECONDS);
 
@@ -254,12 +255,12 @@ public class ShadowDeploymentListener implements InjectionActions {
     }
 
     protected void shadowUpdated(Map<String, Object> desired, Map<String, Object> reported, Integer version) {
-        if (lastVersion != null && lastVersion > version) {
+        if (lastVersion.get() > version) {
             logger.atInfo().kv("SHADOW_VERSION", version)
                     .log("Received an older version of shadow. Ignoring...");
             return;
         }
-        lastVersion = version;
+        lastVersion.set(version);
         //the reported section of the shadow was updated
         if (reported != null && !reported.isEmpty()) {
             syncShadowDeploymentStatus(reported);
--- a/src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
+++ b/src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
@@ -334,9 +334,8 @@ public class ShadowDeploymentListener implements InjectionActions {
         if (lastDeploymentStatus.get() == null) {
             return;
         }
-        HashMap<String, Object> latestDeploymentStatus = populateReportedSectionOfShadow(lastDeploymentStatus.get());
-        if (!reported.get(ARN_FOR_STATUS_KEY).equals(latestDeploymentStatus.get(ARN_FOR_STATUS_KEY))
-                || !reported.get(STATUS_KEY).equals(latestDeploymentStatus.get(STATUS_KEY))) {
+        if (!reported.get(ARN_FOR_STATUS_KEY).equals(lastDeploymentStatus.get().get(DEPLOYMENT_ID_KEY_NAME))
+                || !reported.get(STATUS_KEY).equals(lastDeploymentStatus.get().get(DEPLOYMENT_STATUS_KEY_NAME))) {
             logger.info("Updating reported section of shadow with the latest deployment status");
             updateReportedSectionOfShadowWithDeploymentStatus();
         }
--- a/src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
+++ b/src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
@@ -120,6 +120,7 @@ public class ShadowDeploymentListener implements InjectionActions {
 
     /*
         Subscribe to "$aws/things/{thingName}/shadow/update/accepted" topic to get notified when shadow is updated
+        Subscribe to "$aws/things/{thingName}/shadow/update/rejected" topic to get notified when an update is rejected
         Subscribe to "$aws/things/{thingName}/shadow/get/accepted" topic to retrieve shadow by publishing to get topic
      */
     private void subscribeToShadowTopics() {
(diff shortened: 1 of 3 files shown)


MikeDombo (approved):

src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
```diff
@@ -135,6 +136,14 @@ private void subscribeToShadowTopics() {
...
                         .get(TIMEOUT_FOR_SUBSCRIBING_TO_TOPICS_SECONDS, TimeUnit.SECONDS);
                 logger.info("Subscribed to update named shadow accepted topic");
+
+                iotShadowClient.SubscribeToUpdateNamedShadowRejected(updateNamedShadowSubscriptionRequest,
+                        QualityOfService.AT_LEAST_ONCE,
+                        updateShadowRejected -> handleNamedShadowRejectedEvent(),
+                        (e) -> logger.atError().log("Error processing named shadow update rejected response", e))
+                        .get(TIMEOUT_FOR_SUBSCRIBING_TO_TOPICS_SECONDS, TimeUnit.SECONDS);
+                logger.info("Subscribed to update named shadow rejected topic");
+
```
fengwa-aws, line 146: line 152: why is it `QualityOfService.AT_MOST_ONCE`?

fahadmohammed01: mybad, updated to  QualityOfService.AT_LEAST_ONCE,

src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
```diff
@@ -225,13 +249,22 @@ private Boolean deploymentStatusChanged(Map<String, Object> deploymentDetails) {
...
     protected void shadowUpdated(Map<String, Object> desired, Map<String, Object> reported, Integer version) {
+        if (lastVersion != null && lastVersion > version) {
+            logger.atInfo().kv("SHADOW_VERSION", version)
+                    .log("Received an older version of shadow. Ignoring...");
+            return;
+        }
+        lastVersion = version;
+        //the reported section of the shadow was updated
+        if (reported != null && !reported.isEmpty()) {
+            syncShadowDeploymentStatus(reported);
```
fengwa-aws, line 265: hmmm...don't you want to terminate the code execution here once `syncShadowDeploymentStatus` returns? Otherwise wouldn't it kick off another deployment?

src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
```diff
@@ -95,6 +96,7 @@ public void onConnectionResumed(boolean sessionPresent) {
     };
     private String lastConfigurationArn;
     private Integer lastVersion;
```
tilo-chen, line 98: Should these be atomic as well?

src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
```diff
@@ -303,6 +327,19 @@ protected void shadowUpdated(Map<String, Object> desired, Map<String, Object> re
...
     }
 
+    private void syncShadowDeploymentStatus(Map<String, Object> reported) {
+        // device does not have anything to report
+        if (lastDeploymentStatus.get() == null) {
+            return;
+        }
+        HashMap<String, Object> latestDeploymentStatus = populateReportedSectionOfShadow(lastDeploymentStatus.get());
+        if (!reported.get(ARN_FOR_STATUS_KEY).equals(latestDeploymentStatus.get(ARN_FOR_STATUS_KEY))
+                || !reported.get(STATUS_KEY).equals(latestDeploymentStatus.get(STATUS_KEY))) {
```
fengwa-aws, line 338: Is calling `populateReportedSectionOfShadow()` necessary here? Seems redundant to me. Why not just directly compare `lastDeploymentStatus` and `reported`?

src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
```diff
@@ -303,6 +327,19 @@ protected void shadowUpdated(Map<String, Object> desired, Map<String, Object> re
         }
     }
 
+    private void syncShadowDeploymentStatus(Map<String, Object> reported) {
+        // device does not have anything to report
+        if (lastDeploymentStatus.get() == null) {
+            return;
+        }
```
fengwa-aws, line 335: Do we expect this to happen in normal cases? When would this occur? The handling doesn't seem right.

fahadmohammed01: This would happen when the device receives its first shadow deployment. In that case, the device doesn't have anything to report and there is no need to sync the reported section.

src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
fahadmohammed01: Certain fields in the deployment status is optional like (DeploymentStatusDetails). If a deployment status does not have an optional field, we need to explicitly delete that fields in the shadow document. For this we include the filed name with its value set as null. 

So lastDeploymentStatus will have optional fields names with value as null, while reported may not have them. This prevents us from directly comparing the two.

src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
fahadmohammed01: The device de-dupes deployment based on fleet config arn.

fengwa-aws: Why not just exit here if that's right behavior? That way the code is also more readable. I don't see any benefit of relying the deployment to dedupe it unless there is a scenario when it's a new deployment while the `reported` is also non-null.

src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
fengwa-aws: `syncShadowDeploymentStatus` is only called when `reported` is non-empty. Is that the case for the first shadow deployment?

src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
fengwa-aws: But you're only comparing `ARN_FOR_STATUS_KEY` and `STATUS_KEY`. What I meant is you can just read these two keys from `latestDeploymentStatus`.

src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
fahadmohammed01: Yes, for the first shadow deployment the reported section is empty. This will only trigger when the reported section is not null but the lastDeploymentStatus is null (after a device restart)

src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
fahadmohammed01: Updated likewise

src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
fahadmohammed01: Updated likewise.

src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
fahadmohammed01: A device can get notified of a deployment via two mechanisms:
1. When the device is online, it subscribes to /update/accepted topic to get notified of any changes to the shadow document.
2. When the device is offline for a while. When the device reconnects, it fetches the whole shadow document to see if it missed any deployment notification when it was offline. In this scenario the reported section and desired section can be non-null.

Because shadowUpdated() handles both the scenarios, we cant return after syncShadowDeploymentStatus.

src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
fengwa-aws: Can you add a log then? I don't like we just return without any log. It will make debugging very difficult.

fahadmohammed01: Added log

MikeDombo (approved):

fengwa-aws:
Can you check why the build fails?

fengwa-aws (approved):

PR #869, opened 2021-02-23
fix(dependencyresolver): fail fast when circular dependency is detected

fahadmohammed01:
**Issue #, if available:**
1.  Circular dependencies causes deployment to go into an infinite loop.
2. During dependency resolution, if the resolved version of a component changes. Say during the initial stage of dependency resolution B got resolved to version 1.0.0 and at a later stage  B got resolved to 2.0.0. Current implementation does not remove the dependencies of B=1.0.0.

**Description of changes:**
1. Dependency resolver will detect circular dependency during dependency resolution and fail the deployment. The detailed deployment status will mention circular dependency as the reason.

2. When the version of a component is bumped from v1 to v2. Dependency resolver will remove dependencies of v1. All version constraints placed by the removed components will also be removed.

Unit and Integration tests.

Test Plan:
Testing:
Issue 1:
when root component has circular dependency, deployment completes with status FAILED
```
     /*
     *      group1
     *         \(1.0.0)
     *     |--> A
     *  1.0.0  / \
     *     |__/   \(>1.0)
     *             B2
     */
```

when non-root component has circular dependency, deployment completes with status FAILED
```
    /*
     *      group1
     *         \(1.0.0)
     *          A
     *           \
     *            \(=1.0.0)
     *         |-> B1
     *         |    \
     *         |     \(=1.0.0)
     *      (=1.0.0)  B2
     *         |       \(<=1.1.0)
     *         |        \
     *         |--------C1
     */
```
when a component has more than one depender but there is no cycle in the dependency tree, the deployment completes successfully.
```
        /*
         *       group1
         *          |(1.0.0)
         *          A
         * (1.0.0)/ |  \(>1.0)
         *     B1   |   B2
         *(=1.0.0)\ |  / (=1.0.0)
         *          C1
         */
```

Issue 2:
When a component version is bumped from v1 to v2, all the dependencies of v1 should be removed. All version constraints places by the removed components should also be removed. 

    /*
     *      group1
     *         \(1.0.0)
     *          A
     (>=1.0.0)/   \(1.0)
     *       B     D
                  / (>=2.0.0)
     *           /
     *          B
     *
     *   B will be first resolved to version 1.0.0 which depends on X=2.0.0
     *   X=2.0.0 depends on Y=1.0.0
     *
     *   B will then be resolved to version 2.0.0 which depends on Y=2.0.0
     *   For the test to succeed X should be removed from the dependency tree and the constraint X puts on Y should also be removed.
     */ 

 When a component version is resolved to v1 when resolving dependency tree of group 1 and gets re-resolved to v2 when resolving the dependency tree of group2. All the dependencies of v1 should be removed. 
All version constraints places by the removed components should also be removed. 

    /*
     *      group1                         group2
     *         \(1.0.0)                       \(1.0)
     *          A                              D  
     (>=1.0.0)/                               / (>=2.0.0)
     *       B                               /
     *                                      B          
     *
     *   B will be first resolved to version 1.0.0 which depends on X=2.0.0
     *   X=2.0.0 depends on Y=1.0.0
     *
     *   B will then be resolved to version 2.0.0 which depends on Y=2.0.0
     *   For the test to succeed X should be removed from the dependency tree and the constraint X puts on Y should also be removed.
     */ 

**Checklist:**
 - [ X] Updated or added new unit tests
 - [ X] Updated or added new integration tests

### DIFF
--- a/src/main/java/com/aws/greengrass/componentmanager/DependencyResolver.java
+++ b/src/main/java/com/aws/greengrass/componentmanager/DependencyResolver.java
@@ -6,6 +6,7 @@
 package com.aws.greengrass.componentmanager;
 
 import com.amazon.aws.iot.greengrass.component.common.ComponentType;
+import com.aws.greengrass.componentmanager.exceptions.ComponentVersionNegotiationException;
 import com.aws.greengrass.componentmanager.exceptions.NoAvailableComponentVersionException;
 import com.aws.greengrass.componentmanager.exceptions.PackagingException;
 import com.aws.greengrass.componentmanager.models.ComponentIdentifier;
@@ -221,11 +222,18 @@ public class DependencyResolver {
             ComponentMetadata resolvedVersion = componentResolver.resolve(componentToResolve, versionConstraints);
             logger.atDebug().kv("resolvedVersion", resolvedVersion).log("Resolved component");
             resolvedComponents.put(componentToResolve, resolvedVersion.getComponentIdentifier());
-            resolvedVersion.getDependencies().forEach((k, v) -> {
-                componentNameToVersionConstraints.putIfAbsent(k, new HashMap<>());
-                componentNameToVersionConstraints.get(k).put(componentToResolve, Requirement.buildNPM(v));
-                componentsToResolve.add(k);
-            });
+
+            for (Map.Entry<String, String> dependency : resolvedVersion.getDependencies().entrySet()) {
+                // A circular dependency is present if the dependency is already resolved.
+                if (resolvedComponents.containsKey(dependency.getKey())) {
+                    throw new ComponentVersionNegotiationException("Circular dependency detected for Component "
+                            + dependency.getKey());
+                }
+                componentNameToVersionConstraints.putIfAbsent(dependency.getKey(), new HashMap<>());
+                componentNameToVersionConstraints.get(dependency.getKey()).put(componentToResolve,
+                        Requirement.buildNPM(dependency.getValue()));
+                componentsToResolve.add(dependency.getKey());
+            }
         }
 
         logger.atDebug().setEventType("traverse-dependencies-finish").kv("resolvedComponents", resolvedComponents)
--- a/src/main/java/com/aws/greengrass/componentmanager/DependencyResolver.java
+++ b/src/main/java/com/aws/greengrass/componentmanager/DependencyResolver.java
@@ -226,7 +226,7 @@ public class DependencyResolver {
             for (Map.Entry<String, String> dependency : resolvedVersion.getDependencies().entrySet()) {
                 // A circular dependency is present if the dependency is already resolved.
                 if (resolvedComponents.containsKey(dependency.getKey())) {
-                    throw new ComponentVersionNegotiationException("Circular dependency detected for Component "
+                    throw new ComponentVersionNegotiationException("Circular dependency detected for component "
                             + dependency.getKey());
                 }
                 componentNameToVersionConstraints.putIfAbsent(dependency.getKey(), new HashMap<>());
--- a/src/main/java/com/aws/greengrass/componentmanager/DependencyResolver.java
+++ b/src/main/java/com/aws/greengrass/componentmanager/DependencyResolver.java
@@ -307,7 +307,8 @@ public class DependencyResolver {
                 componentDependencyMap.keySet(), componentDependencyMap::get);
 
         if (result.size() != componentCount) {
-            throw new ComponentVersionNegotiationException("Circular dependency detected for component " + targetComponent);
+            throw new ComponentVersionNegotiationException("Circular dependency detected for component "
+                    + targetComponent);
         }
     }
 
--- a/src/main/java/com/aws/greengrass/componentmanager/DependencyResolver.java
+++ b/src/main/java/com/aws/greengrass/componentmanager/DependencyResolver.java
@@ -19,6 +19,7 @@ import com.aws.greengrass.lifecyclemanager.Kernel;
 import com.aws.greengrass.logging.api.Logger;
 import com.aws.greengrass.logging.impl.LogManager;
 import com.aws.greengrass.util.Coerce;
+import com.aws.greengrass.util.DependencyOrder;
 import com.aws.greengrass.util.Utils;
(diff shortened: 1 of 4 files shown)


src/main/java/com/aws/greengrass/componentmanager/DependencyResolver.java
```diff
@@ -221,11 +222,18 @@ void checkNonExplicitNucleusUpdate(List<String> targetComponents,
...
-            resolvedVersion.getDependencies().forEach((k, v) -> {
-                componentNameToVersionConstraints.putIfAbsent(k, new HashMap<>());
-                componentNameToVersionConstraints.get(k).put(componentToResolve, Requirement.buildNPM(v));
-                componentsToResolve.add(k);
-            });
+
+            for (Map.Entry<String, String> dependency : resolvedVersion.getDependencies().entrySet()) {
+                // A circular dependency is present if the dependency is already resolved.
+                if (resolvedComponents.containsKey(dependency.getKey())) {
+                    throw new ComponentVersionNegotiationException("Circular dependency detected for Component "
```
MikeDombo, line 229: lowercase component here

fahadmohammed01: Updated

MikeDombo (dismissed):

src/main/java/com/aws/greengrass/componentmanager/DependencyResolver.java
```diff
@@ -221,11 +222,18 @@ void checkNonExplicitNucleusUpdate(List<String> targetComponents,
...
             logger.atDebug().kv("resolvedVersion", resolvedVersion).log("Resolved component");
             resolvedComponents.put(componentToResolve, resolvedVersion.getComponentIdentifier());
-            resolvedVersion.getDependencies().forEach((k, v) -> {
-                componentNameToVersionConstraints.putIfAbsent(k, new HashMap<>());
-                componentNameToVersionConstraints.get(k).put(componentToResolve, Requirement.buildNPM(v));
-                componentsToResolve.add(k);
-            });
+
+            for (Map.Entry<String, String> dependency : resolvedVersion.getDependencies().entrySet()) {
+                // A circular dependency is present if the dependency is already resolved.
```
shaguptashaikh, line 227: Is this really the case? can the dependency not be previously resolved if it was a dependency of some other component, i.e. in an independent branch of the dependency tree?

fahadmohammed01: resolvedComponents is only tracking the dependencies of one root component. It does not track resolved components across multiple . But you are right, in the same dependency tree if the component was resolved because it was dependency of some other component it will fail.


Below example which I used to test works, but if A depends on C1 then it fails. Let me fix that.
               group1
                    |(1.0.0)
                   A
       (1.0.0)/   \(>1.0)
              B1     B2
      (1.0.0)\    /(1.0.0)
                 \  /
                  C1

src/main/java/com/aws/greengrass/componentmanager/DependencyResolver.java
```diff
@@ -182,57 +207,112 @@ void checkNonExplicitNucleusUpdate(List<String> targetComponents,
...
+            componentDependencyMap.put(componentName, dependencies);
+            dependencies.stream().filter(dependency -> !componentDependencyMap.containsKey(dependency))
+                    .forEach(componentsToVisit::add);
+        }
+        int componentCount = componentDependencyMap.keySet().size();
+        LinkedHashSet<String> result = new DependencyOrder<String>().computeOrderedDependencies(
+                componentDependencyMap.keySet(), componentDependencyMap::get);
+
+        if (result.size() != componentCount) {
+            throw new ComponentVersionNegotiationException("Circular dependency detected for component "
```
MikeDombo, line 310: would be great if we can list the components in the cycle.

MikeDombo (dismissed):

src/main/java/com/aws/greengrass/componentmanager/DependencyResolver.java
```diff
@@ -182,57 +207,112 @@ void checkNonExplicitNucleusUpdate(List<String> targetComponents,
...
         logger.atDebug().setEventType("traverse-dependencies-finish").kv("resolvedComponents", resolvedComponents)
                 .log("Finish traversing dependencies");
-        return resolvedComponents;
     }
 
+    /*
+     A component version is removed from the dependency tree, remove all dependencies of this version
+     which has an incoming reference count of 1 (i.e no other component has depends on them)
+     */
+    private void removeDependencies(ComponentMetadata removedComponentVersion,
```
shaguptashaikh, line 260: Thanks for sorting this out

src/main/java/com/aws/greengrass/componentmanager/DependencyResolver.java
```diff
@@ -106,18 +122,27 @@
...
         for (String component : targetComponentsToResolve) {
-            resolvedComponents.putAll(resolveComponentDependencies(component, componentNameToVersionConstraints,
+            resolveComponentDependencies(component, componentNameToVersionConstraints,
+                    resolvedComponents, componentIncomingReferenceCount,
                     (name, requirements) -> componentManager.resolveComponentVersion(name, requirements,
-                            document.getDeploymentId())));
+                            document.getDeploymentId()));
+        }
+
+        // detect circular dependencies for target components from the current deployment
```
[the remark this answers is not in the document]
fahadmohammed01: A deployment will not succeed if there is a circular dependency, so we can assume that if are to find a circular dependency it will be in the current deployment.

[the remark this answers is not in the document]
fahadmohammed01: The target components belonging to other deployment/groups has already been checked for circular dependency. If a circular dependency is introduced by a component this deployment brought it (new component or a bump in component version) it would introduce a circular dependency for target components of this deployment as well. Because of this assumption, we only need to check the current deployment/group.

shaguptashaikh (approved):

MikeDombo (approved):

shaguptashaikh (approved):

PR #871, opened 2021-02-25
fix(loader): make loader script executable immediately on startup

MikeDombo:
**Issue #, if available:**

**Description of changes:**
Set the loader script to be executable immediately when the Nucleus starts up if it isn't already set as executable. This is required so that we can `exec` the loader when restarting the Nucleus.

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/lifecyclemanager/KernelAlternatives.java
+++ b/src/main/java/com/aws/greengrass/lifecyclemanager/KernelAlternatives.java
@@ -13,6 +13,7 @@ import com.aws.greengrass.logging.impl.LogManager;
 import com.aws.greengrass.util.CommitableWriter;
 import com.aws.greengrass.util.NucleusPaths;
 import com.aws.greengrass.util.Utils;
+import com.aws.greengrass.util.platforms.Platform;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import lombok.AccessLevel;
 import lombok.Getter;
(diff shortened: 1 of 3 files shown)


avipinku (dismissed):

src/main/java/com/aws/greengrass/lifecyclemanager/KernelAlternatives.java
```diff
@@ -105,7 +107,15 @@ public Path getLoaderPath() {
     }
 
     private Path getLoaderPathFromLaunchDir(Path path) {
-        return path.resolve(KERNEL_DISTRIBUTION_DIR).resolve(KERNEL_BIN_DIR).resolve(LOADER_FILE);
+        Path launchPath = path.resolve(KERNEL_DISTRIBUTION_DIR).resolve(KERNEL_BIN_DIR).resolve(LOADER_FILE);
+        if (Files.exists(launchPath, NOFOLLOW_LINKS) && !launchPath.toFile().canExecute()) {
+            try {
+                Platform.getInstance().setPermissions(OWNER_RWX_ONLY, launchPath);
```
hui-yang, line 113: OWNER_RWX_EVERYONE_RX?

hui-yang: And can we only change in `setupInitLaunchDirIfAbsent`? so it's more explicit and only done once?

avipinku (approved):

hui-yang (approved):

src/main/java/com/aws/greengrass/lifecyclemanager/KernelAlternatives.java
MikeDombo: What if the directory is present, but isn't setup correctly? I put it here so that it is guaranteed to update the permissions to be correct upon launch.

MikeDombo: Updated to put it in `validateLaunchDirSetup` instead.

hui-yang (dismissed):

avipinku (approved):

fahadmohammed01 (approved):

hui-yang (approved):

PR #878, opened 2021-03-01
fix(awsiotmqttclient): more checks to avoid resub if connection closed

tilo-chen:
**Issue #, if available:**

**Description of changes:**
A test case has been consistently failing: it closes the client right after resuming the network. The resub task is submitted right after the client is closed. The thread pool can be closed by then, so the test failed with RejectedExecutionException.

Adding more connection state check to avoid doing resub if connection is closed.

**Checklist:**


MikeDombo (dismissed):

src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
```diff
@@ -252,8 +252,7 @@ int getTimeout() {
      * @param sessionPresent whether the session persisted
      */
     private synchronized void resubscribe(boolean sessionPresent) {
-        // Don't bother spin up a thread if subscription is empty
-        if (!subscriptionTopics.isEmpty()) {
+        if (currentlyConnected.get() && !subscriptionTopics.isEmpty()) {
```
hui-yang, line 255: Do we need to handle exceptions from `currentlyConnected.get()`?

tilo-chen: `AtomicBoolean.get` shouldn't throw right? Actually I removed this one because `resubscribe` is called immediately after `currentlyConnected.set(true)`. The later one in `resubscribeDroppedTopicsTask` should be enough.

MikeDombo (approved):

hui-yang (approved):

PR #884, opened 2021-03-05
feat(shadowmanager): add shadowmanager ipc support

randolphtan121:
…shadowManager to authorizationHandler

**Issue #, if available:**
Shadow-1 

**Description of changes:**
- Taking updated smithy model changes and copying them over to the Nucleus (only shadow related code)
  - `GreengrassCoreIPCService.java`
- Adding ShadowManager to AuhtorizationHandler preset operations map
- Changing iotdevicesdk version to use `1.3.0-SNAPSHOT` which has shadow changes 

These changes are necessary to support ShadowManager component by adding ipc support for `GetThingShadow`, `UpdateThingShadow`, `DeleteThingShadow`, and `ListNamedShadowsForThing`

Currently tested locally with Shadow UAT's using a local nucleus kernel and a local ShadowManager component

These changes are dependent on the updated smithy model changes. An updated iotdevicesdk uber jar has been pushed to GG Maven repo which has the shadow smithy model changes and was built off of iotdevicesdk 1.3.0

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/authorization/AuthorizationHandler.java
+++ b/src/main/java/com/aws/greengrass/authorization/AuthorizationHandler.java
@@ -36,11 +36,15 @@ import static com.aws.greengrass.lifecyclemanager.GreengrassService.ACCESS_CONTR
 import static com.aws.greengrass.lifecyclemanager.GreengrassService.SERVICES_NAMESPACE_TOPIC;
 import static com.aws.greengrass.tes.TokenExchangeService.AUTHZ_TES_OPERATION;
 import static com.aws.greengrass.tes.TokenExchangeService.TOKEN_EXCHANGE_SERVICE_TOPICS;
+import static software.amazon.awssdk.aws.greengrass.GreengrassCoreIPCService.DELETE_THING_SHADOW;
 import static software.amazon.awssdk.aws.greengrass.GreengrassCoreIPCService.GET_SECRET_VALUE;
+import static software.amazon.awssdk.aws.greengrass.GreengrassCoreIPCService.GET_THING_SHADOW;
+import static software.amazon.awssdk.aws.greengrass.GreengrassCoreIPCService.LIST_NAMED_SHADOWS_FOR_THING;
 import static software.amazon.awssdk.aws.greengrass.GreengrassCoreIPCService.PUBLISH_TO_IOT_CORE;
 import static software.amazon.awssdk.aws.greengrass.GreengrassCoreIPCService.PUBLISH_TO_TOPIC;
 import static software.amazon.awssdk.aws.greengrass.GreengrassCoreIPCService.SUBSCRIBE_TO_IOT_CORE;
 import static software.amazon.awssdk.aws.greengrass.GreengrassCoreIPCService.SUBSCRIBE_TO_TOPIC;
+import static software.amazon.awssdk.aws.greengrass.GreengrassCoreIPCService.UPDATE_THING_SHADOW;
 
 /**
  * Main module which is responsible for handling AuthZ for Greengrass. This only manages
@@ -56,6 +60,7 @@ import static software.amazon.awssdk.aws.greengrass.GreengrassCoreIPCService.SUB
 public class AuthorizationHandler  {
     public static final String ANY_REGEX = "*";
     public static final String SECRETS_MANAGER_SERVICE_NAME = "aws.greengrass.SecretManager";
+    public static final String SHADOW_MANAGER_SERVICE_NAME = "aws.greengrass.ShadowManager";
     private static final Logger logger = LogManager.getLogger(AuthorizationHandler.class);
     private final ConcurrentHashMap<String, Set<String>> componentToOperationsMap = new ConcurrentHashMap<>();
     private final ConcurrentHashMap<String, List<AuthorizationPolicy>>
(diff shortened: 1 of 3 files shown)


randolphtan121:
PR paused while adding new shadow manager api (`ListNamedShadowsForThing`)

nikkhilmuthye (approved):

src/main/java/com/aws/greengrass/authorization/AuthorizationHandler.java
```diff
@@ -56,6 +60,7 @@
 public class AuthorizationHandler  {
     public static final String ANY_REGEX = "*";
     public static final String SECRETS_MANAGER_SERVICE_NAME = "aws.greengrass.SecretManager";
+    public static final String SHADOW_MANAGER_SERVICE_NAME = "aws.greengrass.ShadowManager";
```
MikeDombo, line 63: I'd strongly suggest that you consider rewriting this logic so that services can be registered dynamically.

src/main/java/software/amazon/awssdk/aws/greengrass/GeneratedAbstractUpdateThingShadowOperationHandler.java
```diff
@@ -0,0 +1,27 @@
+/*
```
MikeDombo, line 1: Should be possible to add these classes only in the shadow service itself.

randolphtan121: We need a reference to this in the `GreengrassCoreIPCService` when setting the handlers, can see it [here](https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/local-shadow-service-update/src/main/java/software/amazon/awssdk/aws/greengrass/GreengrassCoreIPCService.java#L228-L246).

src/main/java/com/aws/greengrass/authorization/AuthorizationHandler.java
randolphtan121: This was an issue raised by @rbattle before. I don't know how long this effort would take but this can be done in a future PR. In the interest of time we can table that and push this through to unblock ShadowManager.

src/main/java/software/amazon/awssdk/aws/greengrass/GeneratedAbstractUpdateThingShadowOperationHandler.java
rbattle: could we use the `[setOperationHandler](https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/local-shadow-service-update/src/main/java/software/amazon/awssdk/aws/greengrass/GreengrassCoreIPCService.java#L264)` method - but call it from within the shadow manager?

src/main/java/com/aws/greengrass/authorization/AuthorizationHandler.java
rbattle: yes - we've talked with Joan already about the need to have dynamic registration support in the nucleus. Secret Manager also shouldn't be hardcoded in here. This needs to be addressed - but it requires some more design on the plugin interface.

src/main/java/software/amazon/awssdk/aws/greengrass/GeneratedAbstractUpdateThingShadowOperationHandler.java
randolphtan121: Yes we can do that, we should be able to move these generated abstract classes to the ShadowManager and move the `set<Operation>Handlers` to the ShadowManager.

randolphtan121: This PR was updated to reflect this, also relevant [PR](https://github.com/aws-greengrass/aws-greengrass-shadow-manager/pull/42) made to ShadowManager to support this change as well.

src/main/java/software/amazon/awssdk/aws/greengrass/GreengrassCoreIPCService.java
```diff
@@ -66,10 +66,17 @@
...
+
+  public static final String GET_THING_SHADOW = SERVICE_NAMESPACE + "#GetThingShadow";
+
+  public static final String LIST_NAMED_SHADOWS_FOR_THING = SERVICE_NAMESPACE + "#ListNamedShadowsForThing";
+
   static {
     SERVICE_OPERATION_SET = new HashSet();
     SERVICE_OPERATION_SET.add(SUBSCRIBE_TO_IOT_CORE);
-    SERVICE_OPERATION_SET.add(PUBLISH_TO_TOPIC);
     SERVICE_OPERATION_SET.add(PUBLISH_TO_IOT_CORE);
```
rbattle, line 80: did `PUBLISH_TO_TOPIC` get removed somehow?

randolphtan121: It shouldn't have been removed... adding back

rbattle (approved):

nikkhilmuthye (approved):

MikeDombo (approved):

nikkhilmuthye (approved):

rbattle (approved):

PR #887, opened 2021-03-05
fix(mqttclient): add rate limiting for publish request at 100TPS/512KBPS

MikeDombo:
**Issue #, if available:**

**Description of changes:**
Manually including `RateLimiter` from Guava instead of taking a dependency on all of Guava. Implementing rate limiting at 100TPS to avoid IoT Core throttling at 100TPS. Also rate limiting based on bandwidth used since IoT Core limits us to 512 KBPS. Tested that publishing at 100TPS did not trigger IoT Core's throttling at all.

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
+++ b/src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
@@ -20,6 +20,7 @@ import software.amazon.awssdk.crt.mqtt.MqttException;
 import software.amazon.awssdk.crt.mqtt.MqttMessage;
 import software.amazon.awssdk.crt.mqtt.QualityOfService;
 import software.amazon.awssdk.iot.AwsIotMqttConnectionBuilder;
+import vendored.com.google.common.util.concurrent.RateLimiter;
 
 import java.io.Closeable;
 import java.time.Duration;
@@ -75,6 +76,11 @@ class AwsIotMqttClient implements Closeable {
     @Setter
     private static int waitTimeJitterMaxMillis = 10_000;
 
+    // Limit TPS to 100 which is IoT Core's limit per connection
+    private final RateLimiter transactionLimiter = RateLimiter.create(100.0);
+    // Limit bandwidth to 512 KBPS
+    private final RateLimiter bandwidthLimiter = RateLimiter.create(512.0 * 1024);
+
     @Getter(AccessLevel.PACKAGE)
     private final MqttClientConnectionEvents connectionEventCallback = new MqttClientConnectionEvents() {
         @Override
@@ -114,6 +120,13 @@ class AwsIotMqttClient implements Closeable {
         this.ses = ses;
     }
 
+    long getThrottlingWaitTimeMicros() {
+        // Return the worst possible wait time.
+        // Time to wait is independent of how many permits we need because future transactions
+        // will pay this current transaction's cost.  See the JavaDocs for RateLimiter for more info.
+        return Math.max(bandwidthLimiter.microTimeToNextPermit(), transactionLimiter.microTimeToNextPermit());
+    }
+
     // Notes about the CRT MQTT client:
     // client has no timeouts if the connection is dropped, then we do get an exception
     // so we need to retry ourselves. If offline, client waits to be online then tries to subscribe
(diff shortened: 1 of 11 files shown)


src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
```diff
@@ -146,6 +151,7 @@ public void onConnectionResumed(boolean sessionPresent) {
 
     CompletableFuture<Integer> publish(MqttMessage message, QualityOfService qos, boolean retain) {
         return connect().thenCompose((b) -> {
+            rateLimiter.acquire(1);
```
fengwa-aws, line 154: Why not use `tryAcquire`? Is it desirable to be blocked here?

src/main/java/com/google/common/base/Stopwatch.java
```diff
@@ -0,0 +1,244 @@
+/*
```
rbattle, line 1: is it possible to have 3rd party code like this under `src/vendor` instead of `src/main` to indicate it is from a 3rd party (similar to how projects use lib/vendor)?

src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
MikeDombo: I'll double check, but this shouldn't block the spooler thread due to the callback in the compose.

src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
```diff
@@ -146,6 +151,7 @@ public void onConnectionResumed(boolean sessionPresent) {
...
             logger.atTrace().kv(TOPIC_KEY, message.getTopic()).kv(QOS_KEY, qos.name()).kv("retain", retain)
```
rbattle, line 155: this isn't part of your change - but should this log move into the synchronized block? if multiple threads are waiting on the sync block, these logs could be out of order

src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
MikeDombo: I rewrote it so that it throttles in the spooler thread instead of the mqtt client. It now throttles on both TPS and bandwidth used. If there are multiple connections then it will choose to use the connection with the minimum wait time.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -645,7 +646,19 @@ protected void runSpooler() {
...
+                    AwsIotMqttClient connection = getConnection(false);
+                    long minimumWaitTimeMicros = connection.getThrottlingWaitTimeMicros();
+                    for (AwsIotMqttClient client : connections) {
+                        long waitTime = client.getThrottlingWaitTimeMicros();
+                        if (waitTime < minimumWaitTimeMicros) {
+                            connection = client;
+                            minimumWaitTimeMicros = waitTime;
+                        }
+                    }
+                    TimeUnit.MICROSECONDS.sleep(minimumWaitTimeMicros);
```
rbattle, line 659: can you add a comment here about why the sleep was added here. It isn't intuitive to have to sleep here while also still calling `acquire` in the client.

Did you keep those because some other component can use the MQTT client and publish while the spooler is sleeping?

MikeDombo: Acquire is what actually takes the tokens from the bucket. Prior to that I'm just querying when we can next take a token, but not actually taking it out since I need to query multiple limiters to find the minimum wait. Since we're doing the sleeping, the acquire is guaranteed to not block.

rbattle: right - can you add that as a comment to here in the source?

MikeDombo: Yes I'm adding that

MikeDombo: Updated.

rbattle (dismissed):

rbattle (dismissed):

saranyailla (approved):

src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
```diff
@@ -114,6 +120,11 @@ public void onConnectionResumed(boolean sessionPresent) {
         this.ses = ses;
     }
 
+    long getThrottlingWaitTimeMicros() {
```
jbutler, line 123: The bandwidth limiter needs one permit per byte, right? Don't we need to take message size into account here?

MikeDombo: The time to wait is independent of how much you want to send, that just further delays the next time that you can take tokens out of the bucket. Basically the initial request isn't delayed by how big it is, but subsequent requests will be.

jbutler: Nit: prolly worth calling out in a comment just so others don't also assume this is a bug

jbutler (approved):

jbutler (approved):

rbattle (approved):

PR #892, opened 2021-03-12
fix: map components to deployment groups

hui-yang:
**Issue #, if available:**

**Description of changes:**
After deployment, persist a map of each component to the groups that include the component. Previously only information from the last deployment was persisted.

This mapping from components to groups are used by FSS update and by log stream names in LogManager.
FSS update currently sends partial data but there's no impact from customers' perspective.
Log stream names also have partial information but it's low impact because customers don't necessarily look at stream names.

Unit tests and e2e tests

**Checklist:**


MikeDombo (commented):
Can you add context in the description for what this is doing?

MikeDombo (approved):

MikeDombo (approved):

tilo-chen (dismissed):

MikeDombo (approved):

tilo-chen (approved):

PR #894, opened 2021-03-15
feat(windows): lookup user by name and sid

fufranci:
**Issue #, if available:**

**Description of changes:**
Implement `lookupUserByName` and `lookupUserByIdentifier` in `WindowsPlatform`.

Currently, these two methods are not implemented, causing Cli server to crash.

Added unit test

**Checklist:**


src/test/java/com/aws/greengrass/util/platforms/unix/UnixPlatformTest.java
```diff
@@ -8,13 +8,16 @@
...
+import org.junit.jupiter.api.condition.EnabledOnOs;
+import org.junit.jupiter.api.condition.OS;
 import org.junit.jupiter.api.extension.ExtendWith;
 
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.arrayContaining;
 import static org.hamcrest.Matchers.is;
 
 @ExtendWith({GGExtension.class})
+@EnabledOnOs({OS.LINUX, OS.MAC})
```
MikeDombo, line 20: you can just use disabledon windows instead

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsPlatform.java
```diff
@@ -120,13 +126,33 @@ protected void setPermissions(FileSystemPermission permission, Path path,
     }
 
     @Override
-    public UserAttributes lookupUserByName(String user) throws IOException {
-        return null;
+    public WindowsUserAttributes lookupUserByName(String user) throws IOException {
```
MikeDombo, line 129: This is only used to determine if a user exists during setup; not sure we actually need this to be in the platform abstraction. If we can get rid of these, then I'd like that in order to remove the extra dependency.

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsPlatform.java
```diff
@@ -120,13 +126,33 @@ protected void setPermissions(FileSystemPermission permission, Path path,
...
+        }
+
+        return WindowsUserAttributes.builder().principalName(account.name).principalIdentifier(account.sidString)
+                .build();
     }
 
     @Override
-    public UserAttributes lookupUserByIdentifier(String identifier) throws IOException {
-        return null;
+    public WindowsUserAttributes lookupUserByIdentifier(String identifier) throws IOException {
```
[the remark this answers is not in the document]
fufranci: From looking at the Nucleus project, I think I can remove `lookupUserByIdentifier` from the platform abstraction. But do you know is there a good way to verify that there is no other project which depends on the Nucleus project using this method? I know the Cli project depends on this project. I can do a search there.

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsPlatform.java
fufranci: I am a little hesitate to remove this method since I haven't look into the setup code yet. I worried about if we remove this method now, we may need it when we work on the setup code for Windows. What do you think?

src/test/java/com/aws/greengrass/util/platforms/unix/UnixPlatformTest.java
fufranci: Thanks for pointing this out.

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsPlatform.java
MikeDombo: The method we need is userExists and not really this method.

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsPlatform.java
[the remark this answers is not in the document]
MikeDombo: You can search all of GitHub you might be able to scope it to our org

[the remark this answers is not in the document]
MikeDombo: But I also have everything checked out and it isn't used

MikeDombo (approved):

hui-yang (approved):

PR #897, opened 2021-03-16
fix(plugin): plugin components can be re-deployed

fahadmohammed01:
**Issue #, if available:**

Given a device that is running plugin component A
Device receives first deployment that removes the plugin component A
Device receives second deployment that add plugin component A. Second deployment fails because device is not able to load plugin component A.

**Description of changes:**

Root cause:
When a plugin component is loaded into the kernel, the device creates two entries for it in the context.
i . Plugin name in String -> value object
ii. plugin class -> value object

When a component is removed only the first mapping is removed. When the component is added back, plugin class -> object mapping already exists and the device tries to update the object by creating a new one and perform dependency injection. But this happens in two places and one of the places where the kernel tries to create and inject dependencies does not know how to create a Greengrass service instance. This causes the deployment to fail as the plugin component cannot be created.

Fix:
i. During component removal, clean up both the mapping in context that point to the plugin service object
ii. Fix the load component logic to create component /inject dependency once, when there is a existing mapping for the component.

Is the issue is not fixed, customer cannot deploy a plugin component, if the plugin has been previously removed from the device. Customer can recover by restarting the device and deploying the plugin.

 integration tests

**Checklist:**
 - [X ] Updated or added new integration tests


src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
```diff
@@ -375,6 +375,13 @@ public void removeObsoleteServices() throws InterruptedException, ExecutionExcep
...
             }
             servicesToRemove.forEach(serviceName -> {
+                try {
+                    GreengrassService eg = kernel.locate(serviceName);
+                    kernel.getContext().remove(eg.getClass());
+                } catch (ServiceLoadException e) {
+                    logger.atError(MERGE_ERROR_LOG_EVENT_KEY).setCause(e).addKeyValue(SERVICE_NAME_LOG_KEY, serviceName)
+                            .log("Could not locate Greengrass service to remove service");
+                }
                 kernel.getContext().remove(serviceName);
```
MikeDombo, line 385: this should return the item that it removed, right? We can just use that instead of going through locate

src/main/java/com/aws/greengrass/dependency/Context.java
```diff
@@ -171,8 +171,6 @@ public Value getvIfExists(Object tag) {
         parts.compute(clazz, (tagObj, originalValue) -> {
             if (originalValue == null) {
                 originalValue = value;
-            } else {
```
MikeDombo, line 174: what if the value isn't null, don't we want to replace the value?

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
```diff
@@ -336,7 +337,7 @@ public void replaceUnloadableService() throws ServiceLoadException {
                     kernel.getContext().remove(serviceName);
```
hui-yang, line 337: Also change here?

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
fahadmohammed01: remove was returning boolean, updated remove to return the value to avoid calling locate

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
fahadmohammed01: Nice catch, updated!

hui-yang (approved):

src/main/java/com/aws/greengrass/dependency/Context.java
fahadmohammed01: Ideally we would not overwrite a clazz -> object mapping. Replaced compute with put since that is what we are doing.

MikeDombo (approved):

PR #899, opened 2021-03-16
fix(mqtt): use protocol timeout 

MikeDombo:
**Issue #, if available:**

**Description of changes:**
Use the newly added MQTT protocol timeout. Also standardizing connection cleanup in the client connection class.

**Checklist:**
 - [x] If your code makes a remote network call, it was tested with a proxy

### DIFF
--- a/src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
+++ b/src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
@@ -86,8 +86,10 @@ class AwsIotMqttClient implements Closeable {
         @Override
         public void onConnectionInterrupted(int errorCode) {
             currentlyConnected.set(false);
-            // Error code 0 means that the disconnection was intentional, so we don't need to log it
-            if (errorCode != 0) {
+            // Error code 0 means that the disconnection was intentional
+            if (errorCode == 0) {
+                logger.atInfo().log("Connection interrupted");
+            } else {
                 logger.atWarn().kv("error", CRT.awsErrorString(errorCode)).log("Connection interrupted");
             }
             if (resubscribeFuture != null && !resubscribeFuture.isDone()) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
```diff
@@ -165,17 +165,17 @@ long getThrottlingWaitTimeMicros() {
...
                 return connection.publish(message, qos, retain);
             }
         });
     }
 
-    private void throwIfNotConnected() {
-        if (!connected()) {
-            throw new MqttException("Client is not connected");
+    private void throwIfNoConnection() {
+        if (connection == null) {
```
hui-yang, line 177: Why do we make this change?

MikeDombo: Because we do want to attempt to send even if we think we might be offline. We'll just get a future back which will be completed eventually. I saw this error thrown in a UAT but I cannot tell the reason for the error (since the connection could have been null or we thought we were disconnected). So I realized that even if we think we're not connected, there's no harm in creating the request anyway.

hui-yang (approved):

hui-yang (dismissed):

src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
```diff
@@ -239,10 +238,7 @@ void reconnect() throws TimeoutException, ExecutionException, InterruptedExcepti
                 if (error != null) {
                     // Must synchronize since we're messing with the shared connection object and this block
                     // is executed in some other thread
```
rbattle, line 240: can you move/update this comment since the synchronize is inside `connectionCleanup`?

src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
```diff
@@ -335,22 +331,29 @@ synchronized boolean connected() {
...
 
+    @SuppressWarnings("PMD.NullAssignment")
+    private void connectionCleanup() {
+        synchronized (this) {
+            if (connection != null) {
+                connection.close();
+                connection = null;
+            }
+        }
+        connectionEventCallback.onConnectionInterrupted(0);
```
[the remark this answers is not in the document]
MikeDombo: Just means that this is intentional and not a MQTT error

src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
MikeDombo: Updated

rbattle (approved):

youtuyy (approved):

hui-yang (approved):

PR #903, opened 2021-03-25
feat(windows): implement file ownership and permissions

fufranci:
**Issue #, if available:**

**Description of changes:**
* Refactor `setPermissions` so that it is common for all platforms
* However `setPermissions` calls two platform dependent methods: `setOwner` and `setMode`
* Implement `setOwner` and `setMode` for both Unix and Windows

File ownership and permissions were not implemented for Windows

Unit tests for both Unix and Windows platform 

**Checklist:**


src/main/java/com/aws/greengrass/util/FileSystemPermission.java
```diff
@@ -70,6 +82,90 @@
...
+                AclEntryPermission.WRITE_NAMED_ATTRS,
+                AclEntryPermission.WRITE_ATTRIBUTES,
+                AclEntryPermission.SYNCHRONIZE));
+        final Set<AclEntryPermission> executePerms = new HashSet<>(Arrays.asList(
+                AclEntryPermission.READ_DATA,
+                AclEntryPermission.READ_NAMED_ATTRS,
+                AclEntryPermission.EXECUTE,
+                AclEntryPermission.READ_ATTRIBUTES,
+                AclEntryPermission.READ_ACL,
+                AclEntryPermission.SYNCHRONIZE));
```
fufranci, line 113: Please review.

src/main/java/com/aws/greengrass/util/FileSystemPermission.java
```diff
@@ -70,6 +82,90 @@
...
+        }
+        if (ownerExecute) {
+            aclEntries.add(AclEntry.newBuilder()
+                    .setType(AclEntryType.ALLOW)
+                    .setPrincipal(ownerPrincipal)
+                    .setPermissions(executePerms)
+                    .build());
+        }
+
+        // There is no default group concept on Windows. (There is, but is used when mounting as a network share.)
```
fufranci, line 142: `groupRead`, `groupWrite`, `groupExecute` are not used. Windows does not natively has a default group concept. Although administrator can set a default group WHEN a filesystem is being used as a network share that appears as POSIX, but this does not apply here.

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsPlatform.java
```diff
@@ -121,9 +128,28 @@ public void addUserToGroup(String user, String group) throws IOException {
...
+        UserPrincipalLookupService lookupService = path.getFileSystem().getUserPrincipalLookupService();
+        UserPrincipal userPrincipal = lookupService.lookupPrincipalByName(permission.getOwnerUser());
+        FileOwnerAttributeView view = Files.getFileAttributeView(path, FileOwnerAttributeView.class,
+                LinkOption.NOFOLLOW_LINKS);
+
+        logger.atTrace().setEventType(SET_PERMISSIONS_EVENT).kv(PATH, path).kv("owner", permission.getOwnerUser())
+                .log();
+        view.setOwner(userPrincipal);
+
+        // Note that group ownership is not used.
```
fufranci, line 141: I don't see a file can be owned by a group in Windows.

src/main/java/com/aws/greengrass/util/FileSystemPermission.java
```diff
@@ -70,6 +82,90 @@
...
     }
 
+    /**
+     * Convert to a list of Acl entries for use with AclFileAttributeView.setAcl.
+     *
+     * @param path path to apply to
+     * @return List of Acl entries
+     * @throws IOException if any exception occurs while converting to Acl
+     */
+    public List<AclEntry> toAclEntries(Path path) throws IOException {
```
MikeDombo, line 92: move this to the windows implementation since this is entirely platform specific

src/main/java/com/aws/greengrass/util/FileSystemPermission.java
```diff
@@ -70,6 +82,90 @@
...
+        // These sets of permissions are reverse engineered by setting the permission on a file using Windows File
+        // Explorer. Then use AclFileAttributeView.getAcl to examine what were set.
+        final Set<AclEntryPermission> readPerms = new HashSet<>(Arrays.asList(
```
MikeDombo, line 95: save these as static fields instead of creating them on the fly

src/main/java/com/aws/greengrass/util/FileSystemPermission.java
```diff
@@ -70,6 +82,90 @@
...
+                AclEntryPermission.READ_DATA,
+                AclEntryPermission.READ_NAMED_ATTRS,
+                AclEntryPermission.READ_ATTRIBUTES,
+                AclEntryPermission.READ_ACL,
+                AclEntryPermission.SYNCHRONIZE));
+        final Set<AclEntryPermission> writePerms = new HashSet<>(Arrays.asList(
```
MikeDombo, line 101: Is this list of acls the entire list? I feel like windows had many many more options, including specific options for files vs. directories.

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsPlatform.java
```diff
@@ -121,9 +128,28 @@ public void addUserToGroup(String user, String group) throws IOException {
...
+    }
+
+    @Override
+    protected void setMode(FileSystemPermission permission, Path path) throws IOException {
+        List<AclEntry> acl = permission.toAclEntries(path);
+        logger.atTrace().setEventType(SET_PERMISSIONS_EVENT).kv(PATH, path).kv("perm", acl.toString()).log();
+        AclFileAttributeView view = Files.getFileAttributeView(path, AclFileAttributeView.class,
+                LinkOption.NOFOLLOW_LINKS);
+        List<AclEntry> existingAcl = view.getAcl();
```
MikeDombo, line 150: we need to rewrite the existing acl. For example, if someone allowed everyone to read their private key, then we want to get rid of that

src/main/java/com/aws/greengrass/util/FileSystemPermission.java
fufranci: This is not the entire list of available permissions. `writePerms` is the list of permissions that would make File Explorer UI (right click on the file->Properties->Security->username) shows the user has write permission.

The complete list of permission is here:
https://docs.oracle.com/javase/8/docs/api/java/nio/file/attribute/AclEntryPermission.html

See also:
https://docs.oracle.com/javase/8/docs/api/java/nio/file/attribute/AclEntry.html

MikeDombo: write needs DELETE and DELETE_CHILD

MikeDombo: what about WRITE_ACL and WRITE_OWNER?

fufranci: Added all four.

pom.xml
```diff
@@ -550,7 +550,7 @@
                                         <limit>
                                             <counter>INSTRUCTION</counter>
                                             <value>COVEREDRATIO</value>
-                                            <minimum>0.58</minimum>
+                                            <minimum>0.57</minimum>
```
fufranci, line 553: I can't keep 0.58. Welcome any unit test suggestion.

fufranci: Just realized this number is for integration test. Will take a look again.

src/main/java/com/aws/greengrass/util/platforms/unix/UnixPlatform.java
```diff
@@ -376,76 +369,73 @@ public void addUserToGroup(String user, String group) throws IOException {
...
+        if (groupPrincipal != null) {
+            logger.atTrace().setEventType(SET_PERMISSIONS_EVENT).kv(PATH, path).kv("group",
+                    permission.getOwnerGroup()).log();
+            view.setGroup(groupPrincipal);
         }
+    }
 
+    @Override
+    protected void setMode(FileSystemPermission permission, Path path) throws IOException {
+        Set<PosixFilePermission> perms = posixFilePermissions(permission);
```
rbattle, line 393: similar problem as with the owner - we now have to convert the permissions object for every file and allocate a new set everytime.

Throwing this out as a suggestion:

We can do something like have a `FileSystemPermissionView` abstract class which has concrete `Posix` and `Windows` implementations where we can precalculate any platform specific stuff.

Have a platform method like `Platform.getFileSystemPermissionView(FileSystemPermission perm)`

The `posixFilePermissions` method you have would just be a method on the `PosixFileSystemPermissionView`
You could calculate the permissions once when constructing the object and pass the `FileSystemPermissionView` into `setMode`, cast it as a the Posix version and then use those permissions.

src/main/java/com/aws/greengrass/util/platforms/unix/UnixPlatform.java
```diff
@@ -376,76 +369,73 @@ public void addUserToGroup(String user, String group) throws IOException {
...
+            ret.add(PosixFilePermission.OTHERS_READ);
+        }
+        if (permission.isOtherWrite()) {
+            ret.add(PosixFilePermission.OTHERS_WRITE);
+        }
+        if (permission.isOtherExecute()) {
+            ret.add(PosixFilePermission.OTHERS_EXECUTE);
+        }
+
+        return ret;
```
rbattle, line 438: you could convert this to an EnumSet

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsPlatform.java
```diff
@@ -121,9 +158,100 @@ public void addUserToGroup(String user, String group) throws IOException {
...
+
+    /**
+     * Convert to a list of Acl entries for use with AclFileAttributeView.setAcl.
+     *
+     * @param permission permission to convert
+     * @param path path to apply to
+     * @return List of Acl entries
+     * @throws IOException if any exception occurs while converting to Acl
+     */
+    public static List<AclEntry> aclEntries(FileSystemPermission permission, Path path) throws IOException {
```
MikeDombo, line 191: does this need to be public?

src/test/java/com/aws/greengrass/testcommons/testutilities/PlatformTestUtils.java
```diff
@@ -0,0 +1,39 @@
...
+            return INSTANCE;
+        }
+
+        if (PlatformResolver.isWindows) {
+            INSTANCE = new WindowsPlatformTestUtils();
+        } else {
+            INSTANCE = new UnixPlatformTestUtils();
+        }
+
+        logger.atInfo().log("Getting platform test utils instance {}.", INSTANCE.getClass().getName());
```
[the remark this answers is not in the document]
fufranci: Changed to trace and removed period.

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsPlatform.java
fufranci: Yeah, it is also being used by WindowsPlatformTestUtils#hasPermission  which is used by our custom matcher.

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsPlatform.java
```diff
@@ -121,9 +158,100 @@ public void addUserToGroup(String user, String group) throws IOException {
...
+            // On Linux, when we set the file permission for the owner, it applies to the current owner and we don't
+            // need to know who the actual owner is. But on Windows, Acl must be associated with an owner.
+            AclFileAttributeView view = Files.getFileAttributeView(path, AclFileAttributeView.class,
+                    LinkOption.NOFOLLOW_LINKS);
+            ownerPrincipal = view.getOwner();
+        } else {
+            ownerPrincipal = userPrincipalLookupService.lookupPrincipalByName(permission.getOwnerUser());
+        }
+
+        GroupPrincipal everyone = userPrincipalLookupService.lookupPrincipalByGroupName("Everyone");
```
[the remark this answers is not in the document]
fufranci: `lookupPrincipalByGroupName` could throw `IOException`, which will bubble up to higher level logic to handle. Also note that `Everyone` is a well know group which should be present on all Windows.

src/test/java/com/aws/greengrass/testcommons/testutilities/unix/UnixPlatformTestUtils.java
```diff
@@ -0,0 +1,47 @@
...
+        if (view == null) {
+            logger.atError().log("Platform does not have PosixFileAttributeView");
+            return false;
+        }
+
+        try {
+            Set<PosixFilePermission> perms = UnixPlatform.posixFilePermissions(expected);
+            Set<PosixFilePermission> actual = view.readAttributes().permissions();
+            logger.atTrace().log("posix permissions are {} for path {}", PosixFilePermissions.toString(actual), path);
+            return actual.containsAll(perms) && perms.containsAll(actual);
```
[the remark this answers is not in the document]
fufranci: This method is used by our customer matcher (com.aws.greengrass.testcommons.testutilities.Matchers).

src/test/java/com/aws/greengrass/testcommons/testutilities/windows/WindowsPlatformTestUtils.java
```diff
@@ -0,0 +1,46 @@
...
+        if (view == null) {
+            logger.atError().log("Platform does not have AclFileAttributeView");
+            return false;
+        }
+
+        try {
+            List<AclEntry> perms = WindowsPlatform.aclEntries(expected, path);
+            List<AclEntry> actual = view.getAcl();
+            logger.atTrace().log("Acl entries are {} for path {}", actual.toString(), path);
+            return actual.containsAll(perms) && perms.containsAll(actual);
```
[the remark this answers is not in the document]
fufranci: Same here. It is used by our custom matcher.

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsPlatform.java
```diff
@@ -121,9 +158,100 @@ public void addUserToGroup(String user, String group) throws IOException {
...
+
+        UserPrincipal ownerPrincipal;
+        if (permission.getOwnerUser() == null) {
+            // On Linux, when we set the file permission for the owner, it applies to the current owner and we don't
+            // need to know who the actual owner is. But on Windows, Acl must be associated with an owner.
+            AclFileAttributeView view = Files.getFileAttributeView(path, AclFileAttributeView.class,
+                    LinkOption.NOFOLLOW_LINKS);
+            ownerPrincipal = view.getOwner();
+        } else {
+            ownerPrincipal = userPrincipalLookupService.lookupPrincipalByName(permission.getOwnerUser());
```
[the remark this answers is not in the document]
fufranci: `IOException` will be thrown. It is a checked exception. It will bubble up to a higher level to handle. For example, [here](https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/main/src/main/java/com/aws/greengrass/lifecyclemanager/KernelAlternatives.java#L150-L151).

src/main/java/com/aws/greengrass/util/platforms/unix/UnixPlatform.java
```diff
@@ -376,75 +370,89 @@ public void addUserToGroup(String user, String group) throws IOException {
...
-                    return FileVisitResult.CONTINUE;
-                }
+        /**
+         * Convert to a set of PosixFilePermissions for use with Files.setPosixFilePermissions.
+         *
+         * @param permission Permission to convert
+         * @return Set of permissions
+         */
+        public static Set<PosixFilePermission> posixFilePermissions(FileSystemPermission permission) {
+            Set<PosixFilePermission> ret = new HashSet<>();
```
MikeDombo, line 406: I think Rob mean to turn this into an EnumSet instead of HashSet, can you do that instead of the copy stuff at the end?

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsPlatform.java
MikeDombo: it's not really about ownership, but permissions. The group should have some permissions on the file/folder

src/main/java/com/aws/greengrass/util/platforms/unix/UnixPlatform.java
fufranci: My bad. I understood what Rob meant but misunderstood how to use EnumSet. I will fix it in the next commit.

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsPlatform.java
fufranci: The problem is which group? There is no default group AFAIK.

src/main/java/com/aws/greengrass/util/platforms/unix/UnixPlatform.java
```diff
@@ -376,76 +369,73 @@ public void addUserToGroup(String user, String group) throws IOException {
...
+        GroupPrincipal groupPrincipal = Utils.isEmpty(permission.getOwnerGroup()) ? null :
+                lookupService.lookupPrincipalByGroupName(permission.getOwnerGroup());
+        PosixFileAttributeView view = Files.getFileAttributeView(path, PosixFileAttributeView.class,
+                LinkOption.NOFOLLOW_LINKS);
 
-        // noop function that does not set owner
-        CrashableFunction<PosixFileAttributeView, Void, IOException> setOwner = (p) -> null;
+        logger.atTrace().setEventType(SET_PERMISSIONS_EVENT).kv(PATH, path).kv("owner", permission.getOwnerUser())
+                .log();
+        view.setOwner(userPrincipal);
```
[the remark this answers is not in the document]
fufranci: Done

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsPlatform.java
fufranci: Sorry, I was confused with `groupRead`, etc which applies to the default group and there is no default group in Windows. 

Back to your feedback: Would you recommend using the passed in group and allow such group to have some basic read/write permission?

MikeDombo: No, it should have the group permissions that are being requested through setPermission. If the API isn't working for you, can you refactor it so that it does work? 

I believe the api used to just simply be setPermission which took a permission, user, and group.

fufranci: Make sense. I think I can add the permission accordingly [here](https://github.com/aws-greengrass/aws-greengrass-nucleus/pull/903/files#diff-1848e4b7f214f13eb0232bdb3c0d8de3b38e7e84b88b690941d7c4c991392d7fR233).

fufranci: Updated. Please review. Thanks for pointing this out!

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsPlatform.java
```diff
@@ -121,9 +159,153 @@ public void addUserToGroup(String user, String group) throws IOException {
...
+            }
+            if (permission.isOwnerExecute()) {
+                aclEntries.add(AclEntry.newBuilder()
+                        .setType(AclEntryType.ALLOW)
+                        .setPrincipal(ownerPrincipal)
+                        .setPermissions(EXECUTE_PERMS)
+                        .build());
+            }
+
+            // Group
```
MikeDombo, line 232: nice!

MikeDombo (approved):

rbattle (approved):
Nice - thanks for making those updates - this looks great

PR #908, opened 2021-04-06
fix: timestamp of initial jvmOptions; bootstrap config comparison

hui-yang:
**Issue #, if available:**

**Description of changes:**
1. Update timestamp of initial `jvmOptions` config.
During deployments, default config values from recipes are loaded with timestamp `1` [here](https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/003f19ee79627f5a0e281beee285c26aaa311ef6/src/main/java/com/aws/greengrass/componentmanager/KernelConfigResolver.java#L310-L315), and during the config merge, config values with the same timestamp will override current value in [withNewerValue](https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/8a9bd022c7a1da82012eec4a006839d7b7e36f70/src/main/java/com/aws/greengrass/config/Topic.java#L210-L213).
As a result, `jvmOptions` will be reset to default value from recipe, i.e. `""`, incorrectly during deployments.
The timestamp of initial `jvmOptions` should not be set to 1, because it's not the default in recipe. 
Bumping timestamp in this PR.
2. Fix bootstrap step comparison logic to handle config keys in a case-insensitive manner and handle config values as strings. e.g.
```
before. {RequiresPrivilege=true, script=...} # true is of type string
after. {requiresPrivilege=true, script=...}  # true is of type boolean
```

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
+++ b/src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
@@ -275,7 +275,7 @@ public class DeviceConfiguration {
             String jvmOptions = ManagementFactory.getRuntimeMXBean().getInputArguments().stream().sorted()
                     .filter(s -> !s.startsWith(JVM_OPTION_ROOT_PATH)).collect(Collectors.joining(" "));
             kernel.getConfig().lookup(SERVICES_NAMESPACE_TOPIC, getNucleusComponentName(), CONFIGURATION_CONFIG_KEY,
-                    DEVICE_PARAM_JVM_OPTIONS).dflt(jvmOptions);
+                    DEVICE_PARAM_JVM_OPTIONS).withNewerValue(DEFAULT_VALUE_TIMESTAMP + 1, jvmOptions);
 
             kernelAlts.writeLaunchParamsToFile(jvmOptions);
             logger.atInfo().log("Successfully setup Nucleus launch parameters");
--- a/src/main/java/com/aws/greengrass/lifecyclemanager/GenericExternalService.java
+++ b/src/main/java/com/aws/greengrass/lifecyclemanager/GenericExternalService.java
@@ -238,11 +238,43 @@ public class GenericExternalService extends GreengrassService {
         }
         Node serviceOldBootstrap = getConfig().findNode(SERVICE_LIFECYCLE_NAMESPACE_TOPIC,
                 Lifecycle.LIFECYCLE_BOOTSTRAP_NAMESPACE_TOPIC);
-        boolean bootstrapStepChanged =  serviceOldBootstrap == null
-                || !serviceOldBootstrap.toPOJO().equals(newServiceLifecycle.get(lifecycleKey));
-        logger.atDebug().log(String.format("Bootstrap is %srequired: bootstrap step %schanged",
-                bootstrapStepChanged ? "" : "not ", bootstrapStepChanged ? "" : "un"));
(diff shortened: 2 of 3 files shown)


MikeDombo (approved):

MikeDombo (approved):

src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
```diff
@@ -275,7 +275,7 @@ void persistInitialLaunchParams(KernelAlternatives kernelAlts) {
             String jvmOptions = ManagementFactory.getRuntimeMXBean().getInputArguments().stream().sorted()
                     .filter(s -> !s.startsWith(JVM_OPTION_ROOT_PATH)).collect(Collectors.joining(" "));
             kernel.getConfig().lookup(SERVICES_NAMESPACE_TOPIC, getNucleusComponentName(), CONFIGURATION_CONFIG_KEY,
-                    DEVICE_PARAM_JVM_OPTIONS).dflt(jvmOptions);
+                    DEVICE_PARAM_JVM_OPTIONS).withNewerValue(DEFAULT_VALUE_TIMESTAMP + 1, jvmOptions);
```
fengwa-aws, line 278: `DEFAULT_VALUE_TIMESTAMP` is 1, right? so `DEFAULT_VALUE_TIMESTAMP + 1` is 2? Why use 2?

src/main/java/com/aws/greengrass/lifecyclemanager/GenericExternalService.java
```diff
@@ -238,13 +238,45 @@ public boolean isBootstrapRequired(Map<String, Object> newServiceConfig) {
...
+                    .kv("after", newServiceLifecycle.get(lifecycleKey))
+                    .log("Bootstrap is required: bootstrap step changed");
+        } else {
+            logger.atDebug().log("Bootstrap is not required: bootstrap step unchanged");
+        }
         return bootstrapStepChanged;
     }
 
+    private boolean serviceLifecycleBootstrapEquals(Object before, Object after) {
+        if (!(before instanceof Map) || !(after instanceof Map)) {
```
fengwa-aws, line 254: should be `&&`, right?

fengwa-aws: What I meant is, we also need to check the situation when one is map, and the other is not a map.

hui-yang: When one config value is map, and the other is not a map, they definitely are not equal and I think it should be covered with `Objects.equal`

src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
hui-yang: I could also use the current timestamp.
But thinking of the scenario of adding thing to existing thinggroup (ts2) with an active deployment, if the deployment changes jvmOptions (ts1), ts1<ts2, should the desired behavior be thing picking up the updated jvmOptions from the deployment? If yes, we cannot use the current timestamp.

hui-yang: Also I just thought of another case, if nucleus reads in a config.yaml file (with file last modified timestamp), the file can also override jvmOptions if needed.

fengwa-aws (approved):

PR #913, opened 2021-04-07
fix: read procfs (via oshi) instead of ps to find process tree

### LINKED ISSUE #910: getChildPids does not work on Busybox based Linuxes
The  function getChildPids uses the ps command and flags not supported by Busybox

        Process proc = Runtime.getRuntime().exec(new String[]{"ps", "-ax", "-o", "pid,ppid"});

https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/87d89a61342f20e2e84f505af38bd89f68514c2f/src/main/java/com/aws/greengrass/util/platforms/unix/UnixPlatform.java#L489

Busyboxy supports -o and -T and that is it,

cf.  https://busybox.net/downloads/BusyBox.html
cf. https://git.busybox.net/busybox/tree/procps/ps.posix

One way to get the list of all the child processes for a given process that works on on Linux systems is to readdir on 
/proc/{pid}/task/ to get he list of child processes.  You can check the parent process for each by reading the 4th field
of the stat file  /proc/{pid}/task/{tid}/stat.

The process group id is also available in the 5th field.  (In the case of my running core there are 37 child processes in this group for example).

MikeDombo:
**Issue #, if available:**

**Description of changes:**
This change updates the `UnixPlatform` implementation for finding child processes so that it does not find processes using `ps`, instead we will use the oshi library that we already depend on to list all child processes.

Resolves #910.

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/util/platforms/unix/UnixPlatform.java
+++ b/src/main/java/com/aws/greengrass/util/platforms/unix/UnixPlatform.java
@@ -8,7 +8,6 @@ package com.aws.greengrass.util.platforms.unix;
 import com.aws.greengrass.logging.api.LogEventBuilder;
 import com.aws.greengrass.util.Exec;
 import com.aws.greengrass.util.FileSystemPermission;
-import com.aws.greengrass.util.Pair;
 import com.aws.greengrass.util.Permissions;
 import com.aws.greengrass.util.Utils;
 import com.aws.greengrass.util.platforms.Platform;
@@ -21,11 +20,12 @@ import lombok.Getter;
 import lombok.NoArgsConstructor;
 import org.zeroturnaround.process.PidProcess;
 import org.zeroturnaround.process.Processes;
+import oshi.SystemInfo;
+import oshi.software.os.OSProcess;
+import oshi.software.os.OperatingSystem;
 
 import java.io.BufferedReader;
 import java.io.IOException;
-import java.io.InputStreamReader;
-import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.LinkOption;
 import java.nio.file.Path;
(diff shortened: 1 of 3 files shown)


src/main/java/com/aws/greengrass/util/platforms/unix/UnixPlatform.java
```diff
@@ -481,45 +481,37 @@ protected void runCmd(String cmdStr, Consumer<CharSequence> out, String msg)
...
+                .filter(File::isDirectory)) {
+            // For each task, find its children
+            for (File task : tasks.collect(Collectors.toSet())) {
+                String children = new String(Files.readAllBytes(task.toPath().resolve("children")),
+                        StandardCharsets.UTF_8);
+                Set<Integer> childPid =
+                        Arrays.stream(children.split("\\s+"))
+                                .filter(s -> !s.isEmpty())
+                                .map(Integer::parseInt)
+                                .collect(Collectors.toSet());
```
philcali, line 500: I'm doing some reading on this, and this interface is documented as being unreliable because this is meant for CRIU use-cases, specifically: https://man7.org/linux/man-pages/man5/proc.5.html

```
/proc/[pid]/task/[tid]/children (since Linux 3.5)
              A space-separated list of child tasks of this task.  Each
              child task is represented by its TID.

              This option is intended for use by the checkpoint-restore
              (CRIU) system, and reliably provides a list of children
              only if all of the child processes are stopped or frozen.
              It does not work properly if children of the target task
              exit while the file is being read!  Exiting children may
              cause non-exiting children to be omitted from the list.
              This makes this interface even more unreliable than
              classic PID-based approaches if the inspected task and its
              children aren't frozen, and most code should probably not
              use this interface.
```

I'm noticing on my system, the `children` file does not even exist. I suspect we'll need a fallback strategy.

indougnito (approved):

rbattle (dismissed):

rbattle (approved):

indougnito (approved):

PR #916, opened 2021-04-08
fix(tlog): close tlog writer to prevent fd leak

fahadmohammed01:
**Issue #, if available:**
File descriptor leak when rotating tlog file.  

Tested on dev box. Tested by reducing the frequency at which tlogs are rotated. Without the fix, the kernel FD count goes up by 1 after each rotation. 

 The fix can also be verified by checking the metrics in deployment performance tests.

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/config/ConfigurationWriter.java
+++ b/src/main/java/com/aws/greengrass/config/ConfigurationWriter.java
@@ -210,6 +210,7 @@ public class ConfigurationWriter implements Closeable, ChildChanged {
         logger.atDebug(TRUNCATE_TLOG_EVENT).log("existing tlog writer closed");
         // move old tlog
         try {
+            out.close();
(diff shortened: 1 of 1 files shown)


MikeDombo (dismissed):

fengwa-aws (dismissed):

src/main/java/com/aws/greengrass/config/ConfigurationWriter.java
```diff
@@ -210,6 +210,7 @@ private synchronized void truncateTlog() {
         logger.atDebug(TRUNCATE_TLOG_EVENT).log("existing tlog writer closed");
```
tilo-chen, line 210: This log should be after the old writer is actually closed.

fahadmohammed01: moved after close()

tilo-chen (approved):

MikeDombo (approved):

PR #926, opened 2021-04-15
[WIP/DNM] feat: add crypto interface to support non-file-based crypto

MikeDombo:
**Issue #, if available:**

**Description of changes:**

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/crypto/CryptoI.java
+++ b/src/main/java/com/aws/greengrass/crypto/CryptoI.java
@@ -1,0 +1,138 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.aws.greengrass.crypto;
+
+import com.aws.greengrass.config.Subscriber;
+import com.aws.greengrass.config.WhatHappened;
+import com.aws.greengrass.deployment.DeviceConfiguration;
+import com.aws.greengrass.util.Coerce;
+import com.aws.greengrass.util.EncryptionUtils;
+import com.aws.greengrass.util.exceptions.TLSAuthException;
+import software.amazon.awssdk.iot.AwsIotMqttConnectionBuilder;
+
+import java.io.IOException;
+import java.security.GeneralSecurityException;
+import java.security.KeyStore;
+import java.security.PrivateKey;
+import java.security.cert.Certificate;
+import java.security.cert.X509Certificate;
+import java.util.Arrays;
+import java.util.List;
+import javax.inject.Inject;
+import javax.net.ssl.KeyManagerFactory;
+import javax.net.ssl.TrustManagerFactory;
+import javax.net.ssl.X509KeyManager;
+import javax.net.ssl.X509TrustManager;
+import javax.security.auth.x500.X500Principal;
+
+public interface CryptoI extends AutoCloseable {
+    AwsIotMqttConnectionBuilder getIotMqttConnectionBuilder();
+
+    X509KeyManager[] getKeyManagers() throws TLSAuthException;
+
+    X509TrustManager[] getTrustManagers() throws TLSAuthException;
+
+    void setImplementationUpdated(Runnable implChanged);
+
+    class Default implements CryptoI {
+        protected final String certificatePath;
+        protected final String privateKeyPath;
+        protected final String rootCAPath;
+        protected X509KeyManager[] keyManagers = null;
+        protected X509TrustManager[] trustManagers = null;
+        protected Runnable implChanged;
+        @SuppressWarnings("PMD.NullAssignment")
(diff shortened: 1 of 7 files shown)


src/main/java/com/aws/greengrass/crypto/CryptoI.java
```diff
@@ -0,0 +1,141 @@
...
+                }
+                providerChanged.run();
+            }
+        };
+
+        @Inject
+        public Default(DeviceConfiguration deviceConfiguration) {
+            certificatePath = Coerce.toString(deviceConfiguration.getCertificateFilePath().subscribe(subscriber));
+            privateKeyPath = Coerce.toString(deviceConfiguration.getPrivateKeyFilePath().subscribe(subscriber));
+            rootCAPath = Coerce.toString(deviceConfiguration.getRootCAFilePath().subscribe(subscriber));
```
philcali, line 61: How does the subscription trigger work? Will the subscriber unset the managers and run the callback three times or once if the a deployment updated these configuration values?

philcali: Also, why does it only unset the managers if a callback is provided? Isn't that optional?

MikeDombo: No, not optional it should always be set.

Yes it would run 3 times. We can think about some deduping logic.

philcali: > No, not optional it should always be set.

It doesn't appeared to be a required part of the contract. Why is a `providerChanged` runnable required?

src/main/java/com/aws/greengrass/crypto/CryptoI.java
```diff
@@ -0,0 +1,141 @@
...
+import javax.security.auth.x500.X500Principal;
+
+public interface CryptoI {
+    AwsIotMqttConnectionBuilder getIotMqttConnectionBuilder();
+
+    X509KeyManager[] getKeyManagers() throws TLSAuthException;
+
+    X509TrustManager[] getTrustManagers() throws TLSAuthException;
+
+    void setProviderChanged(Runnable providerChanged);
```
philcali, line 37: If we expect implementations to handle when a provider changes, can't this action be correlated? For example: `cleanup` which is invoked by the provider when the crypto implementation is updated?

src/main/java/com/aws/greengrass/crypto/CryptoI.java
MikeDombo: So that the implementation can tell the CryptoProvider that it changed, and thus the provider needs to notify all the users that they need to reconfigure themselves. I can't inject the cryptoprovider directly because it would be a circular dependency.

src/main/java/com/aws/greengrass/crypto/CryptoI.java
MikeDombo: The implementations are only expected to notify the CryptoProvider that _they_ (the implementation) has changed and then the CryptoProvider will notify the users.

MikeDombo: Probably needs to be renamed since I see why you're confused.

philcali (dismissed):

src/main/java/com/aws/greengrass/crypto/CryptoI.java
```diff
@@ -0,0 +1,138 @@
...
+        @Override
+        public synchronized X509TrustManager[] getTrustManagers() throws TLSAuthException {
+            if (trustManagers == null) {
+                trustManagers = createTrustManagers(rootCAPath);
+            }
+            return trustManagers;
+        }
+
+        @Override
+        public void setImplementationUpdated(Runnable implChanged) {
```
fengwa-aws, line 89: What's the use case for this method?

MikeDombo: Allows the implementation to tell the CryptoProvider that something changed in order to trigger the onChange callbacks. Such as the certificate paths updating.

src/main/java/com/aws/greengrass/crypto/CryptoI.java
```diff
@@ -0,0 +1,138 @@
...
+        }
+
+        @Override
+        public AwsIotMqttConnectionBuilder getIotMqttConnectionBuilder() {
+            return AwsIotMqttConnectionBuilder.newMtlsBuilderFromPath(certificatePath, privateKeyPath)
+                    .withCertificateAuthorityFromPath(null, rootCAPath);
+        }
+
+        @Override
+        public synchronized X509KeyManager[] getKeyManagers() throws TLSAuthException {
```
fengwa-aws, line 73: Nit: return a list instead of array?

MikeDombo: All consumers expect an array

PR #931, opened 2021-04-17
perf: use interned strings for MQTT topics for lower memory usage

MikeDombo:
**Issue #, if available:**

**Description of changes:**
Use [String.intern()](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/String.html#intern()) in order to save strings into Java's string pool to avoid string duplication in memory. For MQTT messages with the same topic, which is a fairly common use case, we'd normally end up using quite a bit more memory than just the payload size due to duplicating the topic names.

Additionally, using `String.length()` instead of getting the bytes since getting the bytes requires actually re-encoding the value which is expensive and unnecessary.

**Checklist:**


fengwa-aws (approved):
Nice!
Also add a comment in the code why we use `intern()`?

philcali (approved):

fengwa-aws (approved):

PR #932, opened 2021-04-19
fix(windows): fix IPCEventStreamServiceTest

fufranci:
**Issue #, if available:**

**Description of changes:**
`IPCEventStreamServiceTest` is failing on Windows. This PR fixes it.

Observe in Github action that the test is passing on both Linux and Windows hosts.

Windows result is [here](https://github.com/aws-greengrass/aws-greengrass-nucleus/pull/932/checks?check_run_id=2394782225#step:6:5598).

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/ipc/IPCEventStreamService.java
+++ b/src/main/java/com/aws/greengrass/ipc/IPCEventStreamService.java
@@ -17,7 +17,6 @@ import com.aws.greengrass.util.platforms.Platform;
 import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import lombok.AccessLevel;
-import lombok.NoArgsConstructor;
 import lombok.Setter;
 import software.amazon.awssdk.aws.greengrass.GreengrassCoreIPCService;
 import software.amazon.awssdk.aws.greengrass.GreengrassCoreIPCServiceModel;
@@ -37,7 +36,7 @@ import javax.inject.Inject;
 
 import static com.aws.greengrass.lifecyclemanager.GreengrassService.SETENV_CONFIG_NAMESPACE;
 
-@NoArgsConstructor
+//@NoArgsConstructor
 public class IPCEventStreamService implements Startable, Closeable {
     public static final long DEFAULT_STREAM_MESSAGE_TIMEOUT_SECONDS = 5;
     public static final int DEFAULT_PORT_NUMBER = 8033;
@@ -54,17 +53,17 @@ public class IPCEventStreamService implements Startable, Closeable {
     private RpcServer rpcServer;
 
     @Inject
-    private Kernel kernel;
+    private final Kernel kernel;
 
-    @Inject
-    private GreengrassCoreIPCService greengrassCoreIPCService;
+    //@Inject
+    private final GreengrassCoreIPCService greengrassCoreIPCService;
 
     @Inject
     @Setter(AccessLevel.PACKAGE)
     private AuthenticationHandler authenticationHandler;
 
     @Inject
-    private Configuration config;
+    private final Configuration config;
 
     private SocketOptions socketOptions;
     private EventLoopGroup eventLoopGroup;
--- a/src/main/java/com/aws/greengrass/ipc/IPCEventStreamService.java
+++ b/src/main/java/com/aws/greengrass/ipc/IPCEventStreamService.java
@@ -49,20 +49,18 @@ public class IPCEventStreamService implements Startable, Closeable {
 
     private RpcServer rpcServer;
 
-    @Inject
     private final Kernel kernel;
 
     private final GreengrassCoreIPCService greengrassCoreIPCService;
 
-    @Inject
     private final AuthenticationHandler authenticationHandler;
 
-    @Inject
     private final Configuration config;
 
     private SocketOptions socketOptions;
     private EventLoopGroup eventLoopGroup;
 
+    @Inject
     IPCEventStreamService(Kernel kernel,
                           GreengrassCoreIPCService greengrassCoreIPCService,
                           Configuration config,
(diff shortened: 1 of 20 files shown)


src/main/java/com/aws/greengrass/util/platforms/windows/WindowsPlatform.java
```diff
@@ -366,17 +365,17 @@ public ShellDecorator withShell(String shell) {
 
     @Override
     public String prepareIpcFilepath(Path rootPath) {
-        return NAMED_PIPE;
+        return NAMED_PIPE + rootPath.toAbsolutePath();
```
fufranci, line 368: This change is so that a unit test can for sure connect to the intended IPC server that has the mock setup. Without this change, unit test may connect to a different IPC server that is started by a different test, and for whatever reason, still running.

This change is likely the only one that is absolutely necessary to make the test to pass.

src/main/java/com/aws/greengrass/ipc/IPCEventStreamService.java
```diff
@@ -51,27 +49,26 @@
...
     private SocketOptions socketOptions;
     private EventLoopGroup eventLoopGroup;
 
+    @Inject
     IPCEventStreamService(Kernel kernel,
-                                 GreengrassCoreIPCService greengrassCoreIPCService,
-                                 Configuration config) {
+                          GreengrassCoreIPCService greengrassCoreIPCService,
+                          Configuration config,
+                          AuthenticationHandler authenticationHandler) {
```
fufranci, line 67: This changes to constructor injection. It is not really necessary to make the test to pass. But I think constructor injection is a better pattern.

src/test/java/com/aws/greengrass/ipc/IPCEventStreamServiceTest.java
```diff
@@ -75,35 +69,19 @@
...
-        };
-        when(mockAuthenticationHandler.apply(any(), any())).thenReturn(authenticationData);
-        when(mockAuthorizationHandler.apply(eq(authenticationData))).thenReturn(Authorization.ACCEPT);
-        when(greengrassCoreIPCService.getAuthenticationHandler()).thenReturn(mockAuthenticationHandler);
-        when(greengrassCoreIPCService.getAuthorizationHandler()).thenReturn(mockAuthorizationHandler);
-        when(greengrassCoreIPCService.getServiceModel()).thenReturn(GreengrassCoreIPCServiceModel.getInstance());
-
-        ipcEventStreamService = new IPCEventStreamService(mockKernel, greengrassCoreIPCService, config);
-        NucleusPaths nucleusPaths = mock(NucleusPaths.class);
+    public void setup() throws UnauthenticatedException {
```
fufranci, line 75: I updated the test to setup the mock according to how `IPCEventStreamService` is currently implemented.

MikeDombo:
Can you fix the windows workflow so that it fails the workflow when a test fails? We don't want false confidence that tests are passing.

MikeDombo (approved):

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsPlatform.java
shaguptashaikh: Can this path end up being longer than the max pipe name length? We've faced similar issues with the socket file paths in UATs earlier. Could making the named pipe non static and initializing with a UUID appended to the static prefix also work instead?

fufranci: Good call. I found that the max length of a named pipe is 256. If the the prefix+rootPath is <= 256, I will use that string. Otherwise, I will use the prefix+UUID. The UUID needs to be static because there are a few places that need the same UUID.

fufranci:
> Can you fix the windows workflow so that it fails the workflow when a test fails? We don't want false confidence that tests are passing.

@MikeDombo, I tried removing `continue-on-error: true` from the `Build with Maven (Windows)` step. But the linux build would get cancelled if the Windows build fails. We don't want to do that *yet*. Do you know if there is another way?

MikeDombo:
> > Can you fix the windows workflow so that it fails the workflow when a test fails? We don't want false confidence that tests are passing.
> 
> @MikeDombo, I tried removing `continue-on-error: true` from the `Build with Maven (Windows)` step. But the linux build would get cancelled if the Windows build fails. We don't want to do that _yet_. Do you know if there is another way?

No, ok that's fine.

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsPlatform.java
shaguptashaikh: This should work, although sounds a bit unnecessary to have two code paths. I was suggesting we don't use the rootpath at all, instead could have a default constructor for WindowsPlatform, which would initialize a namedPipe member as `NAMED_PIPE + UUID.randomUUID().toString()` and whenever you have to get the file path you return that so it will be consistent everywhere, since a platform instance gets created only once, but that's to your judgement

shaguptashaikh (dismissed):

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsPlatform.java
```diff
@@ -366,17 +368,22 @@ public ShellDecorator withShell(String shell) {
 
     @Override
     public String prepareIpcFilepath(Path rootPath) {
-        return NAMED_PIPE;
+        String absolutionPath = rootPath.toAbsolutePath().toString();
```
MikeDombo, line 371: absolutePath, not absolution

fufranci: Thanks!

MikeDombo (approved):

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsPlatform.java
fufranci: The problem I am trying to fix is more related to running tests: 
1. The name must be unique to one test only. Otherwise, the test could be connecting to an orphan IPC server.
2. The name must be consistent within the test.

Using UUID as a static suffix cannot solve problem 1. Using a new UUID every time cannot solve problem 2. Using the `rootPath` as the suffix is ideal that it solves both problems. If `rootPath` is too long, I am hoping it is outside the scope of running tests. And therefore static UUID would do.

shaguptashaikh (approved):

abanthiy (approved):

PR #933, opened 2021-04-19
feat: add bootstrap.tlog as a fallback when loading the main tlog fails

MikeDombo:
**Issue #, if available:**

**Description of changes:**
We have been seeing that very rarely the transaction log gets corrupted when a device has a power outage. This change does not fix the corruption since it is happening at the hardware level and there really isn't anything that we can do about it right now. Instead, this change writes out a `bootstrap.tlog` file to be used when the main `config.tlog` cannot be loaded. The bootstrap file is always read before the config, since it is only written once, its timestamps will always be older than those in the main tlog, thus any changes in the main tlog will overwrite it. However, if the main tlog is broken, then there's something to fallback onto.

**Checklist:**


src/main/java/com/aws/greengrass/util/SwappingFileWriter.java
```diff
@@ -0,0 +1,118 @@
...
+        Files.deleteIfExists(writeFilePath);
+        // Sync read and write files if the read file exists
+        if (Files.exists(readFilePath)) {
+            Files.copy(readFilePath, writeFilePath, StandardCopyOption.REPLACE_EXISTING);
+        }
+
+        // Now we know that the read and write files are exactly the same so we can safely begin to simply append
+        // to them without any additional, larger, copies.
+
+        readFile = Files.newBufferedWriter(readFilePath, StandardOpenOption.CREATE,
```
MikeDombo, line 53: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

**Problem**
This line of code assigns a resource that might leak to a class field, causing a resource leak that slows down or crashes your system. For example, when Files.newBufferedWriter throws an exception, the *SwappingFileWriter* object isn’t constructed so you can’t close this resource afterward.

**Fix**
Ensure that if an exception is thrown, the *readFile* resource is closed before the constructor is exited.

**More info**
[View resource management guidelines at oracle.com](https://www.oracle.com/technetwork/articles/java/trywithresources-401775.html) (external link).

src/main/java/com/aws/greengrass/easysetup/GreengrassSetup.java
```diff
@@ -429,6 +429,11 @@ void provision(Kernel kernel) throws IOException, DeviceConfigurationException {
                     kernel.getContext().get(DeviceConfiguration.class).getNucleusVersion());
         }
 
+        // Dump config since we've just provisioned so that the bootstrap config will enable us to
+        // reach the cloud when needed. Must do this now because we normally would never overwrite the bootstrap
+        // file, however we need to do it since we've only just learned about our endpoints, certs, etc.
+        kernel.writeEffectiveConfigAsTransactionLog(kernel.getNucleusPaths().configPath()
```
shaguptashaikh, line 435: Should this go into Kernel./KernelLifecycle so that even if the config crucial for connecting to cloud changes via deployment, we will have the right set to fallback to on restarting after a crash?

MikeDombo: No, we decided that we never want to write to the file, at least initially. So we will not update it in response to a deployment.

shaguptashaikh: So if the config does get changed through deployment then this issue will still be a problem right? Because it's possible that the bootstrap version wouldn't work for connecting the device to cloud anymore

MikeDombo: Potentially yes, we did think about that and we're ok with it for now.

fengwa-aws (commented):
It's a very clean and smart implement. However, I have two concerns regarding consistency and user experience:
1. Depending on which entry of the tlog is corrupted, it's very likely the Greengrass nucleus reads in a mixture of the latest tlog and the default bootstrap tlog. Would this inconsistency cause issues or at minimum confusions?
2. Do we have enough logging to show that a tlog entry (or multiple entries) are corrupted and therefore the default are being used?

MikeDombo:
> 1. Depending on which entry of the tlog is corrupted, it's very likely the Greengrass nucleus reads in a mixture of the latest tlog and the default bootstrap tlog. Would this inconsistency cause issues or at minimum confusions?

Yes we would possibly mix both. There shouldn't be any issue in mixing, at least no issue which would be worse if we didn't do this.

> 2. Do we have enough logging to show that a tlog entry (or multiple entries) are corrupted and therefore the default are being used?

The logs will print that we cannot parse an entry or entries.


The reason for this implementation is that we don't have a method which allows us to read through the tlog looking for corruption without merging while that happens. I think that we might want such a method eventually, but right now it isn't needed because the bootstrap is definitely going to have older timestamps and so any newer values will come from the uncorrupted tlog.

src/main/java/com/aws/greengrass/config/ConfigurationReader.java
```diff
@@ -100,6 +102,26 @@ private static void mergeTLogInto(Configuration c, Path p) throws IOException {
...
+     * @param tlogPath path to the file to validate.
+     * @throws IOException if any entry is invalid.
+     */
+    public static void validateTlog(Path tlogPath) throws IOException {
+        try (BufferedReader in = Files.newBufferedReader(tlogPath)) {
+            // Only validate that the first line is readable. If it is then we will assume the rest of the file
+            // is at least mostly fine since any corruption should only occur during the append and would only affect
+            // the very last entry.
+
+            String l = in.readLine();
```
fengwa-aws, line 117: Can you add more details to the comment? The assumption is we only need to fallback if the beginning of the tlog file is corrupted?

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
```diff
@@ -137,35 +139,96 @@ void initConfigAndTlog(String configFilePath) {
...
+                                .log("Transaction log {} is invalid and no usable backup exists, will attempt to load "
+                                                + "configuration from {}", transactionLogPath, bootstrapTlogPath,
+                                        tlogValidationError);
+                        kernel.getConfig().read(bootstrapTlogPath);
+                        readFromNonTlog = true;
+                    } else {
+                        // There are no files to load from
+                        logger.atError()
+                                .log("Transaction log {} is invalid and no usable backup exists", transactionLogPath,
+                                        tlogValidationError);
```
fengwa-aws, line 209: Should the program exit here? Especially if `externalConfigExists` is false?

MikeDombo: Logging is basically always safe, but I'd be concerned about making us exit at this point. I think it could have some undesirable side-effects.

src/main/java/com/aws/greengrass/config/ConfigurationReader.java
MikeDombo: Right, I thought my comment was clear enough that we're just checking the first line because any other corruption should only affect the last line of the file. Let me know if you have a suggestion for what comment to add.

fengwa-aws (approved):

src/main/java/com/aws/greengrass/util/CommitableFile.java
```diff
@@ -117,14 +117,15 @@ public void abandon() {
...
-     * Close the file and commit the new version.  The old version becomes a backup
+     * Close the file and commit the new version. The old version becomes a backup
      */
     @SuppressWarnings("ConvertToTryWithResources")
     @Override
     public void commit() {
         if (!closed) {
             try {
-                super.close();
+                flush();
```
fahadmohammed01, line 127: is there a difference between flush() and out.flush() ?

src/main/java/com/aws/greengrass/config/ConfigurationReader.java
```diff
@@ -100,6 +102,36 @@ private static void mergeTLogInto(Configuration c, Path p) throws IOException {
         }
     }
 
+    /**
+     * Validate the tlog contents at the given path. Throws an IOException if any entry is invalid.
+     *
+     * @param tlogPath path to the file to validate.
+     * @throws IOException if any entry is invalid.
+     */
+    public static void validateTlog(Path tlogPath) throws IOException {
```
fahadmohammed01, line 111: should we throw exception if the last line of tlog file is corrupted? I think we can treat tlog files that are corrupted only at the end as a valid tlog file.

src/main/java/com/aws/greengrass/util/CommitableFile.java
MikeDombo: No, flush() will redirect to out.flush(). But this way if we add more stuff into flush for some reason, it would get called here.

src/main/java/com/aws/greengrass/config/ConfigurationReader.java
MikeDombo: We are only validating the first line.

fahadmohammed01 (approved):

PR #934, opened 2021-04-20
Add loader.cmd for Windows

NikOrin:
**Issue #, if available:**

**Description of changes:**
New File: **loader.cmd**
Excluded **loader.cmd** from maven license plugin (it wasn't recognizing comment style of _@REM_)
Included **loader.cmd** in zip package

**Known issue:**
If the symlink **target** contains brackets in the name, then the function to flip the symlinks will not correctly set the symlink target.

The existing **loader** shell script does NOT work on Windows platforms. This new **loader.cmd** replicates the behavior in the existing **loader** shell script but for Windows platforms.

1 .Created a zip via the following command:
`mvn -DskipTests clean package`
I had to skip tests because there are existing tests that do not pass on Windows.

2. Deployed Greengrass without starting the Nucleus by running the **<GreengrassCoreRootDirectory>/lib/Greengrass.jar** with *--start false* flag. 

3. Ran **loader.cmd** from Greengrass directory (**\greengrass\v2\alts\current\distory\bin**)

4. Observed output and observed Greengrass logs in **\greengrass\v2\logs\** for proof Greengrass Nucleus running and logging.

5. Stopped Nucleus via Ctrl+C

6. Observed Greengrass logs for termination confirmation:
> com.aws.greengrass.lifecyclemanager.Kernel: Shutting down Nucleus due to external signal. {}

Note: Error seen in greengrass.log
> com.aws.greengrass.deployment.DeviceConfiguration: Unable to set up Nucleus from build recipe file. {}
com.aws.greengrass.componentmanager.exceptions.PackageLoadingException: Failed to load Nucleus recipe
	at com.aws.greengrass.deployment.DeviceConfiguration.initializeNucleusFromRecipe(DeviceConfiguration.java:371)
	at com.aws.greengrass.lifecyclemanager.Kernel.parseArgs(Kernel.java:637)
	at com.aws.greengrass.easysetup.GreengrassSetup.performSetup(GreengrassSetup.java:269)
	at com.aws.greengrass.easysetup.GreengrassSetup.main(GreengrassSetup.java:242)

Is this an existing Windows bug?

7. Observed existing java.exe process end and a new process spin up (the loader script retries 3 times)

Broken deployment testing:
started with _alts/new_ symlinked to a folder without a JAR and _alts/current_ symlinked to a functioning release.
End result:
because JAR couldn't be accessed (because it doesn't exist) the it looped 3 times to start, failed and created a symlink _alts/broken_ pointing towards the broken deployment and _alts/current_ continued to point at the working directory.

New deployment testing:
started with _alts/new_ symlinked to a new deployment and _alts/current_ symlinked to a previous deployment
End result:
_alts/current_ got moved to _alts/old_ and _alts/new_ got moved to _alts/current_

New deployment with an existing _alts/broken_ symlink:
started with _alts/broken, alts/current,_ and _alts/new_.
_alts/current_ symlink was renamed to _alts/old_
_alts/new_ symlink was renamed to _alts/current_
_alts/old_ symlink was renamed to _alts/current_

**Checklist:**

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>com.aws.greengrass</groupId>
     <artifactId>nucleus</artifactId>
-    <version>2.1.0-SNAPSHOT</version>
+    <version>2.2.0-SNAPSHOT</version>
     <packaging>jar</packaging>
 
     <licenses>
@@ -177,7 +177,7 @@
         <dependency>
             <groupId>software.amazon.awssdk.iotdevicesdk</groupId>
             <artifactId>aws-iot-device-sdk</artifactId>
-            <version>1.2.18</version>
+            <version>1.3.0-SNAPSHOT</version>
         </dependency>
         <dependency>
             <groupId>org.bouncycastle</groupId>
@@ -338,6 +338,7 @@
                     </licenseSets>
                     <mapping>
                         <java>JAVA_STYLE</java>
+                        <cmd>BATCH</cmd>
                     </mapping>
                 </configuration>
                 <executions>
--- a/scripts/loader.cmd
+++ b/scripts/loader.cmd
@@ -1,0 +1,138 @@
+@REM -------------------------------
+@REM Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+@REM SPDX-License-Identifier: Apache-2.0
+@REM -------------------------------
+@echo off
+SETLOCAL EnableDelayedExpansion
+SET DIR=%~dp0
+
+@REM Get the root GG directory (generally /greengrass/v2)
+FOR %%I IN ("%DIR%\..\..\..\..") DO SET "GG_ROOT=%%~fI"
+SET LAUNCH_DIR=%GG_ROOT%\alts\current
+@REM SET CONFIG_FILE=""
+
+SET IS_SYMLINK=0
+CALL :directory_is_symlink "%GG_ROOT%\alts\new" IS_SYMLINK
+IF %IS_SYMLINK% EQU 1 (
+    CALL :directory_is_symlink "%GG_ROOT%\alts\old" IS_SYMLINK
+    IF !IS_SYMLINK! EQU 0 (
+        CALL :directory_is_symlink "%LAUNCH_DIR%" IS_SYMLINK
+        IF !IS_SYMLINK! EQU 1 (
+            CALL :flip_links "%LAUNCH_DIR%" "%GG_ROOT%\alts\old"
+        )
+    )
+    CALL :flip_links "%GG_ROOT%\alts\new" "%LAUNCH_DIR%"
+)
+
+CALL :directory_is_symlink "%GG_ROOT%\alts\broken" IS_SYMLINK
+IF !IS_SYMLINK! EQU 1 (
+    CALL :directory_is_symlink "%GG_ROOT%\alts\old" IS_SYMLINK
+    IF !IS_SYMLINK! EQU 1 (
+        CALL :flip_links "%GG_ROOT%\alts\old" "%LAUNCH_DIR%"
+    )
+)
+
+CALL :directory_is_symlink "%GG_ROOT%\alts\old" IS_SYMLINK
+IF !IS_SYMLINK! EQU 1 (
+    CALL :directory_is_symlink "%LAUNCH_DIR%" IS_SYMLINK
+    IF !IS_SYMLINK! EQU 0 (
+        CALL :flip_links "%GG_ROOT%\alts\old" "%LAUNCH_DIR%"
+    )
+)
+
+@REM EXIST works for files, directories, and symlink dirs
+IF NOT EXIST %LAUNCH_DIR% (
+    ECHO FATAL: No Nucelus found!
+    EXIT /B 1
+)
+
+@REM Get JVM_OPTIONS from launch.params if it exists
+IF EXIST %LAUNCH_DIR%\launch.params (
+    FOR /F "delims=" %%A IN (%LAUNCH_DIR%\launch.params) DO SET JVM_OPTIONS=%%A
+)
+
+SET JVM_OPTIONS=%JVM_OPTIONS% -Droot="%GG_ROOT%"
+SET OPTIONS=--setup-system-service false
+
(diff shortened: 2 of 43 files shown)


scripts/loader.cmd
```diff
@@ -0,0 +1,136 @@
...
+@REM Attempt to start the nucleus 3 times
+FOR /L %%i IN (1,1,%MAX_RETRIES%) DO (
+    java -Dlog.store=FILE %JVM_OPTIONS% -jar "%LAUNCH_DIR%\distro\lib\Greengrass.jar" %OPTIONS%
+    SET KERNEL_EXIT_CODE=%ERRORLEVEL%
+
+    IF KERNEL_EXIT_CODE EQU 0 (
+        ECHO Restarting Nucleus
+        %LAUNCH_DIR%\distroy\bin\loader.cmd
+    ) ELSE (
+    IF KERNEL_EXIT_CODE EQU 100 (
```
K-AWS, line 68: What do these error codes translate to?

scripts/loader.cmd
```diff
@@ -0,0 +1,136 @@
...
+ECHO Parent dir: %PARENT_DIR%
+ECHO DIR name: %DIR_NAME%
+
+@REM Get linked dir information using some regex
+@REM Example output:
+@REM    04/08/2021  05:19 PM    <SYMLINKD>     current [C:\greengrass\v2\alts\init]
+FOR /F "delims=" %%I IN ('DIR %PARENT_DIR% ^| FINDSTR /R /C:"\<SYMLINKD\>[ ]*%DIR_NAME%[ ]*\[.*\]$"') DO SET DIR_OUT=%%I
+
+@REM Get second token (B starting from A) when considering [] as delimiters
+FOR /F "tokens=1,2 delims=[]" %%A IN ("%DIR_OUT%") DO SET SOURCE_FILE=%%B
```
K-AWS, line 129: How can we validate that this data is correct? 
How are we handling 8dot3 names?
What if the path has a space in the name?

scripts/loader.cmd
```diff
@@ -0,0 +1,136 @@
...
+
+@REM Split the parent directory from fully qualified path and file name
+FOR %%I IN ("%1") DO SET PARENT_DIR=%%~dpI && SET DIR_NAME=%%~nxI
+ECHO Parent dir: %PARENT_DIR%
+ECHO DIR name: %DIR_NAME%
+
+@REM Get linked dir information using some regex
+@REM Example output:
+@REM    04/08/2021  05:19 PM    <SYMLINKD>     current [C:\greengrass\v2\alts\init]
+FOR /F "delims=" %%I IN ('DIR %PARENT_DIR% ^| FINDSTR /R /C:"\<SYMLINKD\>[ ]*%DIR_NAME%[ ]*\[.*\]$"') DO SET DIR_OUT=%%I
```
K-AWS, line 126: Will this handle the case where GG is installed on something other than C:?

scripts/loader.cmd
```diff
@@ -0,0 +1,136 @@
...
+
+@REM Get second token (B starting from A) when considering [] as delimiters
+FOR /F "tokens=1,2 delims=[]" %%A IN ("%DIR_OUT%") DO SET SOURCE_FILE=%%B
+
+@REM Create new symlink with @param2's name and point towards @param1's source
+MKLINK /D %2 %SOURCE_FILE%
```
K-AWS, line 132: Does %SOURC_FILE% need to get wrapped in parentheses? Will this handle the situation where %SOURCE_FILE% contains spaces?
Can you link across partitions? Windows supports partitions mounted in a folder instead of drive letter.

scripts/loader.cmd
```diff
@@ -0,0 +1,136 @@
...
+
+@REM Remove @param1
+RMDIR %1 || :
```
K-AWS, line 135: What if the directory has contents? Do we want to use quiet mode?

scripts/loader.cmd
NikOrin: the kernel has a `shutdown()` method that accepts these as arguments
https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/36f8d2faf5439e52a4e3c5f739c1df8971e5a0ab/src/main/java/com/aws/greengrass/deployment/bootstrap/BootstrapSuccessCode.java#L9-L11

scripts/loader.cmd
NikOrin: that /C: is not a drive but a flag
https://ss64.com/nt/findstr.html

scripts/loader.cmd
NikOrin: Ah, forgot to wrap that variable, thanks for pointing that out. I have not tested link across partitions. I know you can link across drives though.

scripts/loader.cmd
NikOrin: Both %1 and %2 should be symlinks so RMDIR is not dependent on the folder being empty

scripts/loader.cmd
K-AWS: Okay, across partitions should be the same as across drives.

scripts/loader.cmd
NikOrin: _How can we validate that this data is _correct?__
Like the string and find and regex worked correctly? I can try and add more safe guards against more error scenarios (namely if the source directory is named something like greengrass-v2[2.1.0] or something like that).

_How are we handling 8dot3 names?_
Isn't the only thing about 8dot3 names that they are less than 8 characters and case insensitive? in which case, it would already be handled consistently?

_What if the path has a space in the name?_
As long as we reference `%SOURCE_FILE%` with `"%SOURCE_FILE%"` in quotes, that'll handle any directories with space in the name.

pom.xml
```diff
@@ -329,6 +329,7 @@
                             <useDefaultExcludes>true</useDefaultExcludes>
                             <excludes>
                                 <exclude>*</exclude>
+                                <exclude>scripts/loader.cmd</exclude>
```
MikeDombo, line 332: remove this. 

Try adding `<cmd>BATCH</cmd>` into the `mappings`

NikOrin: Did that and added the configuration to **style.xml** for batch now

MikeDombo (dismissed):

PR #937, opened 2021-04-23
chore: use new pre-released data plane SDK

wikimonkey:
**Issue #, if available:**

**Description of changes:**
Use new pre-release data plane SDK which contains two existing component service APIs and three new client device APIs

client device management components need to integrate with greengrass cloud

existing tests pass

**Checklist:**

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -101,6 +101,11 @@
             <artifactId>greengrassv2</artifactId>
         </dependency>
         <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>greengrassv2-data</artifactId>
+            <version>2.15.x-SNAPSHOT</version>
+        </dependency>
+        <dependency>
             <groupId>org.projectlombok</groupId>
             <artifactId>lombok</artifactId>
             <version>1.18.10</version>
--- a/src/main/java/com/aws/greengrass/componentmanager/ComponentManager.java
+++ b/src/main/java/com/aws/greengrass/componentmanager/ComponentManager.java
@@ -42,7 +42,7 @@ import com.vdurmont.semver4j.SemverException;
 import lombok.AccessLevel;
 import lombok.Setter;
 import software.amazon.awssdk.core.exception.SdkClientException;
-import software.amazon.awssdk.services.greengrassv2.model.ResolvedComponentVersion;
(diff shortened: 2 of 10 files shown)


MikeDombo (commented):
Why are there so many changes? This change should only be commenting out the old SDK and bringing in the new SDK in the POM.

wikimonkey:
> Why are there so many changes? This change should only be commenting out the old SDK and bringing in the new SDK in the POM.

the class name and namespace got changed to distinguish from control plane SDK, otherwise it could have name conflicts.

MikeDombo:
> > Why are there so many changes? This change should only be commenting out the old SDK and bringing in the new SDK in the POM.
> 
> the class name and namespace got changed to distinguish from control plane SDK, otherwise it could have name conflicts.

Why is there a difference? The SDK(s) are the same, they just use different HTTP clients. This renaming is absolutely not necessary. The existing "component service" client _is_ the data plane client. The control plane client is basically not used; it is only used a bit during provisioning.

wikimonkey:
> > > Why are there so many changes? This change should only be commenting out the old SDK and bringing in the new SDK in the POM.
> > 
> > 
> > the class name and namespace got changed to distinguish from control plane SDK, otherwise it could have name conflicts.
> 
> Why is there a difference? The SDK(s) are the same, they just use different HTTP clients. This renaming is absolutely not necessary. The existing "component service" client _is_ the data plane client. The control plane client is basically not used; it is only used a bit during provisioning.

Existing SDK is still used for some places (e.g. testing), so the new data plane SDK is using a slightly different namespace to avoid conflicts I believe.

The reason I rename `GreengrassComponentServiceClientFactory` to `GreengrassServiceClientFactory` is obvious: in GGAD or the other places (e.g. large configuration) we can use this client factory, so it shouldn't be called `component` service I think.

MikeDombo (changes requested):
Do not create new namespaces for the client.

wikimonkey:
> Do not create new namespaces for the client.

Discussed offline with SDK provider, the initiative is to separate control and data plane SDK. SDK team requires new data plane SDK to use different namespace. So if we would like to separate SDK, we can't use the same namespace.

jbutler (approved):

k-khatri (approved):

MikeDombo (approved):

PR #945, opened 2021-04-27
Enable device provisioning via provisioning plugin

abanthiy:
**Issue #, if available:**

**Description of changes:**
Adding nucleus changes for device provisioning using plugins

Unit tests
POC running local plugin

**Checklist:**


src/integrationtests/java/com/aws/greengrass/integrationtests/provisioning/TestDeviceProvisioningPlugin.java
```diff
@@ -0,0 +1,51 @@
...
+
+public class TestDeviceProvisioningPlugin implements DeviceIdentityInterface {
+    private static final String SERVICE_NAME = "aws.greengrass.TestProvisioningPlugin";
+
+    @Override
+    public ProvisionConfiguration updateIdentityConfiguration(ProvisionContext provisionContext) throws RetryableProvisioningException {
+        Map<String, Object> parameterMap = provisionContext.getParameterMap();
+        ProvisionConfiguration.SystemConfiguration systemConfiguration =
+                new ProvisionConfiguration.SystemConfiguration();
+        systemConfiguration.setCertificateFilePath(parameterMap.get("certificateFilePath") == null ? ""
```
MikeDombo, line 23: these should be consts in the deviceconfiguration class

src/main/java/com/aws/greengrass/deployment/IotJobsHelper.java
```diff
@@ -268,17 +276,45 @@ private static void unwrapExecutionException(ExecutionException e)
     }
 
     @Override
-    @SuppressFBWarnings
+    @SuppressWarnings("PMD.CollapsibleIfStatements")
     public void postInject() {
+        deviceConfiguration.onAnyChange((what, node) -> {
+            if (node != null &&  WhatHappened.changed.equals(what)) {
+                // List of configuration nodes that we need to reconfigure for if they change
+                if (node.childOf(DEVICE_MQTT_NAMESPACE) || node.childOf(DEVICE_PARAM_THING_NAME)
```
MikeDombo, line 284: you should ignore mqtt and auth changes. The mqtt client handles that already. Only listen to things that affect what job topics we subscribe to.

src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
```diff
@@ -68,18 +79,19 @@
     public static final String DESIRED_STATUS_CANCELED = "CANCELED";
     public static final String DEPLOYMENT_SHADOW_NAME = "AWSManagedGreengrassV2Deployment";
     @Inject
-    private DeploymentQueue deploymentQueue;
+    private  DeploymentQueue deploymentQueue;
```
MikeDombo, line 82: formatting?

src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
```diff
@@ -100,15 +112,64 @@ public void onConnectionResumed(boolean sessionPresent) {
...
+
     @Override
+    @SuppressWarnings("PMD.CollapsibleIfStatements")
     public void postInject() {
-        if (!deviceConfiguration.isDeviceConfiguredToTalkToCloud()) {
-            logger.atWarn().log("Device not configured to talk to AWS Iot cloud. Single device deployment is offline");
+        deviceConfiguration.onAnyChange((what, node) -> {
+            if (WhatHappened.changed.equals(what) && node != null) {
+                // List of configuration nodes that we need to reconfigure for if they change
+                if (node.childOf(DEVICE_MQTT_NAMESPACE) || node.childOf(DEVICE_PARAM_THING_NAME)
```
MikeDombo, line 141: same as above

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
```diff
@@ -103,9 +128,44 @@ public void launch() {
             kernel.getContext().get(c).startup();
         }
 
+        final Queue<DeviceIdentityInterface> provisioningPlugins = findProvisioningPlugins();
         // Must be called before everything else so that these are available to be
         // referenced by main/dependencies of main
         final Queue<String> autostart = findBuiltInServicesAndPlugins(); //NOPMD
+        // run the provisioning if device is not provisioned
+        if (!kernel.getContext().get(DeviceConfiguration.class).isDeviceConfiguredToTalkToCloud()
```
MikeDombo, line 136: You should be checking the provisioning policy here right? If customers don't want to provision then we should skip this.

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
```diff
@@ -103,9 +128,44 @@ public void launch() {
...
+                && !provisioningPlugins.isEmpty()) {
+            RetryUtils.RetryConfig retryConfig = RetryUtils.RetryConfig.builder()
+                    .maxAttempt(MAX_PROVISIONING_PLUGIN_RETRY_ATTEMPTS)
+                    .retryableExceptions(Collections.singletonList(RetryableProvisioningException.class))
+                    .build();
+            ExecutorService executorService = kernel.getContext().get(ExecutorService.class);
+            executorService.submit(() -> {
```
MikeDombo, line 143: use `execute` if you're going to ignore the returned future. That will resolve the spotbugs error.

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
```diff
@@ -103,9 +128,44 @@ public void launch() {
...
+                provisioningPlugins.stream().forEach((plugin) -> {
+                    String pluginName = plugin.name();
+                    Topics pluginConfig = kernel.getConfig().lookupTopics(SERVICES_NAMESPACE_TOPIC)
```
MikeDombo, line 146: use `findTopics(SERVICES, pluginName, CONFIGURATION)` which would do the same thing as you're doing.

src/main/java/com/aws/greengrass/provisioning/ProvisionConfiguration.java
```diff
@@ -0,0 +1,36 @@
...
+
+package com.aws.greengrass.provisioning;
+
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import lombok.Getter;
+import lombok.Setter;
+
+@Setter
+@Getter
+public class ProvisionConfiguration {
```
MikeDombo, line 14: provide builders for these?

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
```diff
@@ -119,10 +179,36 @@ public void launch() {
...
+
+    }
+
+    @SuppressWarnings("PMD.CloseResource")
+    private Queue<DeviceIdentityInterface> findProvisioningPlugins() {
+        Queue<DeviceIdentityInterface> provisioningPlugins = new LinkedList<>();
+        EZPlugins ezPlugins = kernel.getContext().get(EZPlugins.class);
+        try {
+            ezPlugins.withCacheDirectory(nucleusPaths.pluginPath());
+            ezPlugins.implementing(DeviceIdentityInterface.class, (c) -> {
```
MikeDombo, line 198: if you have the path to the jar you want to load, then use `ezplugins.loadPlugin` instead of `withCacheDirectory`.

src/main/java/com/aws/greengrass/provisioning/ProvisioningPluginFactory.java
```diff
@@ -0,0 +1,12 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.aws.greengrass.provisioning;
+
+public class ProvisioningPluginFactory {
+    public <T> T getPluginInstance(Class pluginClass) throws IllegalAccessException, InstantiationException {
+        return (T)pluginClass.newInstance();
```
MikeDombo, line 10: this requires the plugin to have a default constructor. That's fine, but I just want to make sure you know

src/integrationtests/java/com/aws/greengrass/integrationtests/provisioning/TestDeviceProvisioningPlugin.java
abanthiy: This is a test plugin which is independent of those constants and is dealing with its own parameters. Any change in device configuration constants should not change the values here.

src/main/java/com/aws/greengrass/deployment/IotJobsHelper.java
abanthiy: ok

src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
abanthiy: fixed

src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
abanthiy: Addressed

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
abanthiy: At this point in nucleus, I am not sure if we will even have a provisioning policy set because if device is not connected then there is no deployment from cloud which is where the policy is specified. 

In cases where provisioning policy is available, I pass the policy to the plugin so that it can make a decision of what it wants to do based on the policy, as opposed to nucleus making that decision. If they do not want to provision, they can return empty configuration.

src/main/java/com/aws/greengrass/provisioning/ProvisionConfiguration.java
abanthiy: I was trying to avoid them unless needed. But its probably going to be useful for customers.

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
abanthiy: We do not have the jar path here in nucleus. Was thinking not to take jar path as kernel args, just taking it as argument to easy setup.

src/main/java/com/aws/greengrass/provisioning/ProvisioningPluginFactory.java
abanthiy: Yes, have kept it this way for now. Working on refactoring it to use kernel context.

MikeDombo: This might be a good idea though, this way the plugin cannot get access to classes that we want to keep more hidden.

abanthiy: Ya good point.

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
```diff
@@ -119,10 +181,36 @@ public void launch() {
...
+                try {
+                    provisioningPlugins.add(provisioningPluginFactory.getPluginInstance(c));
+                } catch (InstantiationException | IllegalAccessException e) {
+                    logger.atError().kv("Plugin", c.getName()).setCause(e.getCause())
+                            .log("Error instantiating a provisioning plugin");
+                }
+            });
+
+            // ezPlugins.loadCache();
```
shaguptashaikh, line 209: Remove?

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
```diff
@@ -119,10 +181,36 @@ public void launch() {
...
+        } catch (IOException t) {
+            logger.atError().log("Error launching provisioning plugins", t);
```
shaguptashaikh, line 211: Silent failure, is this a good idea, it will launch nucleus anyway?

src/main/java/com/aws/greengrass/provisioning/ProvisionConfiguration.java
```diff
@@ -0,0 +1,48 @@
...
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+
+@Getter
+@Setter
+@NoArgsConstructor
+@AllArgsConstructor
```
shaguptashaikh, line 18: Are these redundant? Builders should be sufficient I think

src/main/java/com/aws/greengrass/provisioning/ProvisionContext.java
```diff
@@ -0,0 +1,18 @@
...
+
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+
+import java.util.Map;
+
+@AllArgsConstructor
+@Getter
+public class ProvisionContext {
+    String provisioningPolicy;
```
shaguptashaikh, line 16: This should be an enum

src/main/java/com/aws/greengrass/provisioning/ProvisioningHelper.java
```diff
@@ -0,0 +1,84 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+import static com.aws.greengrass.componentmanager.KernelConfigResolver.CONFIGURATION_CONFIG_KEY;
+import static com.aws.greengrass.deployment.DeviceConfiguration.SYSTEM_NAMESPACE_KEY;
+import static com.aws.greengrass.lifecyclemanager.GreengrassService.SERVICES_NAMESPACE_TOPIC;
+
+@AllArgsConstructor
+@NoArgsConstructor
+public class ProvisioningHelper {
```
shaguptashaikh, line 27: We will now have ProvisioningHelper and an existing DeviceProvisioningHelper, should we be renaming one of them to avoid ambiguity? 
Alternately we could avoid adding this class at all if `SystemConfiguration` and `NucleusConfiguration` were to have a `toMap` method which would construct a config map and `KernelLifecycle` could directly invoke that `updateFromMap` with that

src/main/java/com/aws/greengrass/provisioning/ProvisioningHelper.java
```diff
@@ -0,0 +1,84 @@
...
+
+    Kernel kernel;
```
shaguptashaikh, line 29: Make this private?

src/test/java/com/aws/greengrass/lifecyclemanager/KernelTest.java
```diff
@@ -265,6 +265,7 @@ void GIVEN_kernel_WHEN_locate_finds_class_definition_in_config_THEN_create_servi
         try {
             kernel.parseArgs().launch();
         } catch (RuntimeException ignored) {
+            System.out.println("Error captured is: " + ignored.getMessage());
```
shaguptashaikh, line 268: Why do you need this? Use a logger instead?

src/test/java/com/aws/greengrass/lifecyclemanager/KernelLifecycleTest.java
```diff
@@ -134,6 +178,206 @@ void GIVEN_kernel_WHEN_launch_with_autostart_services_THEN_autostarts_added_as_d
...
+                        , eq(UpdateBehaviorTree.UpdateBehavior.MERGE));
+        assertEquals(mockProvisionConfiguration.getNucleusConfiguration(), nucleusConfigCaptor.getValue());
+        assertEquals(mockProvisionConfiguration.getSystemConfiguration(), systemConfigCaptor.getValue());
+
+    }
+
+
+    @SuppressWarnings("PMD.CloseResource")
+    @Test
+    void GIVEN_kernel_WHEN_launch_with_provisioning_plugin_AND_plugin_methods_throw_runtime_Exception_THEN_offline_mode(ExtensionContext context)
```
shaguptashaikh, line 277: I'm still concerned with this behavior, in certain cases this might make sense but more commonly, I think it'll be better to fail the nucleus launch so that the user knows clearly there's a problem and can take steps to fix it, otherwise, the user is not going to know without looking at the logs. We'd talked earlier about adding something like an `onFailure` callback to the provisioning plugin so the plugin author gets some flexibility, where did that land?

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
```diff
@@ -103,9 +127,47 @@ public void launch() {
...
+        if (!kernel.getContext().get(DeviceConfiguration.class).isDeviceConfiguredToTalkToCloud()
+                && !provisioningPlugins.isEmpty()) {
+            logger.atDebug().log("Found provisioning plugins to run");
+            RetryUtils.RetryConfig retryConfig = RetryUtils.RetryConfig.builder()
+                    .maxAttempt(MAX_PROVISIONING_PLUGIN_RETRY_ATTEMPTS)
+                    .retryableExceptions(Collections.singletonList(RetryableProvisioningException.class))
+                    .build();
+            ExecutorService executorService = kernel.getContext().get(ExecutorService.class);
+            executorService.execute(() -> {
+                provisioningPlugins.stream().forEach((plugin) -> {
```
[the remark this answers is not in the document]
shaguptashaikh: Also, is there a test for this?

src/test/java/com/aws/greengrass/lifecyclemanager/KernelLifecycleTest.java
```diff
@@ -134,6 +178,206 @@ void GIVEN_kernel_WHEN_launch_with_autostart_services_THEN_autostarts_added_as_d
         verify(mockMain, times(2)).addOrUpdateDependency(eq(mockOthers), eq(DependencyType.HARD), eq(true));
     }
 
+    @SuppressWarnings("PMD.CloseResource")
```
shaguptashaikh, line 181: Which resource? Can you close it instead of suppressing this or use try with resource?

src/main/java/com/aws/greengrass/deployment/IotJobsHelper.java
```diff
@@ -268,17 +275,45 @@ private static void unwrapExecutionException(ExecutionException e)
...
+
+    private void validateAndSetupCommWithIotJobs(DeviceConfiguration deviceConfiguration)
+            throws DeviceConfigurationException {
 
+        // Not using isDeviceConfiguredToTalkToCloud() in order to provide the detailed error message to user
+        deviceConfiguration.validate();
+        setupCommWithIotJobs();
+    }
+
+    @SuppressFBWarnings
```
shaguptashaikh, line 315: Which warning is this? Could we avoid suppressing warnings unless it's a false alarm?

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
```diff
@@ -103,9 +127,47 @@ public void launch() {
             kernel.getContext().get(c).startup();
         }
 
+        final Queue<DeviceIdentityInterface> provisioningPlugins = findProvisioningPlugins();
```
shaguptashaikh, line 130: I see that the provisioning plugins are also going to be under the `services` config namespace, are these plugins going to be dependencies of `main` ? If so they won't load in `startupAllServices` and are also going to fail or show up in the result of `orderedDependencies`. Should we filter instances of DeviceIdentityInterface from those relevant places?

shaguptashaikh (changes requested):

src/main/java/com/aws/greengrass/deployment/IotJobsHelper.java
```diff
@@ -268,17 +275,45 @@ private static void unwrapExecutionException(ExecutionException e)
     }
 
     @Override
-    @SuppressFBWarnings
+    @SuppressWarnings("PMD.CollapsibleIfStatements")
     public void postInject() {
+        deviceConfiguration.onAnyChange((what, node) -> {
+            if (node != null && what.equals(WhatHappened.childChanged)) {
+                // List of configuration nodes that we need to reconfigure for if they change
+                if (node.childOf(DEVICE_PARAM_THING_NAME) || node.childOf(DEVICE_PARAM_IOT_DATA_ENDPOINT)
```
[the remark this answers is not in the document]
abanthiy: Jobs helper does not depend on cred endpoint. Let me see if it can be put into method.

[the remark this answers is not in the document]
MikeDombo: Isn't it easier to specify what does matter to jobs? The only thing that matters is the thing name isn't it?

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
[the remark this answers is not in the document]
abanthiy: There are no chains but if multiple plugins want to update different parts of config then they can. Ordering of plugins is out of scope of this work.

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
```diff
@@ -53,18 +63,30 @@
...
+import static com.aws.greengrass.componentmanager.KernelConfigResolver.CONFIGURATION_CONFIG_KEY;
+import static com.aws.greengrass.lifecyclemanager.GreengrassService.SERVICES_NAMESPACE_TOPIC;
 import static com.aws.greengrass.util.Utils.close;
 import static com.aws.greengrass.util.Utils.deepToString;
 
 public class KernelLifecycle {
     private static final Logger logger = LogManager.getLogger(KernelLifecycle.class);
     private static final int EXECUTOR_SHUTDOWN_TIMEOUT_SECONDS = 30;
+    // Enum for provision policy will exist in common library package
+    // This will be done as part of re-provisioning
```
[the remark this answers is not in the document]
abanthiy: When we do re-provisioning, we will introduce provisioning policy which will be part of deployment document. So an enum representing provisioning policy, will be made available as part of the common library shared between cloud and device.

src/main/java/com/aws/greengrass/deployment/IotJobsHelper.java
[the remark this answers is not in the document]
abanthiy: I do not want to reuse this across classes as it can be different for different classes.

src/test/java/com/aws/greengrass/lifecyclemanager/KernelTest.java
abanthiy: Oh did not mean to check this in. Was using it for some debugging.

src/test/java/com/aws/greengrass/lifecyclemanager/KernelLifecycleTest.java
abanthiy: I do not think we should fail the launch. Nucleus should still operate in offline mode. Even if we fail nucleus launch, they will have to see logs to find out the problem. And if it is setup as systemd service it will keep restarting with the same issue. The plugin can throw retryable exceptions if it wants to retry otherwise it can fail with a non-retryable error. This should give it enough flexibility to plugin author. We do not want to give nucleus's launch control to plugin.

src/test/java/com/aws/greengrass/lifecyclemanager/KernelLifecycleTest.java
abanthiy: It complains about the mocked EZPlugin. Closing the mocked one does not help. Existing test - "GIVEN_kernel_WHEN_launch_with_autostart_services_THEN_autostarts_added_as_dependencies_of_main" has similar issue.

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
abanthiy: They are not going to be dependencies of main, unless they are annotated with ImplementsService (and they extend GreengrassService). Plugins are under the services config namespace and as far as I understand, there can be components under that namespace which are not direct/indirect dependencies of main. If this is not true then we may need a different namespace for provisioning plugins. For now I do not see a need for that.

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
abanthiy: I have responded to this in other comment below. We can discuss more on slack if you still feel otherwise.

src/main/java/com/aws/greengrass/provisioning/ProvisionConfiguration.java
abanthiy: Since this is going to be used by plugin authors I am just exposing for convenience.

src/main/java/com/aws/greengrass/provisioning/ProvisionContext.java
abanthiy: As commented in the code, enum should exist in common library. Keeping it string here provides more flexibility when the enum is added later.

src/main/java/com/aws/greengrass/provisioning/ProvisioningHelper.java
abanthiy: Renamed to ProvisioningConfigUpdateHelper

src/main/java/com/aws/greengrass/deployment/IotJobsHelper.java
abanthiy: It was coming because of ignoring return value of executorService.submit(). It exists in current code in main as well. Refactored it and removed the warning.

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
shaguptashaikh: Okay, just calling out that this needs to be given some thought. If these plugins are allowed to be updated via deployments then they will become dependencies of main given how deployments work and it will introduce some complications here

src/main/java/com/aws/greengrass/deployment/IotJobsHelper.java
shaguptashaikh: Then could you use `executorService.execute()` instead of `submit()`? Even if it's done this way in existing code we should avoid these patterns if we can.

src/test/java/com/aws/greengrass/lifecyclemanager/KernelLifecycleTest.java
shaguptashaikh: System service should not be setup before provisioning is finished. Retrying is only helpful for intermittent problems. My point is mainly considering that customers writing a plugin and capable of log diving can be different from end users who may not be trained. So I think any failures in critical steps like this should be clear enough. Can you explain a bit what we gain by nucleus operating offline since it won't be able to connect to cloud anyway? I don't see that failing the launch gives any unnecessary control to the plugin, but instead makes the user's life easy. We can discuss offline if you want

src/main/java/com/aws/greengrass/provisioning/ProvisionConfiguration.java
shaguptashaikh: I'd personally prefer keeping only what's needed since it keeps adding to the generated code

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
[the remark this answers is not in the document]
shaguptashaikh: Nit - could we make it a todo then?

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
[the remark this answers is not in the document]
shaguptashaikh: Makes sense that it's out of scope but I'm a bit concerned with the fact that having multiple plugins introduces the possibility of a plugin executing after some other plugin inadvertently overwriting config set by the former plugin. I imagine it will also be difficult for customers to debug

src/main/java/com/aws/greengrass/deployment/IotJobsHelper.java
abanthiy: I tried that and it lead to some failures in tests. I did not debug it then. I will make the change and debug it now.

src/test/java/com/aws/greengrass/lifecyclemanager/KernelLifecycleTest.java
abanthiy: 1. If we want to wait for provisioning to finish, to setup system service later, then that means provisioning needs to be synchronous or easy setup code (or nucleus) needs to wait on its completion. This is one thing we were trying to avoid during our design discussions.
2. I am going to switch calling it non-provisioned mode (from offline mode). Some components may just run in non-provisioned mode. More than non-provisioned mode, it is important to keep greengrass running so it can react to any events (local in case of offline). If nucleus waits on plugin, and the plugin for some reason hangs, this will hang nucleus for that much time (and stop other components from starting). If we put a max timeout on plugin response, we are putting a hard limit on plugin which may not be desirable and takes away the flexibility. 
3. I understand your point and I think we may need a mechanism for plugin to indicate to shutdown on failure. But even in that case if system service is setup then this will be happening again and again.

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
abanthiy: That brings out a good point. Although we do not plan to support cloud deployment right now but when we start supporting reprovisioning, we will need to support updating plugin config via cloud deployments. We need to make sure anything done here does not complicate reprovisioning in future. I will think through, discuss and update.

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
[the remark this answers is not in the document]
abanthiy: As per discussed design, we are going to allow customers to specify a trusted plugins jar using easy setup. Easy setup will then copy that jar to trusted plugins directory. If we were to restrict it to one plugin, it will be difficult to restrict customers to put only one provision plugin jar in the trusted plugins directory OR to put only one DeviceIdentityInterface implementing class in a single jar. If such a scenario happens then nucleus will face a problem of which plugin to pick. 
In the current impl, nucleus does not have to choose. We still depend on customers to configure the plugins correctly. For debugging, we have logs at places to show which plugin is executing.

src/test/java/com/aws/greengrass/lifecyclemanager/KernelLifecycleTest.java
abanthiy: Discussed offline

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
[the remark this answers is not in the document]
abanthiy: As discussed, we will error out if multiple provisioning plugins are found. There is no compelling use case for multiple provisioning plugins right now.

src/integrationtests/java/com/aws/greengrass/integrationtests/provisioning/ProvisionFromTestPlugin.java
```diff
@@ -0,0 +1,262 @@
...
+import static com.aws.greengrass.deployment.DeviceConfiguration.DEVICE_PARAM_THING_NAME;
+import static com.aws.greengrass.testcommons.testutilities.ExceptionLogProtector.ignoreExceptionUltimateCauseOfType;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.containsString;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+@TestMethodOrder(MethodOrderer.OrderAnnotation.class)
```
MikeDombo, line 53: why does order matter? That's always very sketchy

src/integrationtests/java/com/aws/greengrass/integrationtests/provisioning/ProvisionFromTestPlugin.java
```diff
@@ -0,0 +1,262 @@
...
+    @Order(1)
+    @Test
+    @SuppressWarnings("PMD.AvoidDeeplyNestedIfStmts")
+    void GIVEN_nucleus_started_with_testProvisioningPlugin_AND_without_config_params_THEN_device_runs_offline_mode() throws Throwable {
+        URL filepath = getClass().getResource("config_without_test_provisioning_plugin.yaml");
+        ConfigPlatformResolver.initKernelWithMultiPlatformConfig(kernel, filepath);
+        CountDownLatch logLatch =  new CountDownLatch(2);
+        CountDownLatch reverseLatch = new CountDownLatch(1);
+        DeviceConfiguration deviceConfiguration = kernel.getContext().get(DeviceConfiguration.class);
+        TestUtils.createCloseableLogListener((message) -> {
```
MikeDombo, line 90: this should be in a try-with-resources

src/integrationtests/java/com/aws/greengrass/integrationtests/status/IotJobsFleetStatusServiceTest.java
```diff
@@ -162,6 +165,7 @@ void afterEach() {
 
     @Test
     void GIVEN_jobs_deployment_WHEN_deployment_finishes_THEN_status_is_uploaded_to_cloud(ExtensionContext context) throws Exception {
+        ignoreExceptionOfType(context, InvocationTargetException.class);
```
MikeDombo, line 168: new error?

src/integrationtests/resources/com/aws/greengrass/integrationtests/provisioning/resource/AdditionalDeviceProvisioningPluginForJar.java
```diff
@@ -0,0 +1,59 @@
...
+
+        ProvisionConfiguration provisionConfiguration = new ProvisionConfiguration();
+        provisionConfiguration.setSystemConfiguration(systemConfiguration);
+        provisionConfiguration.setNucleusConfiguration(nucleusConfiguration);
+
+        try {
+            if (parameterMap.get("waitTime") != null) {
+                Thread.sleep(Long.valueOf(parameterMap.get("waitTime").toString()));
+            }
+        } catch (InterruptedException ignored) {
```
MikeDombo, line 50: should be thrown up, or at the very least it should re-interrupt.

src/integrationtests/resources/com/aws/greengrass/integrationtests/provisioning/resource/TestDeviceProvisioningPluginForJar.java
```diff
@@ -0,0 +1,59 @@
...
+
+        ProvisionConfiguration provisionConfiguration = new ProvisionConfiguration();
+        provisionConfiguration.setSystemConfiguration(systemConfiguration);
+        provisionConfiguration.setNucleusConfiguration(nucleusConfiguration);
+
+        try {
+            if (parameterMap.get("waitTime") != null) {
+                Thread.sleep(Long.valueOf(parameterMap.get("waitTime").toString()));
+            }
+        } catch (InterruptedException ignored) {
```
MikeDombo, line 50: same

src/main/java/com/aws/greengrass/deployment/IotJobsHelper.java
```diff
@@ -268,17 +275,45 @@ private static void unwrapExecutionException(ExecutionException e)
...
             return;
         }
+    }
 
+    private boolean relevantNodeChanged(Node node) {
+        // List of configuration nodes that we need to reconfigure for if they change
+        return node.childOf(DEVICE_PARAM_THING_NAME) || node.childOf(DEVICE_PARAM_IOT_DATA_ENDPOINT)
+                || node.childOf(DEVICE_PARAM_PRIVATE_KEY_PATH)
+                || node.childOf(DEVICE_PARAM_CERTIFICATE_FILE_PATH) || node.childOf(DEVICE_PARAM_ROOT_CA_PATH)
+                || node.childOf(DEVICE_PARAM_AWS_REGION);
```
MikeDombo, line 305: these aren't relevant. MQTT client reconfigures itself already. The _only_ thing which should trigger this is when the thing name changes because that actually affects the name of the topic which you're subscribing to.

src/main/java/com/aws/greengrass/deployment/IotJobsHelper.java
```diff
@@ -289,8 +324,7 @@ public void postInject() {
 
         logger.dfltKv("ThingName", (Supplier<String>) () ->
                 Coerce.toString(deviceConfiguration.getThingName()));
-
-        executorService.submit(() -> {
+        executorService.execute(() -> {
```
MikeDombo, line 327: save the future, make it a singleton. What if multiple changes come in? This is going to end up executing several times.

src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
```diff
@@ -100,15 +112,68 @@ public void onConnectionResumed(boolean sessionPresent) {
...
+    /**
+     * Constructor for unit testing.
+     * @param deploymentQueue {@link DeploymentQueue}
+     * @param deploymentStatusKeeper {@link DeploymentStatusKeeper}
+     * @param mqttClient {@link MqttClient}
+     * @param executorService {@link ExecutorService}
+     * @param deviceConfiguration {@link DeviceConfiguration}
+     * @param iotShadowClient {@link IotShadowClient}
+     */
+    public ShadowDeploymentListener(DeploymentQueue deploymentQueue, DeploymentStatusKeeper deploymentStatusKeeper,
```
MikeDombo, line 124: make this an injection constructor?

src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
```diff
@@ -100,15 +112,68 @@ public void onConnectionResumed(boolean sessionPresent) {
...
+                } catch (DeviceConfigurationException e) {
+                    logger.atWarn().log("Device not configured to talk to AWS Iot cloud. "
+                            + "Single device deployment is offline: {}", e.getMessage());
+                    return;
+                }
+            }
+        });
+
+        try {
+            connectToShadowService(deviceConfiguration);
```
MikeDombo, line 150: probably same issue as jobs. Make sure we don't have multiple threads trying to connect.

src/main/java/com/aws/greengrass/easysetup/GreengrassSetup.java
```diff
@@ -269,14 +285,21 @@ void performSetup() throws IOException, DeviceConfigurationException, URISyntaxE
...
+            if (Utils.isEmpty(awsRegion)) {
+                awsRegion = Coerce.toString(deviceConfiguration.getAWSRegion());
+            }
+
+            if (Utils.isEmpty(awsRegion)) {
+                throw new RuntimeException("Required input aws region not provided for provisioning");
+            }
+
+            this.deviceProvisioningHelper = new DeviceProvisioningHelper(awsRegion, environmentStage, this.outStream);
+            provision(kernel);
```
MikeDombo, line 302: IotSdkClientFactory stuff definitely needs to happen prior to provisioning.

src/main/java/com/aws/greengrass/iot/IotCloudHelper.java
```diff
@@ -46,9 +47,11 @@
      * @param body        Http body for the request
      * @return Http response corresponding to http request for path
      * @throws AWSIotException when unable to send the request successfully
+     * @throws DeviceConfigurationException when device is not configured with credential endpoint
      */
     public IotCloudResponse sendHttpRequest(final IotConnectionManager connManager, String thingName, final String path,
-                                            final String verb, final byte[] body) throws AWSIotException {
+                                            final String verb, final byte[] body)
+                                            throws AWSIotException, DeviceConfigurationException {
```
MikeDombo, line 54: this is not compatible. Wrap the exception instead.

src/integrationtests/java/com/aws/greengrass/integrationtests/provisioning/ProvisionFromTestPlugin.java
abanthiy: The order matters because the logs seem to get captured by other test in the same test class. Even though I increase the executor timeout to kill the threads, but that still seems to happen. So by running the test which throws CrtRuntimeException at last, I am avoiding that problem. The error only comes when I run the whole suite and not on individual runs so I have chosen this mitigation. I do not see a underlying problem in the code and seems only an issue of how we are using the *ExceptionLogProtector*

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
```diff
@@ -54,17 +67,31 @@
...
 
+import static com.aws.greengrass.componentmanager.KernelConfigResolver.CONFIGURATION_CONFIG_KEY;
+import static com.aws.greengrass.lifecyclemanager.GreengrassService.SERVICES_NAMESPACE_TOPIC;
 import static com.aws.greengrass.util.Utils.close;
 import static com.aws.greengrass.util.Utils.deepToString;
 
 public class KernelLifecycle {
     private static final Logger logger = LogManager.getLogger(KernelLifecycle.class);
-    private static final int EXECUTOR_SHUTDOWN_TIMEOUT_SECONDS = 30;
+    private static final int EXECUTOR_SHUTDOWN_TIMEOUT_SECONDS = 40;
```
MikeDombo, line 77: why? 30s is already quite a lot when we've interrupted the threads.

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
```diff
@@ -123,6 +166,69 @@ public void launch() {
...
+
+    @SuppressWarnings("PMD.AvoidCatchingGenericException")
+    private void executeProvisioningPlugins(List<DeviceIdentityInterface> provisioningPlugins) {
+        logger.atDebug().log("Found provisioning plugins to run");
+        RetryUtils.RetryConfig retryConfig = RetryUtils.RetryConfig.builder()
+                .maxAttempt(MAX_PROVISIONING_PLUGIN_RETRY_ATTEMPTS)
+                .retryableExceptions(Collections.singletonList(RetryableProvisioningException.class))
+                .build();
+        ExecutorService executorService = kernel.getContext().get(ExecutorService.class);
+        executorService.execute(() -> {
```
MikeDombo, line 180: get a future back so we know if they complete or not?

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
```diff
@@ -377,9 +490,13 @@ public void shutdown(int timeoutSeconds) {
...
             logger.atInfo().log("Waiting for executors to shutdown");
-            executorService.awaitTermination(EXECUTOR_SHUTDOWN_TIMEOUT_SECONDS, TimeUnit.SECONDS);
-            scheduledExecutorService.awaitTermination(EXECUTOR_SHUTDOWN_TIMEOUT_SECONDS, TimeUnit.SECONDS);
-            logger.atInfo("executor-service-shutdown-complete").log();
+            boolean executorTerminated = executorService.awaitTermination(timeoutSeconds,
+                    TimeUnit.SECONDS);
+            boolean scheduledExecutorTerminated = scheduledExecutorService.awaitTermination(timeoutSeconds,
+                    TimeUnit.SECONDS);
+            logger.atInfo("executor-service-shutdown-complete")
+                    .kv("executor terminated", executorTerminated)
```
MikeDombo, line 498: not a hard rule, but none of our kvs have spaces

src/main/java/com/aws/greengrass/tes/TokenExchangeService.java
```diff
@@ -131,6 +132,7 @@ private void validateConfig() {
         }
     }
 
+    @SneakyThrows
```
MikeDombo, line 135: really?

src/test/java/com/aws/greengrass/deployment/ShadowDeploymentListenerTest.java
```diff
@@ -0,0 +1,121 @@
...
+    @BeforeEach
+    public void setup() {
+        mockExecutorService = Executors.newSingleThreadExecutor();
+        shadowDeploymentListener = new ShadowDeploymentListener(mockDeploymentQueue, mockDeploymentStatusKeeper,
+                mockMqttClient, mockExecutorService, mockDeviceConfiguration, mockIotShadowClient);
+    }
+
+    @AfterEach
+    public void close() {
+        mockExecutorService.shutdown();
```
MikeDombo, line 64: always use shutdownNow. Shutdown doesn't do anything to stop threads.

src/integrationtests/java/com/aws/greengrass/integrationtests/status/IotJobsFleetStatusServiceTest.java
abanthiy: This is an expected exception according to me, caused due to SSL Handshake failure when communicating with cloud. And it was flaky.

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
abanthiy: I need to revert this default value to 30. I needed 40 for my tests.

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
abanthiy: Where will you be checking that? If we wait on it after starting all the services then launch() can potentially get into waiting indefinitely. And if we put timeout then it will fail after a hard limit, which is not desirable. We probably need a configurable timeout for provisioning plugin?

src/main/java/com/aws/greengrass/deployment/IotJobsHelper.java
abanthiy: These become relevant when device is transitioning from non-provisioned mode to provisioning. In non-provisioning mode, it may have thingName configured but not other params, and when other params get configured the subscriptions need to be made since those do not exist in the mqtt client. However, we need to re-subscribe only when thing name changes.

src/main/java/com/aws/greengrass/iot/IotCloudHelper.java
abanthiy: Compatible with ? Can you describe more?

MikeDombo: any plugin which uses it. This change is not binary compatible

src/main/java/com/aws/greengrass/easysetup/GreengrassSetup.java
```diff
@@ -107,7 +113,9 @@
             + "runs setup steps,\n"
             + "\t\t\t\t\t(optional) provisions resources, and starts the software. If false, the software runs only "
             + "setup\n"
-            + "\t\t\t\t\tsteps and (optional) provisions resources. Defaults to true.\n";
+            + "\t\t\t\t\tsteps and (optional) provisions resources. Defaults to true.\n"
+            + "\n\t--trusted-plugin, -tp\t\t(Optional) Path of a plugin jar file. The plugin will be included as "
+            + "trusted plugin in nucleus. Specify multiple times for including multiple plugins.\n";
```
shaguptashaikh, line 118: If we do not support multiple plugins, why support multiple arguments and have this help text at all? It can confuse customers

src/integrationtests/java/com/aws/greengrass/integrationtests/provisioning/ProvisionFromTestPlugin.java
```diff
@@ -0,0 +1,267 @@
...
+
+            kernel.launch();
+            assertTrue(logLatch.await(2, TimeUnit.SECONDS));
+            assertFalse(reverseLatch.await(5, TimeUnit.SECONDS));
+        }
+    }
+
+    @Order(2)
+    @Test
+    void GIVEN_nucleus_started_with_testProvisioningPlugin_AND_plugin_throws_non_retryable_exception(ExtensionContext context) throws Throwable {
```
shaguptashaikh, line 117: Ni - could you make this into the GIVEN_WHEN_THEN format to make the expectation from this test clear?

src/integrationtests/java/com/aws/greengrass/integrationtests/provisioning/ProvisionFromTestPlugin.java
```diff
@@ -0,0 +1,267 @@
...
+        URL filepath = getClass().getResource("config_without_test_provisioning_plugin.yaml");
+        ConfigPlatformResolver.initKernelWithMultiPlatformConfig(kernel, filepath);
+        CountDownLatch logLatch =  new CountDownLatch(2);
+        CountDownLatch reverseLatch = new CountDownLatch(1);
+        DeviceConfiguration deviceConfiguration = kernel.getContext().get(DeviceConfiguration.class);
+        try (AutoCloseable listener = TestUtils.createCloseableLogListener((message) -> {
+            String messageString = message.getMessage();
+            if (messageString.contains("Device not configured to talk to AWS Iot cloud")) {
+                logLatch.countDown();
+            } else if ("Updated provisioning configuration".equals(messageString)) {
```
shaguptashaikh, line 94: Nit - Looks like these logs you're verifying here and other tests are from Nucleus launch code, could add constants for these and reference them here so you don't need to update tests if these logs change

shaguptashaikh:
No major concerns, only a few minor comments, otherwise LGTM

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -688,6 +688,7 @@ protected void runSpooler() {
                 break;
             } catch (Throwable ex) {
                 logger.atError().log("Unchecked error when publishing from spooler", ex);
+                throw ex;
```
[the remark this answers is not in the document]
abanthiy: Correct. That was the behavior earlier also if you fall into the ExecutionException catch block.

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
```diff
@@ -103,9 +132,23 @@ public void launch() {
...
+        loadPlugins();
+        // run the provisioning if device is not provisioned
+        if (!kernel.getContext().get(DeviceConfiguration.class).isDeviceConfiguredToTalkToCloud()
+                && !provisioningPlugins.isEmpty()) {
+            // Multiple provisioning plugins may need plugin ordering. We do not support plugin ordering right now
+            // There is also no compelling use case right now for multiple provisioning plugins.
+            if (provisioningPlugins.size() > 1) {
+                String errorString = String.format(MULTIPLE_PROVISIONING_PLUGINS_FOUND_EXCEPTION,
+                        provisioningPlugins.toString());
+                throw new RuntimeException(errorString);
```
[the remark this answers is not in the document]
abanthiy: Plugins do not get loaded until loadPlugins is called. EZPLugins scan only once and they scan for built in services/plugins and provisioning plugin in one go only.

src/main/java/com/aws/greengrass/deployment/IotJobsHelper.java
```diff
@@ -289,13 +333,18 @@ public void postInject() {
...
                 Coerce.toString(deviceConfiguration.getThingName()));
-
-        executorService.submit(() -> {
-            subscribeToJobsTopics();
-            logger.atInfo().log("Connection established to IoT cloud");
-            deploymentStatusKeeper.publishPersistedStatusUpdates(DeploymentType.IOT_JOBS);
-            this.fleetStatusService.updateFleetStatusUpdateForAllComponents();
-        });
+        if (subscriptionFuture != null && !subscriptionFuture.isDone()) {
+            subscriptionFuture.cancel(false);
```
MikeDombo, line 337: always use `true`. You must interrupt the existing thread, otherwise it won't stop

src/main/java/com/aws/greengrass/deployment/IotJobsHelper.java
```diff
@@ -412,6 +461,7 @@ public void requestNextPendingJobDocument() {
      */
     public void subscribeToJobsTopics() {
```
MikeDombo, line 462: does this properly handle getting interrupted? This method doesn't throw interrupted exception

src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
```diff
@@ -106,23 +122,86 @@ public void onConnectionResumed(boolean sessionPresent) {
...
         mqttClient.addToCallbackEvents(callbacks);
         deploymentStatusKeeper.registerDeploymentStatusConsumer(DeploymentType.SHADOW,
                 this::deploymentStatusChanged, ShadowDeploymentListener.class.getName());
-        executorService.execute(() -> {
-            subscribeToShadowTopics();
-            // Get the shadow state when kernel starts up by publishing to get topic
-            publishToGetDeviceShadowTopic();
-        });
+        if (subscriptionFuture != null && !subscriptionFuture.isDone()) {
+            subscriptionFuture.cancel(false);
```
MikeDombo, line 195: true

src/main/java/com/aws/greengrass/easysetup/GreengrassSetup.java
MikeDombo: Based on our security discussion today, do we want this to be a parameter? Should we just load from the trusted plugin dir always, since we know that it is protected with root perms?

src/main/java/com/aws/greengrass/tes/CredentialRequestHandler.java
```diff
@@ -157,7 +158,7 @@ private void generateError(HttpExchange exchange, int statusCode) throws IOExcep
      *
      * @return credentials
      */
-    private byte[] getCredentialsBypassCache() {
+    private byte[] getCredentialsBypassCache()  {
```
MikeDombo, line 161: fix format

src/test/java/com/aws/greengrass/lifecyclemanager/KernelLifecycleTest.java
```diff
@@ -96,14 +152,21 @@ void beforeEach() throws IOException {
...
         mockKernelCommandLine = Mockito.spy(new KernelCommandLine(mockKernel));
         kernelLifecycle = new KernelLifecycle(mockKernel, mockKernelCommandLine, mockPaths);
+        kernelLifecycle.setProvisioningConfigUpdateHelper(mockProvisioningConfigUpdateHelper);
+        kernelLifecycle.setProvisioningPluginFactory(mockProvisioningPluginFactory);
     }
 
     @AfterEach
     void afterEach() {
+        if (executorService != null) {
+            executorService.shutdown();
```
MikeDombo, line 168: shutdown now

src/integrationtests/java/com/aws/greengrass/integrationtests/provisioning/ProvisionFromTestPlugin.java
```diff
@@ -0,0 +1,267 @@
...
+
+        CountDownLatch logLatch =  new CountDownLatch(1);
+        CountDownLatch reverseLatch = new CountDownLatch(1);
+        try (AutoCloseable listener = TestUtils.createCloseableLogListener((message) -> {
+                String messageString = message.getMessage();
+                if (messageString.contains("Device not configured to talk to AWS Iot cloud")) {
+                    logLatch.countDown();
+                }
+             });
+             AutoCloseable reverseListener = TestUtils.createCloseableLogListener((message) -> {
```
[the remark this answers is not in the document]
abanthiy: It is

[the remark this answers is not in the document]
MikeDombo: No. This needs to be in the initializer for a try-with-resources so that it gets closed. This isn't being closed

src/main/java/com/aws/greengrass/deployment/IotJobsHelper.java
abanthiy: Individual subscribe methods inside it handle getting interrupted.

MikeDombo: But this method doesn't throw, so how is it being handled?

abanthiy: Those methods absorb the exception and break out of their loop. We can bubble up the exception upto this method.

src/main/java/com/aws/greengrass/iot/IotCloudHelper.java
```diff
@@ -46,12 +47,20 @@
      * @param body        Http body for the request
      * @return Http response corresponding to http request for path
      * @throws AWSIotException when unable to send the request successfully
+     * @throws DeviceConfigurationException when device is not configured with credential endpoint
```
shaguptashaikh, line 50: This is not thrown

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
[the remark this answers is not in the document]
shaguptashaikh: But that was only for when ExecutionException is due to InterruptedException which is expected. Now spooler can break even for intermittent errors and not recover until next Nucleus restart. Or am I missing something?

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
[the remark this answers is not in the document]
shaguptashaikh: Why does EZPlugins scan only once? even if that's a genuine concern, we should be able to change executeProvisioningPlugins to take a single plugin instead of a list

src/main/java/com/aws/greengrass/easysetup/GreengrassSetup.java
abanthiy: Nucleus loads plugins from trusted plugin only. This is just a convenience to copy to the trusted plugin dir. It is not a good customer experience if we expect them to put the plugin into trusted plugins directory after nucleus has started (trusted plugin might be created on nucleus start).

MikeDombo: If it is a convenience, then it isn't required. It isn't hard for the customer to just do the copy into trusted themselves.

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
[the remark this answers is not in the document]
abanthiy: I would think the scan is expensive, thats why once. Updated the execute method to take a single plugin.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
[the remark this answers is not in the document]
abanthiy: Spooler would have earlier kept retrying if it received any runtime exception. This would have lead to infinite loop in case of a non-retryable error. We do not want that. If the exception happens because of some config issue, spooler start will be attempted again when the config update happens (does not require restart) since publish message will be triggered by ShadowListener and IotJobsHelper upon config update.

src/main/java/com/aws/greengrass/easysetup/GreengrassSetup.java
abanthiy: Removing the parameter from easy setup does not make it any more secure. Even with parameter, you need dir (root) permissions to copy jar to the trusted plugin dir.

shaguptashaikh (approved):

MikeDombo (dismissed):

shaguptashaikh (approved):

MikeDombo (approved):

PR #949, opened 2021-04-28
feat(lifecycle): implement resource limit management on Linux using cgroup

fengwa-aws:
**Issue #, if available:**

**Description of changes:**

This change adds a `systemResourceLimits` under `RunWith` configuration, which allows users to configure the resource limits (only memory and cpu for now) for each external service component on Linux. 

This is useful for scenarios where customers want to set a resource limit for components.

Add an integration test 

**Checklist:**
 - [X] Updated or added new integration tests

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -39,7 +39,9 @@ jobs:
       - name: Build with Maven (not Windows)
         env:
           AWS_REGION: us-west-2
-        run: mvn -ntp -U verify
+        run: |
+          sudo -E mvn -ntp -U verify
+          sudo chown -R runner target
         if: matrix.os != 'windows-latest'
       - name: Upload Failed Test Report
         uses: actions/upload-artifact@v1.0.0
@@ -72,8 +74,8 @@ jobs:
       - name: Build benchmark with Maven
         # Changes can break the benchmark, so compile it now to make sure it is buildable
         run: |
-          mvn -ntp -U install -DskipTests
-          mvn -ntp -U -f src/test/greengrass-nucleus-benchmark install
+          sudo mvn -ntp -U install -DskipTests
+          sudo mvn -ntp -U -f src/test/greengrass-nucleus-benchmark install
         if: matrix.os == 'ubuntu-latest'
       - name: Save PR number
         run: |
@@ -83,7 +85,6 @@ jobs:
           mkdir -p target/jacoco-report
           echo ${{ github.event.number }} > ./pr/NR
           echo ${{ github.event.pull_request.head.sha }} > ./pr/SHA
-
           cp -R target/japicmp/default-cli.xml ./pr/japicmp/default-cli.xml
           cp target/jacoco-report/cobertura.xml ./pr/jacoco-report/cobertura.xml
           cp target/jacoco-report/cobertura-it.xml ./pr/jacoco-report/cobertura-it.xml
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/lifecyclemanager/GenericExternalServiceIntegTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/lifecyclemanager/GenericExternalServiceIntegTest.java
@@ -5,10 +5,12 @@
 
 package com.aws.greengrass.integrationtests.lifecyclemanager;
 
+import com.aws.greengrass.config.PlatformResolver;
 import com.aws.greengrass.config.Subscriber;
 import com.aws.greengrass.config.Topic;
 import com.aws.greengrass.config.WhatHappened;
 import com.aws.greengrass.dependency.State;
+import com.aws.greengrass.deployment.DeviceConfiguration;
 import com.aws.greengrass.integrationtests.BaseITCase;
 import com.aws.greengrass.integrationtests.util.ConfigPlatformResolver;
 import com.aws.greengrass.lifecyclemanager.GenericExternalService;
@@ -16,6 +18,7 @@ import com.aws.greengrass.lifecyclemanager.Kernel;
 import com.aws.greengrass.logging.impl.GreengrassLogMessage;
 import com.aws.greengrass.logging.impl.Slf4jLogAdapter;
 import com.aws.greengrass.testcommons.testutilities.NoOpPathOwnershipHandler;
+import com.aws.greengrass.util.platforms.unix.linux.Cgroup;
 import org.apache.commons.lang3.SystemUtils;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
@@ -26,6 +29,7 @@ import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.MethodSource;
 
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.List;
@@ -39,12 +43,16 @@ import java.util.stream.Stream;
 
 import static com.aws.greengrass.componentmanager.KernelConfigResolver.CONFIGURATION_CONFIG_KEY;
 import static com.aws.greengrass.componentmanager.KernelConfigResolver.VERSION_CONFIG_KEY;
+import static com.aws.greengrass.lifecyclemanager.GreengrassService.RUN_WITH_NAMESPACE_TOPIC;
+import static com.aws.greengrass.lifecyclemanager.GreengrassService.SERVICES_NAMESPACE_TOPIC;
 import static com.aws.greengrass.lifecyclemanager.GreengrassService.SERVICE_LIFECYCLE_NAMESPACE_TOPIC;
 import static com.aws.greengrass.lifecyclemanager.GreengrassService.SETENV_CONFIG_NAMESPACE;
+import static com.aws.greengrass.lifecyclemanager.GreengrassService.SYSTEM_RESOURCE_LIMITS_TOPICS;
 import static com.aws.greengrass.testcommons.testutilities.SudoUtil.assumeCanSudoShell;
 import static com.aws.greengrass.testcommons.testutilities.TestUtils.createCloseableLogListener;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.hasItem;
 import static org.hamcrest.Matchers.is;
 import static org.hamcrest.io.FileMatchers.anExistingFile;
@@ -58,9 +66,17 @@ import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 
 
-class GenericExternalServiceTest extends BaseITCase {
+class GenericExternalServiceIntegTest extends BaseITCase {
 
     private Kernel kernel;
+
+    static Stream<Arguments> posixTestUserConfig() {
+        return Stream.of(
+                arguments("config_run_with_user.yaml", "nobody", "nobody"),
+                arguments("config_run_with_user_shell.yaml", "nobody", "nobody"),
+                arguments("config_run_with_privilege.yaml", "nobody", "root")
+        );
+    }
 
     @BeforeEach
     void beforeEach() {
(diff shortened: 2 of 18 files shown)


src/main/java/com/aws/greengrass/util/platforms/unix/LinuxPlatform.java
```diff
@@ -0,0 +1,143 @@
...
+                    Files.createDirectory(cg.getSubsystemGGPath());
+                }
+                if (!Files.exists(cg.getSubsystemComponentPath(componentName))) {
+                    Files.createDirectory(cg.getSubsystemComponentPath(componentName));
+                }
+            }
+        } catch (IOException e) {
+            logger.atError().setCause(e).kv("componentName", componentName)
+                    .log("Failed to create system resource manager");
+        } catch (InterruptedException e) {
```
MikeDombo, line 53: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

src/main/java/com/aws/greengrass/util/platforms/unix/LinuxCgroup.java
```diff
@@ -0,0 +1,49 @@
...
+    public Path getSubsystemGGPath() {
+        return getSubsystemRootPath().resolve("greengrass");
+    }
+
+    public Path getSubsystemComponentPath(String componentName) {
+        return getSubsystemGGPath().resolve(componentName);
+    }
+
+    public String subsystemMountCmd() {
+        return String.format("mount -t cgroup -o %s cgroup /sys/fs/cgroup/%s", this.toString());
```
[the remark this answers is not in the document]
MikeDombo: Also, please don't use `this.toString`. It gets confusing. I'd highly recommend just using `osString` if that's your intention.

src/main/java/com/aws/greengrass/util/platforms/unix/LinuxCgroup.java
```diff
@@ -0,0 +1,49 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.aws.greengrass.util.platforms.unix;
```
MikeDombo, line 6: probably want a new package for cgroup/resource limits

src/main/java/com/aws/greengrass/util/platforms/unix/LinuxPlatform.java
```diff
@@ -0,0 +1,143 @@
...
+            logger.atWarn().kv(STDOUT, inputStreamToString(proc.getInputStream()))
+                    .kv(STDERR, inputStreamToString(proc.getErrorStream()))
+                    .log("Fail to run command with root: {}", command);
+        }
+    }
+
+    private Set<String> getMountedPaths() throws IOException {
+        Set<String> mountedPaths = new HashSet<>();
+
+        Path procMountsPath = Paths.get("/proc/self/mounts");
```
MikeDombo, line 122: how does this work? are `self/mounts` different from what I would get if I ran `mount` to list all the mounts on the system?

src/main/java/com/aws/greengrass/util/platforms/unix/LinuxPlatform.java
```diff
@@ -0,0 +1,143 @@
...
+                    Integer.toString(pid).getBytes());
+            Files.write(LinuxCgroup.CPU.getSubsystemComponentPath(componentName).resolve("cgroup.procs"), Integer
+                    .toString(pid).getBytes());
+        } catch (IOException e) {
+            logger.atError().setCause(e).kv("componentName", componentName).kv("processId", pid)
+                    .log("Failed to add pid to the cgroup");
+        }
+    }
+
+    private void runCmdWithRoot(String command) throws IOException, InterruptedException {
```
MikeDombo, line 106: this will need to return a value so you know if it succeeded

src/main/java/com/aws/greengrass/util/platforms/unix/LinuxPlatform.java
```diff
@@ -0,0 +1,143 @@
...
+        String[] cmd = getUserDecorator().withUser(getPrivilegedUser()).decorate(command);
```
MikeDombo, line 107: Does this still decorate when we're already root? Because we don't want it to do that. `Exec` handles that properly, not sure if calling these directly also covers it.

src/main/java/com/aws/greengrass/util/platforms/unix/LinuxPlatform.java
```diff
@@ -0,0 +1,143 @@
...
+                    .log("Failed to apply resource limits");
+        }
+    }
+
+    public void addProcessToResourceManager(String componentName, Process process) {
+        if (process == null) {
+            logger.atError().log("The process doesn't exist");
+            return;
+        }
+        int pid = PidUtil.getPid(process);
```
MikeDombo, line 94: this pid is only going to be the pid for either sudo or shell and not the customer's process. So not sure if the cgroup will capture the child like it needs to.

src/main/java/com/aws/greengrass/util/platforms/Platform.java
```diff
@@ -199,4 +198,13 @@ protected abstract FileSystemPermissionView getFileSystemPermissionView(FileSyst
     public abstract void setIpcFilePermissions(Path rootPath);
 
     public abstract void cleanupIpcFiles(Path rootPath);
+
+    public abstract void createResourceManager(String componentName);
```
MikeDombo, line 202: Instead of adding more and more into the platform space, I think we might want to split stuff out and add a generic `getPlatformResourceController` or something like that. And then that class can have all these methods. And of course it can be implemented for different platforms still.

src/main/java/com/aws/greengrass/util/platforms/unix/LinuxPlatform.java
shaguptashaikh, line 53: Is silently failing to create cgroup the right failure handling behavior? Should the component still be started up in this case?

src/main/java/com/aws/greengrass/util/platforms/unix/LinuxPlatform.java
```diff
@@ -0,0 +1,143 @@
...
+                int cpuQuotaUs = (int) (cpuPeriodUs * cpu);
+                String cpuQuotaUsStr = Integer.toString(cpuQuotaUs);
+
+                Files.write(LinuxCgroup.CPU.getSubsystemComponentPath(componentName).resolve("cpu.cfs_quota_us"),
+                        cpuQuotaUsStr.getBytes());
+            }
+
+        } catch (IOException e) {
+            logger.atError().setCause(e).kv("componentName", componentName)
+                    .log("Failed to apply resource limits");
```
shaguptashaikh, line 85: Same here and line 100, only logging may not be sufficient for failure handling?

src/main/java/com/aws/greengrass/util/platforms/unix/LinuxPlatform.java
shaguptashaikh: I think this should be verified through integ tests?

src/main/java/com/aws/greengrass/util/platforms/unix/LinuxPlatform.java
shaguptashaikh: Why not use some relevant utility from Exec.java here? Also, why is it `getPrivilegedUser ` instead of the current user that's running nuclues?

src/main/java/com/aws/greengrass/lifecyclemanager/GenericExternalService.java
```diff
@@ -134,19 +133,56 @@ protected GenericExternalService(Topics c, Topics privateSpace, Platform platfor
...
+            systemResourceLimits = deviceConfiguration.findRunWithDefaultSystemResourceLimits();
+        }
 
+        if (systemResourceLimits == null) {
+            systemResourceController.resetResourceLimits(this);
+        } else {
+            Map<String, Object> resourceLimits = new HashMap<>();
+            for (Map.Entry<String, Object> entry : systemResourceLimits.toPOJO().entrySet()) {
+                resourceLimits.put(entry.getKey(), entry.getValue());
+            }
```
MikeDombo, line 152: just use `putAll`?

src/main/java/com/aws/greengrass/util/platforms/SystemResourceController.java
```diff
@@ -0,0 +1,43 @@
...
+import java.util.Map;
+
+public interface SystemResourceController {
+    /**
+     * Remove the resource controller. This method should be called when an existing generic external service is
+     * removed.
+     *
+     * @param component a generic external service
+     */
+    void removeResourceController(GenericExternalService component);
```
MikeDombo, line 19: should the component definitely need to be a generic external service?

src/main/java/com/aws/greengrass/util/platforms/unix/UnixPlatform.java
```diff
@@ -436,6 +438,11 @@ public PosixFileSystemPermissionView(FileSystemPermission permission) {
         }
     }
 
+    @Override
+    public SystemResourceController getSystemResourceController() {
+        return new StubResourceController();
```
MikeDombo, line 443: do we need to call `new` every time this method gets called?

src/main/java/com/aws/greengrass/util/platforms/unix/linux/LinuxPlatform.java
```diff
@@ -0,0 +1,16 @@
...
+
+package com.aws.greengrass.util.platforms.unix.linux;
+
+import com.aws.greengrass.util.platforms.SystemResourceController;
+import com.aws.greengrass.util.platforms.unix.UnixPlatform;
+
+public class LinuxPlatform extends UnixPlatform {
+    @Override
+    public SystemResourceController getSystemResourceController() {
+        return new LinuxSystemResourceController(this);
```
MikeDombo, line 14: same here. Save this object as a field so we don't need to recreate it?

src/main/java/com/aws/greengrass/util/platforms/unix/linux/LinuxSystemResourceController.java
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    @Override
+    public void updateResourceLimits(GenericExternalService component, Map<String, Object> resourceLimit) {
+        try {
+            if (!Files.exists(Cgroup.Memory.getSubsystemComponentPath(component.getServiceName()))) {
+                initializeCgroup(component, Cgroup.Memory);
+            }
+            if (resourceLimit.containsKey(MEMORY_KEY)) {
+                long memoryLimitInKB = (long) resourceLimit.get(MEMORY_KEY);
```
MikeDombo, line 57: use the Coerce methods. Coerce.toLong

src/main/java/com/aws/greengrass/util/platforms/unix/linux/LinuxSystemResourceController.java
```diff
@@ -0,0 +1,174 @@
...
+    public void addComponentProcess(GenericExternalService component, Process process) {
+
+        if (!Files.exists(Cgroup.CPU.getSubsystemComponentPath(component.getServiceName()))
+                || !Files.exists(Cgroup.Memory.getSubsystemComponentPath(component.getServiceName()))) {
+            logger.atInfo().kv(COMPONENT_NAME, component.getServiceName()).log("Resource controller is not enabled");
+            return;
+        }
+
+        try {
+            Set<Integer> childProcesses = platform.getChildPids(process);
```
MikeDombo, line 111: Does this set include the parent pid? I don't remember.

src/main/java/com/aws/greengrass/util/platforms/unix/linux/LinuxSystemResourceController.java
```diff
@@ -0,0 +1,174 @@
...
+            for (Integer pid : childProcesses) {
+                if (pid == null) {
+                    logger.atError().log("The process doesn't exist and is skipped");
+                    continue;
+                }
+
+                Files.write(Cgroup.Memory.getCgroupProcsPath(component.getServiceName()),
```
MikeDombo, line 118: does this actually work? Files.write will overwrite the existing file and not use append, so if there are multiple processes will this work correctly to add them all to the cgroup?

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsPlatform.java
```diff
@@ -162,6 +164,11 @@ public void addUserToGroup(String user, String group) throws IOException {
         // TODO: [P41452086]: Windows support - create user/group, add user to group
     }
 
+    @Override
+    public SystemResourceController getSystemResourceController() {
+        return new StubResourceController();
```
MikeDombo, line 169: same

src/test/java/com/aws/greengrass/lifecyclemanager/KernelTest.java
```diff
@@ -262,6 +263,7 @@ void GIVEN_kernel_WHEN_locate_finds_class_definition_in_config_THEN_create_servi
         // We need to launch the kernel here as this triggers EZPlugins to search the classpath for @ImplementsService
         // it complains that there's no main, but we don't care for this test
         ignoreExceptionUltimateCauseWithMessage(context, "No matching definition in system model for: main");
+        ignoreExceptionOfType(context, AccessDeniedException.class);
```
MikeDombo, line 266: Do we want to ignore this? Seems like it should be fixed somehow.

src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
```diff
@@ -509,6 +511,17 @@ public Topic getRunWithDefaultWindowsUser() {
         return getRunWithTopic().lookup(RUN_WITH_DEFAULT_WINDOWS_USER);
     }
 
+    /**
+     * Find the RunWithDefault.SystemResourceLimits topics.
+     * @return topics
+     */
+    public Topics findRunWithDefaultSystemResourceLimits() {
+        return kernel.getConfig()
+                .findTopics(SERVICES_NAMESPACE_TOPIC, getNucleusComponentName(),
```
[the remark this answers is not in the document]
fengwa-aws: It's intentional to return null if it doesn't exist. It's used to decide whether to reset the system resource limit.

src/main/java/com/aws/greengrass/util/platforms/SystemResourceController.java
MikeDombo: GreengrassService, just in case we want to add anything else which is a sibling to GES instead of a child.

src/main/java/com/aws/greengrass/util/platforms/unix/linux/LinuxSystemResourceController.java
fengwa-aws: probably not...I made an update

src/main/java/com/aws/greengrass/util/platforms/unix/linux/LinuxSystemResourceController.java
fengwa-aws: I tested it and it's working.

src/main/java/com/aws/greengrass/lifecyclemanager/GenericExternalService.java
```diff
@@ -335,6 +352,9 @@ protected synchronized void startup() throws InterruptedException {
                 }
             }
         }, lifecycleProcesses);
+        if (result.getRight() != null) {
+            systemResourceController.addComponentProcess(this, result.getRight().getProcess());
+        }
```
shaguptashaikh, line 357: I think it will be better to do this after the Errored / NothingDone checks below on line 359-364, otherwise it's unnecessarily initializing cgroups in those cases

src/main/java/com/aws/greengrass/lifecyclemanager/GenericExternalService.java
```diff
@@ -366,6 +386,10 @@ private synchronized void handleRunScript() throws InterruptedException {
             }
         }, lifecycleProcesses);
 
+        if (result.getRight() != null) {
+            systemResourceController.addComponentProcess(this, result.getRight().getProcess());
+        }
+
```
shaguptashaikh, line 392: Same here, would be better done on line 400, i.e. `else` -> `else if (result.getRight() != null)`

src/test/java/com/aws/greengrass/lifecyclemanager/KernelTest.java
```diff
@@ -30,6 +30,7 @@
 import java.io.IOException;
 import java.io.StringWriter;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.AccessDeniedException;
```
shaguptashaikh, line 33: Can get rid of this

src/main/java/com/aws/greengrass/util/platforms/unix/linux/LinuxPlatform.java
```diff
@@ -0,0 +1,18 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.aws.greengrass.util.platforms.unix.linux;
+
+import com.aws.greengrass.util.platforms.SystemResourceController;
+import com.aws.greengrass.util.platforms.unix.UnixPlatform;
+
+public class LinuxPlatform extends UnixPlatform {
+    SystemResourceController systemResourceController = new LinuxSystemResourceController(this);;
```
MikeDombo, line 12: double semi-colons.

src/integrationtests/java/com/aws/greengrass/integrationtests/lifecyclemanager/GenericExternalServiceTest.java
```diff
@@ -503,11 +519,75 @@ void GIVEN_posix_default_user_WHEN_runs_THEN_runs_with_default_user(String file,
...
+                .lookup(SYSTEM_RESOURCE_LIMITS_TOPICS, PlatformResolver.getOSInfo(), "cpu").withValue(1.5);
+        //Block until events are completed
+        kernel.getContext().runOnPublishQueueAndWait(() -> {
+        });
+
+        assertResourceLimits(componentName, 10240l * 1024, 1.5);
+
+        // Run with component resource limit
+        kernel.getConfig().lookup(SERVICES_NAMESPACE_TOPIC, componentName,
+                CONFIGURATION_CONFIG_KEY, RUN_WITH_NAMESPACE_TOPIC, SYSTEM_RESOURCE_LIMITS_TOPICS,
```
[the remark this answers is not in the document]
fengwa-aws: `runwith` is under `configuration` after resolution: https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/e13e3039a192b9cbfbb11a78b72f598e9522ef05/src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java#L499

src/main/java/com/aws/greengrass/lifecyclemanager/GenericExternalService.java
```diff
@@ -134,19 +133,53 @@ protected GenericExternalService(Topics c, Topics privateSpace, Platform platfor
                 requestRestart();
             }
         });
+    }
+
+    private void updateSystemResourceLimits() {
+        Topics systemResourceLimits =
+                config.findTopics(CONFIGURATION_CONFIG_KEY, RUN_WITH_NAMESPACE_TOPIC,
```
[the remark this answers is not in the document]
fengwa-aws: You're right regarding generic external services. But for Nucleus, the `runwithDefault` is under `configuration`.

src/main/java/com/aws/greengrass/lifecyclemanager/GenericExternalService.java
fengwa-aws: The null check should have prevented that.

.github/workflows/maven.yml
```diff
@@ -72,8 +74,8 @@ jobs:
       - name: Build benchmark with Maven
         # Changes can break the benchmark, so compile it now to make sure it is buildable
         run: |
-          mvn -ntp -U install -DskipTests
-          mvn -ntp -U -f src/test/greengrass-nucleus-benchmark install
+          sudo mvn -ntp -U install -DskipTests
```
MikeDombo, line 77: why are these sudo now?

src/integrationtests/java/com/aws/greengrass/integrationtests/lifecyclemanager/GenericExternalServiceIntegTest.java
```diff
@@ -503,11 +519,71 @@ void GIVEN_posix_default_user_WHEN_runs_THEN_runs_with_default_user(String file,
...
+        String componentName = "echo_service";
+        // Run with no resource limit
+        ConfigPlatformResolver.initKernelWithMultiPlatformConfig(kernel,
+                getClass().getResource("config_run_with_user.yaml"));
+        kernel.launch();
+
+        // Run with nucleus default resource limit
+        DeviceConfiguration deviceConfiguration = new DeviceConfiguration(kernel);
+        deviceConfiguration.getRunWithTopic()
+                .lookup(SYSTEM_RESOURCE_LIMITS_TOPICS, PlatformResolver.getOSInfo(), "memory").withValue(10240l);
```
MikeDombo, line 534: nit

use capital L for longs

.github/workflows/maven.yml
fengwa-aws: I added this based on @shaguptashaikh change https://github.com/aws-greengrass/aws-greengrass-nucleus/pull/956/files.

shaguptashaikh (approved):

MikeDombo (approved):

.github/workflows/maven.yml
MikeDombo: Since you re-chown it should work without running as sudo

fengwa-aws: According to @shaguptashaikh , "The benckmark step was failing due to some permission issues as well
 and it looks like the main build step creates a bunch of artifacts that the subsequent steps need to use".

MikeDombo: Nothing is created except in the target directory which you're already re owning. I think you'd be OK

PR #952, opened 2021-04-28
fix(windows): fix PluginComponentTest

fufranci:
**Issue #, if available:**

**Description of changes:**

**Checklist:**

### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/lifecyclemanager/PluginComponentTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/lifecyclemanager/PluginComponentTest.java
@@ -10,6 +10,7 @@ import com.aws.greengrass.componentmanager.ComponentStore;
 import com.aws.greengrass.componentmanager.DependencyResolver;
 import com.aws.greengrass.componentmanager.KernelConfigResolver;
 import com.aws.greengrass.componentmanager.exceptions.PackageDownloadException;
+import com.aws.greengrass.componentmanager.exceptions.PackageLoadingException;
 import com.aws.greengrass.componentmanager.exceptions.PackagingException;
 import com.aws.greengrass.componentmanager.models.ComponentIdentifier;
 import com.aws.greengrass.config.Topics;
@@ -221,8 +222,7 @@ class PluginComponentTest extends BaseITCase {
 
     @Test
     void GIVEN_plugin_added_and_removed_WHEN_plugin_added_again_THEN_plugin_is_loaded_into_JVM(ExtensionContext context) throws Exception {
-        ignoreExceptionOfType(context, PackageDownloadException.class);
-        ignoreExceptionOfType(context, SdkClientException.class);
+        ignoreExceptionOfType(context, PackageLoadingException.class);
 
         // launch Nucleus
         kernel.parseArgs();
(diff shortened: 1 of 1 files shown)


src/integrationtests/java/com/aws/greengrass/integrationtests/lifecyclemanager/PluginComponentTest.java
```diff
@@ -221,8 +222,7 @@ void GIVEN_kernel_WHEN_deploy_new_plugin_THEN_plugin_is_loaded_into_JVM(Extensio
 
     @Test
     void GIVEN_plugin_added_and_removed_WHEN_plugin_added_again_THEN_plugin_is_loaded_into_JVM(ExtensionContext context) throws Exception {
-        ignoreExceptionOfType(context, PackageDownloadException.class);
-        ignoreExceptionOfType(context, SdkClientException.class);
+        ignoreExceptionOfType(context, PackageLoadingException.class);
```
fufranci, line 225: I have to add this to make this test to pass. However, I don't think this is the ideal/correct solution.

This exception is generated when the test removes the plugin [here](https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/main/src/integrationtests/java/com/aws/greengrass/integrationtests/lifecyclemanager/PluginComponentTest.java#L244). The ComponentStore wants to delete the component. One of the step to delete the component is to delete the artifacts [here](https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/main/src/main/java/com/aws/greengrass/componentmanager/ComponentStore.java#L249). However the component jar file cannot be deleted. I think it is still being used. 

I think the correct way to fix this test on Windows is to find where is holding on to the jar file. Mostly likely a file resource is not closed yet. But I don't know where is holding on an component artifact.

The complete output of running the test is [here](https://paste.amazon.com/show/fufranci/1619643916).

MikeDombo: It would be open in EZPlugins which keeps a list of all the classloaders. One of the classloaders will probably hold a handle to the file.

MikeDombo: All the classloaders get closed when the context is closed.

MikeDombo: However we cannot close the loader while the nucleus is still running; things would almost certainly break.

You could try `pluginclass.getClassLoader().close()` to close it directly. That _might_ avoid problems, not sure really.

fufranci: I have tried removing strong references [here](https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/main/src/main/java/com/aws/greengrass/dependency/EZPlugins.java#L47) and using try-with-resource [here](https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/main/src/main/java/com/aws/greengrass/dependency/EZPlugins.java#L100) and [here](https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/main/src/main/java/com/aws/greengrass/dependency/EZPlugins.java#L117) and [here](https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/main/src/main/java/com/aws/greengrass/dependency/EZPlugins.java#L149). The test that I am trying to fix does pass (proofing it is class loader that is still holding on to a jar file). However other tests in the same suite start to fail. 

I don't want to mess too much with class loader. Would ignoring the `PackageLoadingException` be OK for the purpose of porting to Windows?

MikeDombo: Can you try my suggestion? We cannot use try with resources like you added because the classloader must stay open even after loading the plugin initially. However for the test purposes you could manually try closing the classloader like I suggest.

MikeDombo (approved):

leaf94 (approved):

PR #961, opened 2021-05-05
fix(deployment): duplicated shadow deployment if restart required

tilo-chen:
**Issue #, if available:**

**Description of changes:**
Added one more check when processing shadow update.

Currently, if a shadow deployment requires restart (aka kernel update workflow), ShadowDeploymentListener will create the same deployment again after restart.

Before the restart, the deployment will have been reported as IN_PROGRESS, which is a non-terminal state. Because it could be due to power off or other unexpected failure, we decided to allow it to create the deployment if the reported status was non-terminal (IN_PROGRESS). But if we're restarting normally, we're still adding the deployment which is duplicated.

If we persist the lastConfigurationArn through restart, then this will be caught by existing logic and won't create duplicated deployment. However, it will also bypass the existing logic to catch unexpected restart. Therefore, we're adding new logic to handle the above scenario.

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
+++ b/src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
@@ -323,7 +323,7 @@ public class ShadowDeploymentListener implements InjectionActions {
                     if (deploymentServiceLocateResult instanceof DeploymentService) {
                         DeploymentTaskMetadata currentDeployment =
                                 ((DeploymentService) deploymentServiceLocateResult).getCurrentDeploymentTaskMetadata();
-                        if (currentDeployment != null && currentDeployment.getDeploymentId().equals(configurationArn)) {
+                        if (currentDeployment != null && configurationArn.equals(currentDeployment.getDeploymentId())) {
                             logger.atInfo().kv(CONFIGURATION_ARN_LOG_KEY_NAME, configurationArn)
                                     .log("Ongoing deployment. Ignoring shadow update at startup");
                             return;
--- a/src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
+++ b/src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
@@ -8,6 +8,10 @@ package com.aws.greengrass.deployment;
 import com.amazon.aws.iot.greengrass.configuration.common.Configuration;
 import com.aws.greengrass.dependency.InjectionActions;
 import com.aws.greengrass.deployment.model.Deployment;
+import com.aws.greengrass.deployment.model.DeploymentTaskMetadata;
+import com.aws.greengrass.lifecyclemanager.GreengrassService;
+import com.aws.greengrass.lifecyclemanager.Kernel;
+import com.aws.greengrass.lifecyclemanager.exceptions.ServiceLoadException;
 import com.aws.greengrass.logging.api.Logger;
 import com.aws.greengrass.logging.impl.LogManager;
 import com.aws.greengrass.mqttclient.MqttClient;
@@ -67,6 +71,8 @@ public class ShadowDeploymentListener implements InjectionActions {
     public static final String GGC_VERSION_KEY = "ggcVersion";
     public static final String DESIRED_STATUS_CANCELED = "CANCELED";
     public static final String DEPLOYMENT_SHADOW_NAME = "AWSManagedGreengrassV2Deployment";
+    @Inject
+    private Kernel kernel;
     @Inject
     private DeploymentQueue deploymentQueue;
     @Inject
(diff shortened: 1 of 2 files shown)


src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
```diff
@@ -308,6 +314,24 @@ protected void shadowUpdated(Map<String, Object> desired, Map<String, Object> re
...
+                                ((DeploymentService) deploymentServiceLocateResult).getCurrentDeploymentTaskMetadata();
+                        if (currentDeployment != null && currentDeployment.getDeploymentId().equals(configurationArn)) {
+                            logger.atInfo().kv(CONFIGURATION_ARN_LOG_KEY_NAME, configurationArn)
+                                    .log("Ongoing deployment. Ignoring shadow update at startup");
+                            return;
+                        }
+                    }
+                } catch (ServiceLoadException e) {
+                    logger.atError().setCause(e).log("Failed to find deployment service");
+                }
```
fahadmohammed01, line 334: we have a dedup logic on line 336, may be update that to store lastConfigurationArn across kernel restarts by storing it as a topic?

src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
```diff
@@ -291,9 +299,10 @@ protected void shadowUpdated(Map<String, Object> desired, Map<String, Object> re
         }
         boolean cancelDeployment = DESIRED_STATUS_CANCELED.equals(desiredStatus);
         synchronized (ShadowDeploymentListener.class) {
+            String lastConfigurationArn = (String) lastConfigurationArnTopic.getOnce();
```
MikeDombo, line 302: use `Coerce.toString(lastConfigurationArnTopic)`

src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
```diff
@@ -67,6 +70,9 @@
     public static final String GGC_VERSION_KEY = "ggcVersion";
     public static final String DESIRED_STATUS_CANCELED = "CANCELED";
     public static final String DEPLOYMENT_SHADOW_NAME = "AWSManagedGreengrassV2Deployment";
+    public static final String SHADOW_LAST_CONFIG_ARN_TOPIC_KEY = "ShadowLastConfigurationArn";
+    @Inject
+    private Kernel kernel;
```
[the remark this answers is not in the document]
tilo-chen: Trying to get CurrentDeploymentTaskMetadata here. AFAIK here we'll have to use kernel to locate DeploymentService to get it...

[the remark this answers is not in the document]
MikeDombo: Look at the diff... you don't user kernel at all except for `getConfig`

src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
```diff
@@ -308,6 +314,24 @@ protected void shadowUpdated(Map<String, Object> desired, Map<String, Object> re
                             .log("Deployment result already reported. Ignoring shadow update at startup");
                     return;
                 }
+                // Ignore if it's the ongoing deployment. This can happen if the last shadow deployment caused restart
+                try {
+                    // Using locate instead of injection here because DeploymentService lacks usable injection
+                    // constructor. Same as in IotJobsHelper.evaluateCancellationAndCancelDeploymentIfNeeded
+                    GreengrassService deploymentServiceLocateResult =
```
MikeDombo, line 321: you don't need to do this, you should be able to just `@inject` the deployment service itself.

src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
```diff
@@ -68,6 +72,8 @@
     public static final String DESIRED_STATUS_CANCELED = "CANCELED";
     public static final String DEPLOYMENT_SHADOW_NAME = "AWSManagedGreengrassV2Deployment";
     @Inject
+    private Kernel kernel;
+    @Inject
```
MikeDombo, line 76: why is this added? How was it working before?

src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
tilo-chen: Sadly, directly `@inject DeploymentService` gives no usable injection constructor error

src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
tilo-chen: Persisting it across restart can avoid the duplicated deployment. But it also prevents us from retrying deployment that failed because of power cycle in the middle of a deployment.
It's still necessary to check the ongoing deployment ID to tell which case we're dealing with.

src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
tilo-chen: this is just to get `DeploymentService.getCurrentDeploymentTaskMetadata`. We didn't need that info. But in order to tell why last deployment didn't finish, (unexpected failure vs. managed restart) we can check the ongoing deployment ID. If it was a managed restart, the ongoing deployment should be the same.

src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
```diff
@@ -308,6 +314,24 @@ protected void shadowUpdated(Map<String, Object> desired, Map<String, Object> re
...
+                // Ignore if it's the ongoing deployment. This can happen if the last shadow deployment caused restart
+                try {
+                    // Using locate instead of injection here because DeploymentService lacks usable injection
+                    // constructor. Same as in IotJobsHelper.evaluateCancellationAndCancelDeploymentIfNeeded
+                    GreengrassService deploymentServiceLocateResult =
+                            kernel.locate(DeploymentService.DEPLOYMENT_SERVICE_TOPICS);
+                    if (deploymentServiceLocateResult instanceof DeploymentService) {
+                        DeploymentTaskMetadata currentDeployment =
+                                ((DeploymentService) deploymentServiceLocateResult).getCurrentDeploymentTaskMetadata();
+                        if (currentDeployment != null && currentDeployment.getDeploymentId().equals(configurationArn)) {
```
MikeDombo, line 326: swap the `.equals` check to prevent possible NPE

MikeDombo (dismissed):

fahadmohammed01 (dismissed):

MikeDombo (approved):

MikeDombo (approved):

abanthiy (approved):

PR #965, opened 2021-05-11
fix(runwith): enable runwith resetting to default when runwith is not set

MikeDombo:
**Issue #, if available:**

**Description of changes:**
Removes null check when converting cloud deployment into internal deployment type. This means that the `posixUser` field will have been set (even if it was set to null), thus when the deployment happens the runWith user will be reset to the default.

Added new test to verify the functionality. Verified that based on the cloud-serialized data, this check removal now works as expected.

**Checklist:**


leaf94 (dismissed):
Thanks a lot for the quick turnaround!

fengwa-aws (commented):
Can we update the code in [KernelConfigResolver](https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/003f19ee79627f5a0e281beee285c26aaa311ef6/src/main/java/com/aws/greengrass/componentmanager/KernelConfigResolver.java#L256-L259) as well? The [RunWith](https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/a3441713a209fea3dd19b4dba01f7d8eb3015890/src/main/java/com/aws/greengrass/deployment/model/RunWith.java#L20) class can also be simplified to remove the `callPosixUser` and `callWindowsUser` fields.

fengwa-aws (dismissed):

PR #966, opened 2021-05-12
fix(fss): make fss local agent more robust

fahadmohammed01:
**Issue #, if available:**

**1: list Installed components returns components that got removed from the device**

The contract between the device and cloud is that when a component is removed from the device, the FSS local agent will update cloud of the removal by reporting status for the components with the fleetConfigArns field as empty. There is a bug in FSS local agent which causes it not send this update. The bug is hit when three events happen in the following order:

i. Kernel restarts
ii. Any Component moves to BROKEN state
iii. A deployment removes a component in BROKEN or FINISHED state.

Root cause:
FSS local agent keeps tracks of components running on the device in two ways.

i. FSS agent keeps track of all components added by a deployment. [STATE 1]
ii. FSS agent keeps track of all components which changed state after the previous FSS update. [STATE 2]

Both the states are not persisted to disk so when the device restarts both states are reset. When the kernel starts up STATE 1 and STATE 2 are empty. When the kernel starts the components STATE 2 starts filling up as all the components will move through the different life-cycle stages to reach RUNNING. But when a component reached BROKEN state, FSS agent will send an update to the cloud and include all the information in STATE 2, then it wipes STATE 2 as its already sent to the cloud. Now the device receives a deployment to remove a component X which is in the BROKEN or FINISHED state. Since X is in BROKEN/FINISHED state it does not makes any state changes before its removed from the mains dependency tree and therefore FSS local agent looses track of the component and is not included in any FSS update afterwards. The data set that list Installed components API looks at is FSS data and hence the component will show up in list Installed components API even after it is removed from the device.

Fix:
Updated FSS local agent to keep state across kernel restarts.

**2: list Installed components returns wrong components version, when a component version is upgraded by a deployment**

Root cause:
FSS local agent will send 2 updates for the component when a version is bumped up (one for old version and one new version). Cloud stores the version based on the order in which the updates is processed. 

Fix:
FSS local agent will only send one update for the component.

Integ tests.

**Checklist:**
 - [ X] Updated or added new unit tests

### DIFF
--- a/src/main/java/com/aws/greengrass/status/FleetStatusService.java
+++ b/src/main/java/com/aws/greengrass/status/FleetStatusService.java
@@ -29,7 +29,6 @@ import org.apache.commons.lang3.RandomUtils;
 import software.amazon.awssdk.crt.mqtt.MqttClientConnectionEvents;
 import software.amazon.awssdk.iot.iotjobs.model.JobStatus;
 
-import javax.inject.Inject;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -43,6 +42,7 @@ import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
+import javax.inject.Inject;
 
 import static com.aws.greengrass.deployment.DeploymentService.COMPONENTS_TO_GROUPS_TOPICS;
 import static com.aws.greengrass.deployment.DeploymentService.DEPLOYMENT_DETAILED_STATUS_KEY;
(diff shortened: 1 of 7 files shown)


src/main/java/com/aws/greengrass/status/FleetStatusService.java
```diff
@@ -353,12 +359,27 @@ private void uploadFleetStatusServiceData(Set<GreengrassService> greengrassServi
...
         long sequenceNumber;
+
         synchronized (greengrassServiceSet) {
+
             // If there are no Greengrass services to be updated, do not send an update.
             if (greengrassServiceSet.isEmpty()) {
                 return;
             }
 
+            //When a component version is bumped up, FSS may have pointers to both old and new service instances
```
MikeDombo, line 370: component versions don't change the instance. The exact same instance is used

fahadmohammed01: True, but if a deployment adds component X with v1 and next deployment removes component X. Then if next  deployment adds a new version of X (v2). Then a new instance is created.

MikeDombo (approved):

MikeDombo (approved):

leaf94 (approved):
Thanks!!!

nikkhilmuthye (approved):

PR #990, opened 2021-05-29
fix(device provisioning): fix duplicate subscriptions to iot jobs and…

abanthiy:
… shadow

**Issue #, if available:**
Mqtt topics for jobs and shadow are subscribed multiple times when provisioning plugin runs since it updates a number of nodes in config.

**Description of changes:**
This PR ensures that the subscriptions are done only once and any previous subscriptions are removed.

It is required for device provisioning to work cleanly.

Tested manually by running Fleet provisioning plugin.

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/deployment/IotJobsHelper.java
+++ b/src/main/java/com/aws/greengrass/deployment/IotJobsHelper.java
@@ -150,6 +150,8 @@ public class IotJobsHelper implements InjectionActions {
 
     private AtomicBoolean isSubscribedToIotJobsTopics = new AtomicBoolean(false);
     private Future<?> subscriptionFuture;
+    private volatile String thingName;
+
     private final Consumer<JobExecutionsChangedEvent> eventHandler = event -> {
         /*
          * This message is received when either of these things happen
@@ -284,6 +286,11 @@ public class IotJobsHelper implements InjectionActions {
 
     @Override
     public void postInject() {
+        mqttClient.addToCallbackEvents(callbacks);
+        this.connection = wrapperMqttConnectionFactory.getAwsIotMqttConnection(mqttClient);
+        // GG_NEEDS_REVIEW: TODO: switch back to IotJobsClient after IoT device sdk updated for jobs namespace
+        this.iotJobsClientWrapper = iotJobsClientFactory.getIotJobsClientWrapper(connection);
+
         deviceConfiguration.onAnyChange((what, node) -> {
             if (node != null && what.equals(WhatHappened.childChanged) && relevantNodeChanged(node)) {
                 try {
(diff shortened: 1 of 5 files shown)


src/main/java/com/aws/greengrass/deployment/IotJobsHelper.java
```diff
@@ -150,6 +150,8 @@
 
     private AtomicBoolean isSubscribedToIotJobsTopics = new AtomicBoolean(false);
     private Future<?> subscriptionFuture;
+    private String thingName;
```
leaf94, line 153: `volatile` might be needed here? As multiple threads could write and read at the same time and a memory inconsistency might happen

abanthiy: The variable is written only in the setupCommWithIotJobs method. Multiple threads cannot write to this, since I am making sure that when setupCommWithIotJobs is invoked, the previous invocation is cancelled.

src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
```diff
@@ -133,20 +138,28 @@ public void onConnectionResumed(boolean sessionPresent) {
...
         this.mqttClient = mqttClient;
         this.executorService = executorService;
         this.deviceConfiguration = deviceConfiguration;
         this.iotShadowClient = iotShadowClient;
+        this.kernel = kernel;
     }
 
     @Override
     public void postInject() {
+        if (iotShadowClient == null) {
```
fahadmohammed01, line 158: why does this need to be in postInject?  shouldn't the shadow client be reconfigured when         deviceConfiguration changes?

src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
```diff
@@ -87,6 +87,11 @@
     public static final String DEVICE_OFFLINE_MESSAGE = "Device not configured to talk to AWS Iot cloud. "
             + "Single device deployment is offline";
     public static final String SUBSCRIBING_TO_SHADOW_TOPICS_MESSAGE = "Subscribing to Iot Shadow topics";
+
+    private static final String SHADOW_UPDATE_ACCEPTED_TOPIC = "$aws/things/{thingName}/shadow/update/accepted";
```
fahadmohammed01, line 91: We switched to named shadows, so the topic prefix would be aws/things/{thingName}/shadow/name/AWSManagedGreengrassV2Deployment

src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
abanthiy: The connection may need to be re-established upon device configuration changes. That is take care by the injected MqttClient object. For IotShadowClient it is abstracted.

src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
abanthiy: Updated

leaf94 (approved):

src/main/java/com/aws/greengrass/deployment/IotJobsHelper.java
```diff
@@ -324,22 +331,31 @@ private void connectToIotJobs(DeviceConfiguration deviceConfiguration)
...
                 Coerce.toString(deviceConfiguration.getThingName()));
-        if (subscriptionFuture != null && !subscriptionFuture.isDone()) {
-            subscriptionFuture.cancel(true);
-        }
+        this.thingName = Coerce.toString(deviceConfiguration.getThingName());
         if (subscriptionFuture == null || subscriptionFuture.isDone()) {
             subscriptionFuture = executorService.submit(() -> {
                 try {
+                    // Wait for all node updates to come through before we subscribe to the topics
+                    Throwable ex = kernel.getContext().runOnPublishQueueAndWait(() -> {});
```
MikeDombo, line 353: use waituntilclear instead of runonpublish queue

src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
```diff
@@ -185,18 +200,25 @@ private void connectToShadowService(DeviceConfiguration deviceConfiguration)
...
+        if (isSubscribedToShadowTopics.get()) {
+            unsubscribeToShadowTopics();
+        }
+        deploymentStatusKeeper.registerDeploymentStatusConsumer(DeploymentType.SHADOW,
+                this::deploymentStatusChanged, ShadowDeploymentListener.class.getName());
+        this.thingName = Coerce.toString(deviceConfiguration.getThingName());
         if (subscriptionFuture == null || subscriptionFuture.isDone()) {
             subscriptionFuture = executorService.submit(() -> {
+                // Wait for all node updates to come through before we subscribe to the topics
+                Throwable ex = kernel.getContext().runOnPublishQueueAndWait(() -> {});
```
MikeDombo, line 216: same

MikeDombo (approved):

PR #991, opened 2021-06-04
fix(sdk-clients): close AWS SDK clients when creating new copies

MikeDombo:
ignore some config change events to prevent us from creating new clients when not needed

**Issue #, if available:**

**Description of changes:**
Greengrass has been creating new SDK clients whenever a configuration changes, however we were not closing the client due to worries about closing a client that someone might be using. This causes an issue because the SDK http clients must be closed in order to remove them from the idle connection reaper thread's mapping. If they are not closed, then they will be referenced by the idle connection reaper forever and leak ~1MB/hr when doing continuous deployments.

An alternative solution would be to disable the idle connection reaper in our SDK clients.

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/util/GreengrassServiceClientFactory.java
+++ b/src/main/java/com/aws/greengrass/util/GreengrassServiceClientFactory.java
@@ -7,6 +7,7 @@ package com.aws.greengrass.util;
 
 import com.aws.greengrass.componentmanager.ClientConfigurationUtils;
 import com.aws.greengrass.config.Node;
+import com.aws.greengrass.config.WhatHappened;
 import com.aws.greengrass.deployment.DeviceConfiguration;
 import com.aws.greengrass.deployment.exceptions.DeviceConfigurationException;
 import com.aws.greengrass.logging.api.Logger;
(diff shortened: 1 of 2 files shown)


src/main/java/com/aws/greengrass/util/GreengrassServiceClientFactory.java
```diff
@@ -70,6 +74,10 @@ private boolean validPath(Node node, String key) {
         return validString(node, key) && Files.exists(Paths.get(key));
     }
 
+    public synchronized GreengrassV2DataClient getGreengrassV2DataClient() {
+        return greengrassV2DataClient;
+    }
```
rbattle, line 79: any chance that a reference to the old client is being used after it is closed?

rbattle: nevermind - I see in the description that was the reason this behavior wasn't added before - so there was a worry about it.

MikeDombo: Yes it is possible. However I checked all the usages to verify that they will always call get again. At worst, they would fail and then retry after getting a new client. That said, the client can only really change during a deployment which would be after the client has been used.

rbattle: yeah - I just checked that myself after I posted that comment - so it shouldn't have any impact

rbattle (approved):

PR #999, opened 2021-06-09
feat(recipe-interp): add {iot:thingName} variable for recipes

MikeDombo:
**Issue #, if available:**

**Description of changes:**
This change adds `{iot:thingName}` variable into recipe resolution. To be clear, this does not get interpolated into default configuration, only the lifecycle section of the recipe (as with all other recipe variables that we support). The purpose is to allow easy access to the thing name.

We should consider following up and allowing a similar variable replacement in our access control policies, especially for MQTT topic pub/sub where thing name is in the topic.

Updated existing config resolver test to use the new variable.

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/componentmanager/KernelConfigResolver.java
+++ b/src/main/java/com/aws/greengrass/componentmanager/KernelConfigResolver.java
@@ -62,6 +62,8 @@ public class KernelConfigResolver {
     public static final String VERSION_CONFIG_KEY = "version";
     public static final String PREV_VERSION_CONFIG_KEY = "previousVersion";
     public static final String CONFIGURATION_CONFIG_KEY = "configuration";
+    static final String IOT_NAMESPACE = "iot";
+    static final String THING_NAME_PATH = "thingName";
     static final String ARTIFACTS_NAMESPACE = "artifacts";
     static final String WORK_NAMESPACE = "work";
     static final String KERNEL_NAMESPACE = "kernel";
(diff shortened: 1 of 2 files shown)


src/main/java/com/aws/greengrass/componentmanager/KernelConfigResolver.java
```diff
@@ -112,20 +114,23 @@ public KernelConfigResolver(ComponentStore componentStore, Kernel kernel, Nucleu
         this.deviceConfiguration = deviceConfiguration;
 
         // More system parameters can be added over time by extending this map with new namespaces/keys
-        HashMap<String, CrashableFunction<ComponentIdentifier, String, IOException>> artifactNamespace =
-                new HashMap<>();
+        Map<String, CrashableFunction<ComponentIdentifier, String, IOException>> artifactNamespace = new HashMap<>();
```
philcali, line 117: nit: `HashMap` shorthand might help here:
```
systemParameters.put(ARTIFACTS_NAMESPACE, new HashMap<String, CrashableFunction<ComponentIdentifier, String, IOException>>() {{
    put(PATH_KEY, (id) -> nucleusPaths.artifactPath(id).toAbsolutePath().toString());
    put(DECOMPRESSED_PATH_KEY, (id) -> nucleusPaths.unarchiveArtifactPath(id).toAbsolutePath().toString());
}});
```
etc. May prevent some namespace -> value setting mistakes for system parameters.

philcali (approved):

PR #1000, opened 2021-06-09
perf(tes): prevent concurrent cloud credential requests through TES

MikeDombo:
**Issue #, if available:**

**Description of changes:**
TES: prevent concurrent cloud credential requests by producing a future which represents the single caller (if any). If any more threads come in and try to get credentials before the request has completed, then they will block on the pending request and return whatever value it ultimately comes back with.

Existing tests passing. New test to verify the functionality by hitting TES with 3 requests at the same time and verifying that only one of them goes to the cloud.

**Checklist:**


philcali (approved):

rbattle (approved):
nice!

PR #1003, opened 2021-06-09
fix(mqtt): decrease connect rate limit to adjust for IoT Core's limits

MikeDombo:
**Issue #, if available:**

**Description of changes:**
When a connect request hits the same host in IoT Core's fleet, the throttle limit is actually 0.1 TPS and not the stated 1 TPS. This discrepancy causes us to fail to connect and then backoff for 2 minutes. This slows down some of our UATs significantly and it is a negative customer experience. To avoid the issue, limit connect calls to 0.1 TPS per client ID instead of 0.25 as before.

**Checklist:**


fahadmohammed01:
So when we re-connect, we wait 10 mins, wouldnt this fail most of the UATs?

MikeDombo:
> So when we re-connect, we wait 10 mins, wouldnt this fail most of the UATs?

No, 10 seconds

fahadmohammed01 (approved):

fengwa-aws (approved):

PR #1005, opened 2021-06-18
fix: fix path validation in gg client factory. reset error message in s3 client factory

abanthiy:
…oning

**Issue #, if available:**
1. Validation for paths was not correct in GreengrassV2Client factory
2. S3 client does not get initialized if device is provisioned after kernel start.

**Description of changes:**
Fixes

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/util/GreengrassServiceClientFactory.java
+++ b/src/main/java/com/aws/greengrass/util/GreengrassServiceClientFactory.java
@@ -84,7 +84,7 @@ public class GreengrassServiceClientFactory {
     }
 
     private boolean validPath(Node node, String key) {
-        return validString(node, key) && Files.exists(Paths.get(key));
+        return validString(node, key) && Files.exists(Paths.get(Coerce.toString(node)));
     }
 
     public synchronized GreengrassV2DataClient getGreengrassV2DataClient() {
--- a/src/main/java/com/aws/greengrass/util/S3SdkClientFactory.java
+++ b/src/main/java/com/aws/greengrass/util/S3SdkClientFactory.java
@@ -34,16 +34,18 @@ public class S3SdkClientFactory {
      * @param credentialsProvider credential provider from TES
      */
     @Inject
-    @SuppressWarnings("PMD.AvoidCatchingGenericException")
+    @SuppressWarnings({"PMD.AvoidCatchingGenericException", "PMD.NullAssignment"})
     public S3SdkClientFactory(DeviceConfiguration deviceConfiguration, LazyCredentialProvider credentialsProvider) {
         this.credentialsProvider = credentialsProvider;
-        try {
-            deviceConfiguration.validate(true);
-        } catch (DeviceConfigurationException e) {
-            configValidationError = e.getMessage();
-        }
-        deviceConfiguration.getAWSRegion().subscribe((what, node) ->
-                this.s3Client = getClientForRegion(Region.of(Coerce.toString(deviceConfiguration.getAWSRegion()))));
+        deviceConfiguration.getAWSRegion().subscribe((what, node) -> {
+            try {
+                deviceConfiguration.validate();
+                configValidationError = null;
+            } catch (DeviceConfigurationException e) {
+                configValidationError = e.getMessage();
+            }
+            this.s3Client = getClientForRegion(Region.of(Coerce.toString(deviceConfiguration.getAWSRegion())));
+        });
     }
 
     /**


src/main/java/com/aws/greengrass/util/GreengrassServiceClientFactory.java
```diff
@@ -84,7 +84,7 @@ private boolean validString(Node node, String key) {
     }
 
     private boolean validPath(Node node, String key) {
-        return validString(node, key) && Files.exists(Paths.get(key));
+        return validString(node, key) && Files.exists(Paths.get(Coerce.toString(node)));
```
MikeDombo, line 87: node can be null

src/main/java/com/aws/greengrass/util/S3SdkClientFactory.java
```diff
@@ -34,16 +46,41 @@
      * @param credentialsProvider credential provider from TES
      */
     @Inject
-    @SuppressWarnings("PMD.AvoidCatchingGenericException")
+    @SuppressWarnings({"PMD.AvoidCatchingGenericException", "PMD.NullAssignment"})
     public S3SdkClientFactory(DeviceConfiguration deviceConfiguration, LazyCredentialProvider credentialsProvider) {
         this.credentialsProvider = credentialsProvider;
+        deviceConfiguration.onAnyChange((what, node) -> {
```
MikeDombo, line 52: this is not needed

src/main/java/com/aws/greengrass/util/S3SdkClientFactory.java
```diff
@@ -34,16 +46,41 @@
...
+            configValidationError = null;
+            if (node != null && what.equals(WhatHappened.childChanged) && relevantNodeChanged(node)) {
+                try {
+                    deviceConfiguration.validate(true);
+                    this.s3Client = getClientForRegion(Region.of(Coerce.toString(deviceConfiguration.getAWSRegion())));
```
MikeDombo, line 57: You must close the client

src/main/java/com/aws/greengrass/util/S3SdkClientFactory.java
```diff
@@ -34,16 +46,41 @@
...
+            }
+        });
         try {
             deviceConfiguration.validate(true);
         } catch (DeviceConfigurationException e) {
             configValidationError = e.getMessage();
         }
-        deviceConfiguration.getAWSRegion().subscribe((what, node) ->
-                this.s3Client = getClientForRegion(Region.of(Coerce.toString(deviceConfiguration.getAWSRegion()))));
+        this.s3Client = getClientForRegion(Region.of(Coerce.toString(deviceConfiguration.getAWSRegion())));
```
MikeDombo, line 70: you must close the client

src/main/java/com/aws/greengrass/util/S3SdkClientFactory.java
```diff
@@ -34,16 +46,41 @@
...
+    }
+
+    private boolean relevantNodeChanged(Node node) {
+        if (this.isClientInitialized.get()) {
+            return node.childOf(DEVICE_PARAM_AWS_REGION);
+        } else {
+            // List of configuration nodes that may change during device provisioning
+            return node.childOf(DEVICE_PARAM_THING_NAME) || node.childOf(DEVICE_PARAM_IOT_DATA_ENDPOINT)
+                    || node.childOf(DEVICE_PARAM_PRIVATE_KEY_PATH)
+                    || node.childOf(DEVICE_PARAM_CERTIFICATE_FILE_PATH) || node.childOf(DEVICE_PARAM_ROOT_CA_PATH)
```
MikeDombo, line 81: none of these matter for s3

src/main/java/com/aws/greengrass/util/S3SdkClientFactory.java
MikeDombo, line 52: this is not needed, the only thing which can affect s3 is when the region changes, but even then it doesn't matter

abanthiy: The validation used here is deviceConfiguration.validate() which validates all these parameters. If the validate() is false (which it will be in case device is not provisioned at the start) then the client will not be initialized. So need to ensure that when the device configuration becomes valid (which can happen with either of these parameter being set), the client is initialized.

src/main/java/com/aws/greengrass/util/GreengrassServiceClientFactory.java
abanthiy: validString takes care of that, if thats false then second one will not get evaluated.

src/main/java/com/aws/greengrass/util/S3SdkClientFactory.java
MikeDombo: Why wouldn't it be initialized? We catch the device validation error in the existing code.

src/main/java/com/aws/greengrass/util/GreengrassServiceClientFactory.java
MikeDombo: No, it won't. Paths.get will throw a InvalidPathException

src/main/java/com/aws/greengrass/util/S3SdkClientFactory.java
abanthiy: Have answered in the comment below

src/main/java/com/aws/greengrass/util/S3SdkClientFactory.java
abanthiy: Updated

src/main/java/com/aws/greengrass/util/S3SdkClientFactory.java
```diff
@@ -53,6 +91,14 @@ public S3SdkClientFactory(DeviceConfiguration deviceConfiguration, LazyCredentia
      * @return s3client
      */
     public S3Client getClientForRegion(Region r) {
+        Iterator<Map.Entry<Region, S3Client>> iterator = clientCache.entrySet().iterator();
+        while (iterator.hasNext()) {
```
MikeDombo, line 95: Why are you doing this?

These clients do not need to be closed

MikeDombo (dismissed):

src/main/java/com/aws/greengrass/util/S3SdkClientFactory.java
```diff
@@ -34,16 +34,18 @@
...
-        deviceConfiguration.getAWSRegion().subscribe((what, node) ->
-                this.s3Client = getClientForRegion(Region.of(Coerce.toString(deviceConfiguration.getAWSRegion()))));
+        deviceConfiguration.getAWSRegion().subscribe((what, node) -> {
+            this.s3Client = getClientForRegion(Region.of(Coerce.toString(deviceConfiguration.getAWSRegion())));
+            try {
+                deviceConfiguration.validate();
+                configValidationError = null;
+            } catch (DeviceConfigurationException e) {
+                configValidationError = e.getMessage();
+            }
```
MikeDombo, line 47: move this before doing the region thing

abanthiy: It does not matter, we are not preventing client creation if validation fails. At least thats how it is right now.

MikeDombo: It somewhat matters because the line above this check may throw an exception and thus the error message won't be updated with what the proper error is.

MikeDombo (approved):

leaf94 (approved):

PR #1007, opened 2021-06-21
fix: local deployment use thing group name

fahadmohammed01:
**Description of changes:**
1. For local deployments, the input from the CLI is thing group name (ex FirstGroup) while kernel uses  thing group resource name (ex thinggroup/FirstGroup) to track packages to thing group.
2. Components deployed via local deployments without a group name are excluded from the thing group removal check. 

### DIFF
--- a/src/main/java/com/aws/greengrass/deployment/DefaultDeploymentTask.java
+++ b/src/main/java/com/aws/greengrass/deployment/DefaultDeploymentTask.java
@@ -36,6 +36,7 @@ import java.util.concurrent.Future;
 
 import static com.aws.greengrass.deployment.DeploymentConfigMerger.DEPLOYMENT_ID_LOG_KEY;
 import static com.aws.greengrass.deployment.DeploymentService.GROUP_TO_ROOT_COMPONENTS_VERSION_KEY;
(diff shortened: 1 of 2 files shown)


shaguptashaikh (approved):

src/main/java/com/aws/greengrass/deployment/DefaultDeploymentTask.java
```diff
@@ -188,6 +189,7 @@ public DeploymentResult call() throws InterruptedException {
             // skip root packages if device does not belong to that group anymore
             if (!groupTopics.getName().equals(deploymentDocument.getGroupName())
                     && (groupTopics.getName().startsWith(DEVICE_DEPLOYMENT_GROUP_NAME_PREFIX)
+                    || groupTopics.getName().equals(LOCAL_DEPLOYMENT_GROUP_NAME)
```
tilo-chen, line 192: Should this be `THING_GROUP_RESOURCE_NAME_PREFIX + LOCAL_DEPLOYMENT_GROUP_NAME`?

fahadmohammed01: No, LOCAL_DEPLOYMENT_GROUP_NAME is used when group name is not specified for a local deployment. It is not a valid thing group and hence does not have the prefix.

tilo-chen (approved):

PR #1010, opened 2021-06-24
fix: make deployment service robust

fahadmohammed01:
**Description of changes:**
Any errors while parsing deployment document should only fail the deployment and not deployment service.

**Checklist:**


src/main/java/com/aws/greengrass/deployment/DeploymentService.java
```diff
@@ -620,12 +620,13 @@ private KernelUpdateDeploymentTask createKernelUpdateDeployment(Deployment deplo
...
     }
 
+    @SuppressWarnings("PMD.AvoidCatchingGenericException")
     private DefaultDeploymentTask createDefaultNewDeployment(Deployment deployment) {
         try {
             logger.atInfo().kv("document", deployment.getDeploymentDocument())
                     .log("Received deployment document in queue");
             parseAndValidateJobDocument(deployment);
-        } catch (InvalidRequestException e) {
+        } catch (Exception e) {
```
abanthiy, line 629: Add it to a different catch block as the messaging is not appropriate. Keep as much specific exception as we can. 
What exception are you seeing btw?

fahadmohammed01: we are seeing null pointer exception because version field for a root component in the deployment document can be null.

src/main/java/com/aws/greengrass/deployment/DeploymentDocumentDownloader.java
```diff
@@ -202,12 +199,14 @@ private void validateHttpExecuteResponse(HttpExecuteResponse executeResponse)
...
+    @SuppressWarnings("PMD.AvoidCatchingGenericException")
+    private DeploymentDocument deserializeDeploymentDoc(String configurationInString)
             throws DeploymentTaskFailureException {
         try {
-            return SerializerFactory.getFailSafeJsonObjectMapper()
+            Configuration configuration =  SerializerFactory.getFailSafeJsonObjectMapper()
                     .readValue(configurationInString, Configuration.class);
-        } catch (JsonProcessingException e) {
+            return DeploymentDocumentConverter.convertFromDeploymentConfiguration(configuration);
+        } catch (Exception e) {
```
abanthiy, line 209: I don't like using generic exception. Would prefer JsonProcessingException separate and InvalidRequestException separate.

src/main/java/com/aws/greengrass/deployment/DeploymentService.java
abanthiy: This is probably fine if you capture the specific exception before this.

src/main/java/com/aws/greengrass/deployment/DeploymentDocumentDownloader.java
fahadmohammed01: Discussed with Amit offline, This adds more context to the error message. Other wise the unhandled errors  gets propagated to Deployment service and would be surfaced to the customers (Like NPE). Catching it here and adding more context will make it more readable.

shaguptashaikh:
Talked offline that we will add unit tests for this parsing logic changes as fast follow since this needs to be urgently merged. Also, cloud side should do early validation and it shouldn't fail this late which will also be added

shaguptashaikh (approved):

src/main/java/com/aws/greengrass/deployment/converter/DeploymentDocumentConverter.java
```diff
@@ -198,17 +200,27 @@ private static String parseGroupNameFromConfigurationArn(Configuration config) {
...
+        }
+        return deploymentPackageConfiguration;
     }
 
     private static DeploymentPackageConfiguration convertComponent(String componentName,
-            ComponentUpdate componentUpdate) {
+            ComponentUpdate componentUpdate) throws InvalidRequestException {
+
+        if (componentName == null || componentName.isEmpty()) {
+            throw new InvalidRequestException("Target component name is empty");
```
leaf94, line 218: nit - Utils.isEmpty() so that is also validates blank

leaf94 (approved):

shaguptashaikh (approved):

PR #1013, opened 2021-06-28
fix: validate artifact uri from component recipe

abanthiy:
**Issue #, if available:**
Artifact Uri in component recipe was not being validated leading to NPE if the URI did not have a scheme.
**Description of changes:**
Adding the validation

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/componentmanager/builtins/ArtifactDownloaderFactory.java
+++ b/src/main/java/com/aws/greengrass/componentmanager/builtins/ArtifactDownloaderFactory.java
@@ -19,6 +19,7 @@ import com.aws.greengrass.util.S3SdkClientFactory;
 
 import java.net.URI;
 import java.nio.file.Path;
+import java.util.Arrays;
 import java.util.List;
 import java.util.stream.Collectors;
 import javax.inject.Inject;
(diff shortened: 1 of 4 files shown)


src/main/java/com/aws/greengrass/componentmanager/builtins/ArtifactDownloaderFactory.java
```diff
@@ -108,22 +111,34 @@ public void checkDownloadPrerequisites(List<ComponentArtifact> artifacts, List<C
...
                 }
+            } catch (InvalidArtifactUriException e) {
+                throw new PackageLoadingException(String
+                        .format("Failed to download due to bad artifact URI: %s", artifact.getArtifactUri()), e);
             }
         }
     }
+
+    private void validateArtifactUri(ComponentArtifact artifact) throws InvalidArtifactUriException {
+        if (artifact.getArtifactUri() == null) {
```
MikeDombo, line 134: Add the component name/version to the error message

abanthiy: With the current method signature that mapping is not available. ComponentArtifact does not store the componentId/name.

MikeDombo: Is it possible to add? I can foresee that that will be confusing without knowing what component has the issue

MikeDombo (approved):

MikeDombo:
What are we doing with this PR?

rbattle (approved):

PR #1017, opened 2021-07-02
fix(install): relink KernelAlternatives to component store during first install

MikeDombo:
**Issue #, if available:**

**Description of changes:**
Currently our installation copies the artifacts into the proper artifact storage, however the KernelAlternatives are not moved to symlink to that location, instead they point to the initial location. This is error prone as customers may delete these files, thus breaking their installation.

Only the `init` path will be changed. The `init` path is only ever used during the setup, and `current` will point to init or whichever is the proper place for it to be pointing. This change ensures that `current` will never be redirected to point to `init` except during the very first setup which is how it works prior to this change.

Manually tested that the symlink points to the proper location inside the unarchive path.

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
+++ b/src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
@@ -324,8 +324,9 @@ public class DeviceConfiguration {
         kernel.getConfig().lookup(SETENV_CONFIG_NAMESPACE, GGC_VERSION_ENV).dflt(nucleusComponentVersion);
     }
 
-    void initializeComponentStore(String nucleusComponentName, Semver componentVersion, Path recipePath,
-                                          Path unpackDir) throws IOException, PackageLoadingException {
+    void initializeComponentStore(KernelAlternatives kernelAlts, String nucleusComponentName,
+                                  Semver componentVersion, Path recipePath,
+                                  Path unpackDir) throws IOException, PackageLoadingException {
         // Copy recipe to component store
         ComponentStore componentStore = kernel.getContext().get(ComponentStore.class);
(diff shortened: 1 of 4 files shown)


src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
```diff
@@ -345,6 +346,9 @@ void initializeComponentStore(String nucleusComponentName, Semver componentVersi
         Permissions.setArtifactPermission(destinationArtifactPath, FileSystemPermission.builder()
                 .ownerRead(true).ownerExecute(true).groupRead(true).groupExecute(true)
                 .otherRead(true).otherExecute(true).build());
+        // Relink the alts path to point to the artifact since we've just installed. This will allow the customer to
+        // delete their unzipped Nucleus distribution.
+        kernelAlts.relinkInitLaunchDir(destinationArtifactPath);
```
shaguptashaikh, line 351: Why does this need to be called in two places? Both `initializeComponentStore` or `setupInitLaunchDirIfAbsent` get invoked during nucleus launch

MikeDombo: The alts directory is initially setup in the constructor of KernelAlternatives and the initial value points to the directory it is launched from because that happens earlier than this call and because this call will not always happen.
This change is needed so that if we start and we install to the component store (which is not a given) then we will correct the alts directory. If we are not installing or we cannot find the launch dir as expected then the behavior is unchanged.

shaguptashaikh (dismissed):

fahadmohammed01 (approved):

shaguptashaikh (approved):

fahadmohammed01 (approved):

PR #1031, opened 2021-07-23
fix: reset lifecycle thread name back to default when thread quits

MikeDombo:
**Issue #, if available:**

**Description of changes:**
Reset lifecycle thread name when the thread quits. If the thread was recycled due to the caching thread pool, then some other task would end up using an improperly named thread which can be confusing.

**Checklist:**


shaguptashaikh (approved):

shaguptashaikh:
Should we add it to the release branch?

MikeDombo:
> Should we add it to the release branch?

It wouldn't hurt....

abanthiy (approved):
Good find!

PR #1032, opened 2021-07-26
fix(resource-mgt): Fixing a few cases for resource limits and hibernation

shaguptashaikh:
**Issue #, if available:**
-

**Description of changes:**
- Change error message for pause/resume requests for non generic components.
- Reconfigure system resource limits on startup/run script execution in case limits got cleaned up during last shutdown
- Add another case in resource limit update test
- Include a fix from mainline in handling names for pooled threads

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/builtin/services/lifecycle/LifecycleIPCEventStreamAgent.java
+++ b/src/main/java/com/aws/greengrass/builtin/services/lifecycle/LifecycleIPCEventStreamAgent.java
@@ -370,7 +370,7 @@ public class LifecycleIPCEventStreamAgent {
                 if (component instanceof GenericExternalService) {
                     target = (GenericExternalService) component;
                 } else {
-                    throw new InvalidArgumentsError("Only external components can be paused.");
+                    throw new InvalidArgumentsError("Only generic components can be paused.");
                 }
 
                 if (State.RUNNING.equals(target.getState())) {
@@ -438,7 +438,7 @@ public class LifecycleIPCEventStreamAgent {
             if (component instanceof GenericExternalService) {
                 target = (GenericExternalService) component;
             } else {
-                throw new InvalidArgumentsError("Only external components can be resumed.");
+                throw new InvalidArgumentsError("Only generic components can be resumed.");
             }
 
             if (target.isPaused()) {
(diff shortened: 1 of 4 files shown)


src/main/java/com/aws/greengrass/builtin/services/lifecycle/LifecycleIPCEventStreamAgent.java
```diff
@@ -352,6 +352,12 @@ public PauseComponentResponse handleRequest(PauseComponentRequest request) {
                     throw new InvalidArgumentsError("Component name is required.");
                 }
 
+                // A component should not be allowed to pause itself because it will never be resumed and can
+                // only be revived via a restart. If it needs it should just exit instead of trying to pause itself.
+                if (componentName.equals(serviceName)) {
```
MikeDombo, line 357: why not? This seems like something people will want...
It can pause itself and be resumed by someone else right?

src/main/java/com/aws/greengrass/builtin/services/lifecycle/LifecycleIPCEventStreamAgent.java
```diff
@@ -370,7 +376,7 @@ public PauseComponentResponse handleRequest(PauseComponentRequest request) {
                 if (component instanceof GenericExternalService) {
```
MikeDombo, line 376: this check includes lambdas, are we good with that?

src/main/java/com/aws/greengrass/builtin/services/lifecycle/LifecycleIPCEventStreamAgent.java
shaguptashaikh: Currently no other component will know about it being paused. We can easily allow this later if we add a mechanism for knowing if a component is paused, either through a state or a new IPC API.

MikeDombo: What's the point of limiting this though? Customers _could_ have their own mechanism, like sending a pubsub message before pausing.

shaguptashaikh: Changed it back to how it was

src/main/java/com/aws/greengrass/builtin/services/lifecycle/LifecycleIPCEventStreamAgent.java
shaguptashaikh: Yes

MikeDombo:
Your changes need to go into main right? This is merging directly to release

MikeDombo (approved):

shaguptashaikh:
> Your changes need to go into main right? This is merging directly to release

Yeah need to merge this asap, will add the new commit to main in separate PR

abanthiy (approved):

rbattle (approved):

PR #1041, opened 2021-08-05
fix: loader script reboot

tilo-chen:
**Issue #, if available:**

**Description of changes:**
Exit immediately after calling `sudo reboot`. Otherwise the script continues to run until the end. It'll create the `alts/broken` symlink and makes nucleus think it failed.

Without the fix, deployment with any component that requests reboot will fail.

Manual.

**Checklist:**

### DIFF
--- a/scripts/loader
+++ b/scripts/loader
@@ -79,6 +79,7 @@ while [ $j -le 3 ] && [ $sigterm_received -eq 0 ];  do
   101)
     echo "Rebooting host"
     sudo reboot
+    exit 0
     ;;
   *)
     echo "Nucleus exited ${kernel_exit_code}. Retrying $j times"


scripts/loader
```diff
@@ -79,6 +79,7 @@ while [ $j -le 3 ] && [ $sigterm_received -eq 0 ];  do
   101)
     echo "Rebooting host"
     sudo reboot
+    exit 101
```
MikeDombo, line 82: This will cause systemd to restart us which may or may not happen before the reboot. I think you need to just pause here

tilo-chen: After calling reboot, systemd will try to stop the service, so I think here systemd won't restart it after exit.
But how about just `exit 0` here, because we finished successfully as far as systemd is concerned?

MikeDombo: That's probably OK, yes

MikeDombo (approved):

shaguptashaikh (approved):

PR #1044, opened 2021-08-06
fix(logging): drop log level for many log messages

### LINKED ISSUE #1002: (com.aws.greengrass.Nucleus): Reduce volume of INFO level logging
**Feature Description**
Thousands of log lines are written into the `greengrass.log` log file for even relatively simple operations such as deployments. The logging volume should be reduced to increase usability of the logs and reduce wear on embedded storage.

**Use Case**
Embedded solutions may often run on devices which not only are storage-constrained, but the storage may also have limited write cycles or poor wear leveling, such as is the case for inexpensive eMMC storage.

While logs are important for monitoring the status of the device and especially crucial for troubleshooting, and log rotation reduces the logs stored to a manageable amount, a large volume of logging will wear down embedded device storage faster.

**Proposed Solution**
Move all current INFO level logging to the DEBUG log level and only hand-pick the most useful log messages back to INFO.

**Other**

* [ ] :wave: I may be able to implement this feature request
* [ ] :warning: This feature might incur a breaking change

MikeDombo:
**Issue #, if available:**
Resolves #1002 

**Description of changes:**
Dropping log level from info to debug or trace when appropriate. This reduces the amount of logs that we and customer need to troll through when debugging. Logs are of course still available to debug or trace level if we need to dig deeper.

In our unit and integ tests run by GitHub this PR reduces the number of log lines by 25% from ~42,000 to 32,000.

Verified E2E tests locally

**Checklist:**

### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/e2e/BaseE2ETestCase.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/e2e/BaseE2ETestCase.java
@@ -25,6 +25,8 @@ import com.aws.greengrass.util.IamSdkClientFactory;
 import com.aws.greengrass.util.IotSdkClientFactory;
 import com.aws.greengrass.util.RegionUtils;
 import com.aws.greengrass.util.Utils;
+import com.aws.greengrass.util.platforms.Platform;
+import com.aws.greengrass.util.platforms.unix.DarwinPlatform;
 import com.vdurmont.semver4j.Semver;
 import org.apache.commons.io.FileUtils;
 import org.junit.jupiter.api.AfterAll;
@@ -226,7 +228,11 @@ public class BaseE2ETestCase implements AutoCloseable {
     }
 
     public static void setDefaultRunWithUser(Kernel kernel) {
-        new DeviceConfiguration(kernel).getRunWithDefaultPosixUser().dflt("nobody");
+        String user = "nobody";
+        if (Platform.getInstance() instanceof DarwinPlatform) {
+            user = System.getProperty("user.name");
+        }
+        new DeviceConfiguration(kernel).getRunWithDefaultPosixUser().dflt(user);
     }
 
     protected void initKernel()
--- a/src/main/java/com/aws/greengrass/authorization/AuthorizationPolicyParser.java
+++ b/src/main/java/com/aws/greengrass/authorization/AuthorizationPolicyParser.java
@@ -47,7 +47,7 @@ public final class AuthorizationPolicyParser {
         Topics allServices = kernel.getConfig().findTopics(SERVICES_NAMESPACE_TOPIC);
 
         if (allServices == null) {
-            logger.atInfo("load-authorization-all-services-component-config-retrieval-error")
+            logger.atWarn("load-authorization-all-services-component-config-retrieval-error")
                     .log("Unable to retrieve services config");
             return primaryAuthorizationPolicyMap;
         }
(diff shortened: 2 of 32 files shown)


src/integrationtests/java/com/aws/greengrass/integrationtests/e2e/BaseE2ETestCase.java
```diff
@@ -226,7 +228,11 @@ protected BaseE2ETestCase() throws Exception {
     }
 
     public static void setDefaultRunWithUser(Kernel kernel) {
-        new DeviceConfiguration(kernel).getRunWithDefaultPosixUser().dflt("nobody");
+        String user = "nobody";
+        if (Platform.getInstance() instanceof DarwinPlatform) {
+            user = System.getProperty("user.name");
```
tilo-chen, line 233: Curious why Mac requires an actual valid user

tilo-chen (approved):

src/integrationtests/java/com/aws/greengrass/integrationtests/e2e/BaseE2ETestCase.java
rbattle: `nobody` is a valid user on the mac too - but the default mac sudoers file sets it up so you need to enter the current users password before switching

src/main/java/com/aws/greengrass/authorization/AuthorizationPolicyParser.java
```diff
@@ -47,7 +47,7 @@
         Topics allServices = kernel.getConfig().findTopics(SERVICES_NAMESPACE_TOPIC);
 
         if (allServices == null) {
-            logger.atInfo("load-authorization-all-services-component-config-retrieval-error")
+            logger.atWarn("load-authorization-all-services-component-config-retrieval-error")
```
rbattle, line 50: :+1:

src/main/java/com/aws/greengrass/builtin/services/lifecycle/LifecycleIPCEventStreamAgent.java
```diff
@@ -236,9 +233,10 @@ public DeferComponentUpdateResponse handleRequest(DeferComponentUpdateRequest re
                 if (deferComponentUpdateRequestFuture == null) {
                     throw new ServiceError("Time limit to respond to PreComponentUpdateEvent exceeded");
                 } else {
+                    log.atDebug().log("Processing deployment deferral from {} for deployment {}", serviceName,
```
rbattle, line 236: question: do we log somewhere that the deployment is deferred?

src/main/java/com/aws/greengrass/deployment/IotJobsHelper.java
```diff
@@ -361,7 +361,6 @@ private void setupCommWithIotJobs() {
                     logger.atWarn().log("Interrupted while subscribing to Iot Jobs topics");
                     return;
                 }
-                logger.atInfo().log("Connection established to IoT cloud");
```
rbattle, line 364: question: do we log somewhere when we are connected? In the MQTT client itself?

MikeDombo: Yeah, in the MQTT client.

src/main/java/com/aws/greengrass/builtin/services/lifecycle/LifecycleIPCEventStreamAgent.java
MikeDombo: Yes, https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/1171b2eb3e39917c1e9273d6d05d2f9eb344f8ba/src/main/java/com/aws/greengrass/lifecyclemanager/UpdateSystemPolicyService.java#L202

rbattle (approved):

PR #1045, opened 2021-08-06
fix(loader): handle exit codes correctly

fahadmohammed01:
**Description of changes:**
Loader.cmd handles the nucleus exit codes properly. 

Tested manually on windows dev desktop

**Checklist:**

### DIFF
--- a/scripts/loader.cmd
+++ b/scripts/loader.cmd
@@ -60,21 +60,24 @@ SET /A MAX_RETRIES=3
 @REM Attempt to start the nucleus 3 times
 FOR /L %%i IN (1,1,%MAX_RETRIES%) DO (
     java -Dlog.store=FILE %JVM_OPTIONS% -jar "%LAUNCH_DIR%\distro\lib\Greengrass.jar" %OPTIONS%
-    SET KERNEL_EXIT_CODE=%ERRORLEVEL%
+    SET KERNEL_EXIT_CODE=!ERRORLEVEL!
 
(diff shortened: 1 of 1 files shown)


MikeDombo (approved):

scripts/loader.cmd
```diff
@@ -60,21 +60,24 @@ SET /A MAX_RETRIES=3
...
 FOR /L %%i IN (1,1,%MAX_RETRIES%) DO (
     java -Dlog.store=FILE %JVM_OPTIONS% -jar "%LAUNCH_DIR%\distro\lib\Greengrass.jar" %OPTIONS%
-    SET KERNEL_EXIT_CODE=%ERRORLEVEL%
+    SET KERNEL_EXIT_CODE=!ERRORLEVEL!
 
-    IF KERNEL_EXIT_CODE EQU 0 (
+    IF !KERNEL_EXIT_CODE! EQU 0 (
         ECHO Restarting Nucleus
-        %LAUNCH_DIR%\distroy\bin\loader.cmd
+        %LAUNCH_DIR%\distro\bin\loader.cmd
```
fufranci, line 67: I don't know when to use `%` and when to use `!`. If the `%` here intentional?

MikeDombo: https://stackoverflow.com/questions/14347038/why-are-my-set-commands-resulting-in-nothing-getting-stored/14347131#14347131

Using `!` is probably what we should have as the default, it expands at runtime instead of at parse time

fufranci (approved):

PR #1049, opened 2021-08-10
fix: do not block local deployment on getting thing group hierarchy w…

shaguptashaikh:
…hen offline

**Issue #, if available:**
-

**Description of changes:**
Currently when a device is offline, a deployment infinitely retries to get thing group hierarchy from cloud. This blocks local deployments too but they should be allowed to go through. As the group hierarchy API call is made on each deployment, this would mean no local deployments would ever go through when once a device is offline. This PR changes that by using finite no of retries for local deployments

See description

Integ tests, added a new for the scenario mentioned in description

**Checklist:**


MikeDombo (dismissed):

src/main/java/com/aws/greengrass/deployment/DefaultDeploymentTask.java
```diff
@@ -173,25 +179,45 @@ public DeploymentResult call() throws InterruptedException {
...
+        int retryCount = isLocalDeployment ? FINITE_RETRY_COUNT : INFINITE_RETRY_COUNT;
+
+        Optional<Set<String>> groupsForDeviceOpt = Optional.empty();
+        AtomicBoolean useLocalMapping = new AtomicBoolean();
+        try {
+            groupsForDeviceOpt = thingGroupHelper.listThingGroupsForDevice(retryCount);
+        } catch (Exception e) {
+            if (isLocalDeployment && ThingGroupHelper.DEVICE_OFFLINE_INDICATIVE_EXCEPTIONS.contains(e.getClass())) {
+                logger.atWarn().setCause(e).log("Failed to get thing group hierarchy, local deployment will proceed");
+                useLocalMapping.set(true);
```
fahadmohammed01, line 200: Do we need this ? setting groupsForDeviceOpt as Optional.empty() should have the desired effect right?

shaguptashaikh: That will end up removing the device from all groups

fahadmohammed01 (approved):

MikeDombo (approved):

PR #1053, opened 2021-08-11
refactor(exec): draft direction

tilo-chen:
**Issue #, if available:**

**Description of changes:**
- Interface change: `new Exec()` -> to `Platform.getInstance().getNewProcessRunner()`

Accommodate Windows run as different user.

**Checklist:**


src/main/java/com/aws/greengrass/lifecyclemanager/ShellRunner.java
```diff
@@ -19,14 +19,7 @@
...
 
-public interface ShellRunner {
-
-    Exec setup(String note, String command, GreengrassService onBehalfOf) throws IOException;
-
-    boolean successful(Exec e, String note, IntConsumer background, GreengrassService onBehalfOf)
-            throws InterruptedException;
-
-    class Default implements ShellRunner {
+public class ShellRunner {
```
MikeDombo, line 22: I don't think this change should be necessary. The ShellRunner should be able to call for a Platform, either static or via injection

src/main/java/com/aws/greengrass/util/platforms/unix/UnixExec.java
```diff
@@ -0,0 +1,140 @@
...
+    @Override
+    public Optional<Integer> exec() throws InterruptedException, IOException  {
+        // Don't run anything if the current thread is currently interrupted
+        if (Thread.currentThread().isInterrupted()) {
+            logger.atWarn().kv("command", this).log("Refusing to execute because the active thread is interrupted");
+            throw new InterruptedException();
+        }
+        final String[] command = getCommand();
+        logger.atTrace().kv("command", (Supplier<String>) () -> String.join(" ", command)).log();
+        process = Runtime.getRuntime().exec(command, environment, dir);
```
MikeDombo, line 110: While we're making changes, we should use ProcessBuilder instead of runtime exec

tilo-chen:
> Did you get all the usages? I think we had quite a few cases of `new Exec()` or one Exec's many static methods

I haven't changed the usage code. Will change all of them if we're good with this interface

src/main/java/com/aws/greengrass/lifecyclemanager/ShellRunner.java
tilo-chen: I'm also considering just move the methods here into `GenericExternalService` because that's the only usage right now, for setting up some Exec options. Not sure what was the original purpose of this interface.

src/main/java/com/aws/greengrass/util/platforms/Platform.java
```diff
@@ -93,6 +93,8 @@ public static synchronized Platform getInstance() {
 
     public abstract SystemResourceController getSystemResourceController();
 
+    public abstract Exec getNewProcessRunner();
```
rbattle, line 96: nit: this should be `createNewProcessRunner()` - we have some poorly named methods in this class (my fault)

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsExec.java
```diff
@@ -0,0 +1,62 @@
...
+import java.lang.reflect.Method;
+import java.nio.file.Path;
+import java.util.Map;
+import javax.annotation.Nullable;
+
+@SuppressWarnings("PMD.AvoidCatchingThrowable")
+public class WindowsExec extends Exec {
+    @Nullable
+    @Override
+    public Path which(String fn) {
```
tilo-chen, line 19: WIP. Should be able to do similar as UnixExec

src/integrationtests/java/com/aws/greengrass/integrationtests/util/ConfigPlatformResolver.java
```diff
@@ -76,7 +76,7 @@ public static void initKernelWithMultiPlatformConfig(Kernel kernel, URL path) th
         }
         if (!isWindows) {
             try {
-                String sysver = Exec.sh("uname -a").toLowerCase();
+                String sysver = Platform.getInstance().createNewProcessRunner().sh("uname -a").toLowerCase();
```
MikeDombo, line 79: Add a TODO for windows if there isn't one.

src/integrationtests/java/com/aws/greengrass/integrationtests/util/ExecTest.java
```diff
@@ -36,21 +41,43 @@ private String readLink(String path) throws IOException {
...
+    void Given_windows_exec_WHEN_commands_executed_using_static_methods_THEN_success() throws InterruptedException,
+            IOException {
+        try (Exec exec = Platform.getInstance().createNewProcessRunner()) {
+            final String command = "powershell -command pwd";
+            String s = exec.cmd(command);
+            assertFalse(s.contains("\n"));
+            assertTrue(s.contains("Path"));
+            assertEquals(s, exec.sh("pwd"));
+            // test change shell
+            s = exec.usingShell("cmd").cmd("cd");
```
MikeDombo, line 72: isn't cmd the default? Changing to powershell would be more interesting, right?

src/main/java/com/aws/greengrass/easysetup/GreengrassSetup.java
```diff
@@ -519,7 +519,7 @@ void provision(Kernel kernel) throws IOException, DeviceConfigurationException {
 
     @SuppressWarnings("PMD.PreserveStackTrace")
     private void setComponentDefaultUserAndGroup(DeviceConfiguration deviceConfiguration) {
-        if (Exec.isWindows) {
+        if (PlatformResolver.isWindows) {
```
MikeDombo, line 522: `Exec.isWindows` was added into the CLI by Francis, let's replace that with a Platform lookup since you're moving exec.

src/main/java/com/aws/greengrass/util/platforms/Exec.java
```diff
@@ -57,74 +55,43 @@
...
-                bg.start();
-                bg.join(2000);
-                addPathEntries(path.toString().trim());
-                // Ensure some level of sanity
-                ensurePresent("/bin", "/usr/bin", "/sbin", "/usr/sbin");
-            }
-        } catch (Throwable ex) {
-            staticLogger.atError().log("Error while initializing PATH", ex);
-        }
-        computePathString();
```
MikeDombo, line 107: you must keep computePathString

src/main/java/com/aws/greengrass/util/platforms/Exec.java
```diff
@@ -140,24 +107,24 @@ public Exec logger(Logger logger) {
         return this;
     }
 
-    public static String cmd(String... command) throws InterruptedException, IOException {
-        return new Exec().withExec(command).execAndGetStringOutput();
+    public String cmd(String... command) throws InterruptedException, IOException {
```
MikeDombo, line 110: can we clean up some of these overrides? They get so complicated; way too many ways to call a command. I'd basically only keep what is used by ShellRunner + one method to run without a shell.

src/integrationtests/java/com/aws/greengrass/integrationtests/util/ConfigPlatformResolver.java
tilo-chen: Yes there is a `if (isWindows)` above this one. Good reminder for me to search whole codebase for "if (!isWindows)"!

src/main/java/com/aws/greengrass/util/platforms/Exec.java
```diff
@@ -360,31 +293,29 @@ public Exec withErr(Consumer<CharSequence> o) {
...
         // Don't run anything if the current thread is currently interrupted
         if (Thread.currentThread().isInterrupted()) {
             logger.atWarn().kv("command", this).log("Refusing to execute because the active thread is interrupted");
             throw new InterruptedException();
         }
         final String[] command = getCommand();
         logger.atTrace().kv("command", (Supplier<String>) () -> String.join(" ", command)).log();
-        ProcessBuilder pb = new ProcessBuilder();
-        pb.environment().putAll(environment);
-        process = pb.directory(dir).command(command).start();
```
MikeDombo, line 387: I'd keep the environment and directory setting here in the super.

```
proc = new ProcessBuilder(); 
<set env>
<set dir>
 createProcess(proc, command).start()
```
^ that will the subclass to override behavior if it wants, but it makes the defaults sane.

src/main/java/com/aws/greengrass/util/platforms/Platform.java
```diff
@@ -70,14 +70,14 @@ public static synchronized Platform getInstance() {
...
-                                                        UserDecorator decorator)
+                                                        UserOptions decorator)
             throws IOException, InterruptedException;
 
     public abstract ShellDecorator getShellDecorator();
 
     public abstract int exitCodeWhenCommandDoesNotExist();
 
-    public abstract UserDecorator getUserDecorator();
+    public abstract UserOptions getUserDecorator();
```
MikeDombo, line 80: Don't change the name of the type, this is backwards incompatible

src/main/java/com/aws/greengrass/util/platforms/UserOptions.java
```diff
@@ -5,21 +5,34 @@
 
 package com.aws.greengrass.util.platforms;
 
+import lombok.Getter;
+
 /**
  * Decorate a command to run as another user or group.
  */
-public interface UserDecorator extends CommandDecorator {
```
MikeDombo, line 11: Undo this change, it is not compatible

src/main/java/com/aws/greengrass/util/platforms/unix/UnixExec.java
```diff
@@ -0,0 +1,91 @@
...
+                sb.append(PATH_SEP);
+            }
+            sb.append(p.toString());
+        });
+        setDefaultEnv(PATH, sb.toString());
+    }
+
+    @Nullable
+    @Override
+    public Path which(String fn) {
```
MikeDombo, line 69: this method might nearly work on windows with only minor tweaks

src/main/java/com/aws/greengrass/util/platforms/unix/UnixExec.java
```diff
@@ -0,0 +1,91 @@
...
+        }
+        return null;
+    }
+
+    @Override
+    protected Process createProcess(String[] command) throws IOException {
+        ProcessBuilder pb = new ProcessBuilder();
+        pb.environment().putAll(environment);
+        process = pb.directory(dir).command(command).start();
+        return process;
```
MikeDombo, line 89: You need to decorate the command with the user info, that cannot happen in the superclass anymore since windows is different.

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsExec.java
```diff
@@ -0,0 +1,62 @@
...
+        process.start(String.join(" ", command));
+        return process;
+    }
+
+    /**
+     * Convert environment Map to lpEnvironment block format.
+     * @return environment block for starting a process
+     */
+    private String computeEnvironmentBlock() {
+        // TODO: is reflection good here or should we copy over the source code
```
MikeDombo, line 41: if the source isn't bad, let's copy it over. No guarantee that this isn't changed in future JDKs

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsExec.java
```diff
@@ -0,0 +1,62 @@
...
+                        "toEnvironmentBlock".equals(method.getName()) && method.getParameterTypes().length == 1
+                                && method.getParameterTypes()[0].equals(Map.class);
+                if (correctMethod) {
+                    method.setAccessible(true);
+                    block = (String) method.invoke(null, environment);
+                    break;
+                }
+            }
+        } catch (Throwable t) {
+            staticLogger.atError().setCause(t)
```
MikeDombo, line 57: why are you using the static logger in a non-static method?

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsRunasProcess.java
```diff
@@ -0,0 +1,89 @@
...
+    @Override
+    public InputStream getErrorStream() {
+        return errorStream;
+    }
+
+    @Override
+    public synchronized int waitFor() throws InterruptedException {
+        while (!hasExited) {
+            wait();
+        }
```
MikeDombo, line 73: not great... but this is all TODO I assume

src/test/java/com/aws/greengrass/util/platforms/windows/WindowsPlatformTest.java
```diff
@@ -53,9 +53,8 @@
 
     @Test
     void GIVEN_command_WHEN_decorate_THEN_is_decorated() {
-        assertThat(new WindowsPlatform.CmdDecorator()
-                        .decorate("echo", "hello"),
-                is(arrayContaining("cmd.exe", "/C", "echo", "hello")));
+        assertThat(new WindowsPlatform.CmdDecorator().decorate("echo", "hello"),
+                is(arrayContaining("powershell", "-Command", "echo", "hello")));
```
MikeDombo, line 57: ps is the default now?

src/integrationtests/java/com/aws/greengrass/integrationtests/util/ConfigPlatformResolver.java
MikeDombo: Yes I saw the if, I mean to add a TODO so we don't skip implementing this feature on windows

MikeDombo: Looking at the context now though, it doesn't matter in this case.

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsExec.java
```diff
@@ -0,0 +1,106 @@
...
+import javax.annotation.Nullable;
+
+@SuppressWarnings("PMD.AvoidCatchingThrowable")
+public class WindowsExec extends Exec {
+    public static final String PATHEXT_KEY = "PATHEXT";
+
+    private static final Logger logger = LogManager.getLogger(WindowsExec.class);
+    private static final List<String> PATHEXT;  // ordered file extensions to try, when no extension is provided
+
+    static {
```
tilo-chen, line 30: Do we need the same hack as UNIX to grab PATH by opening a shell?

src/test/java/com/aws/greengrass/util/platforms/windows/WindowsPlatformTest.java
tilo-chen: My bad. Reverted everything to cmd

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsRunasProcess.java
tilo-chen: Draft for now... Copied this from java.lang.UNIXProcess though.

src/main/java/com/aws/greengrass/util/platforms/Exec.java
tilo-chen: Good idea but the Windows implementation won't use `ProcessBuilder` at all...

src/main/java/com/aws/greengrass/util/platforms/Exec.java
tilo-chen: Agreed there's too many ways. Added TODO to reduce this PR scope a bit

src/main/java/com/aws/greengrass/easysetup/GreengrassSetup.java
tilo-chen: https://github.com/aws-greengrass/aws-greengrass-cli/pull/142

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsExec.java
MikeDombo: No, not clear we still need it on linux either.

src/main/java/com/aws/greengrass/util/Utils.java
```diff
@@ -212,7 +214,7 @@ public static Path homePath(String s) {
     public static String extension(String s) {
         if (s != null) {
             int dp = s.lastIndexOf('.');
-            if (dp > s.lastIndexOf('/')) {
+            if (dp > s.lastIndexOf(File.separatorChar)) {
```
MikeDombo, line 217: not necessarily true, you can use `/` on windows too. We should account for either

MikeDombo: `DefaultFileSystem.getFileSystem().normalize(s)` will ensure the path is properly formatted for the given filesystem

src/main/java/com/aws/greengrass/util/platforms/Exec.java
```diff
@@ -369,22 +314,18 @@ public Exec withErr(Consumer<CharSequence> o) {
...
      * Execute a command.
      *
-     * @returns the process exit code if it is not a background process.
+     * @return the process exit code if it is not a background process.
      * @throws InterruptedException if the command is interrupted while running.
      * @throws IOException if an error occurs while executing.
      */
     @SuppressWarnings("PMD.AvoidRethrowingException")
-    public Optional<Integer> exec() throws InterruptedException, IOException {
+    public Optional<Integer> exec() throws InterruptedException, IOException  {
```
MikeDombo, line 322: fix format

src/main/java/com/aws/greengrass/util/platforms/unix/UnixExec.java
```diff
@@ -0,0 +1,85 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import javax.annotation.Nullable;
+
+@SuppressWarnings("PMD.AvoidCatchingThrowable")
+public class UnixExec extends Exec {
+    private static final Logger logger = LogManager.getLogger(UnixExec.class);
```
MikeDombo, line 21: don't confuse things. This is a static logger, we also keep a logger which is specific to the instance of `Exec`

MikeDombo: Rename to staticLogger to keep that distinction very clear

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsExec.java
```diff
@@ -0,0 +1,106 @@
...
+import java.util.List;
+import java.util.Map;
+import javax.annotation.Nullable;
+
+@SuppressWarnings("PMD.AvoidCatchingThrowable")
+public class WindowsExec extends Exec {
+    public static final String PATHEXT_KEY = "PATHEXT";
+
+    private static final Logger logger = LogManager.getLogger(WindowsExec.class);
+    private static final List<String> PATHEXT;  // ordered file extensions to try, when no extension is provided
```
MikeDombo, line 28: why? Is this some standard?

`PATH` is a valid environment variable on windows AFAIK

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsExec.java
```diff
@@ -0,0 +1,106 @@
...
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import javax.annotation.Nullable;
+
+@SuppressWarnings("PMD.AvoidCatchingThrowable")
+public class WindowsExec extends Exec {
+    public static final String PATHEXT_KEY = "PATHEXT";
+
+    private static final Logger logger = LogManager.getLogger(WindowsExec.class);
```
MikeDombo, line 27: same as unix, rename as staticLogger

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsExec.java
tilo-chen: PATHEXT is for completing the executable file extension when none is provided. See remarks [here](https://docs.microsoft.com/en-us/previous-versions/windows/it-pro/windows-server-2012-r2-and-2012/cc753148(v=ws.11))

MikeDombo: oh, I see

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsExec.java
```diff
@@ -0,0 +1,106 @@
...
+            for (String extCandidate : PATHEXT) {
+                Path f = Paths.get(fn + extCandidate);
+                if (Files.isExecutable(f)) {
+                    return f;
+                }
+            }
+            return null;
+        }
+        // Search in paths
+        for (Path d : paths) {
```
MikeDombo, line 55: so `paths` still needs to be populated then. Not sure it is getting populated correctly for windows or not

MikeDombo: Do we have any tests for `which`? Shouldn't be too tough to add. `which("cmd")` `which("cmd.exe")` and so on

src/main/java/com/aws/greengrass/util/Utils.java
tilo-chen: DefaultFileSystem is not public. Since we already depend on Apache commons, switching to FilenameUtils.getExtension which works for windows

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsExec.java
tilo-chen: Exec super class already adds paths from `System.getenv(PATH)`. Added a test case for which

src/main/java/com/aws/greengrass/util/platforms/unix/UnixExec.java
```diff
@@ -0,0 +1,85 @@
...
+            if (Files.isExecutable(f)) {
+                return f;
+            }
+        }
+        return null;
+    }
+
+    @Override
+    protected Process createProcess() throws IOException {
+        String[] command = getCommand();
```
MikeDombo, line 79: getCommand should be moved into unix right? Since we aren't going to be decorating commands anymore for windows

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsExec.java
```diff
@@ -0,0 +1,112 @@
...
+                if (correctMethod) {
+                    method.setAccessible(true);
+                    block = (String) method.invoke(null, environment);
+                    break;
+                }
+            }
+        } catch (Throwable t) {
+            logger.atError().setCause(t)
+                    .log("Failed to compute environment block. Creating process without env modifications");
+        }
```
shaguptashaikh, line 109: Is it okay to create process fail silently? E.g. if it's for a lifecycle script the only way for users to find out about failures will be logs

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsPlatform.java
```diff
@@ -373,21 +382,46 @@ private static synchronized WindowsUserAttributes loadCurrentUser() throws IOExc
...
+        private static final String CMD_ARG = "/C";
+        private static final String POWERSHELL = "powershell";
+        private static final String POWERSHELL_ARG = "-Command";
+        private String shell;
+        private String arg;
+
+        public CmdDecorator() {
+            shell = CMD;
+            arg = CMD_ARG;
+        }
```
shaguptashaikh, line 399: Nit - could just assign these at line 393 and 394?

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsPlatform.java
```diff
@@ -373,21 +382,46 @@ private static synchronized WindowsUserAttributes loadCurrentUser() throws IOExc
...
             return ret;
         }
 
         @Override
         public ShellDecorator withShell(String shell) {
-            throw new UnsupportedOperationException("changing shell is not supported");
+            switch (shell) {
+                case CMD:
+                    this.shell = CMD;
+                    this.arg = CMD_ARG;
```
shaguptashaikh, line 415: Is this needed again since they're already the defaults?

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsRunasProcess.java
```diff
@@ -0,0 +1,87 @@
...
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import lombok.Setter;
+
+import java.io.InputStream;
+import java.io.OutputStream;
+
+/**
+ * Manage a Windows process running as a specific user.
+ */
+@SuppressWarnings("PMD")  // TODO re-enable warnings
```
shaguptashaikh, line 18: Is this because of the pending TODOs?

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsExec.java
tilo-chen: Good point. This should fail the process creation. Things won't work properly without env vars

src/main/java/com/aws/greengrass/util/exceptions/ProcessCreationException.java
```diff
@@ -0,0 +1,16 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.aws.greengrass.util.exceptions;
+
+import java.io.IOException;
+
+public class ProcessCreationException extends IOException {
```
tilo-chen, line 10: Exec caller code already handles IOException. Planning to use this exception to wrap some runtime exception so that the error can be handled by callers

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsExec.java
```diff
@@ -71,23 +72,34 @@ public Path which(String fn) {
         if (shellDecorator != null) {
             decorated = shellDecorator.decorate(decorated);
         }
+        // First item in the command is the executable. If it's given as absolute path, add quotes around it
```
MikeDombo, line 75: Can we change the API to use `List<String>` instead of `String[]`? That will make it far easier to manipulate.

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsPlatform.java
```diff
@@ -409,17 +404,11 @@ public CmdDecorator() {
 
         @Override
         public ShellDecorator withShell(String shell) {
-            switch (shell) {
```
MikeDombo: why can't I specify withShell("cmd")? Don't break that. If we're going to have overrides like I think we will, then customers could specify either "cmd" or "powershell" explicitly.

src/main/java/com/aws/greengrass/util/Exec.java
```diff
@@ -57,74 +55,44 @@
...
-    private final AtomicBoolean isClosed = new AtomicBoolean(false);
-    Process process;
-    IntConsumer whenDone;
-    Consumer<CharSequence> stdout = NOP;
-    Consumer<CharSequence> stderr = NOP;
-    AtomicInteger numberOfCopiers;
-    private final Map<String, String> environment = new HashMap<>(defaultEnvironment);
-    private String[] cmds;
+    protected final AtomicBoolean isClosed = new AtomicBoolean(false);
+    private Process process;
```
MikeDombo, line 81: why is process private? That should be protected shouldn't it?

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsExec.java
```diff
@@ -0,0 +1,123 @@
...
+    private static boolean isAbsolutePath(String p) {
+        return new File(p).isAbsolute();
+    }
+
+    /**
+     * Convert environment Map to lpEnvironment block format.
+     * @return environment block for starting a process
+     */
+    private String computeEnvironmentBlock() throws IOException {
+        // TODO: copy over the source code so that it's the same regardless of Java version this is running on
```
MikeDombo, line 103: or use JNA

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsPlatform.java
tilo-chen: Removed because cmd is the default. I think it makes sense that user may want to explicitly indicate "cmd" in the code. Reverted.

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsExec.java
tilo-chen: Fantastic! That's better...

src/main/java/com/aws/greengrass/util/Exec.java
tilo-chen: Right now the Exec child classes just implement `createProcess`. The super class uses the returned `Process` object to do waitFor, destroy, etc. The platform-specific code will be in the `Process` class like `WindowsRunasProcess` for Windows.

MikeDombo: The implementation of `close` will be platform specific, and close needs access to the Process

tilo-chen: I see. The PidProcess thing can't shutdown gracefully on windows

MikeDombo: Not likely, no. It should be override-able

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsExec.java
tilo-chen: Array is convenient for varargs which is used heavily in these APIs. I think this is okay.

MikeDombo: Yeah it is alright for varargs, but that's only the input, we can just create a list from the input array. We do a lot of manipulation, and using a list is a lot easier. ProcessBuilder takes a list.

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsExec.java
```diff
@@ -0,0 +1,134 @@
...
+        }
+        // TODO first try to shutdown process and children gracefully
+        // Then force kill if not stopped within timeout
+        Platform platformInstance = Platform.getInstance();
+        try {
+            platformInstance.killProcessAndChildren(process, false, Collections.emptySet(), userDecorator);
+        } catch (InterruptedException e) {
+            // If we're interrupted make sure to kill the process before returning
+            try {
+                platformInstance.killProcessAndChildren(process, true, Collections.emptySet(), userDecorator);
```
MikeDombo, line 111: put this inside a `finally` instead of the `catch` that way it always runs 100% of the time.

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsPlatform.java
```diff
@@ -84,9 +86,13 @@
...
                                                UserDecorator decorator)
             throws IOException, InterruptedException {
+        if (!force) {
+            // TODO support graceful kill
+            throw new UnsupportedOperationException();
+        }
+
         PidProcess pp = Processes.newPidProcess(process);
         ((WindowsProcess) pp).setIncludeChildren(true);
-        ((WindowsProcess) pp).setGracefulDestroyEnabled(true);
```
MikeDombo, line 89: you don't need to remove this and throw unsupported. This does work correctly, but only for GUI apps.

MikeDombo (dismissed):

PR #1056, opened 2021-08-13
fix(exec): switch to ProcessBuilder, fix environment for Windows commands

MikeDombo:
**Issue #, if available:**

**Description of changes:**
Switches Exec to use ProcessBuilder to have nicer access to `environment` as a map instead of string array. Fixes the PATH environment variable to use the correct separator to ensure that environment is properly passed down to child processes.

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/util/Exec.java
+++ b/src/main/java/com/aws/greengrass/util/Exec.java
@@ -24,11 +24,13 @@ import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


tilo-chen (dismissed):

src/main/java/com/aws/greengrass/util/Exec.java
```diff
@@ -219,7 +206,7 @@ private static void ensurePresent(String... fns) {
 
     private static void addPathEntries(String path) {
         if (path != null && path.length() > 0) {
-            for (String f : path.split("[ :,] *")) {
+            for (String f : path.split("[" + PATH_SEP + ",] *")) {
```
tilo-chen, line 209: Nice catch it was splitting by space...

tilo-chen (approved):

src/main/java/com/aws/greengrass/util/Exec.java
MikeDombo: I think that _may_ have been intentional @JamesGosling, but it doesn't work for Windows where space in paths is fine.

shaguptashaikh (approved):

PR #1061, opened 2021-08-23
feat: support Windows service installation

MikeDombo:
**Issue #, if available:**

**Description of changes:**

**Checklist:**


scripts/loader.cmd
```diff
@@ -76,9 +86,14 @@ FOR /L %%i IN (1,1,%MAX_RETRIES%) DO (
         ECHO Rebooting host
         SHUTDOWN /R
         EXIT /B 0
+    ) ELSE (
+    rem normal exit code when using ctrl+c
+    IF !KERNEL_EXIT_CODE! EQU 130 (
```
tilo-chen, line 91: Under what scenario does kernel exit 130? We're not handling 130 on linux.

MikeDombo: As the comment says, this is a normal exit when using ctrl+c to exit. The Windows service uses ctrl+c to shutdown Greengrass and we do not want to retry in that case.

PR #1063, opened 2021-08-27
feat(exec): windows process runner

tilo-chen:
**Issue #, if available:**

**Description of changes:**
- Add vendor library for interacting with windows credential manager
- Add windows run as another user impl

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/util/platforms/windows/WindowsExec.java
+++ b/src/main/java/com/aws/greengrass/util/platforms/windows/WindowsExec.java
@@ -7,7 +7,8 @@ package com.aws.greengrass.util.platforms.windows;
 
 import com.aws.greengrass.util.Exec;
 import com.aws.greengrass.util.Utils;
-import org.zeroturnaround.process.Processes;
+import com.aws.greengrass.util.platforms.Platform;
+import com.aws.greengrass.util.platforms.UserPlatform;
 
 import java.io.File;
 import java.io.IOException;
@@ -15,15 +16,16 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
-import java.util.concurrent.TimeUnit;
 import javax.annotation.Nullable;
 
 @SuppressWarnings("PMD.AvoidCatchingThrowable")
 public class WindowsExec extends Exec {
-    public static final String PATHEXT_KEY = "PATHEXT";
-
+    private static final String PATHEXT_KEY = "PATHEXT";
+    private static final String LOCAL_DOMAIN = ".";
     private static final List<String> PATHEXT;  // ordered file extensions to try, when no extension is provided
+    public static final String SYSTEM_ROOT = "SystemRoot";
 
     static {
         String pathExt = System.getenv(PATHEXT_KEY);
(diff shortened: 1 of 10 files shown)


src/main/java/com/aws/greengrass/util/platforms/windows/WindowsExec.java
```diff
@@ -69,27 +67,33 @@ public Path which(String fn) {
 
     @Override
     public String[] getCommand() {
-        String[] decorated = cmds;
+        String[] decorated = Arrays.copyOf(cmds, cmds.length);
+        for (int i = 0; i < decorated.length; i++) {
+            final String arg = decorated[i];
+            // Space and \t require quoting otherwise will be split to more than one arg
```
tilo-chen, line 73: Equivalent to VERIFICATION_LEGACY: https://github.com/openjdk/jdk/blob/master/src/java.base/windows/classes/java/lang/ProcessImpl.java#L217-L226

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsPlatform.java
```diff
@@ -162,8 +162,9 @@ public void validateDefaultConfiguration(Map<String, Object> proposedDeviceConfi
 
             @Override
             public Optional<RunWith> generate(DeviceConfiguration deviceConfig, Topics config) {
-                // TODO set the actual user name
-                return Optional.of(RunWith.builder().user(System.getProperty("user.name")).build());
+                // TODO do proper checks like UnixRunWithGenerator
+                String user = Coerce.toString(config.find(RUN_WITH_NAMESPACE_TOPIC, POSIX_USER_KEY));
```
MikeDombo, line 166: we need a windows user field, right?

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsExec.java
```diff
@@ -69,27 +67,33 @@ public Path which(String fn) {
...
     @Override
-    protected Process createProcess() {
-        WindowsRunasProcess winProcess = new WindowsRunasProcess(null, userDecorator.getUser());
-        winProcess.setLpEnvironment(computeEnvironmentBlock());
-        winProcess.setLpCurrentDirectory(dir.getAbsolutePath());
-        winProcess.start(String.join(" ", getCommand()));
-        return winProcess;
+    protected Process createProcess() throws IOException {
+        if (needToSwitchUser()) {
+            WindowsRunasProcess winProcess = new WindowsRunasProcess(LOCAL_DOMAIN, userDecorator.getUser());
```
MikeDombo, line 87: what do we do if we can't find the password for the user? Why are we forcing local domain? Why about networked accounts?

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsPlatform.java
tilo-chen: Yes. Can't really just posixUser it's confusing

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsExec.java
tilo-chen: Need to do verification in RunWithGenerator. Check like user exists and we have the credential

PR #1069, opened 2021-09-07
fix(windows): fix lookup of local system user

MikeDombo:
**Issue #, if available:**

**Description of changes:**
"SYSTEM" user cannot be looked up, so this change hard codes the well known username and SID for the user.
Verified that UAT passes which failed previously with this error.

**Checklist:**


src/main/java/com/aws/greengrass/util/platforms/windows/WindowsPlatform.java
```diff
@@ -221,6 +226,14 @@ protected void setOwner(UserPrincipal userPrincipal, GroupPrincipal groupPrincip
         // Note that group ownership is not used.
     }
 
+    @Override
+    public UserPrincipal lookupUserByName(Path path, String name) throws IOException {
+        if (name.endsWith("$")) {
```
rbattle, line 231: can you add a comment explaining why we are doing this here -  link to something like https://docs.microsoft.com/en-us/troubleshoot/windows-server/identity/compatibility-user-accounts-end-dollar-sign here

rbattle (approved):

PR #1071, opened 2021-09-09
fix: improve logging in retryUtil

fahadmohammed01:
**Description of changes:**
RetryUtill only logs errors at info level only the first time. This can be confusing to the customer as the error is not logged but the issue still persists.

From 2.5  IotPolicy needs to include ListThingGroupsForCoreDevice for deployments to succeed. If that is not the case, device will retry indefinitely to call ListThingGroupsForCoreDevice API and fail, but the error will only be logged once at info level.

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/util/RetryUtils.java
+++ b/src/main/java/com/aws/greengrass/util/RetryUtils.java
@@ -18,6 +18,7 @@ public class RetryUtils {
 
     private static final Random RANDOM = new Random();
     private static final int LOG_ON_FAILURE_COUNT = 20;
+
     // Need this to make spotbug check happy
     private RetryUtils() {
     }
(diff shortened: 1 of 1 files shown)


MikeDombo (dismissed):

tilo-chen (approved):

src/main/java/com/aws/greengrass/util/RetryUtils.java
```diff
@@ -53,8 +53,8 @@ private RetryUtils() {
                 }
                 if (retryConfig.retryableExceptions.stream().anyMatch(c -> c.isInstance(e))) {
                     LogEventBuilder logBuild = logger.atDebug(taskDescription);
-                    // Log first failed attempt at info so as not to spam logs
-                    if (attempt == 1) {
+                    // Log first and every 20th failed attempt at info so as not to spam logs
```
leaf94, line 56: Great improvement! What about the magic number 20? 😄

tilo-chen (approved):

MikeDombo (approved):

PR #1075, opened 2021-09-13
feat: windows process creation impl

tilo-chen:
**Issue #, if available:**

**Description of changes:**
- Add WindowsRunasProcess.java which creates and manages windows process via JNA
- Remove JetBrains runas related code

- Remove dependency on JetBrains runas

- CI

**Checklist:**


src/main/java/com/aws/greengrass/util/platforms/windows/WindowsExec.java
```diff
@@ -179,4 +165,8 @@ private boolean needToSwitchUser() throws IOException {
     private static boolean isAbsolutePath(String p) {
         return new File(p).isAbsolute();
     }
+
+    private static boolean isQuoted(String s) {
+        return s.startsWith("\"") && s.endsWith("\"") && !s.endsWith("\\\"");
```
MikeDombo, line 170: `|| s.endsWith("\\\\\"")`
Technically I think you need to do some counting to figure out if the final quote is escaped or not.

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsExec.java
```diff
@@ -80,6 +70,13 @@ public Path which(String fn) {
     @Override
     public String[] getCommand() {
         String[] decorated = Arrays.copyOf(cmds, cmds.length);
+        for (int i = 0; i < decorated.length; i++) {
+            final String arg = decorated[i];
+            // Space and \t require quoting otherwise won't be treated as a single arg
+            if (!isQuoted(arg) && arg.matches(".*[ \t].*")) {
+                decorated[i] = "\"" + arg + "\"";
+            }
+        }
```
MikeDombo, line 79: Where is this logic from, if anywhere? You had quoting logic before from Java didn't you?

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsExec.java
```diff
@@ -100,34 +97,21 @@ protected Process createProcess() throws IOException {
...
-
-        Arrays.fill(cb.array(), (char) 0);  // zero-out temporary buffers
-        Arrays.fill(bb.array(), (byte) 0);
-
-        ProcessBuilder pb = new ProcessBuilder();
-        pb.environment().putAll(environment);
-        Process p = pb.directory(dir).command(args).start();
-        args.clear();  // best effort to clear password presence
+        // Expect username in format: DOMAIN\UserName
+        String[] usernameParts = userDecorator.getUser().split("\\\\");  // split by single backslash
```
MikeDombo, line 101: Single backslash is `\\` right?

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsExec.java
```diff
@@ -100,34 +97,21 @@ protected Process createProcess() throws IOException {
...
+            domain = usernameParts[0];
+            username = usernameParts[1];
+        } else {
+            username = usernameParts[0];
+        }
+
+        WindowsRunasProcess p = new WindowsRunasProcess(domain, username);
+        p.setEnv(environment);
+        p.setCurrentDirectory(dir.getAbsolutePath());
+        p.start(String.join(" ", commands));
```
MikeDombo, line 114: I'd like to pass the commands as an array. Let WindowsRunasProcess figure out the quoting/whatever is needed

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsExec.java
MikeDombo: Also, this logic applies when not using runas as well, right? Is it equally applicable?

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsRunasProcess.java
```diff
@@ -0,0 +1,450 @@
...
+            throw lastErrorProcessCreationException("OpenProcessToken");
+        }
+
+        if (isService(processTokenHandle)) {
+            processAsUser(processTokenHandle, command);
+        } else {
+            processWithLogon(command);
+        }
+
+        Kernel32Util.closeHandleRefs(processTokenHandle);
```
MikeDombo, line 114: Put inside a finally so that we ensure things are closed?

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsRunasProcess.java
```diff
@@ -0,0 +1,450 @@
...
+        enablePrivileges(processTokenHandle.getValue(), Kernel32.SE_TCB_NAME, Kernel32.SE_ASSIGNPRIMARYTOKEN_NAME);
+
+        char[] password = getPassword();
+        // LogonUser
+        boolean logonSuccess = false;
+        WinNT.HANDLEByReference userTokenHandle = new WinNT.HANDLEByReference();
+        int[] logonTypes =
+                {Kernel32.LOGON32_LOGON_INTERACTIVE, Kernel32.LOGON32_LOGON_SERVICE, Kernel32.LOGON32_LOGON_BATCH};
+        for (int logonType : logonTypes) {
+            if (ProcAdvapi32.INSTANCE.LogonUser(username, domain, password, logonType,
```
MikeDombo, line 149: Is this safe to run in parallel? Think about this for all the win api calls here too. We may create multiple processes all at once and we don't want them to conflict

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsRunasProcess.java
```diff
@@ -0,0 +1,450 @@
...
+
+    private void processAsUser(final WinNT.HANDLEByReference processTokenHandle, final String command)
+            throws IOException {
+        enablePrivileges(processTokenHandle.getValue(), Kernel32.SE_TCB_NAME, Kernel32.SE_ASSIGNPRIMARYTOKEN_NAME);
+
+        char[] password = getPassword();
+        // LogonUser
+        boolean logonSuccess = false;
+        WinNT.HANDLEByReference userTokenHandle = new WinNT.HANDLEByReference();
+        int[] logonTypes =
```
MikeDombo, line 146: which type do we expect to have the highest chance of success? Put that one first.

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsRunasProcess.java
```diff
@@ -0,0 +1,450 @@
...
+            throw lastErrorProcessCreationException("CreateProcessAsUser");
+        }
+
+        pid = Kernel32.INSTANCE.GetProcessId(procInfo.hProcess);
+        if (pid == 0) {
+            throw lastErrorProcessCreationException("GetProcessId");
+        }
+
+        redirectStreams();
+        Kernel32Util.closeHandleRefs(userTokenHandle, primaryTokenHandle);
```
MikeDombo, line 221: use finally?

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsRunasProcess.java
```diff
@@ -0,0 +1,450 @@
...
+            throw new IllegalThreadStateException("process hasn't exited");
+        }
+        exited = true;
+        exitCode = exitCodeRef.getValue();
+        return exitCode;
+    }
+
+    @Override
+    public void destroy() {
+        destroyForcibly();
```
MikeDombo, line 374: TODO Graceful @fahadmohammed01

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsRunasProcess.java
tilo-chen: Added sync to Advapi32.INSTANCE for the entire flow.

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsExec.java
tilo-chen: Right. And backslash itself needs escaping...

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsExec.java
tilo-chen: Space and tab is the most basic check from Windows [ProcessImpl](https://github.com/openjdk/jdk/blob/f9b2507f3e86bcb91e8ccfd0a84f31712fd535c2/src/java.base/windows/classes/java/lang/ProcessImpl.java#L226). So this is essentially what JDK does by default.

When not using runas, ProcessBuilder won't double quote the args according to [needsEscaping](https://github.com/openjdk/jdk/blob/f9b2507f3e86bcb91e8ccfd0a84f31712fd535c2/src/java.base/windows/classes/java/lang/ProcessImpl.java#L302).

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsExec.java
tilo-chen: Yes that's better.

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsRunasProcess.java
MikeDombo: Well if it isn't necessary, don't add it.

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsRunasProcess.java
```diff
@@ -0,0 +1,450 @@
...
+
+    private static boolean isService(WinNT.HANDLEByReference processToken) throws ProcessCreationException {
+        // Get required buffer size
+        IntByReference returnLength = new IntByReference();
+        if (!Advapi32.INSTANCE.GetTokenInformation(processToken.getValue(), WinNT.TOKEN_INFORMATION_CLASS.TokenGroups,
+                null, 0, returnLength)) {
+            int lastErrorCode = Kernel32.INSTANCE.GetLastError();
+            if (lastErrorCode != WinError.ERROR_INSUFFICIENT_BUFFER) {
+                throw new ProcessCreationException(String.format("[%s] %s", "GetTokenInformation",
+                        Kernel32Util.formatMessageFromLastErrorCode(lastErrorCode)));
```
rbattle, line 405: throw `lastErrorProcessCreationException("GetTokenInformation")`

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsRunasProcess.java
```diff
@@ -0,0 +1,450 @@
...
+        return this;
+    }
+
+    private static LastErrorException lastErrorRuntimeException() {
+        return new LastErrorException(Kernel32.INSTANCE.GetLastError());
+    }
+
+    private static ProcessCreationException lastErrorProcessCreationException(String context) {
+        return new ProcessCreationException(String.format("[%s] %s", context, Kernel32Util.getLastErrorMessage()));
+    }
```
rbattle, line 395: not sure why we have two types of exceptions that are essentially the same (GetLastError vs getLastErrorMessage)

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsRunasProcess.java
```diff
@@ -0,0 +1,450 @@
...
+        }
+        if (exited) {
+            return exitCode;
+        }
+        IntByReference exitCodeRef = new IntByReference();
+        if (!Kernel32.INSTANCE.GetExitCodeProcess(procInfo.hProcess, exitCodeRef)) {
+            throw lastErrorRuntimeException();
+        }
+        if (WinBase.STILL_ACTIVE == exitCodeRef.getValue()) {
+            throw new IllegalThreadStateException("process hasn't exited");
```
rbattle, line 365: why `IllegalThreadStateException`?

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsRunasProcess.java
```diff
@@ -0,0 +1,450 @@
...
+        currentDirectory = dir;
+    }
+
+    /**
+     * Start the process as given user.
+     *
+     * @param command command line to run
+     * @throws IOException if failed to run subprocess
+     */
+    public synchronized void start(String command) throws IOException {
```
rbattle, line 101: needs more debug logs everywhere (especially after deciding which API method to use)

maybe add trace logs for each JNA call too?

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsExec.java
MikeDombo: What are you escaping for?

MikeDombo: If the user is `DOMAIN\user` then you would need to split on `\` not `\\`

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsRunasProcess.java
MikeDombo: That is what javadoc says the API should do if it hasn't exited

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsRunasProcess.java
tilo-chen: I think we should. At least the last error can be messed up if APIs are called in parallel.

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsRunasProcess.java
```diff
@@ -433,18 +467,40 @@ private static void writefd(final FileDescriptor fd, final WinNT.HANDLE pointer)
...
-        // https://github.com/openjdk/jdk/blob/b17b821/src/java.base/windows/classes/java/lang/ProcessEnvironment.java#L309-L311
-        String systemRootVal = System.getenv(SYSTEM_ROOT);
-        if (systemRootVal != null) {
-            additionalEnv.put(SYSTEM_ROOT, systemRootVal);
-        }
-        return Advapi32Util.getEnvironmentBlock(additionalEnv);
+    private String computeEnvironmentBlock(WinNT.HANDLE userTokenHandle) throws ProcessCreationException {
+        PointerByReference lpEnv = new PointerByReference();
+        // Get user's env vars, inheriting current process env
+        if (!UserEnv.INSTANCE.CreateEnvironmentBlock(lpEnv, userTokenHandle, true)) {
```
MikeDombo, line 475: If we're calling this, do we still need the hack calling `set` that I just wrote?

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsRunasProcess.java
```diff
@@ -433,18 +467,40 @@ private static void writefd(final FileDescriptor fd, final WinNT.HANDLE pointer)
...
+            int splitInd = s.indexOf('=');
+            userEnvMap.put(s.substring(0, splitInd), s.substring(splitInd + 1));
+        }
+
+        if (!UserEnv.INSTANCE.DestroyEnvironmentBlock(lpEnv)) {
+            throw lastErrorProcessCreationException("DestroyEnvironmentBlock");
+        }
+
+        // Set additional envs on top of user default env
+        userEnvMap.putAll(additionalEnv);
```
MikeDombo, line 498: order here matters. We want any settings from Greengrass to overwrite whatever the user has. Maybe use putIfAbsent?

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsExec.java
tilo-chen: Yes the regex for single backslash is `\\`, as a Java string literal, that is `"\\\\"`.

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsRunasProcess.java
tilo-chen: Problem here is that we already called GetLastError above this line, calling it again will give no error. Did it this way because here we want to look at the actual error code and determine what to do.

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsRunasProcess.java
tilo-chen: So I'm using LastErrorException (which is a RuntimeException defined by JNA) in the Java Process interface methods (waitfor, exitValue) because the interface doesn't declare checked exceptions.

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsRunasProcess.java
tilo-chen: additionalEnv comes from Greengrass. So gg env will overwrite user envs this way.

rbattle (dismissed):

rbattle:
I was just looking at MikeDombo's latest change - didn't mean to approve the whole thing 🤦  and can't seem to remove the approval right now

MikeDombo (approved):

MikeDombo:
Replaced by #1081

PR #1077, opened 2021-09-13
feat: define CryptoKeySpi and security service interface

wikimonkey:
**Issue #, if available:**

**Description of changes:**
Define CryptoKeySpi and security service interface.
Basic structure of register/deregister crypto key provider with security service.

Define the interface for implementing Greengrass v2 TPM support.

complete unit tests

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/security/SecurityService.java
+++ b/src/main/java/com/aws/greengrass/security/SecurityService.java
@@ -1,0 +1,92 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.aws.greengrass.security;
+
+import com.aws.greengrass.config.CaseInsensitiveString;
+import com.aws.greengrass.logging.api.Logger;
+import com.aws.greengrass.logging.impl.LogManager;
+import com.aws.greengrass.security.exceptions.KeyLoadingException;
+import com.aws.greengrass.security.exceptions.ServiceProviderConflictException;
+import com.aws.greengrass.security.exceptions.ServiceUnavailableException;
+import lombok.AccessLevel;
+import lombok.Getter;
+
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import javax.net.ssl.KeyManager;
+
+public class SecurityService {
+    private static final Logger logger = LogManager.getLogger(SecurityService.class);
+    private static final String KEY_TYPE = "keyType";
+    private static final String KEY_URI = "keyUri";
+
+    @Getter(AccessLevel.PACKAGE)
+    private final ConcurrentMap<CaseInsensitiveString, CryptoKeySpi> cryptoKeyProviderMap;
+
+    public SecurityService() {
+        this.cryptoKeyProviderMap = new ConcurrentHashMap<>();
+    }
+
+    /**
+     * Register crypto key provider for the key type.
+     *
+     * @param keyProvider Crypto key provider
+     * @throws ServiceProviderConflictException if key type is already registered
+     */
+    public void registerCryptoKeyProvider(CryptoKeySpi keyProvider) throws ServiceProviderConflictException {
+        CaseInsensitiveString keyType = new CaseInsensitiveString(keyProvider.supportedKeyType());
+        logger.atInfo().kv(KEY_TYPE, keyType).log("Register crypto key service provider");
+        CryptoKeySpi provider = cryptoKeyProviderMap.computeIfAbsent(keyType, k -> keyProvider);
+        if (!provider.equals(keyProvider)) {
+            logger.atError().kv(KEY_TYPE, keyType)
+                    .log("Crypto key service provider for the key type is already registered");
+            throw new ServiceProviderConflictException(String.format("Key type %s provider is registered", keyType));
+        }
+    }
+
+    /**
+     * Deregister crypto key provide for the key type.
+     *
+     * @param keyProvider Crypto key provider
+     */
+    public void deregisterCryptoKeyProvider(CryptoKeySpi keyProvider) {
+        CaseInsensitiveString keyType = new CaseInsensitiveString(keyProvider.supportedKeyType());
+        boolean removed = cryptoKeyProviderMap.remove(keyType, keyProvider);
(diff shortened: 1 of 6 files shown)


src/test/java/com/aws/greengrass/security/SecurityServiceTest.java
```diff
@@ -0,0 +1,99 @@
...
+        when(providerB.supportedKeyType()).thenReturn("file");
+        assertThrows(ServiceProviderConflictException.class, () -> service.registerCryptoKeyProvider(providerB));
+    }
+
+    @Test
+    void GIVEN_key_service_provider_registered_WHEN_deregister_THEN_removed() throws Exception {
+        when(mockKeyProvider.supportedKeyType()).thenReturn("FILE");
+        service.registerCryptoKeyProvider(mockKeyProvider);
+        service.deregisterCryptoKeyProvider(mockKeyProvider);
+        assertThat(service.getCryptoKeyProviderMap(), IsMapWithSize.aMapWithSize(0));
```
prateek-y, line 69: de register it twice to validate that case as well?

src/main/java/com/aws/greengrass/security/exceptions/ServiceUnavailableException.java
```diff
@@ -0,0 +1,19 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.aws.greengrass.security.exceptions;
+
+public class ServiceUnavailableException extends Exception {
+
+    static final long serialVersionUID = -3387516993124229948L;
```
prateek-y, line 10: nit: Shouldnt serialVersionUID be unique across classes?

src/test/java/com/aws/greengrass/security/SecurityServiceTest.java
```diff
@@ -0,0 +1,99 @@
...
+    @Mock
+    private CryptoKeySpi mockKeyProvider;
+
+    @Test
+    void GIVEN_key_service_provider_WHEN_register_provider_THEN_succeed() throws Exception {
+        when(mockKeyProvider.supportedKeyType()).thenReturn("FILE");
+        service.registerCryptoKeyProvider(mockKeyProvider);
+        assertThat(service.getCryptoKeyProviderMap(), IsMapWithSize.aMapWithSize(1));
+        assertThat(service.getCryptoKeyProviderMap(), IsMapContaining.hasEntry(new CaseInsensitiveString("file"),
+                mockKeyProvider));
```
prateek-y, line 41: Add a new key type as well to assert that 2nd key type gets added successfully?

prateek-y (commented):
Just some minor comments

src/test/java/com/aws/greengrass/security/SecurityServiceTest.java
wikimonkey: will do

src/test/java/com/aws/greengrass/security/SecurityServiceTest.java
wikimonkey: will do

src/main/java/com/aws/greengrass/security/exceptions/ServiceUnavailableException.java
wikimonkey: I think `serialization id` is intended for distinguishing different version of class. So It doesn't need to be unique across classes.

prateek-y: ahh ok.

prateek-y (approved):
Thanks for kick starting TPM !!

src/main/java/com/aws/greengrass/security/SecurityService.java
```diff
@@ -0,0 +1,92 @@
...
+            throw new ServiceProviderConflictException(String.format("Key type %s provider is registered", keyType));
+        }
+    }
+
+    /**
+     * Deregister crypto key provide for the key type.
+     *
+     * @param keyProvider Crypto key provider
+     */
+    public void deregisterCryptoKeyProvider(CryptoKeySpi keyProvider) {
```
vaibhavmurkute, line 57: Is it expected to return silently if user tries to deregister previously unregistered CryptoKeyProvider? 
How about we return a `boolean` instead of  `void` here, that shows the result of deregister operation?

src/main/java/com/aws/greengrass/security/SecurityService.java
```diff
@@ -0,0 +1,92 @@
...
+        CaseInsensitiveString keyType = new CaseInsensitiveString(keyProvider.supportedKeyType());
+        boolean removed = cryptoKeyProviderMap.remove(keyType, keyProvider);
+        if (!removed) {
+            logger.atInfo().kv(KEY_TYPE, keyType).log("Crypto key service provider is either removed or not the same"
+                    + " provider");
```
vaibhavmurkute, line 62: To avoid confusion, this probably can be simplified to : _Crypto key service provider is either already removed or is unregistered._

src/main/java/com/aws/greengrass/security/SecurityService.java
wikimonkey: It's intended to have API idempotent. Also returning false is probably confusing - whether it's unregistered or not.

src/main/java/com/aws/greengrass/security/SecurityService.java
wikimonkey: removed equals to unregistered in this case. There is case that caller tried to remove a provider with same key type but different instance. In such case, the existing instance shouldn't be removed.

vaibhavmurkute: We can think of it like this:
`register(keyType1, provider1);`    <--- here `CryptoKeyProviderMap` will have entry `{keyType1, provider1}` . This Concurrent map will always hold only _one_ entry for a CaseInsensitive keyType string.

`deregister(keyType1, provider2);`  <---- Note that its the same keyType1 but different provider which was not registered before. `CryptoKeyProviderMap` will not have `provider2` entry value for `keyType1`, because provider2 was not registered previously for keyType1. i.e. provider2 is an _unregistered_ provider.

wikimonkey: Make sense. I will change wording in the next revision. Thanks!

PR #1078, opened 2021-09-13
feat(proxy): add support for https proxies

tomnagy-aws:
**Issue #, if available:** N/A

**Description of changes:**
Adds support for HTTPS proxies. Simply add the certificate of the CA that issued the proxy TLS certificate to the file specified in rootCaPath and make sure the networkProxy url config value uses the https scheme.

Code updates:
- Make sure the root CA truststore system property is used in the CRT SDK's HttpProxyOptions.
- Make sure the root CA truststore is combined with the JVM default truststore and used in the AWS SDK's SdkHttpClient.
- Modify device provisioning helper to preserve contents of truststore if it exists and remove duplicates.

 This is required in case customers operate in an environment where remote calls must traverse an HTTPS proxy

Manually tested with squid v4 and `mvn -U -ntp verify`

**Any additional information or context required to review the change:** No

**Checklist:**
 - [x] Updated the README if applicable
 - [x] If your code makes a remote network call, it was tested with a proxy

### DIFF
--- a/src/main/java/com/aws/greengrass/easysetup/DeviceProvisioningHelper.java
+++ b/src/main/java/com/aws/greengrass/easysetup/DeviceProvisioningHelper.java
@@ -288,8 +288,8 @@ public class DeviceProvisioningHelper {
     }
 
     private void removeDuplicateCertificates(File f) {
-        try (InputStream inputStream = Files.newInputStream(f.toPath())) {
-            String certificates = IoUtils.toUtf8String(inputStream);
+        try {
+            String certificates = new String(Files.readAllBytes(f.toPath()), StandardCharsets.UTF_8);
             Set<String> uniqueCertificates =
                     Arrays.stream(certificates.split(EncryptionUtils.CERTIFICATE_PEM_HEADER))
                             .map(s -> s.trim())
--- a/src/main/java/com/aws/greengrass/easysetup/DeviceProvisioningHelper.java
+++ b/src/main/java/com/aws/greengrass/easysetup/DeviceProvisioningHelper.java
@@ -9,6 +9,7 @@ import com.aws.greengrass.deployment.DeviceConfiguration;
 import com.aws.greengrass.deployment.exceptions.DeviceConfigurationException;
 import com.aws.greengrass.lifecyclemanager.Kernel;
 import com.aws.greengrass.util.CommitableFile;
+import com.aws.greengrass.util.EncryptionUtils;
 import com.aws.greengrass.util.IamSdkClientFactory;
 import com.aws.greengrass.util.IotSdkClientFactory;
 import com.aws.greengrass.util.IotSdkClientFactory.EnvironmentStage;
@@ -69,20 +70,27 @@ import software.amazon.awssdk.services.iot.model.UpdateCertificateRequest;
 import software.amazon.awssdk.services.sts.StsClient;
 import software.amazon.awssdk.services.sts.model.GetCallerIdentityRequest;
 import software.amazon.awssdk.utils.ImmutableMap;
+import software.amazon.awssdk.utils.IoUtils;
 
+import java.io.BufferedWriter;
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.PrintStream;
+import java.net.HttpURLConnection;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.nio.channels.Channels;
 import java.nio.channels.ReadableByteChannel;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.Arrays;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 import java.util.UUID;
+import java.util.stream.Collectors;
 
 /**
  * Provision a device by registering as an IoT thing, creating roles and template first party components.
@@ -261,22 +269,52 @@ public class DeviceProvisioningHelper {
     }
 
     /*
-     * Download root CA to a local file if the file does not exist
+     * Download root CA to a local file.
+     *
+     * To support HTTPS proxies and other custom truststore configurations, append to the file if it exists.
      */
-    private void downloadRootCAToFile(File f) throws IOException {
+    private void downloadRootCAToFile(File f) {
         if (f.exists()) {
-            outStream.printf("Root CA found at \"%s\". Skipping download.%n", f);
-            return;
+            outStream.printf("Root CA file found at \"%s\". Contents will be preserved.%n", f);
         }
         outStream.printf("Downloading Root CA from \"%s\"%n", ROOT_CA_URL);
-        downloadFileFromURL(ROOT_CA_URL, f);
+        try {
+            downloadFileFromURL(ROOT_CA_URL, f, f.exists());
+            removeDuplicateCertificates(f);
+        } catch (IOException e) {
+            // Do not block as the root CA file may have been manually provisioned
+            outStream.printf("Failed to download Root CA - %s%n", e);
+        }
+    }
+
+    private void removeDuplicateCertificates(File f) {
+        try {
+            String certificates = IoUtils.toUtf8String(Files.newInputStream(f.toPath()));
+            Set<String> uniqueCertificates =
+                    Arrays.stream(certificates.split(EncryptionUtils.CERTIFICATE_PEM_HEADER))
+                            .map(s -> s.trim())
+                            .collect(Collectors.toSet());
+
+            try (BufferedWriter bw = Files.newBufferedWriter(f.toPath(), StandardCharsets.UTF_8)) {
+                for (String certificate : uniqueCertificates) {
+                    if (certificate.length() > 0) {
+                        bw.write(EncryptionUtils.CERTIFICATE_PEM_HEADER);
+                        bw.newLine();
+                        bw.write(certificate);
+                        bw.newLine();
+                    }
+                }
+            }
+        } catch (IOException e) {
+            outStream.printf("Failed to remove duplicate certificates - %s%n", e);
+        }
     }
 
     /*
      * Download content from a URL to a local file.
      */
     @SuppressWarnings("PMD.AvoidFileStream")
-    private void downloadFileFromURL(String url, File f) throws IOException {
+    private void downloadFileFromURL(String url, File f, boolean appendFile) throws IOException {
         SdkHttpFullRequest request = SdkHttpFullRequest.builder()
                 .uri(URI.create(url))
                 .method(SdkHttpMethod.GET)
@@ -289,8 +327,13 @@ public class DeviceProvisioningHelper {
         try (SdkHttpClient client = getSdkHttpClient()) {
             HttpExecuteResponse executeResponse = client.prepareRequest(executeRequest).call();
 
+            int responseCode = executeResponse.httpResponse().statusCode();
+            if (responseCode != HttpURLConnection.HTTP_OK) {
+                throw new IOException("Received invalid response code: " + responseCode);
+            }
+
             try (ReadableByteChannel readableByteChannel = Channels.newChannel(executeResponse.responseBody().get());
-                 FileOutputStream fileOutputStream = new FileOutputStream(f)) {
+                 FileOutputStream fileOutputStream = new FileOutputStream(f, appendFile)) {
                 fileOutputStream.getChannel().transferFrom(readableByteChannel, 0, Long.MAX_VALUE);
             }
         }
--- a/src/main/java/com/aws/greengrass/easysetup/DeviceProvisioningHelper.java
+++ b/src/main/java/com/aws/greengrass/easysetup/DeviceProvisioningHelper.java
@@ -74,17 +74,17 @@ import software.amazon.awssdk.utils.IoUtils;
 
 import java.io.BufferedWriter;
 import java.io.File;
-import java.io.FileOutputStream;
 import java.io.IOException;
+import java.io.InputStream;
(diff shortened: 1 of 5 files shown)


src/main/java/com/aws/greengrass/easysetup/DeviceProvisioningHelper.java
```diff
@@ -261,22 +269,52 @@ public void cleanThing(IotClient client, ThingInfo thing, boolean deletePolicies
...
+            removeDuplicateCertificates(f);
+        } catch (IOException e) {
+            // Do not block as the root CA file may have been manually provisioned
+            outStream.printf("Failed to download Root CA - %s%n", e);
+        }
+    }
+
+    private void removeDuplicateCertificates(File f) {
+        try {
+            String certificates = IoUtils.toUtf8String(Files.newInputStream(f.toPath()));
```
MikeDombo, line 292: the input stream must be closed. Use try-with-resources

src/main/java/com/aws/greengrass/easysetup/DeviceProvisioningHelper.java
```diff
@@ -261,22 +269,52 @@ public void cleanThing(IotClient client, ThingInfo thing, boolean deletePolicies
...
+            Set<String> uniqueCertificates =
+                    Arrays.stream(certificates.split(EncryptionUtils.CERTIFICATE_PEM_HEADER))
+                            .map(s -> s.trim())
+                            .collect(Collectors.toSet());
+
+            try (BufferedWriter bw = Files.newBufferedWriter(f.toPath(), StandardCharsets.UTF_8)) {
+                for (String certificate : uniqueCertificates) {
+                    if (certificate.length() > 0) {
+                        bw.write(EncryptionUtils.CERTIFICATE_PEM_HEADER);
+                        bw.newLine();
```
MikeDombo, line 302: is newLine system-specific? ie. Is it \r\n on Windows? I'm not sure if \r\n is allowed in a pem, so you may want to just write \n instead.

src/main/java/com/aws/greengrass/easysetup/DeviceProvisioningHelper.java
```diff
@@ -289,8 +327,13 @@ private void downloadFileFromURL(String url, File f) throws IOException {
...
             HttpExecuteResponse executeResponse = client.prepareRequest(executeRequest).call();
 
+            int responseCode = executeResponse.httpResponse().statusCode();
+            if (responseCode != HttpURLConnection.HTTP_OK) {
+                throw new IOException("Received invalid response code: " + responseCode);
+            }
+
             try (ReadableByteChannel readableByteChannel = Channels.newChannel(executeResponse.responseBody().get());
-                 FileOutputStream fileOutputStream = new FileOutputStream(f)) {
+                 FileOutputStream fileOutputStream = new FileOutputStream(f, appendFile)) {
```
MikeDombo, line 336: if the existing file doesn't end with \n this will not work properly, right? or would it be cleaned up to work correctly by the duplicate removing code?

src/main/java/com/aws/greengrass/util/ProxyUtils.java
```diff
@@ -151,15 +164,27 @@ public static String getProxyPassword(String proxyUrl, String proxyPassword) {
...
      * CRT library clients (like mqtt and http)
      *
      * @param deviceConfiguration contains user specified system proxy values
+     * @param tlsContext contains TLS options for proxy connection if an HTTPS proxy is used
      * @return httpProxyOptions containing user proxy settings, if specified. If not, httpProxyOptions is null.
      */
     @Nullable
-    public static HttpProxyOptions getHttpProxyOptions(DeviceConfiguration deviceConfiguration) {
+    public static HttpProxyOptions getHttpProxyOptions(DeviceConfiguration deviceConfiguration,
+                                                       ClientTlsContext tlsContext) {
```
MikeDombo, line 187: annotate tlsContext as nonnull

src/main/java/com/aws/greengrass/util/ProxyUtils.java
```diff
@@ -169,6 +194,10 @@ public static HttpProxyOptions getHttpProxyOptions(DeviceConfiguration deviceCon
         httpProxyOptions.setHost(ProxyUtils.getHostFromProxyUrl(proxyUrl));
         httpProxyOptions.setPort(ProxyUtils.getPortFromProxyUrl(proxyUrl));
 
+        if ("https".equalsIgnoreCase(getSchemeFromProxyUrl(proxyUrl))) {
+            httpProxyOptions.setTlsContext(tlsContext);
```
MikeDombo, line 198: why not create the context in here since this is where you know the pertinent information?

src/main/java/com/aws/greengrass/util/ProxyUtils.java
```diff
@@ -208,20 +238,62 @@ public static SdkHttpClient getSdkHttpClient() {
...
+     *         a default httpClient builder
      */
     public static ApacheHttpClient.Builder getSdkHttpClientBuilder() {
         ProxyConfiguration proxyConfiguration = getProxyConfiguration();
+        TrustManager[] trustManagers = createTrustManagers();
 
         if (proxyConfiguration != null) {
-            return ApacheHttpClient.builder().proxyConfiguration(proxyConfiguration);
+            return ApacheHttpClient.builder()
+                    .tlsTrustManagersProvider(() -> trustManagers)
```
MikeDombo, line 256: just pass ProxyUtils::createTrustManager instead of a lambda returning the local thing

src/main/java/com/aws/greengrass/util/ProxyUtils.java
```diff
@@ -208,20 +238,62 @@ public static SdkHttpClient getSdkHttpClient() {
...
+                    .proxyConfiguration(proxyConfiguration);
         }
 
-        return ApacheHttpClient.builder();
+        return ApacheHttpClient.builder().tlsTrustManagersProvider(() -> trustManagers);
+    }
+
+    private static TrustManager[] createTrustManagers() {
+        try {
+            List<X509Certificate> certificates = new ArrayList<>(Arrays.asList(getDefaultRootCertificates()));
```
MikeDombo, line 265: create a new list, then use addAll to add the defaults. The trip from array to list and back and forth is expensive.

src/main/java/com/aws/greengrass/util/ProxyUtils.java
```diff
@@ -208,20 +238,62 @@ public static SdkHttpClient getSdkHttpClient() {
...
+            customKeyStore.load(null, null);
+
+            if (Utils.isNotEmpty(rootCAPath)) {
+                certificates.addAll(EncryptionUtils.loadX509Certificates(rootCAPath));
+            }
+
+            for (X509Certificate certificate : certificates) {
+                String name = certificate.getSubjectX500Principal().getName("RFC2253");
+                customKeyStore.setCertificateEntry(name, certificate);
+            }
```
MikeDombo, line 276: add a comment for what this is doing and why

src/main/java/com/aws/greengrass/easysetup/DeviceProvisioningHelper.java
tomnagy-aws: Done

src/main/java/com/aws/greengrass/easysetup/DeviceProvisioningHelper.java
tomnagy-aws: Yes it uses line.separator which is system specific. I've seen a lot of libraries fail to handle windows newlines, so changed to \n.

src/main/java/com/aws/greengrass/easysetup/DeviceProvisioningHelper.java
tomnagy-aws: Looks like it was always writing from the beginning of the file anyway (oops). Changed it so it'll actually append from the end of the file. If the existing file doesn't end with \n and a new cert is appended, it would break cert parsing, but removeDuplicateCertificates does fix it.

src/main/java/com/aws/greengrass/util/ProxyUtils.java
tomnagy-aws: Done

src/main/java/com/aws/greengrass/util/ProxyUtils.java
tomnagy-aws: Done

src/main/java/com/aws/greengrass/util/ProxyUtils.java
tomnagy-aws: Done

src/main/java/com/aws/greengrass/util/ProxyUtils.java
tomnagy-aws: Done

src/main/java/com/aws/greengrass/util/ProxyUtils.java
tomnagy-aws: I didn't want any potential resource leak issues so I incorporated creating/closing into the mqtt client lifecycle.

src/main/java/com/aws/greengrass/easysetup/DeviceProvisioningHelper.java
```diff
@@ -261,15 +269,45 @@ public void cleanThing(IotClient client, ThingInfo thing, boolean deletePolicies
...
+            removeDuplicateCertificates(f);
+        } catch (IOException e) {
+            // Do not block as the root CA file may have been manually provisioned
+            outStream.printf("Failed to download Root CA - %s%n", e);
+        }
+    }
+
+    private void removeDuplicateCertificates(File f) {
+        try (InputStream inputStream = Files.newInputStream(f.toPath())) {
+            String certificates = IoUtils.toUtf8String(inputStream);
```
MikeDombo, line 292: Get rid of this, just use Files.readAllBytes so that we read it into memory. You're going to be writing the file while reading from it which is likely to be problematic.

src/main/java/com/aws/greengrass/util/ProxyUtils.java
```diff
@@ -208,20 +239,66 @@ public static SdkHttpClient getSdkHttpClient() {
...
         ProxyConfiguration proxyConfiguration = getProxyConfiguration();
 
         if (proxyConfiguration != null) {
-            return ApacheHttpClient.builder().proxyConfiguration(proxyConfiguration);
+            return ApacheHttpClient.builder()
+                    .tlsTrustManagersProvider(ProxyUtils::createTrustManagers)
+                    .proxyConfiguration(proxyConfiguration);
+        }
+
+        return ApacheHttpClient.builder().tlsTrustManagersProvider(ProxyUtils::createTrustManagers);
```
MikeDombo, line 260: why are you overriding the trust manager here? When there's no proxy that shouldn't be done, right?

tomnagy-aws: My thinking is it allows customers to add custom CAs (in non-proxy scenarios) via rootCaPath instead of finding and modifying Java's truststore.

src/main/java/com/aws/greengrass/easysetup/DeviceProvisioningHelper.java
tomnagy-aws: Done

MikeDombo (approved):

src/main/java/com/aws/greengrass/easysetup/DeviceProvisioningHelper.java
```diff
@@ -261,15 +269,45 @@ public void cleanThing(IotClient client, ThingInfo thing, boolean deletePolicies
...
+        try {
+            downloadFileFromURL(ROOT_CA_URL, f);
+            removeDuplicateCertificates(f);
+        } catch (IOException e) {
+            // Do not block as the root CA file may have been manually provisioned
+            outStream.printf("Failed to download Root CA - %s%n", e);
+        }
+    }
+
+    private void removeDuplicateCertificates(File f) {
```
shaguptashaikh, line 290: Is this just an optimization? If so, should we just let the original pem file be if it fails for whatever reason? i.e. line 308?

src/main/java/com/aws/greengrass/util/EncryptionUtils.java
```diff
@@ -24,6 +24,9 @@
 
 public final class EncryptionUtils {
 
+    public static final String CERTIFICATE_PEM_HEADER = "-----BEGIN CERTIFICATE-----";
+    public static final String CERTIFICATE_PEM_FOOTER = "-----END CERTIFICATE-----";
```
youtuyy, line 28: why is the footer not used?

src/main/java/com/aws/greengrass/easysetup/DeviceProvisioningHelper.java
tomnagy-aws: We used to skip the CA download step if the root CA file existed, but if the root CA file contains the proxy CA certificate, it won't download the amazon root CA. So we want to always attempt the download and if it succeeds, remove duplicates (in case provisioning is run multiple times), and if it fails, log the error and not block the rest of provisioning as the device may have been manually setup.

src/main/java/com/aws/greengrass/util/EncryptionUtils.java
tomnagy-aws: The header is all that's needed to split and dedupe the root CA file. I included the footer for completeness even though it's not currently used.

shaguptashaikh (approved):

youtuyy (approved):

PR #1079, opened 2021-09-15
feat: implement default file based crypto key service provider

wikimonkey:
**Issue #, if available:**

**Description of changes:**
Implement default file based crypto key service provider.
register it with SecurityService at the construction time so it's always available.

Default file based crypto key provider will keep the existing key loading process same.

Added unit test. The code is not added to use case code path yet.

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/security/SecurityService.java
+++ b/src/main/java/com/aws/greengrass/security/SecurityService.java
@@ -11,25 +11,47 @@ import com.aws.greengrass.logging.impl.LogManager;
 import com.aws.greengrass.security.exceptions.KeyLoadingException;
 import com.aws.greengrass.security.exceptions.ServiceProviderConflictException;
 import com.aws.greengrass.security.exceptions.ServiceUnavailableException;
+import com.aws.greengrass.util.EncryptionUtils;
 import lombok.AccessLevel;
 import lombok.Getter;
 
+import java.io.IOException;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.security.GeneralSecurityException;
+import java.security.KeyStore;
+import java.security.PrivateKey;
+import java.security.cert.Certificate;
+import java.security.cert.X509Certificate;
+import java.util.List;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import javax.net.ssl.KeyManager;
+import javax.net.ssl.KeyManagerFactory;
 
-public class SecurityService {
+public final class SecurityService {
     private static final Logger logger = LogManager.getLogger(SecurityService.class);
     private static final String KEY_TYPE = "keyType";
     private static final String KEY_URI = "keyUri";
+    private static final String CERT_URI = "certificateUri";
 
     @Getter(AccessLevel.PACKAGE)
     private final ConcurrentMap<CaseInsensitiveString, CryptoKeySpi> cryptoKeyProviderMap;
 
+    /**
+     * Constructor of security service.
+     */
     public SecurityService() {
         this.cryptoKeyProviderMap = new ConcurrentHashMap<>();
+
+        // instantiate and register the default file based provider
+        CryptoKeySpi defaultProvider = new DefaultCryptoKeyProvider();
+        try {
+            this.registerCryptoKeyProvider(defaultProvider);
+        } catch (ServiceProviderConflictException e) {
+            // it won't happen, it's programming error if it does
+            throw new RuntimeException("Default crypto key provider has been registered", e);
+        }
     }
 
     /**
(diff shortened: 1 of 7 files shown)


src/main/java/com/aws/greengrass/security/SecurityService.java
```diff
@@ -11,25 +11,47 @@
...
+    private static final String CERT_URI = "certificateUri";
 
     @Getter(AccessLevel.PACKAGE)
     private final ConcurrentMap<CaseInsensitiveString, CryptoKeySpi> cryptoKeyProviderMap;
 
+    /**
+     * Constructor of security service.
+     */
     public SecurityService() {
         this.cryptoKeyProviderMap = new ConcurrentHashMap<>();
```
prateek-y, line 45: nit: Instant initialization?

src/test/java/com/aws/greengrass/security/SecurityServiceTest.java
```diff
@@ -29,80 +37,128 @@
...
+                EncryptionUtilsTest.generateCertificateFile(2048, true, resourcePath.resolve("certificate.pem"));
+        Path privateKeyPath =
+                EncryptionUtilsTest.generatePkCS8PrivateKeyFile(2048, true, resourcePath.resolve("privateKey.pem"));
+
+        KeyManager[] keyManagers =
+                defaultProvider.getKeyManagers(privateKeyPath.toUri().toString(), certPath.toUri().toString());
+        assertThat(keyManagers.length, is(1));
+        X509KeyManager keyManager = (X509KeyManager) keyManagers[0];
+        assertThat(keyManager.getPrivateKey("private-key"), notNullValue());
+        assertThat(keyManager.getPrivateKey("private-key").getAlgorithm(), is("RSA"));
```
prateek-y, line 139: Not sure if there is value in asserting something about the certificate as well. What do you think?

src/test/java/com/aws/greengrass/security/SecurityServiceTest.java
```diff
@@ -29,80 +37,128 @@
...
-        assertThrows(ServiceUnavailableException.class, () -> service.getKeyManagers("file:///path/to/file"));
+    void GIVEN_key_service_provider_not_registered_WHEN_get_key_managers_THEN_throw_exception() {
+        assertThrows(ServiceUnavailableException.class,
+                () -> service.getKeyManagers("pkcs11:object=key-label", "file:///path/to/certificate"));
+    }
+
+    @Test
+    void GIVEN_key_and_cert_uri_WHEN_get_key_managers_from_default_THEN_succeed() throws Exception {
+        Path certPath =
+                EncryptionUtilsTest.generateCertificateFile(2048, true, resourcePath.resolve("certificate.pem"));
```
prateek-y, line 130: Do you know if we support other RSA key lengths, other than 2048?

src/main/java/com/aws/greengrass/security/SecurityService.java
```diff
@@ -89,4 +113,51 @@ private CryptoKeySpi selectCryptoKeyProvider(String keyUri) throws URISyntaxExce
...
+            } catch (GeneralSecurityException | IOException e) {
+                logger.atError().kv(KEY_URI, privateKeyUriStr).kv(CERT_URI, certificateUriStr)
+                        .log("Exception caught during getting key manager");
+                throw new KeyLoadingException("Failed to get key manager", e);
+            }
+        }
+
+        @Override
+        public String supportedKeyType() {
+            return "file";
```
prateek-y, line 156: should this be a list/enum somewhere?

prateek-y (commented):
Minor comments, just one question about RSA lengths we support.

vaibhavmurkute (approved):

src/main/java/com/aws/greengrass/security/SecurityService.java
vaibhavmurkute: +1 for enum

wikimonkey: Because this is a public interface, a enumerable finite list is not applicable I think. Because customer could define its own key type. But I definitely could change it to a constant.

src/test/java/com/aws/greengrass/security/SecurityServiceTest.java
wikimonkey: yes, we do. 1024 or 512 are fine as well.

src/test/java/com/aws/greengrass/security/SecurityServiceTest.java
wikimonkey: will try to add some.

src/main/java/com/aws/greengrass/security/SecurityService.java
wikimonkey: will do. But you know compiler will put instance initialization into constructor as well.

prateek-y (approved):

vaibhavmurkute (dismissed):

vaibhavmurkute (approved):

prateek-y (approved):

PR #1082, opened 2021-09-17
feat: refactor https SSL context creation to use security service API

wikimonkey:
**Issue #, if available:**

**Description of changes:**
Consume security service API to get key managers in creating https SSL context.

Consuming the API could enable the use case to use either file based key or pkcs11 key

component e2e tests passed

**Checklist:**


src/main/java/com/aws/greengrass/componentmanager/ClientConfigurationUtils.java
```diff
@@ -110,23 +120,37 @@ private static void configureClientMutualTLS(ApacheHttpClient.Builder httpBuilde
...
+        URI u;
+        try {
+            u = new URI(path);
+        } catch (URISyntaxException e) {
+            // if can't parse the path string as URI, try it as Path and use URI default provider "file"
+            logger.atDebug().setCause(e)
+                    .kv("path", path).log("can't parse path string as URI");
+            u = Paths.get(path).toUri();
+        }
+        if (Utils.isEmpty(u.getScheme())) {
```
prateek-y, line 150: Which case would hit this check? Would Paths.get(path).toUri() not add a valid scheme.

prateek-y (commented):
LGTM

sushantAtAws (approved):

src/main/java/com/aws/greengrass/componentmanager/ClientConfigurationUtils.java
wikimonkey: The current UNIX/LINUX path would hit this check. The path looks like `/path/to/file`, which will be treated as a valid URI. But it's lack of `scheme` part.
On the other hand, windows path looks like `C:\path\to\file`, which will be parsed as valid URI because of backslash. In this case it hits the path of converting path to URI - `Path.toUri`, which will add `file` scheme.

src/main/java/com/aws/greengrass/componentmanager/ClientConfigurationUtils.java
```diff
@@ -110,23 +120,37 @@ private static void configureClientMutualTLS(ApacheHttpClient.Builder httpBuilde
...
+    KeyManager[] createKeyManagers(String privateKeyPath, String certificatePath) throws TLSAuthException {
         try {
-            List<X509Certificate> certificateChain = EncryptionUtils.loadX509Certificates(Paths.get(certificatePath));
-
-            PrivateKey privateKey = EncryptionUtils.loadPrivateKey(Paths.get(privateKeyPath));
-
-            KeyStore keyStore = KeyStore.getInstance("PKCS12");
-            keyStore.load(null);
-            keyStore.setKeyEntry("private-key", privateKey, null, certificateChain.toArray(new Certificate[0]));
+            return RetryUtils.runWithRetry(SECURITY_SERVICE_RETRY_CONFIG, () -> securityService
```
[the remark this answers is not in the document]
wikimonkey: Yeah, instead of trying to strictly sort services in order. The client could retry on the operations if service is not available. (BTW, the default file based implementation won't throw unavailable exception because it's guaranteed available).  
Infinite retry is because get SSL key manager is critical for the component deployment process. If it's not available, there is no meaning of moving forward. Customer needs to deploy proper key provider service to continue process if it's meant to use the service.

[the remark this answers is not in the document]
prateek-y: This is what I am thinking, let me know if it makes sense. The problem with infinite retry is the case when the security service never comes up due to x reason. Then we just occupy the thread waiting for service and potentially brick greengrass, because no other code can execute in the meanwhile? If it gives up after x retries, then we can atleast inform the user by logging that "Startup failed because PKCS is not available". That way customers can try something different again. x could be a very large value, which kind of mimics infinite retry case, but still gives up and inform the user.

src/main/java/com/aws/greengrass/componentmanager/ClientConfigurationUtils.java
prateek-y: Oh. so new URI() could return a URI with empty scheme? Then should line 150 be inside try to make the obvious?

src/main/java/com/aws/greengrass/componentmanager/ClientConfigurationUtils.java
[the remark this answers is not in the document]
wikimonkey: It makes sense. I was not sure how long is 'long' enough, so started with 'infinite'. In valid case, if the client probes the service in unavailable state, it will only be 'unavailable' for a short while before it initialize properly. So a reasonable retry period should be fine. Let me change to finite retry.

src/main/java/com/aws/greengrass/componentmanager/ClientConfigurationUtils.java
wikimonkey: Yeah, it doesn't hurt. Will do.

prateek-y (approved):

sushantAtAws (approved):

PR #1084, opened 2021-09-21
fix: windows env var casing

tilo-chen:
**Issue #, if available:**

**Description of changes:**
- Our env var map key was case-sensitive, which makes it possible to set both "Path" and "PATH". This can lead to errors. Changing the env var map in Exec and ProcessImplWin32.computeEnvironmentBlock to be case-insensitive.
- Rename ambiguous "path" variable used for log event key.

At least powershell `Start-Process` is confused:

```
Start-Process : Item has already been added. Key in dictionary: 'PATH'  Key being added: 'Path'
```

**Checklist:**


src/main/java/vendored/org/apache/dolphinscheduler/common/utils/process/ProcessImplForWin32.java
```diff
@@ -1056,7 +1057,7 @@ private static String computeEnvironmentBlock(WinNT.HANDLE userTokenHandle, Map<
         }
 
         // The above API returns pointer to a block of null-terminated strings. It ends with two nulls (\0\0).
-        Map<String, String> userEnvMap = new HashMap<>();
+        Map<String, String> userEnvMap = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
```
MikeDombo, line 1060: add a comment for why we're doing this

src/main/java/com/aws/greengrass/util/Exec.java
```diff
@@ -89,6 +89,11 @@
     private Copier stderrc;
     private Copier stdoutc;
 
+    protected Exec() {
+        environment = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
```
MikeDombo, line 93: add a comment for why we're doing this. Also should we apply this to Windows only?

tilo-chen: Fixed. Yes only Windows should be case-insensitive

MikeDombo (approved):

MikeDombo (approved):

shaguptashaikh (approved):

PR #1095, opened 2021-09-28
fix: service listen to windowsUser topic

tilo-chen:
**Issue #, if available:**

**Description of changes:**
- GenericExternalService also requests reinstall if windowsUser config changes, which is same behavior as posixUser
- Update tests

- Fix component not reinstalling on windows when windowsUser config changes.

CI

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/lifecyclemanager/GenericExternalService.java
+++ b/src/main/java/com/aws/greengrass/lifecyclemanager/GenericExternalService.java
@@ -94,6 +94,7 @@ public class GenericExternalService extends GreengrassService {
         this(c, privateSpace, Platform.getInstance());
     }
 
+    @SuppressWarnings("PMD.UselessParentheses")
     protected GenericExternalService(Topics c, Topics privateSpace, Platform platform) {
         super(c, privateSpace);
         this.platform = platform;
@@ -122,7 +123,7 @@ public class GenericExternalService extends GreengrassService {
 
             // Reinstall for changes to the install script or if the package version changed, or runWith user changed
             if (child.childOf(Lifecycle.LIFECYCLE_INSTALL_NAMESPACE_TOPIC) || child.childOf(VERSION_CONFIG_KEY)
-                    || child.childOf(RUN_WITH_NAMESPACE_TOPIC) && !child.childOf(SYSTEM_RESOURCE_LIMITS_TOPICS)) {
+                    || (child.childOf(RUN_WITH_NAMESPACE_TOPIC) && !child.childOf(SYSTEM_RESOURCE_LIMITS_TOPICS))) {
                 logger.atInfo("service-config-change").kv("configNode", child.getFullName())
                         .log("Requesting reinstallation for component");
                 requestReinstall();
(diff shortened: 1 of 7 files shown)


MikeDombo (approved):

src/main/java/com/aws/greengrass/lifecyclemanager/GenericExternalService.java
```diff
@@ -120,9 +120,9 @@ protected GenericExternalService(Topics c, Topics privateSpace, Platform platfor
                 updateSystemResourceLimits();
             }
 
-            // Reinstall for changes to the install script or if the package version changed, or posixUser has changed
+            // Reinstall for changes to the install script or if the package version changed, or runWith user changed
             if (child.childOf(Lifecycle.LIFECYCLE_INSTALL_NAMESPACE_TOPIC) || child.childOf(VERSION_CONFIG_KEY)
-                    || child.childOf(POSIX_USER_KEY)) {
+                    || child.childOf(POSIX_USER_KEY) || child.childOf(WINDOWS_USER_KEY)) {
```
MikeDombo, line 125: shall we just use childOf the runwith namespace instead of specifying posix/windows?

tilo-chen: Child of runWith also has resource limit. That is handled separately on L119

MikeDombo: could do childof runwith && !childof system limits?

shaguptashaikh: This has a small issue where if the customer has set both Windows and Posix user in their deployment config and change one of them which is not applicable to the platform the deployment executes on, it will still end up restarting components, it would be nice to have this only react to the user config that it cares about

shaguptashaikh (dismissed):

src/main/java/com/aws/greengrass/lifecyclemanager/GenericExternalService.java
```diff
@@ -120,9 +120,9 @@ protected GenericExternalService(Topics c, Topics privateSpace, Platform platfor
                 updateSystemResourceLimits();
             }
 
-            // Reinstall for changes to the install script or if the package version changed, or posixUser has changed
+            // Reinstall for changes to the install script or if the package version changed, or runWith user changed
             if (child.childOf(Lifecycle.LIFECYCLE_INSTALL_NAMESPACE_TOPIC) || child.childOf(VERSION_CONFIG_KEY)
-                    || child.childOf(POSIX_USER_KEY)) {
+                    || child.childOf(RUN_WITH_NAMESPACE_TOPIC) && !child.childOf(SYSTEM_RESOURCE_LIMITS_TOPICS)) {
```
MikeDombo, line 125: I hope this is correct, I would add clarifying parenthesis as long as pmd doesn't complain

tilo-chen: It did complain. I added suppression.

MikeDombo (approved):

shaguptashaikh (approved):

PR #1098, opened 2021-09-30
fix(windows): prevent race condition in windows createProcess

junfuchen99:
**Issue #, if available:**

**Description of changes:**
Add a synchronized block in windows createProcess() to remove race condition.

When two processes are created at the same time in windows, they might inherit handles they are not supposed to inherit.
https://www.betaarchive.com/wiki/index.php/Microsoft_KB_Archive/315939

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/util/platforms/windows/WindowsExec.java
+++ b/src/main/java/com/aws/greengrass/util/platforms/windows/WindowsExec.java
@@ -104,7 +104,10 @@ public class WindowsExec extends Exec {
             winPb.environment().clear();
         }
         winPb.environment().putAll(environment);
-        Process process = winPb.directory(dir).command(commands).start();
+        winPb.directory(dir).command(commands);
+        synchronized (Kernel32.INSTANCE) {
+            process = winPb.start();
+        }
         // calling attachConsole right after a process is launched will fail with invalid handle error
         // waiting a bit ensures that we get the process handle from the pid
         try {


src/main/java/com/aws/greengrass/util/platforms/windows/WindowsExec.java
```diff
@@ -133,9 +137,6 @@ public synchronized void close() throws IOException {
 
     @SuppressFBWarnings("SWL_SLEEP_WITH_LOCK_HELD")
     private void stopGracefully() {
-        if (!process.isAlive()) {
```
tilo-chen, line 136: FYI: this is redundant because we just checked isAlive on L123

MikeDombo, line 136: Why remove this?

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsExec.java
```diff
@@ -104,15 +104,19 @@ protected Process createProcess() throws IOException {
             winPb.environment().clear();
         }
         winPb.environment().putAll(environment);
-        Process process = winPb.directory(dir).command(commands).start();
+        winPb.directory(dir).command(commands);
+        Process newProcess;
+        synchronized (Kernel32.INSTANCE) {
+            newProcess = winPb.start();
```
MikeDombo, line 110: Why change to newProcess?

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsExec.java
tilo-chen: We just checked isAlive right before this. On L123

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsExec.java
junfuchen99: just refactored it

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsExec.java
```diff
@@ -103,16 +103,17 @@ protected Process createProcess() throws IOException {
             // Clear the env in this case because later we'll load the given user's env instead
             winPb.environment().clear();
         }
-        winPb.environment().putAll(environment);
```
tilo-chen, line 106: Still need this line

src/main/java/com/aws/greengrass/util/Exec.java
```diff
@@ -347,10 +347,9 @@ public Exec withErr(Consumer<CharSequence> o) {
     /**
      * Create and start the child process in platform-specific ways.
      *
-     * @return child process
      * @throws IOException if IO error occurs
      */
-    protected abstract Process createProcess() throws IOException;
```
MikeDombo, line 353: Why are you changing this? What's the point in breaking compatibility?

MikeDombo (changes requested):
Please do not make any changes except to wrap `process.start()` in a synchronized block. Let's keep this change atomic and targeted to fix the problem that we're seeing.

src/main/java/com/aws/greengrass/util/Exec.java
junfuchen99: it seems redundant to do process=winPb.start(), then return process and assign it again with process=createProcess()

MikeDombo: Keep it the way that it is. The API is more clear that way and it enforce the requirement that `Exec.process` is properly set to the process object. With the changed API you are relying on the implementor to set the field correctly which they may not do because there's no obvious need to do that.

MikeDombo (approved):

tilo-chen (approved):

PR #1099, opened 2021-09-30
fix: add certificate URI input to getKeyPair in CryptoKeySpi

MikeDombo:
**Issue #, if available:**

**Description of changes:**
For PKCS#11, it is difficult or impossible to get the public key for a given private key, so this change adds the certificateUri as an input to `getKeyPair`. This then allows us to implement getKeyPair such that the public key will be extracted from the certificate since the certificate will have been signed by the public key and all the public key information exists in the cert.

Tested with a new test in Secret manager (https://github.com/aws-greengrass/aws-greengrass-secret-manager/pull/60) which verifies that we can encrypt and decrypt secrets using PKCS11.

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/security/SecurityService.java
+++ b/src/main/java/com/aws/greengrass/security/SecurityService.java
@@ -124,15 +124,16 @@ public final class SecurityService {
      * Get JCA KeyManagers, used for Secret manager.
      *
      * @param privateKeyUri private key URI
+     * @param certificateUri certificate URI
      * @return KeyManagers that manage the specified private key
      * @throws ServiceUnavailableException if crypto key provider service is unavailable
      * @throws KeyLoadingException if crypto key provider service fails to load key
      */
-    public KeyPair getKeyPair(URI privateKeyUri)
+    public KeyPair getKeyPair(URI privateKeyUri, URI certificateUri)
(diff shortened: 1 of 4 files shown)


prateek-y (approved):

src/main/java/com/aws/greengrass/security/SecurityService.java
```diff
@@ -241,7 +242,7 @@ public static URI uriFromPossibleFileURIString(String path) {
         }
 
         @Override
-        public KeyPair getKeyPair(URI privateKeyUri)
+        public KeyPair getKeyPair(URI privateKeyUri, URI certificateUri)
```
wikimonkey, line 245: Shouldn't it use public key from certificate instead of generating public key from private key?

MikeDombo: No, why bother? When it is file based we're able to generate a public key from private key, that generated key is guaranteed to be correct. We do not need the certificate in this case at all, but it is part of the interface for PKCS11 where generating public from private isn't feasible.

wikimonkey: Fair, for this use case.

wikimonkey (approved):

PR #1100, opened 2021-09-30
fix: remove potential race condition resulting in NPE

MikeDombo:
**Issue #, if available:**

**Description of changes:**
E2E tests somehow got a NPE when calling `get` on the atomic reference on L715. To help avoiding that, return the boolean we just set.

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
+++ b/src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
@@ -709,10 +709,11 @@ public class DeviceConfiguration {
         try {
             validate(true);
             deviceConfigValidateCachedResult.set(true);
+            return true;
         } catch (DeviceConfigurationException e) {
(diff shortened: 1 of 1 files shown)


J-C-L (approved):

src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
```diff
@@ -709,10 +709,11 @@ public boolean isDeviceConfiguredToTalkToCloud() {
         try {
             validate(true);
             deviceConfigValidateCachedResult.set(true);
```
tilo-chen, line 711: Make it AtomicBoolean? Right now it's `AtomicReference<Boolean>`

MikeDombo: No we don't want that. This is a tri-state: unset, set true, set false.

tilo-chen (approved):

PR #1101, opened 2021-10-04
fix: tighten SYSTEM username lookup to only match computerName$ not just endsWith $

MikeDombo:
**Issue #, if available:**

**Description of changes:**
Update the system name lookup to computerName$ so not any random username ending with $ will work.

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/util/platforms/windows/WindowsPlatform.java
+++ b/src/main/java/com/aws/greengrass/util/platforms/windows/WindowsPlatform.java
@@ -287,16 +287,34 @@ public class WindowsPlatform extends Platform {
         // When running as the SYSTEM user, the name from the "user.name" property is "<hostname>$".
         // This name cannot be looked up normally, but it is well known, so we can account for it easily.
         // https://docs.microsoft.com/en-us/troubleshoot/windows-server/identity/compatibility-user-accounts-end-dollar-sign
-        if (name.endsWith("$")) {
+        if (name.equalsIgnoreCase(getComputerName() + "$")) {
             name = LOCAL_SYSTEM_USERNAME;
         }
         return path.getFileSystem().getUserPrincipalLookupService().lookupPrincipalByName(name);
     }
 
+    private static String getComputerName() {
+        try {
+            return Kernel32Util.getComputerName();
+        } catch (Win32Exception ex) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/aws/greengrass/util/platforms/windows/WindowsPlatform.java
```diff
@@ -287,16 +287,27 @@ public UserPrincipal lookupUserByName(Path path, String name) throws IOException
...
+    private static String getComputerName() {
+        try {
+            return Kernel32Util.getComputerName();
+        } catch (Win32Exception ex) {
+            String env = System.getenv("COMPUTERNAME");
+            if (env != null) {
+                return env;
+            }
+        }
+        return "";
```
leaf94, line 305: Some details: 
1. In catch block, I think we are trying again with `System.getenv("COMPUTERNAME");` could we add a comment on when do we expect the exception to be thrown? Also should we log.warn it?
2. Could you also add a comment on why we default to empty String and possible implications?

leaf94 (commented):
Thank you for super fast and great fix Michael!

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsPlatform.java
MikeDombo: Updated

tilo-chen (approved):

rbattle (approved):
nice comments

src/main/java/com/aws/greengrass/util/platforms/windows/WindowsPlatform.java
```diff
@@ -287,16 +287,34 @@ public UserPrincipal lookupUserByName(Path path, String name) throws IOException
...
+            String env = System.getenv("COMPUTERNAME");
+            if (env != null) {
+                return env;
+            }
+        }
+        // Default to empty string if we can't find any name because we use getComputerName()+"$", so it needs
+        // to be appendable. The check to see if the user is SYSTEM will return false since just $ won't match the
+        // username, but that is the proper behavior. We do not want to simply fail, instead it will continue to lookup
+        // the user through the regular Windows APIs which will then either work or generate a proper error.
+        return "";
```
leaf94, line 312: Fantastic!!! I did have the question about why not re-throw up and it answers my question completely.

leaf94 (approved):

PR #1104, opened 2021-10-05
chore: add warning log for resource limits on non-linux platforms

J-C-L:
**Description of changes:**
Adding a warning when users attempt to use resource limits on non-linux platforms

Resource limits only supported on linux

**Checklist:**


src/main/java/com/aws/greengrass/util/platforms/StubResourceController.java
```diff
@@ -20,12 +23,12 @@ public void removeResourceController(GreengrassService component) {
 
     @Override
     public void updateResourceLimits(GreengrassService component, Map<String, Object> resourceLimit) {
```
MikeDombo, line 25: how often are these logs going to trigger? I don't want our log file flooded by these warnings which are not all that useful

MikeDombo: @shaguptashaikh

src/main/java/com/aws/greengrass/util/platforms/StubResourceController.java
```diff
@@ -20,12 +23,12 @@ public void removeResourceController(GreengrassService component) {
...
-        // no op
+        LOGGER.warn("System resource limits for components not supported on this platform.");
```
rbattle, line 26: nit: you don't need the "." the logger will append one

src/main/java/com/aws/greengrass/util/platforms/StubResourceController.java
J-C-L: Looks like it's only during startup and initial handling of run script, but yes, want to hear from @shaguptashaikh

src/main/java/com/aws/greengrass/util/platforms/StubResourceController.java
J-C-L: GTK. Thanks.

src/main/java/com/aws/greengrass/util/platforms/StubResourceController.java
shaguptashaikh: This log will be produced on each restart of a component that has been configured with system resource limits on a platform that doesn't support system resource limits.

shaguptashaikh: I don't know if there will be a way to not log it on each restart but still account for all cases like, customer puts resource limits on an already running component

MikeDombo: Sounds to me like this will log too much stuff.

MikeDombo: I don't think that adding these logs really adds much value at all. But if we really want it, then I'd suggest a one-shot sort of thing so that it only logs once per GG startup.

J-C-L: Does our system know the difference between 'first time startup' and restart? Where is the right place to put a one time check?

PR #1105, opened 2021-10-05
refactor: use char array for password. log pid after process is created

tilo-chen:
**Issue #, if available:**

**Description of changes:**
- use char array for password during windows process creation
- log pid after process is created

- reduce password presence as much as possible

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/util/Exec.java
+++ b/src/main/java/com/aws/greengrass/util/Exec.java
@@ -11,6 +11,8 @@ import com.aws.greengrass.util.platforms.Platform;
 import com.aws.greengrass.util.platforms.ShellDecorator;
 import com.aws.greengrass.util.platforms.UserDecorator;
 import lombok.Getter;
+import org.zeroturnaround.process.Processes;
+import vendored.org.apache.dolphinscheduler.common.utils.process.ProcessImplForWin32;
 
 import java.io.BufferedReader;
 import java.io.Closeable;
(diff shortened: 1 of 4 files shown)


src/main/java/com/aws/greengrass/util/Exec.java
```diff
@@ -314,6 +316,11 @@ public Exec withErr(Consumer<CharSequence> o) {
             throw new InterruptedException();
         }
         process = createProcess();
+
+        logger.debug("Created process with pid {}",
+                process instanceof ProcessImplForWin32 ? ((ProcessImplForWin32) process).getPid()
+                        : Processes.newPidProcess(process).getPid());
```
MikeDombo, line 322: instead of the casting around, let's just add a `getPid` method onto Exec and implement it in the subclasses

src/main/java/com/aws/greengrass/util/Exec.java
```diff
@@ -74,6 +74,7 @@
 
     protected final AtomicBoolean isClosed = new AtomicBoolean(false);
     protected Process process;
+    protected int pid;
```
MikeDombo, line 77: I'd highly recommend making this a method, not a field. `getPid`, then subclasses can handle that however they want. This way is forcing a particular implementation.

MikeDombo (approved):

shaguptashaikh (approved):

PR #1109, opened 2021-10-07
refactor: lazy loading for greengrass data client to avoid blocking

k-khatri:
**Issue #, if available:**
GG-37592
**Description of changes:**
pkcs11 crypto key service is not running during the factory creation so configuring it during the get call.

pkcs plugin won't work without this

manual testing

**Checklist:**


MikeDombo:
Verify that there are no additional places that call the newly-blocking methods

PR #1124, opened 2021-10-25
Minor fixes - Provisioning IAM:GetPolicy fallback, Resource limits race condition, Windows bootstrap script dependency on powershell

shaguptashaikh:
**Issue #, if available:**
-

**Description of changes:**
1. Fix easysetup to not fail on unauthorized error in iam::getPolicy since we have fallback steps in case the IAM getPolicy permissions are not available for provisioning 
2. Make resource limit not enabled log at debug level to reduce noise in logging
3. Fix race condition in adding pid to cgroup, i.e. sometimes a child process is created right when existing pids are being added to the resource limit cgroups, so the child pid is never added to the cgroup by greengrass lifecycle or by linux
4. Nucleus bootstrap command currently uses powershell introducing a hard dependency on it, moving to cmd to avoid that. It includes the same set of steps as linux/earlier powershell script i.e. on bootstrap, remove contents of <root>/alts/current, create a new launch.params file with the new component configuration for launch params, and create a new symlink for distro targeting the newly downloaded nucleus artifact

1. Setup should not fail if the iam::getpolicy permissions are not available but the permissions for the fallback steps i.e. get user policy or createPolicy are present
2. Log for resource limit not enabled case is distracting to customers
3. All component pids started by Greengrass should be added to resource limit cgroups for resource limits to be applied correctly.

- Unit tests for provisioning fix
- UATs for resource limits passed consistently when run ~15 times
- UATs for OTA workflow pass on windows

**Checklist:**

### DIFF
--- a/conf/recipe.yaml
+++ b/conf/recipe.yaml
@@ -59,13 +59,4 @@ Manifests:
       bootstrap:
         RequiresPrivilege: true
         script: >-
-          powershell -command "& { Set-StrictMode -Version Latest;
-          $ErrorActionPreference = \"Stop\";
-          $PSDefaultParameterValues['*:ErrorAction']=\"Stop\";
-          $KernelRoot = \"{kernel:rootPath}\";
-          $UnpackDir = \"{artifacts:decompressedPath}\aws.greengrass.nucleus\";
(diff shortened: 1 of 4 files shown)


MikeDombo (approved):
LGTM, title must conform to conventional commits

junfuchen99 (approved):

PR #1126, opened 2021-10-29
consider empty or pkcs11-type uri case for certificateFilePath validations

vaibhavmurkute:
**Description of changes:**
- Modifies validations for `certificateFilePath` to allow the case when certificateFilePath value could be empty (for PKCS11-type keys).
- Requires `certificateFilePath`  to be non-empty only if privateKey path is file-type.

 - `certificateFilePath` is ignored while building MQTT client if privateKey and certificate are stored onto an HSM (PKCS11) device. Certificate is accessed using the same label as the privateKey stored in PKCS11. So user may not configure certificateFilePath in case of PKCS11-type keys
- for reference: https://github.com/aws-greengrass/aws-greengrass-pkcs11-provider/pull/13

- unit tests
- integration tests

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
+++ b/src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
@@ -776,7 +776,7 @@ public class DeviceConfiguration {
         if (Utils.isEmpty(thingName)) {
             errors.add(DEVICE_PARAM_THING_NAME + CANNOT_BE_EMPTY);
         }
-        if (Utils.isEmpty(certificateFilePath)) {
+        if (!Utils.isPkcs11Uri(privateKeyPath) && Utils.isEmpty(certificateFilePath)) {
             errors.add(DEVICE_PARAM_CERTIFICATE_FILE_PATH + CANNOT_BE_EMPTY);
         }
         if (Utils.isEmpty(privateKeyPath)) {
--- a/src/main/java/com/aws/greengrass/security/SecurityService.java
+++ b/src/main/java/com/aws/greengrass/security/SecurityService.java
@@ -199,8 +199,24 @@ public final class SecurityService {
         return uriFromPossibleFileURIString(Coerce.toString(deviceConfiguration.getPrivateKeyFilePath()));
     }
 
+    /**
+     * Get URI for configured certificate-path.
+     * Configured certificate-path can be empty for PKCS11-type keys.
+     *
+     * @return URI of device-identity certificate
+     */
     public URI getDeviceIdentityCertificateURI() {
-        return uriFromPossibleFileURIString(Coerce.toString(deviceConfiguration.getCertificateFilePath()));
+        try {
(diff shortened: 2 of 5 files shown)


src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
```diff
@@ -776,7 +776,7 @@ private void validateDeviceConfiguration(String thingName, String certificateFil
         if (Utils.isEmpty(thingName)) {
             errors.add(DEVICE_PARAM_THING_NAME + CANNOT_BE_EMPTY);
         }
-        if (Utils.isEmpty(certificateFilePath)) {
+        if (!Utils.isPkcs11Uri(privateKeyPath) && Utils.isEmpty(certificateFilePath)) {
```
MikeDombo, line 779: Nucleus *must not* be aware of PKCS11

src/main/java/com/aws/greengrass/security/SecurityService.java
```diff
@@ -199,8 +199,24 @@ public URI getDeviceIdentityPrivateKeyURI() {
...
+     * @return URI of device-identity certificate
+     */
     public URI getDeviceIdentityCertificateURI() {
-        return uriFromPossibleFileURIString(Coerce.toString(deviceConfiguration.getCertificateFilePath()));
+        try {
+            String certPath = Coerce.toString(deviceConfiguration.getCertificateFilePath());
+            if (Utils.isNotEmpty(certPath)) {
+                return uriFromPossibleFileURIString(Coerce.toString(deviceConfiguration.getCertificateFilePath()));
+            }
+            // certificate path can be empty in case of PKCS11 keys
```
MikeDombo, line 214: The certificate must exist in the TPM, so why shouldn't we require that the URI to the cert in the TPM be specified?

src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
prateek-y: We can rely on security service to do the right thing, just make it optional here?

src/main/java/com/aws/greengrass/security/SecurityService.java
prateek-y: Because the field name is confusing. certificateFilePath
and java requires you to have same label on cert as private key anyway

vaibhavmurkute: We require customers to use the same label when loading privateKey and certificate onto the TPM. Greengrass PKCS11 provider needs only this label to access these keys off of TPM, using KeyStore. Certificate-uri is redundant since we can use the same privateKey label, extracted from `privateKeyPath` uri. So customers may not specify `certificateFilePath` in config.

Here is the corresponding PR for PKCS11 provider: https://github.com/aws-greengrass/aws-greengrass-pkcs11-provider/pull/13

PR #1132, opened 2021-11-12
feat: add interface to enable blocking service state transitions by a plugin

MikeDombo:
**Issue #, if available:**

**Description of changes:**
Add an interface to allow a plugin to prevent the state transition of Greengrass services. If any registered allower disallows the state transition, then the state transition will not be allowed to proceed.

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
+++ b/src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
@@ -67,6 +67,7 @@ public class Lifecycle {
     // The maximum number of ERRORED before transitioning the service state to BROKEN.
     private static final int MAXIMUM_CONTINUAL_ERROR = 3;
     private static final long DEFAULT_ERROR_RESET_TIME_IN_SEC = Duration.ofHours(1).getSeconds();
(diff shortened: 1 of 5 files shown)


src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
```diff
@@ -433,6 +436,16 @@ private void handleCurrentStateNew(Optional<State> desiredState) throws Interrup
...
+            // Wait until we're allowed to transition into INSTALLED
+            while (!stateTransitionAllowerService.isStateTransitionAllowed(greengrassService, State.NEW,
+                    State.INSTALLED)) {
+                try {
+                    Thread.sleep(STATE_TRANSITION_ALLOWED_POLLING_TIME_MILLIS);
+                } catch (InterruptedException e) {
+                    Thread.currentThread().interrupt();
+                    return;
+                }
+            }
```
philcali, line 448: Possible to throw this in a method that takes the `from` -> `to` args? Looks the same as the one below.

philcali (approved):

philcali (approved):

PR #1133, opened 2021-11-15
fix(win32): set invalid handle value properly on 32 bit systems

MikeDombo:
**Issue #, if available:**

**Description of changes:**
Properly set the invalid handle value on 32 bit Windows systems. Based on https://github.com/java-native-access/jna/blob/master/contrib/platform/src/com/sun/jna/platform/win32/WinBase.java#L51-L53

Without this change, our code believes that it does not need to create the standard out/error pipes because it thinks that we're passing in a valid handle already, meaning that the process will write to a file which exists instead of pipes that we need to create first.

Verified that logs are provided on both 32 and 64 bit JVMs. Without this change only 64 bit JVMs work properly.

**Checklist:**

### DIFF
--- a/src/main/java/vendored/org/apache/dolphinscheduler/common/utils/process/ProcessImplForWin32.java
+++ b/src/main/java/vendored/org/apache/dolphinscheduler/common/utils/process/ProcessImplForWin32.java
@@ -95,7 +95,11 @@ public class ProcessImplForWin32 extends Process {
 
     private static final int OFFSET_WRITE = 1;
 
-    private static final WinNT.HANDLE JAVA_INVALID_HANDLE_VALUE = new WinNT.HANDLE(Pointer.createConstant(-1));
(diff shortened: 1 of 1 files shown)


rightpeter (dismissed):

src/main/java/vendored/org/apache/dolphinscheduler/common/utils/process/ProcessImplForWin32.java
```diff
@@ -95,7 +95,8 @@
 
     private static final int OFFSET_WRITE = 1;
 
-    private static final WinNT.HANDLE JAVA_INVALID_HANDLE_VALUE = new WinNT.HANDLE(Pointer.createConstant(-1));
+    private static final WinNT.HANDLE JAVA_INVALID_HANDLE_VALUE =
+            new WinNT.HANDLE(Pointer.createConstant(POINTER_SIZE == 8 ? -1L : 4294967295L));
```
philcali, line 99: `4294967295L` is too magical. Can that be hoisted up as a static and then link to some docs?

philcali (approved):

tilo-chen (approved):

PR #1134, opened 2021-11-15
feat: add optional setting to enable interpolation in configuration during deployments

jmrtinb:
**Issue #, if available:**
N/A

**Description of changes:**
Added a new setting, `interpolateConfiguration` to nucleus config that allows configuration to be interpolated during a deployment. Extended interpolation to handle lists.

We want to be able to deploy the LegacySubscriptionRouter in a thinggroup deployment, but have it create subscriptions to device specific topics. For example, we may wish to create a subscription to the thing's shadow topics which currently would either require subscribing all greengrass devices to all thing's shadow changes (through a wildcard subscription) which is obviously untenable, or deploy the LegacySubscriptionRouter component individually to each device. The latter option however means that we have to build up a separate infrastructure for keeping those deployments up to date any time we need to add a new subscription across to the devices in our fleet. By allowing interpolation inside configuration, we could create a single fleet-wide deployment which references `{iot:thingName}`.

I have done some preliminary testing using our real world scenario, as well having added several unit tests to confirm the behavior of the interpolation. To keep the change backwards compatible, the new behavior defaults to off.

**Checklist:**
 - [X] Updated the README if applicable
 - [X] You confirm that the change is backwards compatible

### DIFF
--- a/src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
+++ b/src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
@@ -85,7 +85,7 @@ import static java.nio.file.StandardCopyOption.REPLACE_EXISTING;
 /**
  * Class for providing device configuration information.
  */
-@SuppressWarnings("PMD.DataClass")
+@SuppressWarnings({"PMD.DataClass", "PMD.ExcessivePublicCount"})
 @SuppressFBWarnings("IS2_INCONSISTENT_SYNC")
 public class DeviceConfiguration {
 
@@ -99,6 +99,7 @@ public class DeviceConfiguration {
     public static final String DEVICE_PARAM_PRIVATE_KEY_PATH = "privateKeyPath";
     public static final String DEVICE_PARAM_CERTIFICATE_FILE_PATH = "certificateFilePath";
     public static final String DEVICE_PARAM_ROOT_CA_PATH = "rootCaPath";
+    public static final String DEVICE_PARAM_INTERPOLATE_COMPONENT_CONFIGURATION = "interpolateComponentConfiguration";
     public static final String SYSTEM_NAMESPACE_KEY = "system";
     public static final String PLATFORM_OVERRIDE_TOPIC = "platformOverride";
     public static final String DEVICE_PARAM_AWS_REGION = "awsRegion";
@@ -583,6 +584,10 @@ public class DeviceConfiguration {
     public Topic getRootCAFilePath() {
         return kernel.getConfig().lookup(SYSTEM_NAMESPACE_KEY, DEVICE_PARAM_ROOT_CA_PATH).dflt("")
                 .addValidator(deTildeValidator);
+    }
+
+    public Topic getInterpolateComponentConfiguration() {
+        return getTopic(DEVICE_PARAM_INTERPOLATE_COMPONENT_CONFIGURATION).dflt(false);
     }
 
     public Topic getIotDataEndpoint() {
--- a/src/test/java/com/aws/greengrass/componentmanager/KernelConfigResolverTest.java
+++ b/src/test/java/com/aws/greengrass/componentmanager/KernelConfigResolverTest.java
@@ -387,6 +387,213 @@ class KernelConfigResolverTest {
     }
 
     @Test
+    void GIVEN_deployment_with_artifact_and_interpolate_config_set_WHEN_config_resolution_requested_THEN_artifact_path_should_be_interpolated_in_deployment_configuration()
+            throws Exception {
+        // GIVEN
+        ComponentIdentifier rootComponentIdentifier =
+                new ComponentIdentifier(TEST_INPUT_PACKAGE_A, new Semver("1.2", Semver.SemverType.NPM));
+        List<ComponentIdentifier> packagesToDeploy = Arrays.asList(rootComponentIdentifier);
+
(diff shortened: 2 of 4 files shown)


src/test/java/com/aws/greengrass/componentmanager/KernelConfigResolverTest.java
```diff
@@ -386,6 +386,213 @@ void GIVEN_deployment_with_artifact_WHEN_config_resolution_requested_THEN_artifa
...
+        Map<String, Object> servicesConfig = (Map<String, Object>) resolvedConfig.get(SERVICES_NAMESPACE_TOPIC);
+
+        Path jarPath = Paths.get("/packages/artifacts").toAbsolutePath();
+        assertThat("{configuration:/jarPath} should be replace by the package's artifact path",
+                getServiceRunCommand(TEST_INPUT_PACKAGE_A, servicesConfig),
+                equalTo("java -jar " + jarPath + "/test.jar -x arg"));
+    }
+
+    @Test
+    void GIVEN_interpolate_config_set_WHEN_config_resolution_requested_THEN_self_referential_config_resolved_one_level()
```
MikeDombo, line 440: Do you think this is the appropriate/best/most understandable behavior to only resolve a single level instead of trying to recursively resolve until the interpolation value is found or cannot be resolved?

MikeDombo:
Have you used this build in your own testing and found that this works for your use case of using the thing name?

src/test/java/com/aws/greengrass/componentmanager/KernelConfigResolverTest.java
jmrtinb: That's a good question - I would suspect for most usescases resolving a single level like this is enough. I had started with this way because the alternative gets a bit more complicated, not just with dealing with figuring out when to stop resolving, but also either detecting loops or disallowing a component from referencing it's own config (which would mean needing slightly different codepaths between interpolation for lifecycle vs config).

With the behavior as implemented you can still sort of get two levels of resolution if the first is in the config and the second is in the lifecycle, and at least for usecases I have that's enough - e.g. I can have `{iot:ThingName}` in the config for an environment variable in a lambda component, and that component interpolates `{configuration:/lambdaExecutionParameters}` (where those env vars are set) into it's setenv, and because configuration is interpolated first, and then lifecycle is interpolated after that first interpolation is applied, that works.

MikeDombo:
I also messed up the build by merging from main into your branch here. If you could, please rebase your change on the latest `main` and then update this PR again so that we can have passing tests. Thank you.

src/test/java/com/aws/greengrass/componentmanager/KernelConfigResolverTest.java
MikeDombo: Yes that sounds good to me. If we have a requirement later on it sounds like we could add that ability without issue (not a one-way door).

jmrtinb:
> Have you used this build in your own testing and found that this works for your use case of using the thing name?

Yeah, I've been using a similar build for some testing for a little bit, and just yesterday tried using this specific build as well.

> I also messed up the build by merging from main into your branch here. If you could, please rebase your change on the latest main and then update this PR again so that we can have passing tests. Thank you.

Will do

MikeDombo:
Please update your commit message to conform to convention commits format, for example "feat: enable interpolation in configuration during deployments".

In discussion with the team, we decided that this does not need to be an optional feature.

shaguptashaikh (commented):
LGTM. Thanks for your contribution and effort implementing the tests for this. 

One minor thing that could be improved if possible is the name of this flag `interpolateConfiguration`. It can be confusing to customers and developers since currently interpolation in nucleus context is used to describe component configuration getting evaluated and replaced in the recipe. Something like `embedSystemConfigInComponentConfig` might convey what it means but is terribly long.

src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
```diff
@@ -99,6 +99,7 @@
     public static final String DEVICE_PARAM_PRIVATE_KEY_PATH = "privateKeyPath";
     public static final String DEVICE_PARAM_CERTIFICATE_FILE_PATH = "certificateFilePath";
     public static final String DEVICE_PARAM_ROOT_CA_PATH = "rootCaPath";
+    public static final String DEVICE_PARAM_INTERPOLATE_CONFIGURATION = "interpolateConfiguration";
```
MikeDombo, line 102: `interpolateComponentConfiguration`

`DEVICE_PARAM_INTERPOLATE_COMPONENT_CONFIGURATION`

src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
```diff
@@ -585,6 +586,10 @@ public Topic getRootCAFilePath() {
                 .addValidator(deTildeValidator);
     }
 
+    public Topic getInterpolateConfiguration() {
```
MikeDombo, line 589: `getInterpolateComponentConfiguration`

MikeDombo (approved):

MikeDombo (approved):

shaguptashaikh (approved):

rightpeter (approved):

MikeDombo (approved):

PR #1136, opened 2021-11-18
fix(WindowsExec): env var loading order

tilo-chen:
**Issue #, if available:**

**Description of changes:**
- Add a new field `defaultEnvironment` to ProcessBuilderForWin32 and accessible via setter.
- In ProcessImplForWin32, apply default env before the env returned by CreateEnvironmentBlock, so that the default does not override user's env.

- Fix default PATH always overriding PATH in user's env
- If we have more default env vars in the future, it'll be handled correctly as well.

- Added a new integ test

**Checklist:**

### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/util/ExecIntegTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/util/ExecIntegTest.java
@@ -1,0 +1,110 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.aws.greengrass.integrationtests.util;
+
+import com.aws.greengrass.integrationtests.BaseITCase;
+import com.aws.greengrass.util.Exec;
+import com.aws.greengrass.util.platforms.Platform;
+import com.sun.jna.platform.win32.Advapi32Util;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.EnabledOnOs;
+import org.junit.jupiter.api.condition.OS;
+
+import java.io.IOException;
+import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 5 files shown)


MikeDombo (commented):
Can we get a test for this to show the order of overwriting is correct?

src/integrationtests/java/com/aws/greengrass/integrationtests/util/ExecTest.java
```diff
@@ -37,29 +32,10 @@
...
 import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
-import static org.mockito.Mockito.lenient;
-import static org.mockito.Mockito.mock;
 
 class ExecTest {
 
-    @TempDir
-    protected Path tempDir;
-    private static Kernel kernel;
```
tilo-chen, line 47: Removed because these tests don't need kernel

MikeDombo (approved):

MikeDombo (approved):

rightpeter (approved):

src/integrationtests/java/com/aws/greengrass/integrationtests/util/ExecIntegTest.java
```diff
@@ -0,0 +1,113 @@
...
+        try (Exec exec = Platform.getInstance().createNewProcessRunner()) {
+            String output = exec.cd("C:\\").withUser(WINDOWS_TEST_UESRNAME)
+                    .withShell("echo", "%" + TEST_ENV_ENTRY + "%").execAndGetStringOutput();
+            assertEquals(expectedVal, output);
+        }
+
+        // Set user-level env var via registry. should override the system-level setting
+        // FYI: Usually, user-level setting has higher precedence than system-level. PATH is special.
+        // User PATH is appended to system-level PATH by windows.
+        expectedVal = "user account env var";
```
rightpeter, line 87: Nit pick if we can give it different name instead of using expectedVal.

PR #1137, opened 2021-11-23
fix: do not fail deployment when missing permission to get thing grou…

shaguptashaikh:
…p hierarchy

**Issue #, if available:**

**Description of changes:**
`greengrass:ListThingGroupsForCoreDevice` call was recently added in 2.5.0 version for cleaning up components previously deployed to thing groups that the device no longer belongs to and avoiding version conflicts. However, users might be setting up devices with pre-existing policy that may not include this permission, so devices if set up with latest nucleus will fail any deployments. This change makes it backwards compatible by proceeding with the deployment but logs a warning so customers will know to update their policy to avail the cleanup feature. 

Manually tested that without this fix a device without this permission fails deployments. But with this fix, it only logs the warning and deployment proceeds

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/deployment/DefaultDeploymentTask.java
+++ b/src/main/java/com/aws/greengrass/deployment/DefaultDeploymentTask.java
@@ -22,6 +22,7 @@ import com.aws.greengrass.logging.api.Logger;
 import com.aws.greengrass.util.Coerce;
 import com.vdurmont.semver4j.Semver;
 import lombok.Getter;
+import software.amazon.awssdk.services.greengrassv2data.model.GreengrassV2DataException;
 
 import java.io.IOException;
 import java.util.ArrayList;
@@ -194,6 +195,18 @@ public class DefaultDeploymentTask implements DeploymentTask {
         Optional<Set<String>> groupsForDeviceOpt;
         try {
             groupsForDeviceOpt = thingGroupHelper.listThingGroupsForDevice(retryCount);
+        } catch (GreengrassV2DataException e) {
+            if (e.statusCode() == 403) {
+                // Getting group hierarchy requires permission to call the ListThingGroupsForCoreDevice API which
+                // may not be configured on existing IoT Thing policy in use for current device, log a warning in
+                // that case and move on.
+                logger.atWarn().setCause(e).log("Failed to get thing group hierarchy. Deployment will proceed. "
+                        + "To automatically clean up unused components, please add "
+                        + "greengrass:ListThingGroupsForCoreDevice permission to your IoT Thing policy.");
+                groupsForDeviceOpt = getPersistedMembershipInfo();
+            } else {
+                throw new DeploymentTaskFailureException("Error fetching thing group information", e);
+            }
         } catch (Exception e) {
             if (isLocalDeployment && ThingGroupHelper.DEVICE_OFFLINE_INDICATIVE_EXCEPTIONS.contains(e.getClass())) {
                 logger.atWarn().setCause(e).log("Failed to get thing group hierarchy, local deployment will proceed");


J-C-L (approved):

src/main/java/com/aws/greengrass/deployment/DefaultDeploymentTask.java
```diff
@@ -198,6 +199,14 @@ public DeploymentResult call() throws InterruptedException {
             if (isLocalDeployment && ThingGroupHelper.DEVICE_OFFLINE_INDICATIVE_EXCEPTIONS.contains(e.getClass())) {
                 logger.atWarn().setCause(e).log("Failed to get thing group hierarchy, local deployment will proceed");
                 groupsForDeviceOpt = getPersistedMembershipInfo();
+            } else if (e instanceof GreengrassV2DataException && ((GreengrassV2DataException) e).statusCode() == 403) {
```
MikeDombo, line 202: positive that this is the exception type? May want to use the super class (SdkServiceException?)

rightpeter (dismissed):

src/main/java/com/aws/greengrass/deployment/DefaultDeploymentTask.java
MikeDombo: or get more specific and use AccessDeniedException

src/main/java/com/aws/greengrass/deployment/DefaultDeploymentTask.java
```diff
@@ -194,6 +195,18 @@ public DeploymentResult call() throws InterruptedException {
         Optional<Set<String>> groupsForDeviceOpt;
         try {
             groupsForDeviceOpt = thingGroupHelper.listThingGroupsForDevice(retryCount);
+        } catch(GreengrassV2DataException e) {
+            if (e.statusCode() == 403) {
```
MikeDombo, line 199: Just use AccessDeniedException?

src/main/java/com/aws/greengrass/deployment/DefaultDeploymentTask.java
```diff
@@ -194,6 +195,18 @@ public DeploymentResult call() throws InterruptedException {
...
+                // Getting group hierarchy requires permission to call the ListThingGroupsForCoreDevice API which
+                // may not be configured on existing IoT Thing policy in use for current device, log a warning in
+                // that case and move on.
+                logger.atWarn().setCause(e).log("Failed to get thing group hierarchy. Deployment will proceed. "
+                        + "To automatically clean up unused components, please add "
+                        + "greengrass:ListThingGroupsForCoreDevice permission to your IoT Thing policy.");
+                groupsForDeviceOpt = getPersistedMembershipInfo();
+            } else {
+                throw new DeploymentTaskFailureException("Error fetching thing group information", e);
```
MikeDombo, line 208: As an aside, what gets reported as the error in the deployment status? Do we follow the whole exception chain and generate a message or what are we doing? If we're only taking the top exception of "Error fetch thing group information" I think we could improve that experience.

MikeDombo (approved):

src/main/java/com/aws/greengrass/deployment/DefaultDeploymentTask.java
shaguptashaikh: It looks like we only include the top level message due to the way the task failure exception is wrapped around the original error, so yeah we should definitely improve it. Will take that up separately since this is actually avoiding a deployment failure right now

src/main/java/com/aws/greengrass/deployment/DefaultDeploymentTask.java
shaguptashaikh: `GreengrassV2DataException` is the reported error and it includes status code unfortunately

junfuchen99 (approved):

src/main/java/com/aws/greengrass/deployment/DefaultDeploymentTask.java
```diff
@@ -194,6 +195,18 @@ public DeploymentResult call() throws InterruptedException {
         Optional<Set<String>> groupsForDeviceOpt;
         try {
             groupsForDeviceOpt = thingGroupHelper.listThingGroupsForDevice(retryCount);
+        } catch (GreengrassV2DataException e) {
+            if (e.statusCode() == 403) {
```
JamieHunter, line 199: Not blocking / follow-up checkin: Let's use constants rather than magic numbers (they should already exist).

src/main/java/com/aws/greengrass/deployment/DefaultDeploymentTask.java
```diff
@@ -194,6 +195,18 @@ public DeploymentResult call() throws InterruptedException {
...
+                // Getting group hierarchy requires permission to call the ListThingGroupsForCoreDevice API which
+                // may not be configured on existing IoT Thing policy in use for current device, log a warning in
+                // that case and move on.
+                logger.atWarn().setCause(e).log("Failed to get thing group hierarchy. Deployment will proceed. "
```
JamieHunter, line 203: Not blocking / follow-up checkin: I'm almost inclined to report this at error level. the choice to continue is for robustness, at the cost of correct behavior.

src/main/java/com/aws/greengrass/deployment/DefaultDeploymentTask.java
```diff
@@ -194,6 +195,18 @@ public DeploymentResult call() throws InterruptedException {
...
+                        + "To automatically clean up unused components, please add "
+                        + "greengrass:ListThingGroupsForCoreDevice permission to your IoT Thing policy.");
+                groupsForDeviceOpt = getPersistedMembershipInfo();
+            } else {
+                throw new DeploymentTaskFailureException("Error fetching thing group information", e);
+            }
         } catch (Exception e) {
             if (isLocalDeployment && ThingGroupHelper.DEVICE_OFFLINE_INDICATIVE_EXCEPTIONS.contains(e.getClass())) {
                 logger.atWarn().setCause(e).log("Failed to get thing group hierarchy, local deployment will proceed");
```
JamieHunter, line 212: Not blocking / follow-up checkin: This feels wrong - device may go offline, doesn't mean it's a local deployment.

JamieHunter (approved):
I'm good for check-in as is for fast sev-2 fix, but have some comments to address in a follow-up change.

src/main/java/com/aws/greengrass/deployment/DefaultDeploymentTask.java
MikeDombo: We know this is a local deployment at this point. There is a check above for isLocalDeployment

JamieHunter: I see it now, yes.

PR #1138, opened 2021-11-29
fix: credential provider should not use TES service directly

MikeDombo:
**Issue #, if available:**

**Description of changes:**
LazyCredentialProvider was using TokenExchangeService directly which caused TES to be created and stored in the Context. However, since TES wasn't depended on by anything, it was not known by the Nucleus to be a component which shouldn't be removed. Therefore, when a deployment came it, that deployment would remove the config namespace for TES, however TES still had pointers to that now-removed namespace. Because of this, the in-memory representation that TES had about itself differed from the representation of any component outside of TES. In this case, TES believed that it was version 0.0.0, while anyone outside of TES (not going through TES's outdated pointers) would see that the version was properly set to 2.0.3. 
The fix in this change is to have the LazyCredentialProvider _not_ use TES, but instead go to the source for the credentials. TES was only redirecting the calls, so now this is more clear and will resolve the issue of TES being constructed and holding onto outdated pointers.

**Checklist:**

### DIFF
--- a/src/test/java/com/aws/greengrass/tes/CredentialRequestHandlerTest.java
+++ b/src/test/java/com/aws/greengrass/tes/CredentialRequestHandlerTest.java
@@ -7,6 +7,11 @@ package com.aws.greengrass.tes;
 
 import com.aws.greengrass.authorization.AuthorizationHandler;
 import com.aws.greengrass.authorization.exceptions.AuthorizationException;
+import com.aws.greengrass.config.Subscriber;
+import com.aws.greengrass.config.Topic;
+import com.aws.greengrass.config.WhatHappened;
+import com.aws.greengrass.dependency.Context;
+import com.aws.greengrass.deployment.DeviceConfiguration;
 import com.aws.greengrass.deployment.exceptions.AWSIotException;
 import com.aws.greengrass.iot.IotCloudHelper;
 import com.aws.greengrass.iot.IotConnectionManager;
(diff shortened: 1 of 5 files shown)


shaguptashaikh (dismissed):

src/test/java/com/aws/greengrass/tes/CredentialRequestHandlerTest.java
```diff
@@ -95,6 +97,9 @@
     @Mock
     OutputStream mockStream;
 
+    @Mock(answer = RETURNS_DEEP_STUBS)
```
tilo-chen, line 100: This is neat!

src/test/java/com/aws/greengrass/tes/TokenExchangeServiceTest.java
```diff
@@ -108,14 +103,9 @@ void setup() {
         Topics servicesTopics = Topics.of(context, SERVICES_NAMESPACE_TOPIC, null);
         Topic componentTypeTopic = Topic.of(context, SERVICE_TYPE_TOPIC_KEY, ComponentType.NUCLEUS.name());
         Topic componentStoreSizeLimitTopic = Topic.of(context, COMPONENT_STORE_MAX_SIZE_BYTES, 10_000_000_000L);
-        Topic thingName = Topic.of(context, SYSTEM_NAMESPACE_KEY, "abc");
```
tilo-chen, line 111: Now these are removed from TES tests, do CredentialRequestHandler tests cover changing related device configs?

MikeDombo: updated

tilo-chen (approved):

shaguptashaikh (approved):

PR #1148, opened 2021-12-23
feat: pubsub wildcard support

youtuyy:
**Issue #, if available:**

**Description of changes:**
Support MQTT style wildcard in PubSub subscriptions (using + and #) 
Implement a SubscriptionTrie to manage topic insert/search
Update device sdk to 1.6.0-PUBSUB-SNAPSHOT that includes model change for `JsonMessage` and `BinaryMessage` to include `eventTopic`

Now we can only subscribe to some specific topic which does not scale well. With the change we can subscribe to a set of topics using + and # wildcards. e.g. to subscribe to a set of shadow topics for different shadows and operations

mvn -U -ntp verify

Referenced [POC](https://github.com/aws-greengrass/aws-greengrass-nucleus/commit/492407c4972ad4cce1de9117cff4de00bb46c84b), current [MqttTopic](https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/e8be1f23516d22d3e38726d753e752cd08707f76/src/main/java/com/aws/greengrass/mqttclient/MqttTopic.java) implementation and v1 [routing](https://github.com/aws-greengrass/aws-greengrass-lambda-manager/blob/main/src/main/java/com/aws/greengrass/lambdamanager/system/v1subscription/RouteTrie.java) implementation in lambda manager.
The PubSub wildcard will differ from [MQTT Spec](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html) in several ways, including
-  "#" does note match parent level
-  "$" has no special meaning

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/builtin/services/pubsub/PubSubIPCEventStreamAgent.java
+++ b/src/main/java/com/aws/greengrass/builtin/services/pubsub/PubSubIPCEventStreamAgent.java
@@ -121,6 +121,7 @@ public class PubSubIPCEventStreamAgent {
             // Still technically successful, just no one was subscribed
             return new PublishToTopicResponse();
         }
+        // TODO: include topic name in message after updating Smithy model
         SubscriptionResponseMessage message = new SubscriptionResponseMessage();
         PublishEvent publishedEvent = PublishEvent.builder().topic(topic).build();
         if (jsonMessage.isPresent()) {
--- a/src/main/java/com/aws/greengrass/builtin/services/pubsub/PubSubIPCEventStreamAgent.java
+++ b/src/main/java/com/aws/greengrass/builtin/services/pubsub/PubSubIPCEventStreamAgent.java
@@ -121,12 +121,12 @@ public class PubSubIPCEventStreamAgent {
             // Still technically successful, just no one was subscribed
             return new PublishToTopicResponse();
         }
+        // TODO: include topic in message when new sdk is ready
         SubscriptionResponseMessage message = new SubscriptionResponseMessage();
         PublishEvent publishedEvent = PublishEvent.builder().topic(topic).build();
         if (jsonMessage.isPresent()) {
             JsonMessage message1 = new JsonMessage();
             message1.setMessage(jsonMessage.get());
-            message1.setEventTopic(topic);
             message.setJsonMessage(message1);
             try {
                 publishedEvent.setPayload(SERIALIZER.writeValueAsBytes(jsonMessage.get()));
@@ -138,7 +138,6 @@ public class PubSubIPCEventStreamAgent {
         if (binaryMessage.isPresent()) {
             BinaryMessage binaryMessage1 = new BinaryMessage();
             binaryMessage1.setMessage(binaryMessage.get());
-            binaryMessage1.setEventTopic(topic);
             message.setBinaryMessage(binaryMessage1);
             publishedEvent.setPayload(binaryMessage.get());
         }
--- a/src/main/java/com/aws/greengrass/builtin/services/pubsub/PubSubIPCEventStreamAgent.java
+++ b/src/main/java/com/aws/greengrass/builtin/services/pubsub/PubSubIPCEventStreamAgent.java
@@ -34,8 +34,6 @@ import software.amazon.awssdk.eventstreamrpc.model.EventStreamJsonMessage;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Consumer;
 import javax.inject.Inject;
 
@@ -50,7 +48,7 @@ public class PubSubIPCEventStreamAgent {
     private static final String GLOB_WILDCARD = "*";
     private static final ObjectMapper SERIALIZER = new ObjectMapper();
     @Getter(AccessLevel.PACKAGE)
-    private final Map<String, Set<Object>> listeners = new ConcurrentHashMap<>();
+    private final SubscriptionTrie listeners = new SubscriptionTrie();
 
     private final OrderedExecutorService orderedExecutorService;
     private final AuthorizationHandler authorizationHandler;
@@ -89,12 +87,7 @@ public class PubSubIPCEventStreamAgent {
      * @param serviceName name of the service unsubscribing.
      */
     public void unsubscribe(String topic, Consumer<PublishEvent> cb, String serviceName) {
-        AtomicBoolean removed = new AtomicBoolean(false);
-        listeners.computeIfPresent(topic, (s, objects) -> {
-            removed.set(objects.remove(cb));
-            return objects.isEmpty() ? null : objects;
-        });
-        if (removed.get()) {
+        if (listeners.remove(topic, cb)) {
             log.atDebug().kv(COMPONENT_NAME, serviceName).log("Unsubscribed from topic {}", topic);
         }
     }
@@ -164,7 +157,8 @@ public class PubSubIPCEventStreamAgent {
 
     private void handleSubscribeToTopicRequest(String topic, String serviceName, Object handler) {
         // TODO: [P32540011]: All IPC service requests need input validation
-        if (listeners.computeIfAbsent(topic, k -> ConcurrentHashMap.newKeySet()).add(handler)) {
+        validateSubTopic(topic);
+        if (listeners.add(topic, handler)) {
             log.atDebug().kv(COMPONENT_NAME, serviceName).log("Subscribed to topic {}", topic);
         }
     }
@@ -225,9 +219,8 @@ public class PubSubIPCEventStreamAgent {
 
         @Override
         protected void onStreamClosed() {
-            if (Utils.isNotEmpty(subscribeTopic) && listeners.containsKey(subscribeTopic)) {
-                listeners.computeIfPresent(subscribeTopic,
-                        (s, objects) -> objects.remove(this) && objects.isEmpty() ? null : objects);
+            if (Utils.isNotEmpty(subscribeTopic)) {
+                listeners.remove(subscribeTopic, this);
             }
         }
 
@@ -250,6 +243,13 @@ public class PubSubIPCEventStreamAgent {
         @Override
         public void handleStreamEvent(EventStreamJsonMessage streamRequestEvent) {
             // NA
+        }
+    }
+
+    private void validateSubTopic(String topic) {
+        // TODO: regex validation? size limit?
+        if (Utils.isEmpty(topic)) {
+            throw new InvalidArgumentsError("Subscribe topic must not be empty");
         }
     }
 
--- a/src/main/java/com/aws/greengrass/builtin/services/pubsub/PubSubIPCEventStreamAgent.java
+++ b/src/main/java/com/aws/greengrass/builtin/services/pubsub/PubSubIPCEventStreamAgent.java
@@ -121,12 +121,12 @@ public class PubSubIPCEventStreamAgent {
             // Still technically successful, just no one was subscribed
             return new PublishToTopicResponse();
         }
-        // TODO: include topic name in message after updating Smithy model
         SubscriptionResponseMessage message = new SubscriptionResponseMessage();
         PublishEvent publishedEvent = PublishEvent.builder().topic(topic).build();
         if (jsonMessage.isPresent()) {
             JsonMessage message1 = new JsonMessage();
             message1.setMessage(jsonMessage.get());
+            message1.setEventTopic(topic);
             message.setJsonMessage(message1);
             try {
                 publishedEvent.setPayload(SERIALIZER.writeValueAsBytes(jsonMessage.get()));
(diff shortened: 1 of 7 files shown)


src/test/java/com/aws/greengrass/builtin/services/pubsub/PubSubIPCEventStreamAgentTest.java
```diff
@@ -303,6 +305,15 @@ void GIVEN_subscribed_consumer_WHEN_publish_binary_message_THEN_publishes_And_TH
         assertEquals(0, pubSubIPCEventStreamAgent.getListeners().size());
     }
 
+    @Test
```
MikeDombo, line 308: Let's get a test for the wildcard stuff.

src/test/java/com/aws/greengrass/builtin/services/pubsub/SubscriptionTrieTest.java
```diff
@@ -0,0 +1,126 @@
...
+    @Test
+    public void GIVEN_subscription_wildcard_WHEN_remove_topic_THEN_no_matches() {
+        Object cb1 = new Object();
+        Object cb2 = new Object();
+        String topic = "foo/+/bar/#";
+        trie.add(topic, cb1);
+        trie.add(topic, cb2);
+        assertTrue(trie.containsKey(topic));
+        assertThat(trie.get(topic), containsInAnyOrder(cb1, cb2));
+        assertEquals(trie.size(), 2);
```
MikeDombo, line 117: These assertions are backwards. Order is expected and then the actual value.

src/main/java/com/aws/greengrass/builtin/services/pubsub/SubscriptionTrie.java
```diff
@@ -0,0 +1,179 @@
...
+import java.util.LinkedHashSet;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * Trie to manage subscriptions.
+ */
+public class SubscriptionTrie {
+    private static final String TOPIC_LEVEL_SEPARATOR = "/";
```
MikeDombo, line 20: These should all be characters, not strings.

src/main/java/com/aws/greengrass/builtin/services/pubsub/SubscriptionTrie.java
```diff
@@ -0,0 +1,179 @@
...
+     * @param topic topic
+     * @return a set of callback objects
+     */
+    public Set<Object> get(String topic) {
+        String[] topicLevels = topic.split(TOPIC_LEVEL_SEPARATOR);
+        Set<Object> result = new LinkedHashSet<>();
+        Set<SubscriptionTrie> paths = this.getMatchingPaths(topicLevels[0], result);
+
+        for (int iter = 1; iter < topicLevels.length && !paths.isEmpty(); iter++) {
+            Set<SubscriptionTrie> newPaths = new LinkedHashSet<>();
```
MikeDombo, line 164: Why create a new set here? Why not just append to the result set?

src/test/java/com/aws/greengrass/builtin/services/pubsub/SubscriptionTrieTest.java
```diff
@@ -0,0 +1,126 @@
...
+import static org.hamcrest.Matchers.contains;
+import static org.hamcrest.Matchers.containsInAnyOrder;
+import static org.hamcrest.Matchers.empty;
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.not;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.params.provider.Arguments.arguments;
+
+public class SubscriptionTrieTest {
```
rbattle, line 29: is there any test for size()?

src/test/java/com/aws/greengrass/builtin/services/pubsub/SubscriptionTrieTest.java
youtuyy: thanks

src/test/java/com/aws/greengrass/builtin/services/pubsub/PubSubIPCEventStreamAgentTest.java
youtuyy: added

src/main/java/com/aws/greengrass/builtin/services/pubsub/SubscriptionTrie.java
youtuyy: Not sure how this can be done. the new set is for storing all current matching paths and result is for callbacks

src/test/java/com/aws/greengrass/builtin/services/pubsub/SubscriptionTrieTest.java
youtuyy: tested in the cases for add and removing topics so I didn't add a specific one
GIVEN_subscription_WHEN_remove_topic_THEN_no_matches and GIVEN_subscription_wildcard_WHEN_remove_topic_THEN_no_matches

src/main/java/com/aws/greengrass/builtin/services/pubsub/SubscriptionTrie.java
MikeDombo: You already have a result set, right? So why don't you just add to that set instead of creating a new set, adding to that set, then adding that set into the result set

src/main/java/com/aws/greengrass/builtin/services/pubsub/PubSubIPCEventStreamAgent.java
```diff
@@ -128,6 +121,7 @@ private PublishToTopicResponse handlePublishToTopicRequest(String topic, String
             // Still technically successful, just no one was subscribed
             return new PublishToTopicResponse();
         }
+        // TODO: include topic name in message after updating Smithy model
```
MikeDombo, line 124: do this in the same commit

src/main/java/com/aws/greengrass/builtin/services/pubsub/SubscriptionTrie.java
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    /**
+     * Add a topic and a set of callbacks.
+     *
+     * @param topic topic
+     * @param cbs   callbacks
+     */
+    public void put(String topic, Set<Object> cbs) {
+        SubscriptionTrie[] current = {this};
```
MikeDombo, line 123: this array isn't necessary is it?

src/main/java/com/aws/greengrass/builtin/services/pubsub/SubscriptionTrie.java
```diff
@@ -0,0 +1,179 @@
...
+
+    private Set<SubscriptionTrie> getMatchingPaths(String topicLevel, Set<Object> result) {
+        Set<SubscriptionTrie> paths = new LinkedHashSet<>();
+
+        SubscriptionTrie childPath = this.children.get(topicLevel);
+        if (childPath != null) {
+            paths.add(childPath);
+        }
+
+        SubscriptionTrie childPlusPath = this.children.get(String.valueOf(SINGLE_LEVEL_WILDCARD));
```
MikeDombo, line 138: Store these as both strings and characters so that you have the proper type already created; calling String.valueOf on every single lookup will add up and hurt performance.

`final SINGLE_LEVEL_WILDCARD_STR = "+";`
`final SINGLE_LEVEL_WILDCARD = SINGLE_LEVEL_WILDCARD_STR.charAt(0);`

src/test/java/com/aws/greengrass/builtin/services/pubsub/SubscriptionTrieTest.java
```diff
@@ -0,0 +1,128 @@
...
```
MikeDombo, line 29: Add ExtendWith GGExtension.

src/test/java/com/aws/greengrass/builtin/services/pubsub/SubscriptionTrieTest.java
```diff
@@ -0,0 +1,128 @@
...
+        for (String topic : topics) {
+            assertThat(trie.get(topic), containsInAnyOrder(cb1, cb2));
+        }
+    }
+
+    static Stream<Arguments> subscriptionMatch() {
+        return Stream.of(
+                arguments("foo", singletonList("foo")),
+                arguments("foo/bar", singletonList("foo/bar")),
+//                // multilevel wildcard #
```
tilo-chen, line 54: nit: extra `//`

src/main/java/com/aws/greengrass/builtin/services/pubsub/SubscriptionTrie.java
```diff
@@ -0,0 +1,179 @@
...
+ */
+public class SubscriptionTrie {
+    private static final char TOPIC_LEVEL_SEPARATOR = '/';
+    private static final char SINGLE_LEVEL_WILDCARD = '+';
+    private static final char MULTI_LEVEL_WILDCARD = '#';
+
+    private final Map<String, SubscriptionTrie> children = new ConcurrentHashMap<>();
+    @SuppressWarnings("PMD.UnusedPrivateField")
+    @SuppressFBWarnings("URF_UNREAD_FIELD")
+    private final String value;
```
tilo-chen, line 27: Why is `value` field unused? If `callbacks` is basically the "value" for the trie node, can we remove this field?

src/main/java/com/aws/greengrass/builtin/services/pubsub/SubscriptionTrie.java
```diff
@@ -0,0 +1,179 @@
...
+        return sub.callbacks.removeAll(cbs);
+    }
+
+    /**
+     * Size of trie.
+     *
+     * @return size
+     */
+    public int size() {
+        int[] size = {this.callbacks.size()};
```
tilo-chen, line 95: Creating an array doesn't seem necessary.

src/main/java/com/aws/greengrass/builtin/services/pubsub/SubscriptionTrie.java
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    /**
+     * Add a topic callback.
+     *
+     * @param topic topic
+     * @param cb    callback
+     * @return true
+     */
+    public boolean add(String topic, Object cb) {
```
tilo-chen, line 109: For `remove`, we have overloaded method for a single or a set of `cb`. I think it's a good idea to be consistent and overload `put` instead of calling this `add`

src/main/java/com/aws/greengrass/builtin/services/pubsub/SubscriptionTrie.java
tilo-chen: So make `getMatchingPaths` take another argument `newPaths`. Instead of returning the new paths from `getMatchingPaths `, just put them in the given newPaths argument. Similar to how the `result` argument work.

Nit: consider rename `iter` -> `level`?

src/main/java/com/aws/greengrass/builtin/services/pubsub/PubSubIPCEventStreamAgent.java
```diff
@@ -50,7 +48,7 @@
     private static final String GLOB_WILDCARD = "*";
     private static final ObjectMapper SERIALIZER = new ObjectMapper();
     @Getter(AccessLevel.PACKAGE)
-    private final Map<String, Set<Object>> listeners = new ConcurrentHashMap<>();
```
tilo-chen, line 53: This was a `ConcurrentHashMap`. Does this mean our new Trie operations must be thread-safe as well?

src/main/java/com/aws/greengrass/builtin/services/pubsub/SubscriptionTrie.java
tilo-chen: Oh in that case, can we just use regular for loop (like `for (SubscriptionTrie child : children.values())`)? So we don't need to create anything extra just to sum up the sizes.

src/main/java/com/aws/greengrass/builtin/services/pubsub/SubscriptionTrie.java
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    /**
+     * Size of trie.
+     *
+     * @return size
+     */
+    public int size() {
+        int size = this.callbacks.size();
+        for (SubscriptionTrie child: children.values()) {
```
MikeDombo, line 86: nit: formatting

MikeDombo (dismissed):

MikeDombo (approved):

tilo-chen (dismissed):

youtuyy:
Rebased and reverted the change to include eventTopic in the message that depends on new sdk to unblock testing and shadow integration before sdk is debugged. Added TODO in PubSubIPCEventStreamAgent.java

MikeDombo (approved):

tilo-chen (approved):

PR #1149, opened 2022-01-07
fix: enable FSS to react to DeviceConfig updates and bring itself online

J-C-L:
**Description of changes:**
Enable FSS to react to DeviceConfig updates and bring itself online

Currently if device provisioning isn't finished, when FSS starts up, FSS will not be able to start and the device won't be picked up by greengrass cloud. Customers have to restart greengrass to have FSS startup.
This fix enables FSS to watch for changes and bring itself up when provisioning completes.

Manually installed greengrass with a delay in fleet provisioning. Confirmed that device shows up in greengrass cloud after some time. 

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
+++ b/src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
@@ -725,12 +725,12 @@ public class DeviceConfiguration {
      * Reports if device provisioning values have changed.
      *
      * @param node what may have changed during device provisioning
-     * @param initialSetupDone has initial setup has been done for a given service
+     * @param checkThingNameOnly has initial setup has been done for a given service
      * @return true if any device provisioning values have changed before initial service setup
      *         or if the thing name has changed after
      */
-    public static boolean  provisionInfoNodeChanged(Node node, Boolean initialSetupDone) {
-        if (initialSetupDone) {
+    public static boolean provisionInfoNodeChanged(Node node, Boolean checkThingNameOnly) {
+        if (checkThingNameOnly) {
             return node.childOf(DEVICE_PARAM_THING_NAME);
         } else {
             // List of configuration nodes that may change during device provisioning
--- a/src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
+++ b/src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
@@ -721,6 +721,26 @@ public class DeviceConfiguration {
         }
     }
 
+    /**
+     * Reports if device provisioning values have changed.
+     *
+     * @param node what may have changed during device provisioning
+     * @param initialSetupDone has initial setup has been done for a given service
+     * @return true if any device provisioning values have changed before initial service setup
+     *         or if the thing name has changed after
+     */
+    public static boolean  provisionInfoNodeChanged(Node node, Boolean initialSetupDone) {
+        if (initialSetupDone) {
+            return node.childOf(DEVICE_PARAM_THING_NAME);
+        } else {
+            // List of configuration nodes that may change during device provisioning
+            return node.childOf(DEVICE_PARAM_THING_NAME) || node.childOf(DEVICE_PARAM_IOT_DATA_ENDPOINT)
+                    || node.childOf(DEVICE_PARAM_PRIVATE_KEY_PATH)
+                    || node.childOf(DEVICE_PARAM_CERTIFICATE_FILE_PATH) || node.childOf(DEVICE_PARAM_ROOT_CA_PATH)
+                    || node.childOf(DEVICE_PARAM_AWS_REGION);
+        }
+    }
+
     private Topic getTopic(String parameterName) {
         return kernel.getConfig()
                 .lookup(SERVICES_NAMESPACE_TOPIC, getNucleusComponentName(), CONFIGURATION_CONFIG_KEY, parameterName);
--- a/src/main/java/com/aws/greengrass/deployment/IotJobsHelper.java
+++ b/src/main/java/com/aws/greengrass/deployment/IotJobsHelper.java
@@ -283,7 +283,7 @@ public class IotJobsHelper implements InjectionActions {
         this.iotJobsClientWrapper = iotJobsClientFactory.getIotJobsClientWrapper(connection);
 
         deviceConfiguration.onAnyChange((what, node) -> {
-            if (node != null && what.equals(WhatHappened.childChanged)
+            if (node != null && WhatHappened.childChanged.equals(what)
                     && deviceConfiguration.provisionInfoNodeChanged(node, this.isSubscribedToIotJobsTopics.get())) {
                 try {
                     connectToIotJobs(deviceConfiguration);
--- a/src/main/java/com/aws/greengrass/deployment/IotJobsHelper.java
+++ b/src/main/java/com/aws/greengrass/deployment/IotJobsHelper.java
@@ -5,7 +5,6 @@
 
 package com.aws.greengrass.deployment;
 
-import com.aws.greengrass.config.Node;
 import com.aws.greengrass.config.WhatHappened;
 import com.aws.greengrass.dependency.InjectionActions;
 import com.aws.greengrass.deployment.exceptions.DeviceConfigurationException;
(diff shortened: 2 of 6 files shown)


src/main/java/com/aws/greengrass/status/FleetStatusService.java
```diff
@@ -156,46 +161,60 @@ public FleetStatusService(Topics topics, MqttClient mqttClient, DeploymentStatus
...
+
+                    config.getContext().addGlobalStateChangeListener(this::handleServiceStateChange);
+
+                    this.deploymentStatusKeeper.registerDeploymentStatusConsumer(IOT_JOBS,
+                        this::deploymentStatusChanged, FLEET_STATUS_SERVICE_TOPICS);
+                    this.deploymentStatusKeeper.registerDeploymentStatusConsumer(LOCAL,
+                        this::deploymentStatusChanged, FLEET_STATUS_SERVICE_TOPICS);
+                    this.deploymentStatusKeeper.registerDeploymentStatusConsumer(SHADOW,
+                        this::deploymentStatusChanged, FLEET_STATUS_SERVICE_TOPICS);
+                    schedulePeriodicFleetStatusDataUpdate(false);
```
rbattle, line 208: do these add multiple listeners / consumers if the config changes multiple times?

src/main/java/com/aws/greengrass/status/FleetStatusService.java
```diff
@@ -66,6 +68,8 @@
     static final String FLEET_STATUS_SEQUENCE_NUMBER_TOPIC = "sequenceNumber";
     static final String FLEET_STATUS_LAST_PERIODIC_UPDATE_TIME_TOPIC = "lastPeriodicUpdateTime";
     private static final int MAX_PAYLOAD_LENGTH_BYTES = 128_000;
+    public static final String DEVICE_OFFLINE_MESSAGE = "Device not configured to talk to AWS Iot cloud. "
```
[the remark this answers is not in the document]
J-C-L: Good tip. This is the first sentence of a 2 line string though, so I think I need the "." after "cloud", but not after "offline" on line 72. Do you agree?

src/main/java/com/aws/greengrass/status/FleetStatusService.java
```diff
@@ -156,46 +161,60 @@ public FleetStatusService(Topics topics, MqttClient mqttClient, DeploymentStatus
...
+
+        deviceConfiguration.onAnyChange((what, node) -> {
+            if (node != null && what.equals(WhatHappened.childChanged)
+                    && deviceConfiguration.provisionInfoNodeChanged(node, this.isFSSSetupComplete.get())) {
+                try {
+                    // Not using isDeviceConfiguredToTalkToCloud() in order to provide the detailed error
+                    // message to user
+                    deviceConfiguration.validate();
+
+                    Topics configurationTopics = deviceConfiguration.getStatusConfigurationTopics();
```
tilo-chen, line 183: Similar to `IotJobsHelper#connectToIotJobs`, we'll want to wrap these setup logic in a method and also run it once here in `postInject` even if there's callback from deviceConfiguration change. Because deviceConfig could be already setup and there's no change. Need to handle that case.

src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
```diff
@@ -721,6 +721,26 @@ public boolean isDeviceConfiguredToTalkToCloud() {
...
+     * Reports if device provisioning values have changed.
+     *
+     * @param node what may have changed during device provisioning
+     * @param initialSetupDone has initial setup has been done for a given service
+     * @return true if any device provisioning values have changed before initial service setup
+     *         or if the thing name has changed after
+     */
+    public static boolean  provisionInfoNodeChanged(Node node, Boolean initialSetupDone) {
+        if (initialSetupDone) {
+            return node.childOf(DEVICE_PARAM_THING_NAME);
```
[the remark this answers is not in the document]
rbattle: never mind - `deviceConfiguration.onAnyChange` listens only to nucleus config or the system config nodes, not anything generic

src/main/java/com/aws/greengrass/status/FleetStatusService.java
J-C-L: Good point. Will wrap setup logic in a separate function and call it outside of the watcher.

src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
```diff
@@ -721,6 +721,26 @@ public boolean isDeviceConfiguredToTalkToCloud() {
...
 
+    /**
+     * Reports if device provisioning values have changed.
+     *
+     * @param node what may have changed during device provisioning
+     * @param initialSetupDone has initial setup has been done for a given service
+     * @return true if any device provisioning values have changed before initial service setup
+     *         or if the thing name has changed after
+     */
+    public static boolean  provisionInfoNodeChanged(Node node, Boolean initialSetupDone) {
```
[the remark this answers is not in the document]
J-C-L: spacing fixed.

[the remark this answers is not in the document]
J-C-L: name changed.

src/main/java/com/aws/greengrass/status/FleetStatusService.java
J-C-L: good question. will look into this.

J-C-L: Added flag to check if initial setup has already been done, and ensured these are only called once.

src/main/java/com/aws/greengrass/status/FleetStatusService.java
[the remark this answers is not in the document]
rbattle: still have the small nit pick that it should be "IoT" , not "Iot"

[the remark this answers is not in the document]
J-C-L: missed that. thanks for the reminder. fixed now.

src/main/java/com/aws/greengrass/deployment/IotJobsHelper.java
```diff
@@ -290,7 +283,8 @@ public void postInject() {
         this.iotJobsClientWrapper = iotJobsClientFactory.getIotJobsClientWrapper(connection);
 
         deviceConfiguration.onAnyChange((what, node) -> {
-            if (node != null && what.equals(WhatHappened.childChanged) && relevantNodeChanged(node)) {
+            if (node != null && what.equals(WhatHappened.childChanged)
```
tilo-chen, line 286: nit: it's a better habit to do `WhatHappened.childChanged.equals(what)` so that in case `what` is null we won't get null pointer exception. Same for line 176 in FSS.

MikeDombo:
Please name the PR in accordance with conventional commits.

src/main/java/com/aws/greengrass/deployment/IotJobsHelper.java
J-C-L: Changed in both places

tilo-chen (dismissed):

src/main/java/com/aws/greengrass/status/FleetStatusService.java
```diff
@@ -173,46 +173,57 @@ public void postInject() {
...
+            logger.atWarn().kv("errorMessage", e.getMessage()).log(DEVICE_OFFLINE_MESSAGE);
+        }
+    }
+
+    private void setUpFSS() throws DeviceConfigurationException {
+        // Not using isDeviceConfiguredToTalkToCloud() in order to provide the detailed error message to user
+        deviceConfiguration.validate();
+        if (!isFSSSetupComplete.get()) {
+            initialFSSsetup();
+        }
```
rbattle, line 197: The way the atomic boolean is used here is susceptible to a time of check/time of set bug. You should use:
```
if (!isFSSSetupComplete.checkAndSet(false, true)) {
  initialFSSsetup();
}
```
You could then move the contents of `initialFSSsetup` into if check here instead - that way there's no separate method modifying the atomic bool.

Super nit: you are inconsistent with capitalization for `isFSSSetup` and `initialFSSsetup`

J-C-L: Do you mean compareAndSet ? I don't see a checkAndSet method in the AtomicBoolean documentation.
I'll make the change as suggested.

Will also make the capitalization consistent.

tilo-chen (approved):

rbattle (approved):

PR #1151, opened 2022-01-11
Feat: Only interpolate system parameter in component configuration section, not configuration parameter

rightpeter:
**Description of changes:**
Before, nucleus interpolates component config after aggregating config from recipe, deployment update and running config. Running config is dynamic which exposes the risk of attacker can manipulate the running config in a way to be interpolated in next deployment.

This change let nucleus only interpolate system parameter for component configuration, not configuration parameter. For example:
* `{configuration:/paramA}` won't be interpolated in configuration section
* `{iot:thingName}` will be interpolated

This makes the component configuration section interpolation behavior static.

Security risk: enabling interpolation for configuration parameter under configuration section can cause issue like
corrupted running config triggers unexpected behavior by having special combination of variable designator.

Unit test

**Checklist:**
 - [X] You confirm that the change is backwards compatible


MikeDombo:
Where are we with this?

rightpeter:
> Where are we with this?

We decided first support interpolation by "reserved keys". Need a bit adjustment to this PR.

PR #1164, opened 2022-02-23
feat: support calling V1 TES when useTesV1 Nucleus config is true

MikeDombo:
**Issue #, if available:**

**Description of changes:**
This change adds a new configuration in the Nucleus component config called "useTesV1". When true, the TES URL path is changed to be the path which the Greengrass V1 cloud offers. The customer can use this in conjunction with setting the `iotCredEndpoint` to the Greengrass ats data endpoint in order to retrieve credentials from the V1 TES. If a customer chooses to do this, they must have a Greengrass group setup with the role associated, a core created with the Greengrass V2 thing's certificate ARN, and must have a deployment created for the group.

This is necessary for customers who want to use Greengrass in a VPC with VPCE as IoT Credential Provider doesn't yet support VPCE.

Added basic E2E test code (not committed) which created a Greengrass group with role and then verified that we got the credentials from it.

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/tes/CredentialRequestHandler.java
+++ b/src/main/java/com/aws/greengrass/tes/CredentialRequestHandler.java
@@ -5,9 +5,11 @@
 
 package com.aws.greengrass.tes;
 
+import com.amazon.aws.iot.greengrass.component.common.SerializerFactory;
 import com.aws.greengrass.authorization.AuthorizationHandler;
 import com.aws.greengrass.authorization.Permission;
 import com.aws.greengrass.authorization.exceptions.AuthorizationException;
+import com.aws.greengrass.config.Topic;
 import com.aws.greengrass.deployment.DeviceConfiguration;
 import com.aws.greengrass.deployment.exceptions.AWSIotException;
 import com.aws.greengrass.iot.IotCloudHelper;
@@ -19,13 +21,14 @@ import com.aws.greengrass.logging.api.Logger;
 import com.aws.greengrass.logging.impl.LogManager;
 import com.aws.greengrass.util.Coerce;
 import com.aws.greengrass.util.DefaultConcurrentHashMap;
+import com.aws.greengrass.util.Utils;
+import com.fasterxml.jackson.annotation.JsonAlias;
 import com.fasterxml.jackson.core.JsonProcessingException;
-import com.fasterxml.jackson.databind.DeserializationFeature;
-import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.sun.net.httpserver.HttpExchange;
 import com.sun.net.httpserver.HttpHandler;
 import lombok.AccessLevel;
+import lombok.NoArgsConstructor;
 import lombok.Setter;
 import software.amazon.awssdk.auth.credentials.AwsCredentials;
 import software.amazon.awssdk.auth.credentials.AwsSessionCredentials;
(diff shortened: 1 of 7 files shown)


src/main/java/com/aws/greengrass/tes/TokenExchangeService.java
```diff
@@ -37,6 +38,9 @@
     private static final String TES_CONFIG_ERROR_STR = "%s parameter is either empty or not configured for TES";
     // randomly choose a port
     private static final int DEFAULT_PORT = 0;
+    // Use a time in the far future so that when we use withNewerValue, the configuration change will always happen,
+    // no matter what the current system clock is (as long as the clock isn't ahead of 3000)
+    private static final long FUTURE = Instant.parse("3000-01-01T00:00:00Z").toEpochMilli();
```
jbutler, line 43: Can you break this into a separate commit? I'd prefer having a commit message that adds a little more context

src/main/java/com/aws/greengrass/tes/CredentialRequestHandler.java
```diff
@@ -115,16 +126,31 @@ public CredentialRequestHandler(final IotCloudHelper cloudHelper, final IotConne
...
         deviceConfiguration.getRootCAFilePath().subscribe((why, newv) -> clearCache());
         deviceConfiguration.getPrivateKeyFilePath().subscribe((why, newv) -> clearCache());
+        Topic tesV1Topic = deviceConfiguration.getNucleusConfig().lookup("useTesV1");
+        tesV1Topic.subscribe((why, newv) -> {
+            if (tesV1Topic.getOnce() == null) {
+                setIotCredentialsPath(Coerce.toString(deviceConfiguration.getIotRoleAlias()));
+            } else {
+                if (Coerce.toBoolean(tesV1Topic)) {
+                    this.iotCredentialsPath = "/greengrass/assumeRoleForGroup";
+                }
```
jbutler, line 151: If the customer sets `useTesV1` to true and then to false, won't this fail to set the credential path back to V2 TES?

src/main/java/com/aws/greengrass/tes/TokenExchangeService.java
MikeDombo: Different commit, same PR, done.

src/main/java/com/aws/greengrass/tes/CredentialRequestHandler.java
MikeDombo: Done.

src/main/java/com/aws/greengrass/tes/CredentialRequestHandler.java
```diff
@@ -115,16 +126,29 @@ public CredentialRequestHandler(final IotCloudHelper cloudHelper, final IotConne
 
         deviceConfiguration.getIotRoleAlias().subscribe((why, newv) -> {
             String iotRoleAlias = Coerce.toString(newv);
-            clearCache();
-            setIotCredentialsPath(iotRoleAlias);
+            if (Utils.isNotEmpty(iotRoleAlias)) {
+                clearCache();
+                setIotCredentialsPath(iotRoleAlias);
+            }
```
jbutler, line 132: I don't think this is quite right - `useTesV1` needs to be definitive otherwise you get into situations where the order in which config is updated or read changes the behavior, right? Role alias should only be processed if `useTesV1` is absent or false.

Example:
* Customer configures `useTesV1 == true` -> TES switches to V1
* Customer updates `iotRoleAlias` -> TES switches back to V2
* Customer restarts Nucleus -> TES processes `useTesV1` last and uses V1

jbutler: It may be a stop gap solution, but once we ship it, we can't remove it.

I can see this being an issue after a customer migrates from V1 back to V2. From an ops perspective, we need to be able to know what GG is doing based on just the config. Should be simple to fix, no?

MikeDombo: Updated

jbutler (approved):

src/test/java/com/aws/greengrass/tes/CredentialRequestHandlerTest.java
```diff
@@ -78,8 +78,8 @@
     // Set in the far future so it doesn't expire
     private static final String EXPIRATION = "2030-08-19T07:35:15Z";
     private static final String RESPONSE_STR =
-            "{\"credentials\":" + "{\"accessKeyId\":\"" + ACCESS_KEY_ID + "\"," + "\"secretAccessKey\":\""
-                    + SECRET_ACCESS_KEY + "\"," + "\"sessionToken\":\"" + SESSION_TOKEN + "\"," + "\"expiration\":\""
+            "{\"credentials\":" + "{\"AccessKeyId\":\"" + ACCESS_KEY_ID + "\"," + "\"secretAccessKey\":\""
+                    + SECRET_ACCESS_KEY + "\"," + "\"sesSionToken\":\"" + SESSION_TOKEN + "\"," + "\"expiration\":\""
```
shaguptashaikh, line 82: why?

MikeDombo: Because V1 TES uses different capitalization, I wanted to ensure that we don't care how it is capitalized.

shaguptashaikh (approved):
LGTM, I'm just a little unsure if 'useTesV1' is very user friendly, don't think customers would be pretty clear on what `TES` means

MikeDombo:
> LGTM, I'm just a little unsure if 'useTesV1' is very user friendly, don't think customers would be pretty clear on what `TES` means

Any alternative suggestions?

shaguptashaikh:
> > LGTM, I'm just a little unsure if 'useTesV1' is very user friendly, don't think customers would be pretty clear on what `TES` means
> 
> Any alternative suggestions?

Nothing great, maybe something like `useV1CredentialEndpoint` or `useV1TokenExchangeService`

PR #1170, opened 2022-03-08
feat: support configuring procd system service for OpenWRT

weichihl:
**Issue #, if available:**

**Description of changes:**
Support configuring `procd` system service for OpenWRT.

OpenWRT does not use `systemd` to manage its system services. Instead, it uses `procd`. To support that, we need to add another implementation of `SystemServiceUtils`.

UT passed.
It's also tested by configuring Greengrass as a system service on a device with OpenWRT OS. Checked if Nucleus is still running after reboot or updating configuration.

**Checklist:**

### DIFF
--- a/assembly.xml
+++ b/assembly.xml
@@ -20,6 +20,7 @@
                 <include>loader.cmd</include>
                 <include>loader</include>
                 <include>greengrass.service.template</include>
+                <include>greengrass.service.procd.template</include>
                 <include>greengrass.xml.template</include>
                 <include>greengrass.exe</include>
             </includes>
--- a/src/main/java/com/aws/greengrass/util/orchestration/SystemServiceUtilsFactory.java
+++ b/src/main/java/com/aws/greengrass/util/orchestration/SystemServiceUtilsFactory.java
@@ -12,6 +12,7 @@ import com.aws.greengrass.logging.impl.LogManager;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import lombok.AllArgsConstructor;
 
+import java.io.File;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Paths;
@@ -41,8 +42,17 @@ public class SystemServiceUtilsFactory {
                 return context.get(SystemdUtils.class);
             }
         } catch (IOException e) {
-            logger.atError().log("Unable to determine init process type");
+            logger.atDebug().log("Unable to determine init process type");
         }
+
+        if (new File("/sbin/procd").isFile()) {
+            logger.atDebug().log("Detected procd on the device");
+            return context.get(ProcdUtils.class);
+        } else {
+            logger.atDebug().log("No procd is detected");
+        }
+
+        logger.atError().log("Unable to determine system service type");
         return context.get(InitUtils.class);
     }
 }
--- a/scripts/greengrass.service.procd.template
+++ b/scripts/greengrass.service.procd.template
@@ -1,0 +1,28 @@
+#!/bin/sh /etc/rc.common
+
+USE_PROCD=1
+
+START=99
+STOP=01
+
+start_service() {
+    procd_open_instance
+
+    # Retry count would be reset to 0 after 3600 seconds.
+    # Retry after 10 seconds.
+    # Always retry.
+    procd_set_param respawn 3600 10 0
+
+    procd_set_param pidfile REPLACE_WITH_GG_LOADER_PID_FILE
+
+    # Redirect stdout to system log.
+    procd_set_param stdout 1
+
+    # Redirect stderr to system log.
+    procd_set_param stderr 1
+
+    # We pass environment variables of Java here because Java may not be installed through opkg. If it's manually installed, we need to provide these variables.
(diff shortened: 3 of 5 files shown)


assembly.xml
```diff
@@ -20,6 +20,7 @@
                 <include>loader.cmd</include>
                 <include>loader</include>
                 <include>greengrass.service.template</include>
+                <include>greengrass.service.procd.template</include>
```
MikeDombo, line 23: https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/b6e384085e4428d97724c370f7a419057967c6e6/src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java#L402

Include here too

src/main/java/com/aws/greengrass/util/orchestration/SystemServiceUtilsFactory.java
```diff
@@ -41,8 +43,41 @@ public synchronized SystemServiceUtils getInstance() {
                 return context.get(SystemdUtils.class);
             }
         } catch (IOException e) {
-            logger.atError().log("Unable to determine init process type");
+            logger.atDebug().log("No systemd is detected");
```
MikeDombo, line 46: do not change this, detect which init system to use properly

src/main/java/com/aws/greengrass/util/orchestration/SystemServiceUtilsFactory.java
```diff
@@ -41,8 +43,41 @@ public synchronized SystemServiceUtils getInstance() {
...
         }
+
+        try {
+            String systemServiceProcessName = getProcessName(1);
```
MikeDombo, line 50: can you detect this another way, similar to what we do with systemd can we infer that procd is in use without actually trying to read pid 1?

src/main/java/com/aws/greengrass/util/orchestration/SystemServiceUtilsFactory.java
weichihl: The `procd` is a script instead of an executable file located in the file system. So it can't be detected in the same way as systemd. Besides, the solution has a bug. How can we tell which system management is used if there is a `systemd` in `/sbin/init` in some specific OpenWRT port? From my point of view, we should read pid 1 in a Linux-like platform to detect which system management is because that process is actually used.

src/main/java/com/aws/greengrass/util/orchestration/SystemServiceUtilsFactory.java
weichihl: The `procd` is a script instead of an executable file located in the file system. So it can't be detected in the same way as systemd. Besides, the solution has a bug. How can we tell which system management is used if there is a `systemd` in `/sbin/init` in some specific OpenWRT port? From my point of view, we should read pid 1 in a Linux-like platform to detect which system management is because that process is actually used.

src/main/java/com/aws/greengrass/util/orchestration/SystemServiceUtilsFactory.java
weichihl: Will do

src/main/java/com/aws/greengrass/util/orchestration/SystemServiceUtilsFactory.java
MikeDombo: You can check that it exists or not, right? Not that it is executable

weichihl: Consider this situation. If an OpenWRT port has a `systemd` in `/sbin/init`. We can't say it uses `systemd` to manage its system services. On the other way, if there is a `procd` script in some place of init process, we can't say it uses `procd` to manage system services because it might actually use `systemd`. The safest way is to check pid 1 to know which one is actually being used.

MikeDombo: I don't think that's a real concern unless you can point to an example of that in reality. Why would a system (and especially openwrt) have both systemd and procd?

weichihl: Indeed, it totally makes sense to have only one system service management. I'll check how to detect procd without using read pid 1 status.

shaguptashaikh (approved):

src/main/java/com/aws/greengrass/util/orchestration/SystemServiceUtilsFactory.java
```diff
@@ -43,6 +44,15 @@ public synchronized SystemServiceUtils getInstance() {
         } catch (IOException e) {
             logger.atError().log("Unable to determine init process type");
```
MikeDombo, line 45: My comment was maybe a bit wrong, this shouldn't be an error yet because we haven't identified that the system _isn't_ procd, so we still may be able to determine the init type.

MikeDombo (approved):

rbattle (approved):

PR #1171, opened 2022-03-08
fix(deployment): parse non-target group component requirements as requirements, not version

MikeDombo:
**Issue #, if available:**

**Description of changes:**
A local deployment allows you to specify version requirements for root components such as ">0.0.0". This works fine, but subsequent non-local deployments failed in `getNonTargetGroupToRootPackagesMap` due to trying to parse `>0.0.0` as a version instead of a version requirement. We ultimately built a requirement from the version in DependencyResolver, so now we just properly construct the version requirement earlier on.

**Checklist:**

### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/deployment/DeploymentTaskIntegrationTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/deployment/DeploymentTaskIntegrationTest.java
@@ -818,7 +818,7 @@ class DeploymentTaskIntegrationTest extends BaseITCase {
         groupToRootComponentsTopics.lookupTopics("CustomerApp")
                 .replaceAndWait(ImmutableMap.of(GROUP_TO_ROOT_COMPONENTS_VERSION_KEY, "1.0.0"));
         groupToRootComponentsTopics.lookupTopics("YellowSignal")
-                .replaceAndWait(ImmutableMap.of(GROUP_TO_ROOT_COMPONENTS_VERSION_KEY, "1.0.0"));
+                .replaceAndWait(ImmutableMap.of(GROUP_TO_ROOT_COMPONENTS_VERSION_KEY, ">=1.0.0"));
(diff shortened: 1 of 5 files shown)


src/integrationtests/java/com/aws/greengrass/integrationtests/deployment/DeploymentTaskIntegrationTest.java
```diff
@@ -818,7 +818,7 @@ void GIVEN_services_running_WHEN_service_added_and_deleted_THEN_add_remove_servi
         groupToRootComponentsTopics.lookupTopics("CustomerApp")
                 .replaceAndWait(ImmutableMap.of(GROUP_TO_ROOT_COMPONENTS_VERSION_KEY, "1.0.0"));
         groupToRootComponentsTopics.lookupTopics("YellowSignal")
-                .replaceAndWait(ImmutableMap.of(GROUP_TO_ROOT_COMPONENTS_VERSION_KEY, "1.0.0"));
+                .replaceAndWait(ImmutableMap.of(GROUP_TO_ROOT_COMPONENTS_VERSION_KEY, ">=1.0.0"));
```
J-C-L, line 821: Is this replacing a test? Looks like we'll now check that it works with a version range, but no longer check that it works with an exact version. Should we add a check to have both?

MikeDombo: No, there are plenty of tests to cover the previous situation, including the line above this

shaguptashaikh (approved):

src/integrationtests/java/com/aws/greengrass/integrationtests/deployment/DeploymentTaskIntegrationTest.java
J-C-L: Sounds good.

J-C-L (approved):

PR #1172, opened 2022-03-10
feat: change wildcard behaviour in Authz

vabhasin:
**Issue #, if available:**

**Description of changes:**
Change behavior of wildcard evaluation inside Authorization. This was previously implemented where we evaluated both MQTT wildcards (`+,#`) and glob wildcard (`*`). https://github.com/aws-greengrass/aws-greengrass-nucleus/pull/1152

Changes: 
- MQTT wildcards will not be evaluated and will be treated as string literals.
- Glob wildcard will not respect MQTT levels now and match across levels. (eg - `ab/*/c` will match `ab/1/2/c` and `ab/12/c`)

The change was made so that wildcards in different policies are consistent with each other and particularly match IoT Core policy's behavior.

Unit tests
Design change Discussion: https://quip-amazon.com/9tvKA4JH4Tag/Authorization-Edge-Workflow-CX-Review

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/authorization/WildcardTrie.java
+++ b/src/main/java/com/aws/greengrass/authorization/WildcardTrie.java
@@ -5,7 +5,6 @@
 
 package com.aws.greengrass.authorization;
 
-import com.aws.greengrass.authorization.AuthorizationHandler.ResourceLookupPolicy;
 import com.aws.greengrass.util.DefaultConcurrentHashMap;
 
 import java.util.HashMap;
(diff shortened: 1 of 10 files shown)


src/main/java/com/aws/greengrass/authorization/WildcardTrie.java
```diff
@@ -15,70 +14,28 @@
...
-            initial.isMQTTWildcard = true;
-            return;
-        }
-        if (subject.startsWith(MQTT_SINGLELEVEL_WILDCARD + MQTT_LEVEL_SEPARATOR)) {
-            WildcardTrie initial = this.children.get(MQTT_SINGLELEVEL_WILDCARD);
-            initial.isMQTTWildcard = true;
-            initial.add(subject.substring(1), true);
-        }
-
         add(subject, true);
```
MikeDombo, line 39: how is the first entry a terminal?

src/test/java/com/aws/greengrass/authorization/AuthorizationHandlerTest.java
```diff
@@ -117,13 +115,13 @@ private AuthorizationPolicy getStarResourceAuthZPolicy() {
...
 
-    private AuthorizationPolicy getWildcardResourceAuthZPolicy() {
+    private AuthorizationPolicy getStarWildcardResourceAuthZPolicy() {
         return AuthorizationPolicy.builder()
                 .policyId("Id1")
                 .policyDescription("Test policy")
                 .principals(new HashSet<>(Arrays.asList("compA")))
                 .operations(new HashSet<>(Arrays.asList("OpA")))
-                .resources(new HashSet<>(Arrays.asList("abc*/+/*xyz*4/#")))
+                .resources(new HashSet<>(Arrays.asList("abc*/xyz*4/*")))
```
MikeDombo, line 124: keep some with + and #
Customers can still use those in policies, they just have a literal meaning

src/test/java/com/aws/greengrass/authorization/WildcardTrieTest.java
```diff
@@ -16,282 +16,72 @@
...
-        assertTrue(rt2.matchesMQTT("a/12/34/d"));
-        assertFalse(rt2.matchesMQTT("a/2/3/4/d"));
-
-        assertTrue(rt2.matchesStandard("a/12/34/d"));
-        assertFalse(rt2.matchesStandard("a/2/3/4/d"));
-
-        rt2.add("**/abc");
-        assertTrue(rt2.matchesMQTT("78/abc"));
-        assertFalse(rt2.matchesMQTT("7/8/abc"));
+        rt1.add("*qwe*90*");
```
MikeDombo, line 51: glob matches include matching 0 characters. Do we have a test for that? 
Such as "a*" should match "ab" and "a"

vabhasin: Yes

MikeDombo (dismissed):

src/main/java/com/aws/greengrass/authorization/WildcardTrie.java
vabhasin: It starts with assuming terminal, unless it encounters a `*` where in it adds a node with all the previous characters and mark it terminal false

src/test/java/com/aws/greengrass/authorization/AuthorizationHandlerTest.java
vabhasin: You're right I should just keep them as is

src/main/java/com/aws/greengrass/authorization/WildcardTrie.java
MikeDombo: Is it ever false, can the parameter be removed?

vabhasin: Yes, it is needed and would be false.

youtuyy:
Should this pr be closed as replaced by #1173 ?

vabhasin:
> 

Yes, we don't need this

PR #1173, opened 2022-03-14
fix: change glob wildcard behavior in AuthZ

vabhasin:
**Issue #, if available:**

**Description of changes:**
Change behavior of wildcard evaluation inside Authorization. This was previously implemented where we evaluated both MQTT wildcards (`+,#`) and glob wildcard (`*`) inside AuthZ policy for both MQTT and PubSub IPC. https://github.com/aws-greengrass/aws-greengrass-nucleus/pull/1152

We want only glob wildcard in PubSub IPC, MQTT IPC already had MQTT wildcards, so we will keep it as is along with addition of glob wildcard into it

Changes: 
- PubSub IPC AuthZ: MQTT wildcards will not be evaluated and will be treated as string literals. Glob wildcard will be evaluated
- MQTT IPC AuthZ: MQTT wildcards will be evaluated as wildcards. Glob wildcard will be evaluated
- Glob wildcard will not respect MQTT levels now and match across levels. (eg - `ab/*/c` will match `ab/1/2/c` and `ab/12/c`)

The change was made so that wildcards in different policies are consistent with each other and particularly match IoT Core policy's behavior.

Unit tests
Design change Discussion: https://quip-amazon.com/9tvKA4JH4Tag/Authorization-Edge-Workflow-CX-Review

**Checklist:**


src/main/java/com/aws/greengrass/authorization/WildcardTrie.java
```diff
@@ -18,7 +18,7 @@
  * - isTerminalLevel: If the node is the last level before a valid use "#" wildcard (eg: "abc/123/#", 123/ would be the
  *   terminalLevel).
  * - isWildcard: If current Node is a valid glob wildcard (*)
- * - isWildcard: If current Node is a valid MQTT wildcard (#, +)
+ * - isMQTTWildcard: If current Node is a valid MQTT wildcard (#, +)
```
MikeDombo, line 21: why do we have MQTT wildcards here? I thought we were only globbing?

MikeDombo (dismissed):

jbutler:
Thank you for including a good PR description - it gives good context for the change.

src/integrationtests/resources/com/aws/greengrass/integrationtests/ipc/pubsub.yaml
```diff
@@ -133,4 +133,4 @@ services:
               - 'aws.greengrass#SubscribeToTopic'
               - 'aws.greengrass#PublishToTopic'
             resources:
-              - /to*/#
```
youtuyy, line 136: Will you also add some more integ tests using the new authz wildcard? e.g. for mqtt with different policies than https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/main/src/integrationtests/resources/com/aws/greengrass/integrationtests/ipc/mqttproxy.yaml

vabhasin: Yes, good point. I should add at least one case using each wildcard in mqtt

MikeDombo (approved):

youtuyy (dismissed):

src/test/java/com/aws/greengrass/authorization/WildcardTrieTest.java
```diff
@@ -16,51 +16,111 @@
...
+        assertTrue(rt.matchesMQTT("nowildcard"));
+        assertFalse(rt.matchesMQTT("topic"));
+
+        // Test wildcards in middle
+        rt.add("abc*xy*z");
+        assertTrue(rt.matchesStandard("abc123xyabc!@/<>#$%^&*()_+-=z" ));
+        assertTrue(rt.matchesStandard("abcxyz"));
+        assertTrue(rt.matchesStandard("abcxy/ 7895z"));
+        assertTrue(rt.matchesStandard("abc123xy90zABCz"));
+        assertTrue(rt.matchesStandard("abc123xy90zABCxyABCz"));
```
jbutler, line 34: Good test cases. Also add one that has a partial match after the first glob. E.g. `abc123x123xyz`

src/test/java/com/aws/greengrass/authorization/WildcardTrieTest.java
```diff
@@ -16,51 +16,111 @@
...
+        rt.add("nowildcard");
+        assertTrue(rt.matchesStandard("nowildcard"));
+        assertFalse(rt.matchesStandard("topic"));
+
+        assertTrue(rt.matchesMQTT("nowildcard"));
+        assertFalse(rt.matchesMQTT("topic"));
+
+        // Test wildcards in middle
+        rt.add("abc*xy*z");
+        assertTrue(rt.matchesStandard("abc123xyabc!@/<>#$%^&*()_+-=z" ));
```
jbutler, line 30: This input contains *. What would the resource string look like to allow that? Does IoT Core allow it? Add test case for it

src/test/java/com/aws/greengrass/authorization/WildcardTrieTest.java
```diff
@@ -16,51 +16,111 @@
...
+
+        assertTrue(rt.matchesMQTT("abc123xyabc!@/<>#$%^&*()_+-=z" ));
+        assertTrue(rt.matchesMQTT("abcxyz"));
+        assertTrue(rt.matchesMQTT("abcxy/ 7895z"));
+        assertTrue(rt.matchesMQTT("abc123xy90zABCz"));
+        assertTrue(rt.matchesMQTT("abc123xy90zABCxyABCz"));
+        assertFalse(rt.matchesMQTT("ab789xyz123"));
+        assertFalse(rt.matchesMQTT("abc789xy56z123"));
+        assertFalse(rt.matchesMQTT("abc123xy90zABCzz0"));
+        assertFalse(rt.matchesMQTT("abc123yx90z"));
```
jbutler, line 49: Is there a reason you've duplicated these assertions?

src/test/java/com/aws/greengrass/authorization/WildcardTrieTest.java
```diff
@@ -237,25 +297,31 @@ void testMultipleWildcardsMatching() {
...
-        assertFalse(rt.matchesMQTT("xyzzzz/89/fan2/bedroom/light/5"));
+        assertTrue(rt.matchesMQTT("xyzzzz/89/fan2/bedroom/light/5"));
 
         assertFalse(rt.matchesStandard("xyzzzz/89/bedroom/light/5"));
         assertFalse(rt.matchesStandard("xyzzzz/89/fan2/bedroom/light/5"));
         assertTrue(rt.matchesStandard("xyzzzz/+/bedroom/#"));
 
         rt.add("12/*#");
         assertTrue(rt.matchesMQTT("12/345#"));
+        assertTrue(rt.matchesMQTT("12/34/5#"));
```
jbutler, line 308: These aren't valid topic filters according to the spec. We probably shouldn't even be accepting them at all

http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718106

vabhasin: We don't check that here as it's a general resource, but that check is implemented in Pubsub and MQTT IPC code when operation is called for that particular resource.

src/test/java/com/aws/greengrass/authorization/WildcardTrieTest.java
vabhasin: The resource one above this line will allow it `rt.add("abc*xy*z")`.

youtuyy: I remember secret arn contains some special characters like _  Those are covered in this case right?

youtuyy (approved):

src/test/java/com/aws/greengrass/authorization/WildcardTrieTest.java
jbutler: Had a chat offline about this.

In general I am very uncomfortable with this code allowing actions on invalid resources. This code is aware of MQTT semantics. `matchesMQTT` should NEVER return true on invalid MQTT.

That said, fixing this will require some larger refactoring which can be done later. This method is being supported for legacy reasons, and we want to ensure that nothing else uses the MQTT style policy moving forward. Rather than give callers the option to specify the policy type, we are instead going to force this to happen during trie creation. Only MQTT proxy operations will use this policy type, and it will be hard coded inside the auth module itself rather than being provided by the caller as part of the `isAuthorized` call.

src/test/java/com/aws/greengrass/authorization/WildcardTrieTest.java
jbutler: Documenting another offline discussion with Vaibhav: IAM would escape this with `${*}`. There's also the `?` wildcard for matching a single character

https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_variables.html#policy-vars-infotouse

I'd also like to see that same behavior here. We can do this incrementally, but at a minimum we should require `?` and `$` to be escaped

Note: this would be backward incompatible. Need to think about this would impact customers who might be subscribing to `$aws` topics

PR #1174, opened 2022-03-14
fix(tes): use override value to ensure the value is always updated

MikeDombo:
**Issue #, if available:**

**Description of changes:**
Adds `overrideValue` to the API of `Topic` which will ensure that the topic has its value changed to the provided value by using the existing node's modtime. When modtimes are equal, the config will accept the incoming change.

This change is necessary in TES because when the system clock is in the future compared to the real world, when the system clock is unskewed TES may try to update the config value, but the value won't stick due to the timestamps. This change ensures that for TES where the timestamp doesn't actually matter, that the latest value always wins. We are going to work on bigger options for detecting and addressing bad clocks.

Added unit test.

**Checklist:**


shaguptashaikh (approved):

src/main/java/com/aws/greengrass/tes/TokenExchangeService.java
```diff
@@ -94,6 +95,8 @@ protected void startup() {
             logger.atInfo().log("Started server at port {}", server.getServerPort());
             // Get port from the server, in case no port was specified and server started on a random port
             setEnvVariablesForDependencies(server.getServerPort());
+            // Store the actual port being used in the config so that the CLI can show the value for debugging
```
tilo-chen, line 98: Huh so the actual port could be different than the configured PORT_TOPIC?

tilo-chen (approved):

src/main/java/com/aws/greengrass/tes/TokenExchangeService.java
MikeDombo: port topic contains the user's requested port (0 by default). Since 0 is the default, the port topic doesn't contain the real port since the OS will choose a random open port.

PR #1175, opened 2022-03-15
feat: batched subscriber

jcosentino11:
**Issue #, if available:**

**Description of changes:**

Adds a `BatchedSubscriber` utility that allows callers to run a callback *after* a batch of changes have fired, rather than executing potentially expensive work per each change.

Thanks @MikeDombo for suggesting this feature

One use case is in mqtt bridge, where we wish to restart the bridge if a config change is detected. If multiple config changes happen in quick succession, we only want to trigger the restart once.  (development effort is in progress, and this change is intended to be used in mqtt bridge)

Unit tests

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/util/BatchedSubscriber.java
+++ b/src/main/java/com/aws/greengrass/util/BatchedSubscriber.java
@@ -11,39 +11,51 @@ import com.aws.greengrass.config.Subscriber;
 import com.aws.greengrass.config.Topic;
 import com.aws.greengrass.config.Topics;
 import com.aws.greengrass.config.WhatHappened;
+import lombok.NonNull;
 
-import java.util.Objects;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.BiPredicate;
 import java.util.function.Consumer;
 
 /**
- * Utility class that allows a subscription to only fire
- * on the last change in a batch.
+ * {@link BatchedSubscriber} is a subscriber that fires once for a <i>batch</i> of changes
+ * (and on subscription initialization).
  *
- * <p>This is useful in scenarios such as configuration changes
- * where you may only want to perform expensive work once within
- * a short span.
+ * <br><br><p>A <i>batch</i> is defined as all the elements in a {@link Topic} or {@link Topics}' publish queue,
+ * with the last <i>batch</i> element being the most recent topic change.
+ *
+ * <br><br><p>By default, commonly ignored changes, like {@link WhatHappened#timestampUpdated} and
+ * {@link WhatHappened#interiorAdded}, will NOT be added to a <i>batch</i>
+ * (see {@link BatchedSubscriber#BASE_EXCLUSION}).
+ *
+ * <br><br><p>To be precise, a {@link BatchedSubscriber} will trigger its {@link BatchedSubscriber#callback}
+ * after the following events:
+ * <ul>
+ *     <li>when {@link WhatHappened#initialized} is fired on initial subscription</li>
+ *     <li>when the last <i>batch</i> element is popped from the topic's publish queue</li>
+ * </ul>
  */
 public final class BatchedSubscriber implements ChildChanged, Subscriber {
 
-    private static final BiPredicate<WhatHappened, Node> BASE_EXCLUSION = (what, child) ->
+    public static final BiPredicate<WhatHappened, Node> BASE_EXCLUSION = (what, child) ->
             what == WhatHappened.timestampUpdated || what == WhatHappened.interiorAdded;
 
-    private final Node node;
     private final AtomicInteger numRequestedChanges = new AtomicInteger();
 
+    private final Node node;
     private final BiPredicate<WhatHappened, Node> exclusions;
     private final Consumer<WhatHappened> callback;
 
     /**
      * Constructs a new BatchedSubscriber.
      *
+     * <p>Defaults to using {@link BatchedSubscriber#BASE_EXCLUSION} for excluding changes from a <i>batch</i>.
+     *
      * @param topic    topic to subscribe to
-     * @param callback action to perform after a batch of changes and on initialization
+     * @param callback action to perform after a <i>batch</i> of changes and on initialization
      */
     public BatchedSubscriber(Topic topic, Consumer<WhatHappened> callback) {
-        this(topic, null, callback);
+        this(topic, BASE_EXCLUSION, callback);
     }
 
     /**
@@ -51,9 +63,10 @@ public final class BatchedSubscriber implements ChildChanged, Subscriber {
      *
      * @param topic      topic to subscribe to
      * @param exclusions predicate for ignoring a subset topic changes
-     * @param callback   action to perform after a batch of changes and on initialization
+     * @param callback   action to perform after a <i>batch</i> of changes and on initialization
      */
-    public BatchedSubscriber(Topic topic, BiPredicate<WhatHappened, Node> exclusions,
+    public BatchedSubscriber(Topic topic,
+                             BiPredicate<WhatHappened, Node> exclusions,
                              Consumer<WhatHappened> callback) {
         this((Node) topic, exclusions, callback);
     }
@@ -61,11 +74,13 @@ public final class BatchedSubscriber implements ChildChanged, Subscriber {
     /**
      * Constructs a new BatchedSubscriber.
      *
+     * <p>Defaults to using {@link BatchedSubscriber#BASE_EXCLUSION} for excluding changes from a <i>batch</i>.
+     *
      * @param topics   topics to subscribe to
-     * @param callback action to perform after a batch of changes and on initialization
+     * @param callback action to perform after a <i>batch</i> of changes and on initialization
      */
     public BatchedSubscriber(Topics topics, Consumer<WhatHappened> callback) {
-        this(topics, null, callback);
+        this(topics, BASE_EXCLUSION, callback);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/main/java/com/aws/greengrass/util/BatchedSubscriber.java
```diff
@@ -0,0 +1,148 @@
...
+     * @param onInitialization callback to perform after the subscription has been added
+     * @param excludeFilter    custom filter for ignoring changes
+     * @return batched subscriber
+     */
+    public static BatchedSubscriber subscribe(Topic topic,
+                                              Runnable afterBatch, Runnable onInitialization,
+                                              BiPredicate<WhatHappened, Topic> excludeFilter) {
+        AtomicInteger numRequestedChanges = new AtomicInteger();
+        AtomicInteger totalCalls = new AtomicInteger();
+        Subscriber subscriber = (what, child) -> {
```
MikeDombo, line 69: instead of creating this subscriber and atomic integers inline, store them as fields of this class. Have BatchedSubscriber implement the ChildChanged interface so you can then just do `topic.subscribe(this)`, `topic.remove(this)`

src/main/java/com/aws/greengrass/util/BatchedSubscriber.java
```diff
@@ -0,0 +1,148 @@
...
+     * @param afterBatch       callback to perform after a batch of changes
+     * @param onInitialization callback to perform after the subscription has been added
+     * @param excludeFilter    custom filter for ignoring changes
+     * @return batched subscriber
+     */
+    public static BatchedSubscriber subscribe(Topic topic,
+                                              Runnable afterBatch, Runnable onInitialization,
+                                              BiPredicate<WhatHappened, Topic> excludeFilter) {
+        AtomicInteger numRequestedChanges = new AtomicInteger();
+        AtomicInteger totalCalls = new AtomicInteger();
```
MikeDombo, line 68: unused

src/main/java/com/aws/greengrass/util/BatchedSubscriber.java
```diff
@@ -0,0 +1,148 @@
...
+     *
+     * @param topics           topics
+     * @param afterBatch       callback to perform after a batch of changes
+     * @param onInitialization callback to perform after the subscription has been added
+     * @param excludeFilter    custom filter for ignoring changes
+     * @return batched subscriber
+     */
+    public static BatchedSubscriber subscribe(Topics topics,
+                                              Runnable afterBatch, Runnable onInitialization,
+                                              BiPredicate<WhatHappened, Node> excludeFilter) {
```
MikeDombo, line 120: I'd rather not split out the batch and initialization callbacks. Combine them and pass into the callback either WhatHappened.initialized or WhatHappened.changed accordingly.

Also, rather than a predicate for filtering I'd suggest a simple `WhatHappened[]` of exclusions.

src/test/java/com/aws/greengrass/util/BatchedSubscriberTest.java
```diff
@@ -0,0 +1,175 @@
...
+            assertTrue(testComplete.await(5L, TimeUnit.SECONDS));
+            topics.context.waitForPublishQueueToClear();
+        } finally {
+            s.unsubscribe();
+            topics.context.close();
+        }
+
+        assertEquals(expectedNumChanges, numTimesCalled.get());
+    }
+}
```
MikeDombo, line 175: [nit] New line at end of file

src/main/java/com/aws/greengrass/util/BatchedSubscriber.java
jcosentino11: To provide some context to `excludeFilter`: the use-case I had in mind was to ignore a subset of config changes:
```
// treat mqttTopicMapping property changes differently than rest
var excludeFilter = (what, child) -> child.childOf("mqttTopicMapping");
```
and I threw in `WhatHappened` in the filter so that caller could have maximum flexibility on the subscription.


Thoughts on having a `Predicate<Node>` for filtering in addition to WhatHappened[] exclusions?

MikeDombo: Your current implementation is fine for that yes. Don't add the array

src/main/java/com/aws/greengrass/util/BatchedSubscriber.java
```diff
@@ -0,0 +1,99 @@
...
+     * Callback to perform after a batch of changes fires.
+     */
+    public interface Callback {
+        /**
+         * Perform the subscriber action.
+         *
+         * @param what {@link WhatHappened#initialized} on subscription initialization,
+         *             otherwise a pass-through from the subscription.
+         */
+        void run(WhatHappened what);
```
MikeDombo, line 46: why do you include the `what` here? Is this useful since it will just have the last change?

src/main/java/com/aws/greengrass/util/BatchedSubscriber.java
```diff
@@ -0,0 +1,99 @@
...
+
+    @Override
+    public void childChanged(WhatHappened what, Node child) {
+        onChange(what, child);
+    }
+
+    @Override
+    public void published(WhatHappened what, Topic t) {
+        onChange(what, t);
+    }
```
MikeDombo, line 78: looks like you got rid of the unsubscribe call. Can you add that back please? That's fairly important

src/main/java/com/aws/greengrass/util/BatchedSubscriber.java
```diff
@@ -0,0 +1,99 @@
...
+        void run(WhatHappened what);
+    }
+
+    /**
+     * Constructs a new BatchedSubscriber.
+     *
+     * @param callback action to perform after a batch of changes
+     */
+    public BatchedSubscriber(Callback callback) {
+        this((what, node) -> false, callback);
```
MikeDombo, line 55: set exclusions to null here

src/main/java/com/aws/greengrass/util/BatchedSubscriber.java
```diff
@@ -0,0 +1,99 @@
...
+
+    /**
+     * Constructs a new BatchedSubscriber.
+     *
+     * @param exclusions exclude changes based on what happened
+     * @param callback   action to perform after a batch of changes
+     */
+    public BatchedSubscriber(BiPredicate<WhatHappened, Node> exclusions, Callback callback) {
+        Objects.requireNonNull(exclusions);
+        this.exclusions = (what, node) -> BASE_EXCLUSION.test(what, node) || exclusions.test(what, node);
```
MikeDombo, line 66: if `exclusions` is null, then use the base, otherwise use what the user gave you. They many not want to exclude anything

src/main/java/com/aws/greengrass/util/BatchedSubscriber.java
jcosentino11: Yep, I did try to keep it in originally, but there didn't seem like a clean way to get the Topic/Topics reference to use for unsubscribing, plus the caller of BatchedSubscriber would already have all the info they'd need to unsubscribe themselves.  However, writing this out just now, I suppose I could get the Topic/Topics reference during WhatHappened.initialized event, will try that out

MikeDombo: They actually don't have enough to unsubscribe because you need to give the unsubscribe call a reference to the callable; which is made by BatchedSubscriber, so the user doesn't actually have that.

If you store a `Node` reference that may help you since both Topic and Topics are Node. If needed, you can use `instanceof` and then cast to the appropriate value.

src/main/java/com/aws/greengrass/util/BatchedSubscriber.java
```diff
@@ -0,0 +1,155 @@
...
+    private final Node node;
+    private final AtomicInteger numRequestedChanges = new AtomicInteger();
+
+    private final BiPredicate<WhatHappened, Node> exclusions;
+    private final Callback callback;
+
+    /**
+     * Callback to perform after a batch of changes fires.
+     */
+    public interface Callback {
```
MikeDombo, line 41: this interface exists already, it is a `Consumer<WhatHappened>`, reuse the builtin Java interface since this doesn't add anything

MikeDombo (commented):
One last comment, then this is good by me.

src/main/java/com/aws/greengrass/util/BatchedSubscriber.java
```diff
@@ -0,0 +1,142 @@
...
+ * <p>This is useful in scenarios such as configuration changes
+ * where you may only want to perform expensive work once within
+ * a short span.
+ */
+public final class BatchedSubscriber implements ChildChanged, Subscriber {
+
+    private static final WhatHappened[] DEFAULT_IGNORED_CHANGES = {
+            WhatHappened.timestampUpdated,
+            WhatHappened.interiorAdded,
+            WhatHappened.initialized
```
MikeDombo, line 33: remove initialized. I want it to run on initialization by default. On initialization it should run the onChange callback immediately. Undo the last change.

MikeDombo (dismissed):

MikeDombo:
LGTM!

src/main/java/com/aws/greengrass/util/BatchedSubscriber.java
```diff
@@ -0,0 +1,146 @@
...
+    private final Consumer<WhatHappened> callback;
+
+    /**
+     * Constructs a new BatchedSubscriber.
+     *
+     * @param topic    topic to subscribe to
+     * @param callback action to perform after a batch of changes and on initialization
+     */
+    public BatchedSubscriber(Topic topic, Consumer<WhatHappened> callback) {
+        this(topic, null, callback);
```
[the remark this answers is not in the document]
jcosentino11: Yeah this sounds more flexible, I had a similar approach earlier that missed the mark a bit (https://github.com/aws-greengrass/aws-greengrass-nucleus/pull/1175/commits/1e9fd1c271505fc78174d103fbc96bdfc7bf0879#diff-b0dd07995d2c4bd84e07f0de7a1ed6d35a3e453e86c6677dc4b8764efc62db4eR66).   Will adjust 👍

MikeDombo (approved):

rbattle (approved):

PR #1178, opened 2022-03-23
fix: don't initialize S3 client on startup

rbattle:
**Issue #, if available:**

**Description of changes:**
Don't construct an S3 client until it is requested. Removes S3 client from cache if region changes.

The HTTP client initialization for the S3 client takes ~5s on rpi3.

It is not necessary to pre-construct an S3 client as that just increase the
service initialization time.

Tested load time on rpi3

This changes behavior of the `getS3Client` to also throw a `DeviceConfigurationException` if the config is invalid.
Previously this getter would always return the last valid client, or null if there never was one.

S3Downloader appears to be the only place that uses this factory.

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/util/S3SdkClientFactory.java
+++ b/src/main/java/com/aws/greengrass/util/S3SdkClientFactory.java
@@ -8,7 +8,6 @@ package com.aws.greengrass.util;
 import com.aws.greengrass.deployment.DeviceConfiguration;
 import com.aws.greengrass.deployment.exceptions.DeviceConfigurationException;
 import com.aws.greengrass.tes.LazyCredentialProvider;
-import lombok.Getter;
 import software.amazon.awssdk.regions.Region;
 import software.amazon.awssdk.services.s3.S3Client;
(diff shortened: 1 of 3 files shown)


src/main/java/com/aws/greengrass/util/S3SdkClientFactory.java
```diff
@@ -20,12 +19,13 @@
 /**
  * S3 accessor that uses Token Exchange Service to get credentials for customer's S3.
  */
-@Getter
 public class S3SdkClientFactory {
-    private static final Map<Region, S3Client> clientCache = new ConcurrentHashMap<>();
-    private S3Client s3Client;
+    static final Map<Region, S3Client> clientCache = new ConcurrentHashMap<>();
```
MikeDombo, line 23: why not private anymore?

MikeDombo (approved):

leaf94:
This is fantastic! I'm sure this 5 seconds saved would be much appreciated by drivers in the car who is waiting for Nucleus to start! : )

PR #1179, opened 2022-03-23
fix: bubble up more complete error messages from deployment workflow

J-C-L:
**Issue #, if available:**

**Description of changes:**
We currently only send last error message from deployment workfkow to FSS. This does not give customers enough meaningful information. This change will enable customers to see the chain of error messages instead.

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/deployment/DeploymentService.java
+++ b/src/main/java/com/aws/greengrass/deployment/DeploymentService.java
@@ -312,7 +312,8 @@ public class DeploymentService extends GreengrassService {
                     deploymentDirectoryManager.persistLastSuccessfulDeployment();
                 } else {
                     if (result.getFailureCause() != null) {
-                        statusDetails.put(DEPLOYMENT_FAILURE_CAUSE_KEY, result.getFailureCause().getMessage());
+                        Throwable failureCause = result.getFailureCause();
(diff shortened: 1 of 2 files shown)


src/main/java/com/aws/greengrass/deployment/DeploymentService.java
```diff
@@ -655,7 +656,7 @@ private DefaultDeploymentTask createDefaultNewDeployment(Deployment deployment)
                     .kv("DeploymentType", deployment.getDeploymentType().toString())
                     .log("Invalid document for deployment");
             HashMap<String, String> statusDetails = new HashMap<>();
-            statusDetails.put("error", e.getMessage());
+            statusDetails.put(DEPLOYMENT_FAILURE_CAUSE_KEY, e.getMessage());
```
MikeDombo, line 659: why change this key? This may have unintended consequences.

tilo-chen: FSS only checks this and another key, it won't look at the key called "error": https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/main/src/main/java/com/aws/greengrass/status/FleetStatusService.java#L515-L516

But right we need to make sure nothing else is relying on the "error" key.

kucnavya:
Tracking this PR with additional changes in https://github.com/aws-greengrass/aws-greengrass-nucleus/pull/1211

PR #1187, opened 2022-03-28
perf: static list of service implementors

MikeDombo:
**Issue #, if available:**

**Description of changes:**
Stops scanning the classloader of Nucleus for `@ImplementsService`, rather we have a static list of builtin services. This is useful for performance as scanning the classloader can take a while. It is also useful for static builds (Graal, JamaicaVM) and Android. 

Also falls back to reading the `java.home` system property if we cannot find the Jar path by finding the code location from the classloader. This is to support JamaicaVM.

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
+++ b/src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
@@ -6,12 +6,14 @@
 package com.aws.greengrass.lifecyclemanager;
 
 import com.amazon.aws.iot.greengrass.component.common.DependencyType;
+import com.aws.greengrass.componentmanager.plugins.docker.DockerApplicationManagerService;
 import com.aws.greengrass.config.ConfigurationReader;
 import com.aws.greengrass.config.ConfigurationWriter;
 import com.aws.greengrass.config.Topics;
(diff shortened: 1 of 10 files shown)


src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
```diff
@@ -366,6 +372,18 @@ void initConfigAndTlog() {
         } catch (IOException t) {
             logger.atError().log("Error finding built in service plugins", t);
         }
+
+        for (Class<? extends GreengrassService> cl : Arrays
+                .asList(DockerApplicationManagerService.class, UpdateSystemPolicyService.class,
+                        DeploymentService.class, FleetStatusService.class, TelemetryAgent.class,
+                        TokenExchangeService.class)) {
```
rbattle, line 379: nit: make this a constant like `Class<? extends GreengrassService> BUILTIN_SERVICE = Arrays.asList(....)` so it's more obvious to add to it if we ever add a new service?

rbattle (approved):

rbattle (dismissed):

rbattle (approved):

tilo-chen (approved):

PR #1188, opened 2022-03-29
feat: implement receive mode in pubsub

youtuyy:
**Issue #, if available:**

**Description of changes:**
- Implement the new `ReceiveMode` flag for `SubscribeToTopicRequest` in PubSub agent. 
- Refactor SubscriptionTrie to store the new SubscriptionCallback. 
- Update device SDK snapshot. (which also removes eventTopic from `JsonMessage/BinaryMessage` and adds `MessageContext`)

`ReceiveMode` currently has two modes: 
- `RECEIVE_ALL_MESSAGES`: current behavior, all messages will be received by any subscriber.
- `RECEIVE_MESSAGES_FROM_OTHERS`: messaged will not be sent back to the same component that publishes it.

If ReceiveMode is NOT specified, RECEIVE_MESSAGES_FROM_OTHERS will be used if the subscription topic uses wildcard, and RECEIVE_ALL_MESSAGES will be used otherwise.

The new RECEIVE_MESSAGES_FROM_OTHERS mode can prevent unwanted circular routes.

mvn -U -ntp verify

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/builtin/services/pubsub/PubSubIPCEventStreamAgent.java
+++ b/src/main/java/com/aws/greengrass/builtin/services/pubsub/PubSubIPCEventStreamAgent.java
@@ -21,8 +21,10 @@ import software.amazon.awssdk.aws.greengrass.GeneratedAbstractSubscribeToTopicOp
 import software.amazon.awssdk.aws.greengrass.model.BinaryMessage;
 import software.amazon.awssdk.aws.greengrass.model.InvalidArgumentsError;
 import software.amazon.awssdk.aws.greengrass.model.JsonMessage;
+import software.amazon.awssdk.aws.greengrass.model.MessageContext;
 import software.amazon.awssdk.aws.greengrass.model.PublishToTopicRequest;
 import software.amazon.awssdk.aws.greengrass.model.PublishToTopicResponse;
+import software.amazon.awssdk.aws.greengrass.model.ReceiveMode;
 import software.amazon.awssdk.aws.greengrass.model.SubscribeToTopicRequest;
 import software.amazon.awssdk.aws.greengrass.model.SubscribeToTopicResponse;
 import software.amazon.awssdk.aws.greengrass.model.SubscriptionResponseMessage;
@@ -31,6 +33,7 @@ import software.amazon.awssdk.eventstreamrpc.OperationContinuationHandlerContext
 import software.amazon.awssdk.eventstreamrpc.StreamEventPublisher;
 import software.amazon.awssdk.eventstreamrpc.model.EventStreamJsonMessage;
 
+import java.util.HashSet;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
@@ -76,7 +79,19 @@ public class PubSubIPCEventStreamAgent {
      * @param serviceName name of the service subscribing.
      */
     public void subscribe(String topic, Consumer<PublishEvent> cb, String serviceName) {
-        handleSubscribeToTopicRequest(topic, serviceName, cb);
+        subscribe(topic, cb, serviceName, null);
+    }
+
+    /**
+     * Handle the subscription request from internal plugin services with given receive mode.
+     *
+     * @param topic       topic name.
+     * @param cb          callback to be called for each published message
+     * @param serviceName name of the service subscribing.
+     * @param receiveMode the mode of messages sending back to the originating component.
+     */
+    public void subscribe(String topic, Consumer<PublishEvent> cb, String serviceName, ReceiveMode receiveMode) {
+        handleSubscribeToTopicRequest(topic, serviceName, receiveMode, cb);
     }
 
     /**
@@ -87,9 +102,19 @@ public class PubSubIPCEventStreamAgent {
      * @param serviceName name of the service unsubscribing.
      */
     public void unsubscribe(String topic, Consumer<PublishEvent> cb, String serviceName) {
-        if (listeners.remove(topic, cb)) {
-            log.atDebug().kv(COMPONENT_NAME, serviceName).log("Unsubscribed from topic {}", topic);
-        }
+        unsubscribe(topic, cb, serviceName, null);
+    }
+
+    /**
+     * Unsubscribe from a topic for internal plugin services with given receive mode.
+     *
+     * @param topic       topic name.
+     * @param cb          callback to remove from subscription
+     * @param serviceName name of the service unsubscribing.
+     * @param receiveMode the mode of messages sending back to the originating component.
+     */
+    public void unsubscribe(String topic, Consumer<PublishEvent> cb, String serviceName, ReceiveMode receiveMode) {
+        handleUnsubscribeToTopicRequest(topic, serviceName,receiveMode, cb);
     }
 
     /**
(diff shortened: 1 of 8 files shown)


src/main/java/com/aws/greengrass/builtin/services/pubsub/PubSubIPCEventStreamAgent.java
```diff
@@ -76,7 +79,19 @@ public PublishToTopicOperationHandler getPublishToTopicHandler(OperationContinua
...
+
+    /**
+     * Handle the subscription request from internal plugin services with given receive mode.
+     *
+     * @param topic       topic name.
+     * @param cb          callback to be called for each published message
+     * @param serviceName name of the service subscribing.
+     * @param receiveMode the mode of messages sending back to the originating component.
+     */
+    public void subscribe(String topic, Consumer<PublishEvent> cb, String serviceName, ReceiveMode receiveMode) {
```
MikeDombo, line 93: Add a SubscriptionRequest object to encapsulate these options so that if we need to add more again, we don't need to add more subscribe overloads in the future.

```java
SubscriptionRequest {
String topic;
Consumer<PublishEvent> cb;
String serviceName;
ReceiveMode receiveMode;
}
```

src/main/java/com/aws/greengrass/builtin/services/pubsub/SubscriptionCallback.java
```diff
@@ -0,0 +1,22 @@
...
+import lombok.Builder;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+import software.amazon.awssdk.aws.greengrass.model.ReceiveMode;
+
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+@Builder
+public class SubscriptionCallback {
```
MikeDombo, line 18: I think this should use `@Value` instead of all those others.

src/main/java/com/aws/greengrass/builtin/services/pubsub/SubscriptionTrie.java
```diff
@@ -20,13 +20,13 @@
     private static final String MULTI_LEVEL_WILDCARD = "#";
 
     private final Map<String, SubscriptionTrie> children = new ConcurrentHashMap<>();
-    private final Set<Object> callbacks;
+    private final Set<SubscriptionCallback> subscriptionCallbacks;
```
MikeDombo, line 23: just make this a generic class `Set<K>`, `public class <K> SubscriptionTrie<K>`

src/main/java/com/aws/greengrass/builtin/services/pubsub/PubSubIPCEventStreamAgent.java
```diff
@@ -115,18 +140,31 @@ private PublishToTopicResponse handlePublishToTopicRequest(String topic, String
...
             log.atDebug().kv(COMPONENT_NAME, serviceName).log("No one subscribed to topic {}. Returning.", topic);
             // Still technically successful, just no one was subscribed
             return new PublishToTopicResponse();
         }
+        Set<Object> cbs = new HashSet<>();
+        contexts.forEach(context -> {
+            // With RECEIVE_MESSAGES_FROM_OTHERS mode, message will not be sent back to its source component.
+            if (serviceName.equals(context.getSourceComponent()) && ReceiveMode.RECEIVE_MESSAGES_FROM_OTHERS
+                    .equals(context.getReceiveMode())) {
+                log.atDebug().kv(COMPONENT_NAME, serviceName)
```
MikeDombo, line 154: trace I think

src/test/java/com/aws/greengrass/builtin/services/pubsub/PubSubIPCEventStreamAgentTest.java
```diff
@@ -421,23 +536,60 @@ void GIVEN_subscribed_consumer_WHEN_invalid_topic_THEN_throws() {
...
+        // Default is RECEIVE_MESSAGES_FROM_OTHERS if mode is not set for wildcard topic
         pubSubIPCEventStreamAgent.subscribe(TEST_WILDCARD_TOPIC, consumer, TEST_SERVICE);
 
         assertEquals(1, pubSubIPCEventStreamAgent.getListeners().size());
         assertTrue(pubSubIPCEventStreamAgent.getListeners().containsKey(TEST_WILDCARD_TOPIC));
         assertEquals(1, pubSubIPCEventStreamAgent.getListeners().get(TEST_WILDCARD_TOPIC).size());
 
         pubSubIPCEventStreamAgent.publish("Test/A/Topic/B/C", "ABCDEF".getBytes(), TEST_SERVICE);
-        assertTrue(countDownLatch.await(10, TimeUnit.SECONDS));
+        assertFalse(countDownLatch.await(10, TimeUnit.SECONDS));
```
MikeDombo, line 551: you sure about this change?

src/main/java/com/aws/greengrass/builtin/services/pubsub/PubSubIPCEventStreamAgent.java
youtuyy: yeah that's better

src/test/java/com/aws/greengrass/builtin/services/pubsub/PubSubIPCEventStreamAgentTest.java
youtuyy: yeah it's the case where internal plugin subscribe() with a wild carded topic. For existing usage of subscribe(), there is no receiveMode set, so its receiveMode will be set to RECEIVE_MESSAGES_FROM_OTHERS for a wildcarded topic and the message will not be published

src/main/java/com/aws/greengrass/builtin/services/pubsub/SubscriptionCallback.java
youtuyy: right. I was thinking if any option would need to become configurable but actually not. Keeping only `@Value` and `@Builder`

src/main/java/com/aws/greengrass/builtin/services/pubsub/PubSubIPCEventStreamAgent.java
youtuyy: Added to both subscribe and unsubscribe

src/main/java/com/aws/greengrass/builtin/services/pubsub/SubscribeRequest.java
```diff
@@ -0,0 +1,19 @@
...
+import lombok.Value;
+import software.amazon.awssdk.aws.greengrass.model.ReceiveMode;
+
+@Builder
+@Value
+public class SubscribeRequest {
+    String topic;
+    Object callback;
+    String serviceName;
+    ReceiveMode receiveMode;
```
MikeDombo, line 18: do you need to exclude receiveMode from `equals` and `hashcode`? Ensure that unsubscribe actually works when using the old API (not providing a SubscribeRequest object, but instead just proving a callback, topic, etc)

youtuyy: Currently it is implemented that the old API unsubscribe() has the same steps as the old subscribe() so their SubscribeRequest will be the same: it is treated as having no receive mode and it evaluates the mode based on if it's using wildcard, and generates a `SubscriptionCallback` object that also includes the ReceiveMode. The trie will remove that subscription given the topic and `SubscriptionCallback` so ReceiveMode will also be compared when removing. I've also thought about if we should just ignore the ReceiveMode when unsubscribe but I thought we would want to check it..
`GIVEN_subscribed_consumer_no_receive_mode_WHEN_publish_binary_message_THEN_publishes_And_THEN_unsubscribes` and `GIVEN_subscribed_consumer_to_wildcard_and_no_receive_mode_WHEN_publish_to_subtopic_THEN_not_publishes_And_THEN_unsubscribes` tests the old subscribe() and unsubscribe().

nikkhilmuthye (approved):

src/test/java/com/aws/greengrass/builtin/services/pubsub/PubSubIPCEventStreamAgentTest.java
rbattle: looks fine - the next test case `GIVEN_subscribed_consumer_to_wildcard_and_receive_all_mode_WHEN_publish_to_subtopic_THEN_publishes_And_THEN_unsubscribes` is what this test was previously doing

rbattle (approved):
lgtm

MikeDombo (approved):

PR #1193, opened 2022-03-31
feat: add escaping to AuthZ resources

vabhasin:
**Issue #, if available:**

**Description of changes:**
This change adds escaping of special characters (`'*', '$', '?'`) similar to what IoT Core policy(See under [title](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_variables.html#policy-vars-infotouse) "Special Characters") follows.
To escape a special character it should be inside the following  escape sequence `${}` (eg: `${*}`). 
Any occurrence of `'${'` is only valid if it holds a single valid special character (`'*', '$', '?'`) inside it and ends with `}`. (eg: `${*}` is valid, `${c}` is invalid, `${c` is invalid, `${*bc}` is invalid)

- `*` when escaped will not behave as a wild card while matching but as a normal string literal
- `$` when escaped will not behave as escaping character but as normal string literal
- Right now, `?` is not supported as a single Character wildcard in AuthZ so it can only be used when escaped.
- Any other invalid occurrence of `'${'` will be treated as invalid resource and rejected with a failure.

**Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/authorization/AuthorizationHandler.java
+++ b/src/main/java/com/aws/greengrass/authorization/AuthorizationHandler.java
@@ -324,7 +324,8 @@ public class AuthorizationHandler  {
         // now start adding the policies as permissions
         for (AuthorizationPolicy policy : policies) {
             try {
-                addPermission(componentName, policy.getPrincipals(), policy.getOperations(), policy.getResources());
+                addPermission(componentName, policy.getPolicyId(), policy.getPrincipals(), policy.getOperations(),
+                        policy.getResources());
                 logger.atDebug("load-authorization-config")
                         .log("loaded authorization config for {} as policy {}", componentName, policy);
             } catch (AuthorizationException e) {
@@ -408,6 +409,7 @@ public class AuthorizationHandler  {
     }
 
     private void addPermission(String destination,
+                               String policyId,
                                Set<String> principals,
                                Set<String> operations,
                                Set<String> resources) throws AuthorizationException {
@@ -428,7 +430,13 @@ public class AuthorizationHandler  {
                                             .build());
                         } catch (AuthorizationException e) {
                             logger.atError("load-authorization-config-add-resource-error").setCause(e)
-                                    .log("Error while adding permission for resource {}", resource);
+                                    .kv("policyId", policyId)
+                                    .kv("component", principal)
+                                    .kv("operation", operation)
+                                    .kv("IPC service", destination)
+                                    .kv("resource", resource)
+                                            .log("Error while adding permission for component {} "
+                                                    + "to IPC Service {}", principal, destination);
                         }
                     }
                 }
--- a/src/main/java/com/aws/greengrass/authorization/AuthorizationHandler.java
+++ b/src/main/java/com/aws/greengrass/authorization/AuthorizationHandler.java
@@ -419,12 +419,17 @@ public class AuthorizationHandler  {
                             Permission.builder().principal(principal).operation(operation).resource(null).build());
                 } else {
                     for (String resource : resources) {
-                        authModule.addPermission(destination,
-                                Permission.builder()
-                                        .principal(principal)
-                                        .operation(operation)
-                                        .resource(resource)
-                                        .build());
+                        try {
+                            authModule.addPermission(destination,
+                                    Permission.builder()
+                                            .principal(principal)
+                                            .operation(operation)
+                                            .resource(resource)
+                                            .build());
+                        } catch (AuthorizationException e) {
+                            logger.atError("load-authorization-config-add-resource-error").setCause(e)
+                                    .log("Error while adding permission for resource {}", resource);
+                        }
                     }
                 }
             }
--- a/src/main/java/com/aws/greengrass/authorization/AuthorizationModule.java
+++ b/src/main/java/com/aws/greengrass/authorization/AuthorizationModule.java
@@ -10,10 +10,7 @@ import com.aws.greengrass.authorization.exceptions.AuthorizationException;
 import com.aws.greengrass.util.DefaultConcurrentHashMap;
 import com.aws.greengrass.util.Utils;
 
-import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.HashSet;
-import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.CopyOnWriteArraySet;
@@ -36,8 +33,6 @@ public class AuthorizationModule {
                     new DefaultConcurrentHashMap<>(WildcardTrie::new)));
     Map<String, Map<String, Map<String, Set<String>>>> rawResourceList = new DefaultConcurrentHashMap<>(
             () -> new DefaultConcurrentHashMap<>(() -> new DefaultConcurrentHashMap<>(CopyOnWriteArraySet::new)));
-    List<Character> specialChars = new ArrayList<>(Arrays.asList(wildcardChar, escapeChar,
-            singleCharWildcard));
 
     /**
      * Add permission for the given input set.
@@ -83,7 +78,7 @@ public class AuthorizationModule {
                 if (actualChar == nullChar) {
                     throw new AuthorizationException("Resource not allowed, incorrect escape sequence used");
                 }
-                if (!specialChars.contains(actualChar)) {
+                if (!isSpecialChar(actualChar)) {
                     throw new AuthorizationException("Resource not allowed, Only special characters "
                             + "('*', '$', '?') can be escaped");
                 }
@@ -100,6 +95,10 @@ public class AuthorizationModule {
             sb.append(currentChar);
         }
         return sb.toString();
+    }
+
+    boolean isSpecialChar(char actualChar) {
+        return actualChar == wildcardChar || actualChar == escapeChar || actualChar == singleCharWildcard;
     }
 
 
--- a/src/main/java/com/aws/greengrass/authorization/AuthorizationModule.java
+++ b/src/main/java/com/aws/greengrass/authorization/AuthorizationModule.java
@@ -76,11 +76,11 @@ public class AuthorizationModule {
             if (currentChar == escapeChar && i + 1 < length && resource.charAt(i + 1) == '{') {
                 char actualChar = getActualChar(resource.substring(i));
                 if (actualChar == nullChar) {
-                    throw new AuthorizationException("Resource not allowed, incorrect escape sequence used");
+                    throw new AuthorizationException("Resource can not contain invalid escape sequence");
                 }
                 if (!isSpecialChar(actualChar)) {
-                    throw new AuthorizationException("Resource not allowed, Only special characters "
-                            + "('*', '$', '?') can be escaped");
+                    throw new AuthorizationException("Resource can not have escaping for normal characters, "
+                            + "only special characters ('*', '$', '?') can be escaped");
                 }
                 // skip next 3 characters as they are accounted for in escape sequence
                 i = i + 3;
@@ -89,8 +89,7 @@ public class AuthorizationModule {
                 continue;
             }
             if (currentChar == singleCharWildcard) {
-                throw new AuthorizationException("Resource not allowed, '?' inside a resource can only be "
-                        + "used with escaping");
+                throw new AuthorizationException("Resource can not contain character '?' without escaping");
             }
             sb.append(currentChar);
         }
--- a/src/main/java/com/aws/greengrass/authorization/AuthorizationModule.java
+++ b/src/main/java/com/aws/greengrass/authorization/AuthorizationModule.java
@@ -48,52 +48,47 @@ public class AuthorizationModule {
             throw new AuthorizationException("Invalid arguments");
         }
         String resource = permission.getResource();
-        String validatedResource = validateResource(resource);
+        validateResource(resource);
         resourceAuthZCompleteMap.get(destination).get(permission.getPrincipal()).get(permission.getOperation()).add(
                 resource);
         rawResourceList.get(destination).get(permission.getPrincipal()).get(permission.getOperation()).add(
-                validatedResource);
+                resource);
     }
 
     /**
      * Only allow '?' if it's escaped. You can only escape special characters ('*', '$', '?').
      * Any occurrence of '${' is only valid if it holds a single valid special character ('*', '$', '?') inside it
      * and ends with '}'. (eg: "${*}" is valid, "${c}" is invalid, "${c" is invalid, ${*bc} is invalid)
-     * @param resource resource to be parsed
+     * @param resource resource to be validated
      */
-    private String validateResource(String resource) throws AuthorizationException {
+    private void validateResource(String resource) throws AuthorizationException {
         if (resource == null) {
-            return null;
+            return;
         }
         // resource as null is ok, but it should not be empty
         if (Utils.isEmpty(resource)) {
             throw new AuthorizationException("Resource cannot be empty");
         }
         int length = resource.length();
-        StringBuilder sb = new StringBuilder(length);
         for (int i = 0; i < length; i++) {
             char currentChar = resource.charAt(i);
             if (currentChar == escapeChar && i + 1 < length && resource.charAt(i + 1) == '{') {
                 char actualChar = getActualChar(resource.substring(i));
                 if (actualChar == nullChar) {
-                    throw new AuthorizationException("Resource can not contain invalid escape sequence");
+                    throw new AuthorizationException("Resource contains invalid escape sequence. "
+                            + "Only ${*}, ${$}, or ${?} are allowed");
                 }
                 if (!isSpecialChar(actualChar)) {
-                    throw new AuthorizationException("Resource can not have escaping for normal characters, "
-                            + "only special characters ('*', '$', '?') can be escaped");
+                    throw new AuthorizationException("Resource contains invalid escape "
+                            + "sequence: ${" + actualChar + "}. Only ${*}, ${$}, or ${?} are allowed");
                 }
                 // skip next 3 characters as they are accounted for in escape sequence
                 i = i + 3;
-                // add escaped character to resource string and continue to next character
-                sb.append(actualChar);
-                continue;
             }
             if (currentChar == singleCharWildcard) {
-                throw new AuthorizationException("Resource can not contain character '?' without escaping");
+                throw new AuthorizationException("Invalid resource: '?' character must be escaped");
             }
-            sb.append(currentChar);
         }
-        return sb.toString();
     }
 
     boolean isSpecialChar(char actualChar) {
--- a/src/main/java/com/aws/greengrass/authorization/AuthorizationModule.java
+++ b/src/main/java/com/aws/greengrass/authorization/AuthorizationModule.java
@@ -10,12 +10,20 @@ import com.aws.greengrass.authorization.exceptions.AuthorizationException;
 import com.aws.greengrass.util.DefaultConcurrentHashMap;
 import com.aws.greengrass.util.Utils;
 
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.CopyOnWriteArraySet;
 
 import static com.aws.greengrass.authorization.AuthorizationHandler.ANY_REGEX;
+import static com.aws.greengrass.authorization.WildcardTrie.escapeChar;
+import static com.aws.greengrass.authorization.WildcardTrie.getActualChar;
+import static com.aws.greengrass.authorization.WildcardTrie.nullChar;
+import static com.aws.greengrass.authorization.WildcardTrie.singleCharWildcard;
+import static com.aws.greengrass.authorization.WildcardTrie.wildcardChar;
 
 /**
  * Simple permission table which stores permissions. A permission is a
@@ -28,6 +36,8 @@ public class AuthorizationModule {
                     new DefaultConcurrentHashMap<>(WildcardTrie::new)));
     Map<String, Map<String, Map<String, Set<String>>>> rawResourceList = new DefaultConcurrentHashMap<>(
             () -> new DefaultConcurrentHashMap<>(() -> new DefaultConcurrentHashMap<>(CopyOnWriteArraySet::new)));
+    List<Character> specialChars = new ArrayList<>(Arrays.asList(wildcardChar, escapeChar,
+            singleCharWildcard));
 
     /**
      * Add permission for the given input set.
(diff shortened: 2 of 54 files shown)


src/main/java/com/aws/greengrass/authorization/AuthorizationModule.java
```diff
@@ -28,6 +36,8 @@
                     new DefaultConcurrentHashMap<>(WildcardTrie::new)));
     Map<String, Map<String, Map<String, Set<String>>>> rawResourceList = new DefaultConcurrentHashMap<>(
             () -> new DefaultConcurrentHashMap<>(() -> new DefaultConcurrentHashMap<>(CopyOnWriteArraySet::new)));
+    List<Character> specialChars = new ArrayList<>(Arrays.asList(wildcardChar, escapeChar,
+            singleCharWildcard));
```
rbattle, line 40: nit: can this be made a static constant?

src/main/java/com/aws/greengrass/authorization/AuthorizationModule.java
```diff
@@ -42,17 +52,57 @@ public void addPermission(String destination, Permission permission) throws Auth
...
+        int length = resource.length();
+        StringBuilder sb = new StringBuilder(length);
+        for (int i = 0; i < length; i++) {
+            char currentChar = resource.charAt(i);
+            if (currentChar == escapeChar && i + 1 < length && resource.charAt(i + 1) == '{') {
+                char actualChar = getActualChar(resource.substring(i));
+                if (actualChar == nullChar) {
+                    throw new AuthorizationException("Resource not allowed, incorrect escape sequence used");
+                }
+                if (!specialChars.contains(actualChar)) {
```
rbattle, line 86: if this is the only place it's being used, why not make a method like:

```
boolean isSpecialChar(char actualChar) {
  return actualChar == wildcardChar || actualChar == escapeChar || actualChar == singleCharWildcard;
}
```

There's only 3 characters to check - no need create an array list and store that.

src/main/java/com/aws/greengrass/authorization/WildcardTrie.java
```diff
@@ -149,6 +161,23 @@ private WildcardTrie add(String subject, boolean isTerminal) {
         return current;
     }
 
+    /**
+     * The method tries to parse the given string using escape sequence ${c} (where c is a character to be escaped)
+     * and returns the character c if the pattern is matched. In any other scenario it returns null character ('\0')
+     *
+     * @param str string provided to get
+     */
+    public static char getActualChar(String str) {
```
rbattle, line 170: why public? can this be package private?

src/main/java/com/aws/greengrass/authorization/WildcardTrie.java
```diff
@@ -141,6 +145,14 @@ private WildcardTrie add(String subject, boolean isTerminal) {
...
                 return current.add(subject.substring(i + 1), true);
             }
+            if (currentChar == escapeChar) {
+                char actualChar = getActualChar(subject.substring(i));
+                if (actualChar != nullChar) {
+                    sb.append(actualChar);
+                    i = i + 3;
+                    continue;
+                }
+            }
```
rbattle, line 155: you have AuthModule tests that validate resources that end with `$` but you don't have any tests that validate the WildCardTrie if it ends in $/if the `actualChar == null`

src/main/java/com/aws/greengrass/authorization/AuthorizationModule.java
vabhasin: Yes, I can make a method for this. Thanks

src/main/java/com/aws/greengrass/authorization/WildcardTrie.java
vabhasin: Are you saying something like 
add("abc$") should match "abc$"

src/main/java/com/aws/greengrass/authorization/WildcardTrie.java
vabhasin: done.

rbattle (approved):

src/main/java/com/aws/greengrass/authorization/AuthorizationModule.java
```diff
@@ -42,17 +47,61 @@ public void addPermission(String destination, Permission permission) throws Auth
...
         }
-        // resource as null is ok, but it should not be empty
         String resource = permission.getResource();
-        if (resource != null && Utils.isEmpty(resource)) {
-            throw new AuthorizationException("Resource cannot be empty");
-        }
+        String validatedResource = validateResource(resource);
         resourceAuthZCompleteMap.get(destination).get(permission.getPrincipal()).get(permission.getOperation()).add(
-                permission.getResource());
+                resource);
```
youtuyy, line 53: How is this map used? Why it doesn't need the validated resource?

src/main/java/com/aws/greengrass/authorization/AuthorizationModule.java
```diff
@@ -42,17 +47,61 @@ public void addPermission(String destination, Permission permission) throws Auth
...
+        StringBuilder sb = new StringBuilder(length);
+        for (int i = 0; i < length; i++) {
+            char currentChar = resource.charAt(i);
+            if (currentChar == escapeChar && i + 1 < length && resource.charAt(i + 1) == '{') {
+                char actualChar = getActualChar(resource.substring(i));
+                if (actualChar == nullChar) {
+                    throw new AuthorizationException("Resource not allowed, incorrect escape sequence used");
+                }
+                if (!isSpecialChar(actualChar)) {
+                    throw new AuthorizationException("Resource not allowed, Only special characters "
```
youtuyy, line 82: nit: only

youtuyy: I also wonder if those exception should say"Invalid arguments" instead of "Resource not allowed"

src/main/java/com/aws/greengrass/authorization/AuthorizationModule.java
vabhasin: This map maintains the combination of Destination, principal, operation and the resource trie with it.
ValidateResource also parses and removes escaping to give out raw resource. But the add method for the trie needs the original resource as it needs that info to differentiate real wildcard and escaped wildcard chars

src/main/java/com/aws/greengrass/authorization/AuthorizationModule.java
vabhasin: For Customers, its the policy they are adding thats failing. I think this tells why it is failing.

MikeDombo: I'm not sure about throwing here. Logging, sure, but not about throwing. 

I do not like how we handle auth policy issues right now in that it can prevent ALL policies from working, not only the broken policies.

vabhasin: I had talked to Joe about this, I think he said we want to fail as early as possible maybe at deployment so the user knows the problem . I could have understood it wrong. Joe hasn't reviewed this and said he will look at this on Monday. We can discuss it more then. But I agree this prevents all policies from working.

MikeDombo: It doesn't cause the deployment to fail, so I don't think this is good.

src/main/java/com/aws/greengrass/authorization/AuthorizationModule.java
youtuyy: Thanks for explaining.

src/main/java/com/aws/greengrass/authorization/AuthorizationModule.java
youtuyy: I think this validation is in addPermission() which is called when loading/updating configs. So I think the log should point out that they need to fix that resource argument but I just feel "Resource not allowed" sounds like it's not authorized..

vabhasin: Oh okay, so  what will happen when this error is thrown, that particular merge-config just won't work?

MikeDombo: Trace the calls. Who calls this and what do they do with the exception?

vabhasin: > I think this validation is in addPermission() which is called when loading/updating configs. So I think the log should point out that they need to fix that resource argument but I just feel "Resource not allowed" sounds like it's not authorized..
hmm...makes sense @youtuyy . Any suggestions, like "Invalid resource in AccessControl, resource contains incorrect escape sequence" or something?

youtuyy: yeah probably just something like "Resource cannot contain invalid escape sequence + some details.. " also depending on the path the exception is propagated. You can also check with some integ tests what will be logged and thrown

src/main/java/com/aws/greengrass/authorization/AuthorizationModule.java
```diff
@@ -42,17 +47,61 @@ public void addPermission(String destination, Permission permission) throws Auth
...
+                validatedResource);
+    }
+
+    /**
+     * Only allow '?' if it's escaped. You can only escape special characters ('*', '$', '?').
+     * Any occurrence of '${' is only valid if it holds a single valid special character ('*', '$', '?') inside it
+     * and ends with '}'. (eg: "${*}" is valid, "${c}" is invalid, "${c" is invalid, ${*bc} is invalid)
+     * @param resource resource to be parsed
+     */
+    private String validateResource(String resource) throws AuthorizationException {
```
MikeDombo, line 64: We allow `*` as a resource. Is that allowed by this?

vabhasin: Yes

src/main/java/com/aws/greengrass/authorization/AuthorizationModule.java
vabhasin: > Trace the calls. Who calls this and what do they do with the exception?

@MikeDombo The only time the policies are loaded is initially and when there's a merge/update config. The behavior right now is, any other resources included in that same policy(with an invalid resource) will not work and throw exception which will be logged by AuthorizationHandler. But a separate policy that is before/after the invalid policy will still be loaded and will work.
I can change it to reject a particular resource inside the policy, and continue with the next resource, so only that resource is rejected.

src/main/java/com/aws/greengrass/authorization/AuthorizationHandler.java
```diff
@@ -419,12 +419,17 @@ private void addPermission(String destination,
...
+                        try {
+                            authModule.addPermission(destination,
+                                    Permission.builder()
+                                            .principal(principal)
+                                            .operation(operation)
+                                            .resource(resource)
+                                            .build());
+                        } catch (AuthorizationException e) {
+                            logger.atError("load-authorization-config-add-resource-error").setCause(e)
+                                    .log("Error while adding permission for resource {}", resource);
```
MikeDombo, line 431: log more info here. What is the principal and component name?

src/main/java/com/aws/greengrass/authorization/WildcardTrie.java
MikeDombo: `$` is a valid MQTT topic character. For example `$aws`. How does this affect customers currently allowing subscription/publishing on $aws?

vabhasin: Only `$` is still valid, its only when the resource has `$` followed by `{` we start checking if its using escape sequence correctly.

MikeDombo: But it is intended have special meaning, such that if I wanted a literal $, I would put ${$}

src/main/java/com/aws/greengrass/authorization/AuthorizationModule.java
```diff
@@ -42,17 +47,54 @@ public void addPermission(String destination, Permission permission) throws Auth
...
+                }
+                if (!isSpecialChar(actualChar)) {
+                    throw new AuthorizationException("Resource can not have escaping for normal characters, "
+                            + "only special characters ('*', '$', '?') can be escaped");
+                }
+                // skip next 3 characters as they are accounted for in escape sequence
+                i = i + 3;
+            }
+            if (currentChar == singleCharWildcard) {
+                throw new AuthorizationException("Resource can not contain character '?' without escaping");
```
youtuyy, line 88: Will we support '?' wildcard in the future? If so maybe we can say "currently unsupported. reserved for future use" ?

src/main/java/com/aws/greengrass/authorization/WildcardTrie.java
vabhasin: That makes sense as we have escaping allowed for `$`. But acc to IoT Core policy's behavior. A `$` doesn't always need needs escaping and we are trying to match that.

src/main/java/com/aws/greengrass/authorization/AuthorizationHandler.java
```diff
@@ -419,12 +419,17 @@ private void addPermission(String destination,
                             Permission.builder().principal(principal).operation(operation).resource(null).build());
```
youtuyy, line 419: So in this case it still propagates the exception? why are we handling differently?

vabhasin: I believe that part will never have exception but if in case it does, that exception should propagate up and log `Error while loading policy` instead of `Error while adding resource` because we are not really going through each resource.
So I though it's ok to keep it that way.

src/main/java/com/aws/greengrass/authorization/AuthorizationModule.java
vabhasin: Yes, we can and this could be changed, but we are not sure of that. So I can add just "currently unsupported" part maybe

src/main/java/com/aws/greengrass/authorization/AuthorizationHandler.java
```diff
@@ -419,12 +421,19 @@ private void addPermission(String destination,
...
+                        try {
+                            authModule.addPermission(destination,
+                                    Permission.builder()
+                                            .principal(principal)
+                                            .operation(operation)
+                                            .resource(resource)
+                                            .build());
+                        } catch (AuthorizationException e) {
+                            logger.atError("load-authorization-config-add-resource-error").setCause(e)
+                                    .log("PolicyId {}: Error while adding permission for principal {} "
```
MikeDombo, line 433: we want to include the destination too. Make it impossible for a customer to not know where to look to fix it

MikeDombo: And don't necessarily use the terms which we use in code, this needs to be customer friendly.

vabhasin: policyId should tell them where to look. Anyways I can add it, no harm in it.
Also, Can I use the term service for the Destination Component?
`PolicyId {}: Error while adding permission for component {principal} to perform operation {op} ({destination}) on resource {resource}`

MikeDombo: https://docs.aws.amazon.com/greengrass/v2/developerguide/interprocess-communication.html#ipc-authorization-policies
We use "IPC service" in the docs it seems

MikeDombo (approved):

youtuyy (approved):

src/main/java/com/aws/greengrass/authorization/AuthorizationHandler.java
```diff
@@ -419,12 +421,19 @@ private void addPermission(String destination,
...
+                                            .principal(principal)
+                                            .operation(operation)
+                                            .resource(resource)
+                                            .build());
+                        } catch (AuthorizationException e) {
+                            logger.atError("load-authorization-config-add-resource-error").setCause(e)
+                                    .log("PolicyId {}: Error while adding permission for component {} "
+                                                    + "to perform operation {} (IPC Service: {}) on resource {} ",
+                                            policyId, principal, operation, destination, resource);
+                        }
```
rbattle, line 436: why not add these as k/v - 
e.g.:
```
logger.atError(...).setCause(e)
  .kv("policyId", policyId)
  .kv("component", principal)
  .kv("operation", operation)
  .kv("IPC service", destination) // what does IPC service mean here? 
  .kv("resource',. resource)
  .log("Error while adding permission for component");
```

src/main/java/com/aws/greengrass/authorization/AuthorizationModule.java
```diff
@@ -42,17 +47,55 @@ public void addPermission(String destination, Permission permission) throws Auth
...
+        if (Utils.isEmpty(resource)) {
+            throw new AuthorizationException("Resource cannot be empty");
+        }
+        int length = resource.length();
+        for (int i = 0; i < length; i++) {
+            char currentChar = resource.charAt(i);
+            if (currentChar == escapeChar && i + 1 < length && resource.charAt(i + 1) == '{') {
+                char actualChar = getActualChar(resource.substring(i));
+                if (actualChar == nullChar) {
+                    throw new AuthorizationException("Resource can not contain invalid escape sequence");
```
rbattle, line 78: nit: 

is it possible to add the invalid escape sequence to the message?

like `"Resource contains invalid escape sequence: ${c}"`

or if that's a pain,  can we say something like:

`"Invalid escape sequence. Only ${*}, ${$}, or ${?} are allowed."`

src/main/java/com/aws/greengrass/authorization/AuthorizationModule.java
```diff
@@ -42,17 +47,55 @@ public void addPermission(String destination, Permission permission) throws Auth
...
+                }
+                if (!isSpecialChar(actualChar)) {
+                    throw new AuthorizationException("Resource can not have escaping for normal characters, "
```
rbattle, line 81: nit: same here: can we clean up these error messages:
`"Invalid escape sequence: ${" + actualChar + "}". Only ${*}, ${$}, or ${?} are allowed."`

src/main/java/com/aws/greengrass/authorization/AuthorizationModule.java
```diff
@@ -42,17 +47,55 @@ public void addPermission(String destination, Permission permission) throws Auth
...
+                            + "only special characters ('*', '$', '?') can be escaped");
+                }
+                // skip next 3 characters as they are accounted for in escape sequence
+                i = i + 3;
+            }
+            if (currentChar == singleCharWildcard) {
+                throw new AuthorizationException("Resource can not contain character '?' without escaping. "
+                        + "'?' as a wildcard is currently unsupported");
```
rbattle, line 89: Nit: `Invalid resource: ? must be escaped.`

rbattle: @FThompsonAWS

src/main/java/com/aws/greengrass/authorization/AuthorizationModule.java
rbattle: @FThompsonAWS

src/main/java/com/aws/greengrass/authorization/AuthorizationModule.java
rbattle: @FThompsonAWS do you have any thoughts on the exception messages here?

rbattle (commented):
some nit picks on the exception messages - I've tagged @FThompsonAWS since I know docs team likes to see these messages

rbattle (approved):

youtuyy (dismissed):

youtuyy (approved):

rbattle (dismissed):

src/main/java/com/aws/greengrass/authorization/AuthorizationHandler.java
```diff
@@ -419,12 +421,23 @@ private void addPermission(String destination,
...
+                                            .build());
+                        } catch (AuthorizationException e) {
+                            logger.atError("load-authorization-config-add-resource-error").setCause(e)
+                                    .kv("policyId", policyId)
+                                    .kv("component", principal)
+                                    .kv("operation", operation)
+                                    .kv("IPC service", destination)
+                                    .kv("resource", resource)
+                                            .log("Error while adding permission for component {} "
+                                                    + "to IPC Service {}", principal, destination);
```
FThompsonAWS, line 439: What causes the error here? Can we explain it to the customer, including information about what step to take to resolve the error?

src/main/java/com/aws/greengrass/authorization/AuthorizationModule.java
```diff
@@ -42,17 +47,55 @@ public void addPermission(String destination, Permission permission) throws Auth
...
+            throw new AuthorizationException("Resource cannot be empty");
+        }
+        int length = resource.length();
+        for (int i = 0; i < length; i++) {
+            char currentChar = resource.charAt(i);
+            if (currentChar == escapeChar && i + 1 < length && resource.charAt(i + 1) == '{') {
+                char actualChar = getActualChar(resource.substring(i));
+                if (actualChar == nullChar) {
+                    throw new AuthorizationException("Resource contains invalid escape sequence. "
+                            + "Only ${*}, ${$}, or ${?} are allowed");
```
FThompsonAWS, line 79: Can we change this to `"Resource contains an empty escape sequence: ${}. You can use ${*}, ${$}, or ${?}."`

src/main/java/com/aws/greengrass/authorization/AuthorizationModule.java
```diff
@@ -42,17 +47,55 @@ public void addPermission(String destination, Permission permission) throws Auth
...
+                }
+                if (!isSpecialChar(actualChar)) {
+                    throw new AuthorizationException("Resource contains invalid escape "
+                            + "sequence: ${" + actualChar + "}. Only ${*}, ${$}, or ${?} are allowed");
```
FThompsonAWS, line 83: Can we change this to `"Resource contains an invalid escape sequence: ${<actualChar>}. You can use ${*}, ${$}, or ${?}."`

src/main/java/com/aws/greengrass/authorization/AuthorizationModule.java
```diff
@@ -42,17 +47,55 @@ public void addPermission(String destination, Permission permission) throws Auth
...
+                }
+                // skip next 3 characters as they are accounted for in escape sequence
+                i = i + 3;
+            }
+            if (currentChar == singleCharWildcard) {
+                throw new AuthorizationException("Invalid resource: '?' character must be escaped");
```
FThompsonAWS, line 89: Can we change this to `Resource contains invalid character: '?'. Use an escape sequence: ${?}.`

src/main/java/com/aws/greengrass/authorization/AuthorizationHandler.java
vabhasin: The error reason will follow this message. The reason comes from the 3 cases that we talk about below

src/main/java/com/aws/greengrass/authorization/AuthorizationModule.java
vabhasin: This might not necessarily be an "empty" escape sequence, but we can change the second line.

src/main/java/com/aws/greengrass/authorization/AuthorizationModule.java
vabhasin: Sure.

src/main/java/com/aws/greengrass/authorization/AuthorizationModule.java
FThompsonAWS: Got it. `"Resource contains an invalid escape sequence. You can use ${*}, ${$}, or ${?}."`

src/main/java/com/aws/greengrass/authorization/AuthorizationModule.java
FThompsonAWS: Can we change this to `"Resource contains invalid character: '?'. Use an escape sequence: ${?}. The '?' character isn't supported as a wildcard."`

src/main/java/com/aws/greengrass/authorization/AuthorizationModule.java
FThompsonAWS: Minor change: `"Resource contains invalid character: '?'. Use an escape sequence: ${?}. The '?' character isn't supported as a wildcard."`

FThompsonAWS (approved):

youtuyy (approved):

rbattle (approved):

PR #1210, opened 2022-04-25
feat: add handleRequestAsync to IPC OperationContinuationHandler

MikeDombo:
**Issue #, if available:**

**Description of changes:**
This change adds a new method `handleRequestAsync` to the `OperationContinuationHandler` of our IPC server. This method returns a completable future which ultimately returns the result. It is optional to implement this method. The reason to use it, is that the IPC request can be handled by a different thread, and thus unblock the SDK's main thread, allowing it to process other requests.

This change enables us to not block the main SDK thread while doing something which would otherwise block. This is needed for our CDA APIs which perform cloud calls which we want to return a result from during the initial IPC request (non-streaming).

Manually updated PubSub server to use this method and run the existing handler in a different thread, showing that it does work asynchronously the same way that it worked synchronously.

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/software/amazon/awssdk/eventstreamrpc/OperationContinuationHandler.java
+++ b/src/main/java/software/amazon/awssdk/eventstreamrpc/OperationContinuationHandler.java
@@ -5,12 +5,6 @@
 
 package software.amazon.awssdk.eventstreamrpc;
 
-import java.nio.charset.StandardCharsets;
-
-import java.util.ArrayList;
-import java.util.List;
-import java.util.concurrent.CompletableFuture;
-
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import software.amazon.awssdk.crt.eventstream.Header;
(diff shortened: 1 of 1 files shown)


naikvidish (approved):

src/main/java/software/amazon/awssdk/eventstreamrpc/OperationContinuationHandler.java
```diff
@@ -189,128 +199,155 @@ final public CompletableFuture<Void> sendStreamEvent(final StreamingResponseType
...
                 initialRequest = serviceModel.fromJson(getRequestClass(), bytes);
                 //call into business logic
-                final ResponseType result = handleRequest(initialRequest);
-                if (result != null) {
-                    if (!getResponseClass().isInstance(result)) {
-                        throw new RuntimeException("Handler for operation [" + getOperationName()
-                                + "] did not return expected type. Found: " + result.getClass().getName());
+                CompletableFuture<ResponseType> resultFuture = handleRequestAsync(initialRequest);
+                if (resultFuture == null) {
+                    resultFuture = CompletableFuture.completedFuture(handleRequest(initialRequest));
```
graebm, line 290: the old implementation would `throw new RuntimeException("Operation handler returned null response!");` if this returned null, the new implementation will just `NullPointerException` in the next few lines ... which is fine. But just pointing it out

src/main/java/software/amazon/awssdk/eventstreamrpc/OperationContinuationHandler.java
```diff
@@ -189,128 +199,155 @@ final public CompletableFuture<Void> sendStreamEvent(final StreamingResponseType
...
+                            if (ex != null) {
+                                LOGGER.error(ex.getClass().getName() + " sending response message: " + ex.getMessage());
+                            } else {
+                                LOGGER.trace("Response successfully sent");
+                            }
+                        });
+                        invokeAfterHandleRequest();
+                    } else {
+                        //not streaming, but null response? we have a problem
+                        throw new RuntimeException("Operation handler returned null response!");
```
graebm, line 312: This runtime used to be synchronous from the `onContinuationMessage` callback, but now it happens as part of  CompletableFuture.handle. I'm not sure where these exceptions go. Do they just vanish? Was this something we expected to catch from whatever calls `onContinuationMessage`? Or did that exception just vanish as well?

graebm (commented):
this seems as similar to the old implementation as you can get. I made some comments where they're trivially different and I'm not sure if there would be any repurcussions

src/main/java/software/amazon/awssdk/eventstreamrpc/OperationContinuationHandler.java
MikeDombo: No, this is different. The null completable future is part of the specification of the method. If the completable future _resolves_ to null, then the same exception will be logged that the handler returned null.

src/main/java/software/amazon/awssdk/eventstreamrpc/OperationContinuationHandler.java
MikeDombo: They were being caught and handled in the place that I extracted to the handleAndSendError. They are still handled that way, and in fact these exceptions aren't really possible anyway.

naikvidish (approved):

MikeDombo:
note: approved by @graebm as well.

PR #1215, opened 2022-04-30
fix: avoid NPE in getting v2 data client

junfuchen99:
**Issue #, if available:**

**Description of changes:**
1. Adds subscription to iot data and credential endpoints in `GreengrassServiceClientFactory` so that it re-validates cloud connection when any cloud-related configs change.
2. Validate configs when getting v2 data client or config validation error, instead of whenever a config node change, to  reduce duplicate callbacks.

`GreengrassServiceClientFactory` has incomplete subscription to device configuration on cloud-related configs, so that it may miss the updates from `ProvisioningConfigUpdateHelper::updateNucleusConfiguration` at the end of fleet provisioning and still considers device in an offline state even though it's online.

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/util/GreengrassServiceClientFactory.java
+++ b/src/main/java/com/aws/greengrass/util/GreengrassServiceClientFactory.java
@@ -22,6 +22,7 @@ import software.amazon.awssdk.services.greengrassv2data.GreengrassV2DataClient;
 import software.amazon.awssdk.services.greengrassv2data.GreengrassV2DataClientBuilder;
 
 import java.net.URI;
+import java.util.Optional;
 import java.util.concurrent.atomic.AtomicBoolean;
 import javax.inject.Inject;
 
@@ -62,9 +63,10 @@ public class GreengrassServiceClientFactory {
                     DEVICE_PARAM_PRIVATE_KEY_PATH) || validString(node, DEVICE_PARAM_GG_DATA_PLANE_PORT)
                     || validString(node, DEVICE_PARAM_IOT_CRED_ENDPOINT) || validString(node,
                     DEVICE_PARAM_IOT_DATA_ENDPOINT)) {
-                logger.atDebug().kv("modifiedNode", node.getFullName()).kv("changeType", what)
-                        .log("Queued re-validation of Greengrass v2 data client.");
-                deviceConfigChanged.set(true);
+                logger.atTrace().kv("what", what).kv("node", node.getFullName()).log();
+                if (deviceConfigChanged.compareAndSet(false, true)) {
+                    logger.atDebug().log("Queued re-validation of Greengrass v2 data client");
+                }
                 cleanClient();
             }
         });
@@ -99,19 +101,19 @@ public class GreengrassServiceClientFactory {
      * Validate again if the device config has changed.
      *
      */
-    public String getConfigValidationError() {
+    public Optional<String> getConfigValidationError() {
         if (deviceConfigChanged.compareAndSet(true, false)) {
             validateConfiguration();
         }
-        return configValidationError;
+        return Optional.ofNullable(configValidationError);
     }
 
     /**
      * Initializes and returns GreengrassV2DataClient.
-     *
+     * Note that this method can return null if there is a config validation error.
      */
     public synchronized GreengrassV2DataClient getGreengrassV2DataClient() {
-        if (getConfigValidationError() != null) {
+        if (getConfigValidationError().isPresent()) {
             logger.atWarn().log("Failed to validate config for Greengrass v2 data client: {}", configValidationError);
             return null;
         }
--- a/src/main/java/com/aws/greengrass/util/GreengrassServiceClientFactory.java
+++ b/src/main/java/com/aws/greengrass/util/GreengrassServiceClientFactory.java
@@ -40,6 +40,7 @@ public class GreengrassServiceClientFactory {
     private static final Logger logger = LogManager.getLogger(GreengrassServiceClientFactory.class);
     private final DeviceConfiguration deviceConfiguration;
     private GreengrassV2DataClient greengrassV2DataClient;
+    // stores the result of last validation; null <=> successful
     private volatile String configValidationError;
 
     /**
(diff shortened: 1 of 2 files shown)


src/main/java/com/aws/greengrass/util/GreengrassServiceClientFactory.java
```diff
@@ -48,23 +53,29 @@ public class GreengrassServiceClientFactory {
...
+            return false;
+        }, (what) -> {
+            validateConfigQueued.set(false);
+            cleanClient();
+        }));
     }
 
     @SuppressWarnings("PMD.NullAssignment")
     private void validateConfiguration() {
+        logger.atDebug().log("Validating device configs for Greengrass v2 data client.");
```
MikeDombo, line 78: this isn't a useful log

junfuchen99: will remove it

src/main/java/com/aws/greengrass/util/GreengrassServiceClientFactory.java
```diff
@@ -87,12 +98,24 @@ private boolean validString(Node node, String key) {
         return node != null && node.childOf(key) && Utils.isNotEmpty(Coerce.toString(node));
     }
 
+    /**
+     * Retrieve configValidationError.
+     * Validate again if the device config has changed.
+     *
+     */
+    public String getConfigValidationError() {
```
MikeDombo, line 106: no one is call this, so we still won't know what the issue is

src/main/java/com/aws/greengrass/util/GreengrassServiceClientFactory.java
```diff
@@ -48,23 +53,29 @@ public class GreengrassServiceClientFactory {
     @Inject
     public GreengrassServiceClientFactory(DeviceConfiguration deviceConfiguration) {
         this.deviceConfiguration = deviceConfiguration;
-        deviceConfiguration.onAnyChange((what, node) -> {
+        deviceConfiguration.onAnyChange(new BatchedSubscriber((what, node) -> {
```
MikeDombo, line 56: I'd undo this, just clear the client when stuff changes and do the validation when someone actually asks for the client.

src/main/java/com/aws/greengrass/util/GreengrassServiceClientFactory.java
junfuchen99: will add a log

src/main/java/com/aws/greengrass/util/GreengrassServiceClientFactory.java
```diff
@@ -69,6 +76,7 @@ private void validateConfiguration() {
             deviceConfiguration.validate(true);
             configValidationError = null;
         } catch (DeviceConfigurationException e) {
+            logger.atWarn().log("Failed to validate config for Greengrass v2 data client", e);
```
MikeDombo, line 79: we don't need the stacktrace, only log the actual message

src/main/java/com/aws/greengrass/util/GreengrassServiceClientFactory.java
```diff
@@ -70,6 +77,7 @@ private void validateConfiguration() {
             configValidationError = null;
         } catch (DeviceConfigurationException e) {
             configValidationError = e.getMessage();
+            logger.atWarn().log("Failed to validate config for Greengrass v2 data client: {}", configValidationError);
```
MikeDombo, line 80: Not sure I want the log here, I think putting it within the getter may be better, so that the warning and subsequent null pointer (or other error) are right next to each other

junfuchen99: updated

MikeDombo (dismissed):

src/main/java/com/aws/greengrass/util/GreengrassServiceClientFactory.java
```diff
@@ -87,12 +94,25 @@ private boolean validString(Node node, String key) {
...
+            validateConfiguration();
+        }
+        return configValidationError;
+    }
+
     /**
      * Initializes and returns GreengrassV2DataClient.
      *
      */
     public synchronized GreengrassV2DataClient getGreengrassV2DataClient() {
```
rbattle, line 113: since you are making changes here, can you update the JavaDoc to indicate that this can return `null` if there is a config validation error

src/main/java/com/aws/greengrass/util/GreengrassServiceClientFactory.java
```diff
@@ -87,12 +94,25 @@ private boolean validString(Node node, String key) {
         return node != null && node.childOf(key) && Utils.isNotEmpty(Coerce.toString(node));
     }
 
+    /**
+     * Retrieve configValidationError.
+     * Validate again if the device config has changed.
+     *
+     */
+    public String getConfigValidationError() {
```
rbattle, line 102: make this Optional<String> or add a javadoc that this will return `null` if there is no error

src/main/java/com/aws/greengrass/util/GreengrassServiceClientFactory.java
```diff
@@ -54,13 +59,15 @@ public GreengrassServiceClientFactory(DeviceConfiguration deviceConfiguration) {
...
             if (validString(node, DEVICE_PARAM_AWS_REGION) || validString(node, DEVICE_PARAM_ROOT_CA_PATH)
                     || validString(node, DEVICE_PARAM_CERTIFICATE_FILE_PATH) || validString(node,
-                    DEVICE_PARAM_PRIVATE_KEY_PATH) || validString(node, DEVICE_PARAM_GG_DATA_PLANE_PORT)) {
-                validateConfiguration();
+                    DEVICE_PARAM_PRIVATE_KEY_PATH) || validString(node, DEVICE_PARAM_GG_DATA_PLANE_PORT)
+                    || validString(node, DEVICE_PARAM_IOT_CRED_ENDPOINT) || validString(node,
+                    DEVICE_PARAM_IOT_DATA_ENDPOINT)) {
+                logger.atDebug().kv("modifiedNode", node.getFullName()).kv("changeType", what)
+                        .log("Queued re-validation of Greengrass v2 data client.");
+                deviceConfigChanged.set(true);
```
rbattle, line 67: couple nits:
1. could this spam a bunch of logs? Would it be better to do something like:

```
logger.atTrace().kv("what", what).kv("node", node.getFullName()).log()
if (deviceConfigChanged.compareAndSet(false, true)) {
  logger.atDebug().log("Queued re-validation of Greengrass v2 data client.");
}
```

2. Don't add the "." at the end, otherwise we get "client.." in the logs

src/main/java/com/aws/greengrass/util/GreengrassServiceClientFactory.java
junfuchen99: I checked that only `GreengrassRepositoryDownloader::checkDownloadable` calls `getConfigValidationError`. Will make this optional and modify the usage.

src/main/java/com/aws/greengrass/util/GreengrassServiceClientFactory.java
junfuchen99: Will update.

src/main/java/com/aws/greengrass/util/GreengrassServiceClientFactory.java
junfuchen99: will update

rbattle (approved):

MikeDombo (approved):

PR #1217, opened 2022-05-03
feat: only interpolate system parameter in component configuration section, not configuration parameter

rightpeter:
**Description of changes:**
Before, nucleus interpolates component config after aggregating config from recipe, deployment update and running config. Running config is dynamic which exposes the risk of attacker can manipulate the running config in a way to be interpolated in next deployment.

This change let nucleus only interpolate system parameter for component configuration, not configuration parameter. For example:
* `{configuration:/paramA}` won't be interpolated in configuration section
* `{iot:thingName}` will be interpolated
This makes the component configuration section interpolation behavior static.

Security risk: enabling interpolation for configuration parameter under configuration section can cause issue like
corrupted running config triggers unexpected behavior by having special combination of variable designator.

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/componentmanager/KernelConfigResolver.java
+++ b/src/main/java/com/aws/greengrass/componentmanager/KernelConfigResolver.java
@@ -162,8 +162,8 @@ public class KernelConfigResolver {
                 Object existingConfiguration = ((Map) servicesConfig.get(resolvedComponentsToDeploy.getName()))
                         .get(CONFIGURATION_CONFIG_KEY);
 
-                Object interpolatedConfiguration = interpolate(existingConfiguration, resolvedComponentsToDeploy,
-                        componentRecipe.getDependencies().keySet(), servicesConfig);
+                Object interpolatedConfiguration = interpolateSystemParametersOnly(existingConfiguration,
+                        resolvedComponentsToDeploy, componentRecipe.getDependencies().keySet(), servicesConfig);
 
                 ((Map) servicesConfig.get(resolvedComponentsToDeploy.getName()))
                         .put(CONFIGURATION_CONFIG_KEY, interpolatedConfiguration);
@@ -238,7 +238,7 @@ public class KernelConfigResolver {
         }
 
         Map<String, Object> resolvedConfiguration = resolveConfigurationToApply(optionalConfigUpdate.orElse(null),
-                componentRecipe, document, componentIdentifier);
+                componentRecipe, document);
 
         // merge resolved param and resolved configuration for backward compatibility
         resolvedServiceConfig
@@ -315,7 +315,7 @@ public class KernelConfigResolver {
     @SuppressWarnings("PMD.ConfusingTernary")
     private Map<String, Object> resolveConfigurationToApply(
             @Nullable ConfigurationUpdateOperation configurationUpdateOperation, ComponentRecipe componentRecipe,
-            DeploymentDocument document, ComponentIdentifier componentIdentifier) {
+            DeploymentDocument document) {
 
         // try read the running service config
         try (Context context = new Context()) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/aws/greengrass/componentmanager/KernelConfigResolver.java
```diff
@@ -505,22 +543,22 @@ private String replace(String stringValue, ComponentIdentifier componentIdentifi
...
-            if (CONFIGURATION_NAMESPACE.equals(namespace)) {
+            if (shouldInterpolateConfiguration && CONFIGURATION_NAMESPACE.equals(namespace)) {
                 Optional<String> configReplacement =
                         lookupConfigurationValueForComponent(targetComponent, key, resolvedKernelServiceConfig);
                 if (configReplacement.isPresent()) {
                     stringValue = stringValue.replace(matcher.group(), configReplacement.get());
                 }
-
-            } else if (systemParameters.containsKey(namespace)) {
+            } else if (!Objects.isNull(resolvedKernelServiceConfig) && systemParameters.containsKey(namespace)) {
```
MikeDombo, line 561: Why is this needed now?

src/test/java/com/aws/greengrass/componentmanager/KernelConfigResolverTest.java
```diff
@@ -481,66 +481,6 @@ void GIVEN_deployment_with_artifact_and_interpolate_config_set_WHEN_config_resol
                 equalTo("java -jar " + "{artifacts:{randomNameSpace:" + artifactsPath + "}}" + "/test.jar -x arg"));
     }
 
-    @Test
```
MikeDombo, line 484: Keep this test, just disable it with a junit annotation

src/main/java/com/aws/greengrass/componentmanager/KernelConfigResolver.java
```diff
@@ -555,8 +593,8 @@ private Optional<String> lookupConfigurationValueForComponent(String componentNa
 
         Map componentResolvedConfig;
 
-        if (resolvedKernelServiceConfig.containsKey(componentName) && ((Map) resolvedKernelServiceConfig
-                .get(componentName)).containsKey(CONFIGURATION_CONFIG_KEY)) {
+        if (!Objects.isNull(resolvedKernelServiceConfig) && resolvedKernelServiceConfig.containsKey(componentName)
```
MikeDombo, line 596: Why is this now required?

src/main/java/com/aws/greengrass/componentmanager/KernelConfigResolver.java
```diff
@@ -416,6 +417,22 @@ private List<String> pointerToPath(JsonPointer pointer) {
...
+     * @param componentIdentifier         target component id
+     * @param dependencies                name set of component's dependencies
+     * @param resolvedKernelServiceConfig resolved kernel configuration under "Services" key
+     * @return the interpolated lifecycle object
+     * @throws IOException for directory issues
+     */
+    public Object interpolateSystemParametersOnly(Object configValue, ComponentIdentifier componentIdentifier,
+                                                 Set<String> dependencies,
+                                                  Map<String, Object> resolvedKernelServiceConfig) throws IOException {
+        return interpolate(configValue, componentIdentifier, dependencies, resolvedKernelServiceConfig, false);
```
MikeDombo, line 433: I don't really like the idea of adding this boolean. Perhaps pass in the map of interpolatable things instead, and in this case copy the map and remove configuration.

src/main/java/com/aws/greengrass/componentmanager/KernelConfigResolver.java
rightpeter: Just do the validation before calling. Is there any concern to do so?

MikeDombo: Well it is a new check which was not considered necessary before. If the value is null it seems that we have some massive problems. I think we would want it to fail here.

MikeDombo: Can the value ever be null?

src/test/java/com/aws/greengrass/componentmanager/KernelConfigResolverTest.java
rightpeter: Changed it to Disabled.

src/main/java/com/aws/greengrass/componentmanager/KernelConfigResolver.java
rightpeter: This boolean make it explicit that if it's interpolate all parameters or only the system parameters which is more deterministic and readable. If we copy the map and remove configuration, it's more dynamic and expose the interpolation behavior to the caller. What's the benefit of doing so?

src/main/java/com/aws/greengrass/componentmanager/KernelConfigResolver.java
rightpeter: With our current code, not likely. The `resolvedKernelServiceConfig` was initialized at the very beginning.

But `resolve` is a public method, in the future it's possible that someone pass an empty one.

MikeDombo: If prefer that you undo this.

src/main/java/com/aws/greengrass/componentmanager/KernelConfigResolver.java
MikeDombo: A free floating boolean is more difficult to understand than using an enum or even what I'm suggesting with the map. I expect that this change will be temporary, so I'm OK with this for now. If it stays longer, I'd like a more explicit option.

shaguptashaikh (approved):

MikeDombo (approved):

PR #1218, opened 2022-05-05
fix: handle null ggv2dataclient

J-C-L:
**Issue #, if available:**

**Description of changes:**
Updating users of GreengrassV2DataClient getGreengrassV2DataClient() to handle a null response.

We've seen NPEs when the client is not retrieved.

**Documentation Checklist:**

**Compatibility Checklist:**
- [X ] I confirm that the change is backwards compatible.

### DIFF
--- a/src/main/java/com/aws/greengrass/componentmanager/ComponentServiceHelper.java
+++ b/src/main/java/com/aws/greengrass/componentmanager/ComponentServiceHelper.java
@@ -11,6 +11,7 @@ import com.aws.greengrass.deployment.exceptions.DeviceConfigurationException;
 import com.aws.greengrass.logging.api.Logger;
 import com.aws.greengrass.logging.impl.LogManager;
 import com.aws.greengrass.util.GreengrassServiceClientFactory;
+import com.aws.greengrass.util.RetryUtils;
 import com.vdurmont.semver4j.Requirement;
 import com.vdurmont.semver4j.Semver;
 import org.apache.commons.lang3.Validate;
@@ -22,6 +23,8 @@ import software.amazon.awssdk.services.greengrassv2data.model.ResolveComponentCa
 import software.amazon.awssdk.services.greengrassv2data.model.ResolvedComponentVersion;
 import software.amazon.awssdk.services.greengrassv2data.model.ResourceNotFoundException;
 
+import java.time.Duration;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.Map;
 import java.util.stream.Collectors;
@@ -30,6 +33,7 @@ import javax.inject.Inject;
 public class ComponentServiceHelper {
 
     protected static final Logger logger = LogManager.getLogger(ComponentServiceHelper.class);
+    public static final int CLIENT_RETRY_COUNT = 3;
 
     private final GreengrassServiceClientFactory clientFactory;
     private final PlatformResolver platformResolver;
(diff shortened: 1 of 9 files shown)


src/main/java/com/aws/greengrass/componentmanager/ComponentServiceHelper.java
```diff
@@ -65,8 +72,24 @@ ResolvedComponentVersion resolveComponentVersion(String componentName, Semver lo
...
+        RetryUtils.RetryConfig clientExceptionRetryConfig =
+                RetryUtils.RetryConfig.builder().initialRetryInterval(Duration.ofSeconds(30))
+                        .maxRetryInterval(Duration.ofSeconds(30)).maxAttempt(CLIENT_RETRY_COUNT)
+                        .retryableExceptions(Arrays.asList(DeviceConfigurationException.class)).build();
+
+        try (GreengrassV2DataClient greengrasV2DataClient = RetryUtils.runWithRetry(clientExceptionRetryConfig, () -> {
+            GreengrassV2DataClient client = clientFactory.getGreengrassV2DataClient();
+            if (client == null) {
+                String errorMessage = clientFactory.getConfigValidationError().isPresent()
+                        ? clientFactory.getConfigValidationError().get() : "Could not get GreengrassV2DataClient";
```
rbattle, line 85: nit: you could combine this as `clientFactory.getConfigValidationError().orElse("Could not get GreengrassV2DataClient")`

src/main/java/com/aws/greengrass/componentmanager/ComponentServiceHelper.java
```diff
@@ -65,8 +72,24 @@ ResolvedComponentVersion resolveComponentVersion(String componentName, Semver lo
...
-        try {
-            result = clientFactory.getGreengrassV2DataClient().resolveComponentCandidates(request);
+
+        RetryUtils.RetryConfig clientExceptionRetryConfig =
+                RetryUtils.RetryConfig.builder().initialRetryInterval(Duration.ofSeconds(30))
+                        .maxRetryInterval(Duration.ofSeconds(30)).maxAttempt(CLIENT_RETRY_COUNT)
+                        .retryableExceptions(Arrays.asList(DeviceConfigurationException.class)).build();
+
+        try (GreengrassV2DataClient greengrasV2DataClient = RetryUtils.runWithRetry(clientExceptionRetryConfig, () -> {
+            GreengrassV2DataClient client = clientFactory.getGreengrassV2DataClient();
```
rbattle, line 82: would it be better to keep and deprecate this API and add a new one that `throws DeviceConfigurationException`?

I'm not sure what a useful name for that API would be though

J-C-L: Was wondering if we should add a new API. What's our process for deprecation APIs without breaking folks using them?

src/main/java/com/aws/greengrass/componentmanager/ComponentServiceHelper.java
J-C-L: That's cleaner. Updated.

src/main/java/com/aws/greengrass/componentmanager/ComponentServiceHelper.java
J-C-L: Spoke offline. Will add @Deprecated and create a new API.

src/test/java/com/aws/greengrass/componentmanager/ComponentServiceHelperTest.java
```diff
@@ -60,7 +62,6 @@ class ComponentServiceHelperTest {
     @BeforeEach
     void beforeEach() {
         PlatformResolver platformResolver = new PlatformResolver(null);
-        lenient().when(clientFactory.getGreengrassV2DataClient()).thenReturn(client);
```
kucnavya, line 63: Would it be useful to have some tests still cover this code path and test `getGreengrassV2DataClient` ?

src/main/java/com/aws/greengrass/deployment/DeploymentDocumentDownloader.java
```diff
@@ -173,6 +175,7 @@ private GetDeploymentConfigurationResponse getDeploymentConfiguration(String dep
                     "Failed to contact Greengrass cloud or unable to parse response.", e);
         }
 
+
```
kucnavya, line 178: nit: remove new line

kucnavya (approved):

src/test/java/com/aws/greengrass/componentmanager/ComponentServiceHelperTest.java
J-C-L: Not sure which code paths you mean. No coverage was removed in this PR. I did notice that nothing in GreengrassServiceClientFactory is tested. So yeah, we should probably add those tests as a follow up.

kucnavya: I meant that `getGreengrassV2DataClient()` is now replaced with `fetchGreengrassV2DataClient()` and tests are modified to use the new method. We could add some more tests that cover calls to `getGreengrassV2DataClient()`. But that can be a follow up.

Issue #1239, opened 2022-06-23
(EasySetup): --setup-system-service true and --start false doesn't behave as expected


jmrtinb:
**Describe the bug**
When setting `--setup-system-service true` and `--start false` the service is still started right away.

**To Reproduce**
Install greengrass using the options `--setup-system-service true` and `--start false`.

**Expected behavior**
Based on the documentation for these flags:
```
--setup-system-service,  -ss  (Optional) Specify true or false. If true, then the AWS IoT Greengrass Core software sets
itself up as a system service that runs when this device boots. The system service name is greengrass.
Defaults to false.
--start, -s  (Optional) Specify true or false. If true, the AWS IoT Greengrass Core software runs setup steps,
(optional) provisions resources, and starts the software. If false, the software runs only setup
steps and (optional) provisions resources. Defaults to true.
```
I expected `If false, the software runs only setup steps and (optional) provisions resources.` to mean that the service would be setup, but not started.

**Actual behavior**
Greengrass service is started right away.

**Environment**
 - OS: Ubuntu 20.04
 - JDK version: OpenJDK Runtime Environment (build 1.8.0_312-8u312-b07-0ubuntu1~20.04-b07)
 - Nucleus version: 2.5.6

MikeDombo:
Closing as the pull request has been merged. This will be included in our next release (no ETA).

PR #1241, opened 2022-06-24
feat: respect --start false when using --setup-system-service true

jmrtinb:
**Issue #, if available:**
#1239

**Description of changes:**
This change makes it so that if `--start false` is set with `--setup-system-service true` then the system service is only installed, but not enabled or started.

Current behavior is unintuitive and doesn't really match the help printout for setup which doesn't call out that `--start false` currently doesn't do anything if `--setup-system-service true` is set. I have a usecase where a system reboot happens right after installing greengrass so I want to avoid starting up greengrass until afterwards.

As far as I can tell there weren't existing tests for easysetup, and I don't really have the expertise for creating those kinds of tests from scratch (esp. for handling other platforms). I did manually test the changes with systemd, but I don't easily have a way of testing the other platforms.

**Documentation Checklist:**

These changes should make the behavior more closely match what's already described in the help printout for easysetup.

**Compatibility Checklist:**

Not sure if this is considered a breaking change. It does change the behavior of setup when both `--start false` and `--setup-system-service true` are set, but the old behavior doesn't match the documentation.

### DIFF
--- a/src/main/java/com/aws/greengrass/util/orchestration/ProcdUtils.java
+++ b/src/main/java/com/aws/greengrass/util/orchestration/ProcdUtils.java
@@ -30,7 +30,7 @@ public class ProcdUtils implements SystemServiceUtils {
     private static final String PROCD_SERVICE_TEMPLATE = "greengrass.service.procd.template";
 
     @Override
-    public boolean setupSystemService(KernelAlternatives kernelAlternatives) {
+    public boolean setupSystemService(KernelAlternatives kernelAlternatives, boolean start) {
         logger.atInfo(LOG_EVENT_NAME).log("Start procd setup");
         try {
             kernelAlternatives.setupInitLaunchDirIfAbsent();
@@ -56,7 +56,9 @@ public class ProcdUtils implements SystemServiceUtils {
             SystemServiceUtils.runCommand(logger, LOG_EVENT_NAME,SERVICE_CONFIG_FILE_PATH + " reload", false);
             SystemServiceUtils.runCommand(logger, LOG_EVENT_NAME,SERVICE_CONFIG_FILE_PATH + " stop", false);
             SystemServiceUtils.runCommand(logger, LOG_EVENT_NAME,SERVICE_CONFIG_FILE_PATH + " enable", false);
-            SystemServiceUtils.runCommand(logger, LOG_EVENT_NAME,SERVICE_CONFIG_FILE_PATH + " start", false);
+            if (start) {
+                SystemServiceUtils.runCommand(logger, LOG_EVENT_NAME,SERVICE_CONFIG_FILE_PATH + " start", false);
+            }
 
             logger.atInfo(LOG_EVENT_NAME).log("Successfully set up procd service");
             return true;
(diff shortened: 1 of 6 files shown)


src/main/java/com/aws/greengrass/util/orchestration/ProcdUtils.java
```diff
@@ -55,8 +55,10 @@ public boolean setupSystemService(KernelAlternatives kernelAlternatives) {
             // configured in "/bin/sh". So we launch this service through System V style.
             SystemServiceUtils.runCommand(logger, LOG_EVENT_NAME,SERVICE_CONFIG_FILE_PATH + " reload", false);
             SystemServiceUtils.runCommand(logger, LOG_EVENT_NAME,SERVICE_CONFIG_FILE_PATH + " stop", false);
-            SystemServiceUtils.runCommand(logger, LOG_EVENT_NAME,SERVICE_CONFIG_FILE_PATH + " enable", false);
-            SystemServiceUtils.runCommand(logger, LOG_EVENT_NAME,SERVICE_CONFIG_FILE_PATH + " start", false);
+            if (start) {
+                SystemServiceUtils.runCommand(logger, LOG_EVENT_NAME,SERVICE_CONFIG_FILE_PATH + " enable", false);
```
MikeDombo, line 59: service should always be enabled, that way it will start on the next reboot.

src/main/java/com/aws/greengrass/util/orchestration/SystemdUtils.java
```diff
@@ -49,8 +49,10 @@ public boolean setupSystemService(KernelAlternatives kernelAlternatives) {
             SystemServiceUtils.runCommand(logger, LOG_EVENT_NAME,"systemctl daemon-reload", false);
             SystemServiceUtils.runCommand(logger, LOG_EVENT_NAME,"systemctl unmask greengrass.service", false);
             SystemServiceUtils.runCommand(logger, LOG_EVENT_NAME,"systemctl stop greengrass.service", false);
-            SystemServiceUtils.runCommand(logger, LOG_EVENT_NAME,"systemctl start greengrass.service", false);
-            SystemServiceUtils.runCommand(logger, LOG_EVENT_NAME,"systemctl enable greengrass.service", false);
+            if (start) {
+                SystemServiceUtils.runCommand(logger, LOG_EVENT_NAME,"systemctl start greengrass.service", false);
+                SystemServiceUtils.runCommand(logger, LOG_EVENT_NAME,"systemctl enable greengrass.service", false);
```
MikeDombo, line 54: likewise, we should enable, but not necessarily start immediately.

src/main/java/com/aws/greengrass/util/orchestration/ProcdUtils.java
MikeDombo: On Windows the service is always enabled, no additional command.

MikeDombo: I understand that point of view, yes. Though I think that from a documentation and expectation perspective it would make more sense to have it enabled, but not started.

jmrtinb: Ok yeah, if windows is effectively enabled but not started then I think it definitely makes sense to be consistent.

MikeDombo (approved):
Thanks for adding this!

shaguptashaikh (approved):

PR #1242, opened 2022-06-24
feat: add IPC interface and agent for publishing component metrics

kucnavya:
**Issue #, if available:**

**Description of changes:**
Add IPC operation handlers for publishing component metrics 
Edit Telemetry metrics payload schema

To allow 1P component to publish metrics

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/builtin/services/telemetry/ComponentMetricIPCEventStreamAgent.java
+++ b/src/main/java/com/aws/greengrass/builtin/services/telemetry/ComponentMetricIPCEventStreamAgent.java
@@ -1,0 +1,197 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.aws.greengrass.builtin.services.telemetry;
+
+import com.aws.greengrass.authorization.AuthorizationHandler;
+import com.aws.greengrass.authorization.Permission;
+import com.aws.greengrass.authorization.exceptions.AuthorizationException;
+import com.aws.greengrass.logging.api.Logger;
+import com.aws.greengrass.logging.impl.LogManager;
+import com.aws.greengrass.telemetry.impl.Metric;
+import com.aws.greengrass.telemetry.impl.MetricFactory;
+import com.aws.greengrass.telemetry.models.TelemetryAggregation;
+import com.aws.greengrass.telemetry.models.TelemetryUnit;
+import com.aws.greengrass.util.Utils;
+import lombok.AccessLevel;
+import lombok.Getter;
+import org.apache.commons.lang3.EnumUtils;
+import software.amazon.awssdk.aws.greengrass.GeneratedAbstractPutComponentMetricOperationHandler;
+import software.amazon.awssdk.aws.greengrass.model.InvalidArgumentsError;
+import software.amazon.awssdk.aws.greengrass.model.PutComponentMetricRequest;
+import software.amazon.awssdk.aws.greengrass.model.PutComponentMetricResponse;
+import software.amazon.awssdk.aws.greengrass.model.ServiceError;
+import software.amazon.awssdk.aws.greengrass.model.UnauthorizedError;
+import software.amazon.awssdk.eventstreamrpc.OperationContinuationHandlerContext;
+import software.amazon.awssdk.eventstreamrpc.model.EventStreamJsonMessage;
+
+import java.nio.charset.StandardCharsets;
+import java.time.Instant;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Pattern;
+import javax.inject.Inject;
+
+import static com.aws.greengrass.ipc.common.ExceptionUtil.translateExceptions;
+import static com.aws.greengrass.ipc.modules.ComponentMetricIPCService.PUT_COMPONENT_METRIC_SERVICE_NAME;
+
+public class ComponentMetricIPCEventStreamAgent {
+    private static final Logger logger = LogManager.getLogger(ComponentMetricIPCEventStreamAgent.class);
+    private static final Pattern SERVICE_NAME_REGEX = Pattern.compile("^aws\\.");
+    private static final String NON_ALPHANUMERIC_REGEX = "[^A-Za-z0-9]";
+    private static final String SERVICE_NAME = "ServiceName";
+
+    @Getter(AccessLevel.PACKAGE)
+    private final Map<String, MetricFactory> metricFactoryMap = new HashMap<>();
+    private final AuthorizationHandler authorizationHandler;
+
+    @Inject
+    ComponentMetricIPCEventStreamAgent(AuthorizationHandler authorizationHandler) {
+        this.authorizationHandler = authorizationHandler;
+    }
+
+    public PutComponentMetricOperationHandler getPutComponentMetricHandler(
+            OperationContinuationHandlerContext context) {
+        return new PutComponentMetricOperationHandler(context);
+    }
+
+    class PutComponentMetricOperationHandler extends GeneratedAbstractPutComponentMetricOperationHandler {
+        private final String serviceName;
+
+        protected PutComponentMetricOperationHandler(OperationContinuationHandlerContext context) {
+            super(context);
+            serviceName = context.getAuthenticationData().getIdentityLabel();
+        }
+
+        @Override
+        protected void onStreamClosed() {
+            // NA
+        }
+
+        @SuppressWarnings({"PMD.PreserveStackTrace", "PMD.AvoidCatchingGenericException"})
+        @Override
+        public PutComponentMetricResponse handleRequest(PutComponentMetricRequest componentMetricRequest) {
+            return translateExceptions(() -> {
+                logger.atDebug().kv(SERVICE_NAME, serviceName)
+                        .log("Received putComponentMetricRequest from component " + serviceName);
+
+                // Authorize request for given operation
+                try {
+                    doAuthorization(this.getOperationModelContext().getOperationName(), serviceName);
+                } catch (AuthorizationException e) {
+                    logger.atError().kv(SERVICE_NAME, serviceName)
+                            .log("{} is not authorized to perform operation", serviceName);
+                    throw new UnauthorizedError(e.getMessage());
+                }
+
+                //validate - metric name length, value is non negative etc etc
+                List<software.amazon.awssdk.aws.greengrass.model.Metric> metricList =
+                        componentMetricRequest.getMetrics();
+                try {
+                    validateComponentMetricRequest(metricList);
+                } catch (IllegalArgumentException e) {
+                    logger.atError().kv(SERVICE_NAME, serviceName)
+                            .log("invalid component metric request from %s", serviceName);
+                    throw new InvalidArgumentsError(e.getMessage());
+                }
+
+                // Perform translations on metrics List
+                try {
+                    final String metricNamespace = serviceName;
+                    translateAndEmit(metricList, metricNamespace);
+                } catch (IllegalArgumentException e) {
+                    logger.atError().kv(SERVICE_NAME, serviceName)
+                            .log("invalid component metric request from %s", serviceName);
+                    throw new InvalidArgumentsError(e.getMessage());
+                } catch (Exception ex) {
+                    logger.atError().kv(SERVICE_NAME, serviceName)
+                            .log("error while emitting metrics from %s", serviceName);
+                    throw new ServiceError(ex.getMessage());
+                }
+
+                return new PutComponentMetricResponse();
+            });
+        }
+
+        @Override
+        public void handleStreamEvent(EventStreamJsonMessage streamRequestEvent) {
+            // NA
+        }
+
+        // Translate request metrics to telemetry metrics and emit them
+        private void translateAndEmit(List<software.amazon.awssdk.aws.greengrass.model.Metric> componentMetrics,
+                                      String metricNamespace) {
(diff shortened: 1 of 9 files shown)


src/main/java/com/aws/greengrass/builtin/services/telemetry/ComponentMetricIPCEventStreamAgent.java
```diff
@@ -0,0 +1,187 @@
...
+    private static final String NON_ALPHANUMERIC_REGEX = "[^A-Za-z0-9]";
+
+    @Getter(AccessLevel.PACKAGE)
+    private final Map<String, MetricFactory> metricFactoryMap = new HashMap<>();
+
+    @Inject
+    @Setter(AccessLevel.PACKAGE)
+    private AuthorizationHandler authorizationHandler;
+
+    @Inject
```
nikkhilmuthye, line 53: why do you have the `@Inject` here if you already have injected the `AuthorizationHandler` above?

src/main/java/com/aws/greengrass/builtin/services/telemetry/ComponentMetricIPCEventStreamAgent.java
```diff
@@ -0,0 +1,187 @@
...
+        @Override
+        protected void onStreamClosed() {
+            // NA
+        }
+
+        @SuppressWarnings("PMD.PreserveStackTrace")
+        @Override
+        public PutComponentMetricResponse handleRequest(PutComponentMetricRequest componentMetricRequest) {
+            return translateExceptions(() -> {
+                logger.atDebug().log("Received putComponentMetricRequest from component " + serviceName);
```
nikkhilmuthye, line 80: instead of doing a string operations. do this:
```suggestion
                logger.atDebug().kv("ServiceName", serviceName).log("Received putComponentMetricRequest from component " + serviceName);
```

src/main/java/com/aws/greengrass/builtin/services/telemetry/ComponentMetricIPCEventStreamAgent.java
```diff
@@ -0,0 +1,187 @@
...
+
+                // Authorize request for given operation
+                try {
+                    doAuthorization(this.getOperationModelContext().getOperationName(), serviceName);
+                } catch (AuthorizationException e) {
+                    throw new UnauthorizedError(e.getMessage());
```
nikkhilmuthye, line 86: log error everywhere in this file.

src/main/java/com/aws/greengrass/builtin/services/telemetry/ComponentMetricIPCEventStreamAgent.java
```diff
@@ -0,0 +1,178 @@
...
+            });
+        }
+
+        @Override
+        public void handleStreamEvent(EventStreamJsonMessage streamRequestEvent) {
+            // NA
+        }
+    }
+
+    // Translate request metrics to telemetry metrics and emit them
```
nikkhilmuthye, line 118: Add a java doc here instead of the a normal comment. As well as all the other functions.

src/main/java/com/aws/greengrass/builtin/services/telemetry/ComponentMetricIPCEventStreamAgent.java
```diff
@@ -0,0 +1,178 @@
...
+    private void translateAndEmit(List<software.amazon.awssdk.aws.greengrass.model.Metric> componentMetrics,
+                                  String metricNamespace) {
+        final MetricFactory metricFactory =
+                metricFactoryMap.computeIfAbsent(metricNamespace, k -> new MetricFactory(metricNamespace));
+
+        componentMetrics.forEach(metric -> {
+            logger.atDebug().log("Translating component metric to Telemetry metric" + metric.getName());
```
nikkhilmuthye, line 125: use `kv` here as well

src/main/java/com/aws/greengrass/builtin/services/telemetry/ComponentMetricIPCEventStreamAgent.java
```diff
@@ -0,0 +1,178 @@
...
+            Metric telemetryMetric = getTelemetryMetric(metric, metricNamespace);
```
nikkhilmuthye, line 126: You can just use the `builder` here instead of having another function.

src/main/java/com/aws/greengrass/builtin/services/telemetry/ComponentMetricIPCEventStreamAgent.java
```diff
@@ -0,0 +1,178 @@
...
+            throw new IllegalArgumentException("Invalid telemetry unit: No matching TelemetryUnit type found");
+        }
+        return telemetryEnum;
+    }
+
+    // Validate metric request
+    // Check name, unit and value arguments
+    private void validateComponentMetricRequest(List<software.amazon.awssdk.aws.greengrass.model.Metric> metrics) {
+        for (software.amazon.awssdk.aws.greengrass.model.Metric metric : metrics) {
+            if (Utils.isEmpty(metric.getName()) || metric.getName().length() <= 0 || Utils.isEmpty(
```
nikkhilmuthye, line 159: Add in the metric name validation. It should not be more than 32 bytes

src/main/java/com/aws/greengrass/builtin/services/telemetry/ComponentMetricIPCEventStreamAgent.java
kucnavya: Mainly did this for readability, but I'm okay adding the code here if you feel strongly about it.

src/main/java/com/aws/greengrass/builtin/services/telemetry/ComponentMetricIPCEventStreamAgent.java
kucnavya: I can add java doc for public methods but we might not need them for all the private and internal methods which aren't consumed externally. What do you think?

src/main/java/com/aws/greengrass/builtin/services/telemetry/ComponentMetricIPCEventStreamAgent.java
kucnavya: Good call. Modify the inject above for `AuthorizationHandler`

src/main/java/com/aws/greengrass/builtin/services/telemetry/ComponentMetricIPCEventStreamAgent.java
kucnavya: done.

src/main/java/com/aws/greengrass/builtin/services/telemetry/ComponentMetricIPCEventStreamAgent.java
nikkhilmuthye: It just keeps the entire class much more cleaner to follow one style of doc.

src/main/java/com/aws/greengrass/builtin/services/telemetry/ComponentMetricIPCEventStreamAgent.java
nikkhilmuthye: I don't think we should have a function for a one liner. but I don't feel strongly about it. So I'll leave it up to you.

src/main/java/com/aws/greengrass/builtin/services/telemetry/ComponentMetricIPCEventStreamAgent.java
```diff
@@ -0,0 +1,198 @@
...
+                        String.format("Invalid argument found in PutComponentMetricRequest"));
+            }
+        }
+    }
+
+    // Validate if request is authorized for given operation
+    // Also validate if serviceName is of format "aws.*"
+    private void doAuthorization(String opName, String serviceName) throws AuthorizationException {
+        if (authorizationHandler.isAuthorized(PUT_COMPONENT_METRIC_SERVICE_NAME,
+                Permission.builder().operation(opName).principal(serviceName).resource("*").build())
```
shaguptashaikh, line 191: Why is resource always *?

src/main/java/com/aws/greengrass/builtin/services/telemetry/ComponentMetricIPCEventStreamAgent.java
```diff
@@ -0,0 +1,198 @@
...
+        if (telemetryEnum == null || telemetryEnum.toString().isEmpty()) {
+            throw new IllegalArgumentException("Invalid telemetry unit: No matching TelemetryUnit type found");
+        }
+        return telemetryEnum;
+    }
+
+    // Validate metric request
+    // Check name, unit and value arguments
+    private void validateComponentMetricRequest(List<software.amazon.awssdk.aws.greengrass.model.Metric> metrics) {
+        if (metrics == null || metrics.isEmpty()) {
```
shaguptashaikh, line 173: Could use `Utils.isEmpty()`

src/main/java/com/aws/greengrass/builtin/services/telemetry/ComponentMetricIPCEventStreamAgent.java
```diff
@@ -0,0 +1,198 @@
...
+            throw new IllegalArgumentException(
+                    String.format("Null or Empty list of metrics found in PutComponentMetricRequest"));
+        }
+        for (software.amazon.awssdk.aws.greengrass.model.Metric metric : metrics) {
+            if (Utils.isEmpty(metric.getName()) || metric.getName().length() <= 0
```
shaguptashaikh, line 178: `Utils.isEmpty(metric.getName())` is sufficient here, same for the unit string

src/test/java/com/aws/greengrass/builtin/services/telemetry/ComponentMetricIPCEventStreamAgentTest.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    @AfterEach
+    void afterEach() {
+        pool.shutdownNow();
+    }
+
+
+    @Test
+    void GIVEN_put_component_metric_request_WHEN_handle_request_called_THEN_component_passes_authorization()
```
shaguptashaikh, line 77: Is only one test enough here? We should add more tests for other authorization cases and validation, emitmetrics etc

src/main/java/com/aws/greengrass/builtin/services/telemetry/ComponentMetricIPCEventStreamAgent.java
```diff
@@ -0,0 +1,197 @@
...
+            }
+        }
+    }
+
+    // Validate if request is authorized for given operation
+    // Also validate if serviceName is of format "aws.*"
+    private void doAuthorization(String opName, String serviceName) throws AuthorizationException {
+        if (authorizationHandler.isAuthorized(PUT_COMPONENT_METRIC_SERVICE_NAME,
+                Permission.builder().operation(opName).principal(serviceName).resource("*").build())
+                && SERVICE_NAME_REGEX.matcher(serviceName).find()) {
```
rbattle, line 191: nit: probably better to switch the order here so you check the name first and then check the authHandler. When a non aws service calls, it will log in the auth handler and customers could grant themselves the permissions and be confused as to why this is failing.

src/main/java/com/aws/greengrass/builtin/services/telemetry/ComponentMetricIPCEventStreamAgent.java
```diff
@@ -0,0 +1,197 @@
...
+        @SuppressWarnings({"PMD.PreserveStackTrace", "PMD.AvoidCatchingGenericException"})
+        @Override
+        public PutComponentMetricResponse handleRequest(PutComponentMetricRequest componentMetricRequest) {
+            return translateExceptions(() -> {
+                logger.atDebug().kv(SERVICE_NAME, serviceName)
+                        .log("Received putComponentMetricRequest from component " + serviceName);
+
+                // Authorize request for given operation
+                try {
+                    doAuthorization(this.getOperationModelContext().getOperationName(), serviceName);
```
rbattle, line 83: is the operation always going to be "putComponentMetric" here?

src/main/java/com/aws/greengrass/builtin/services/telemetry/ComponentMetricIPCEventStreamAgent.java
rbattle: I assume if the operation is putComponentMetric, the resource is the metric? and we allow all aws components to put data for any metric in the recipe?

If that is the case, then it could be useful to put the metric name in the resource instead of "*"

rbattle (approved):

src/main/java/com/aws/greengrass/builtin/services/telemetry/ComponentMetricIPCEventStreamAgent.java
nikkhilmuthye: Yes. We spoke about this. We will add that in the next PR.

nikkhilmuthye (approved):

src/main/java/com/aws/greengrass/builtin/services/telemetry/ComponentMetricIPCEventStreamAgent.java
kucnavya: It would always be `aws.greengrass#PutComponentMetric`

shaguptashaikh (approved):

PR #1243, opened 2022-06-28
fix: save deployment queue during nucleus shutdown; re-implement DeploymentQueue to handle de-duplication

indougnito:
**Issue #, if available:**

**Description of changes:**

This PR does three things, primarily:

1) modify shutdown logic in KernelLifecycle so the config transaction log is not closed until after service shutdown

2) since the config transaction log is now available during shutdown, update DeploymentService so that the deployment queue is saved to the config during shutdown, and then loaded from the config during startup.

3) re-implement DeploymentService so it handles de-duplication of deployments in the queue.

Without 1 & 2, it is possible to "lose" deployments in the queue because of a kernel shutdown (queued local deployments, in general; or, queued cloud deployments that were received while online, but where the kernel restarts in an offline state).

3 avoids duplication of queued deployments which might be introduced by 2 in certain scenarios.

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/deployment/DeploymentQueue.java
+++ b/src/main/java/com/aws/greengrass/deployment/DeploymentQueue.java
@@ -14,6 +14,7 @@ import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Queue;
 
 /**
@@ -93,11 +94,10 @@ public class DeploymentQueue {
         }
 
         // is the internal queue id already in use?
-        final String enqueuedDeploymentId = deploymentIdQueue.stream()
+        final Optional<String> enqueuedDeploymentId = deploymentIdQueue.stream()
                 .filter(enqueuedId -> enqueuedId.equals(offeredDeploymentInternalId))
-                .findAny()
-                .orElse(null);
-        if (enqueuedDeploymentId != null) {
+                .findAny();
+        if (enqueuedDeploymentId.isPresent()) {
             // internal queue id is already in use
             final Deployment enqueuedDeployment = deploymentMap.get(enqueuedDeploymentId);
             if (enqueuedDeployment == null) {
--- a/src/main/java/com/aws/greengrass/deployment/DeploymentQueue.java
+++ b/src/main/java/com/aws/greengrass/deployment/DeploymentQueue.java
@@ -6,76 +6,162 @@
 package com.aws.greengrass.deployment;
 
 import com.aws.greengrass.deployment.model.Deployment;
-import com.aws.greengrass.deployment.model.Deployment.DeploymentType;
 import com.aws.greengrass.logging.api.Logger;
 import com.aws.greengrass.logging.impl.LogManager;
 
-import java.util.Iterator;
-import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.Queue;
 
+/**
+ * <p>DeploymentQueue is a thread-safe deployment queue that automatically de-duplicates by deployment id, and
+ * also de-duplicates shadow deployments, i.e. there can be at-most-one shadow deployment enqueued.</p>
+ *
+ * <p>DeploymentQueue is implemented internally as a deployment id queue, along with a map of id -> deployment.</p>
+ *
+ * <p>When an offered deployment has a unique deployment id, it is enqueued normally.</p>
+ *
+ * <p>When an offered deployment has the same deployment id as an already-enqueued element:
+ *     - if the offered deployment meets replacement criteria, then the enqueued element is replaced by the offered
+ *       element, preserving queue order.
+ *     - otherwise, the offered deployment is ignored.</p>
+ *
+ * <p>Replacement criteria are as follows:
+ *     - offered.getDeploymentStage().getPriority() > enqueued.getDeploymentStage().getPriority(), OR
+ *     - offered.isCancelled() == true, OR
+ *     - offered.getDeploymentType().equals(SHADOW)</p>
+ *
+ * <p>When an offered deployment is a shadow deployment:
+ *     - if there is already a shadow deployment in the queue, then replace it and preserve queue order.
+ *     - otherwise, enqueue normally.</p>
+ */
 public class DeploymentQueue {
 
+    /**
+     * For the internal queue, shadow deployments use a special queue id. Effectively, this means there can be at most
+     * one shadow deployment in the queue.
+     */
+    private static final String SHADOW_DEPLOYMENT_QUEUE_ID = Deployment.DeploymentType.SHADOW.name();
+
+    private static final String DEPLOYMENT_ID_LOG_KEY = "DeploymentId";
+    private static final String DISCARDED_DEPLOYMENT_ID_LOG_KEY = "DiscardedDeploymentId";
     private static final Logger logger = LogManager.getLogger(DeploymentQueue.class);
-    private final ConcurrentLinkedQueue<Deployment> deploymentsQueue = new ConcurrentLinkedQueue();
 
     /**
-     * Add a deployment to the queue.
-     *
-     * @param deployment deployment
+     * Internal queue of deployment id's. For shadow deployments, the internal queue id is not the same as the
+     * actual deployment id.
      */
-    public synchronized boolean offer(Deployment deployment) {
-        //For shadow deployment when desired state is reverted, it can result in scheduling a deployment which is
-        // same as
-        if (!DeploymentType.SHADOW.equals(deployment.getDeploymentType()) && deploymentsQueue.contains(deployment)) {
+    private final Queue<String> deploymentIdQueue = new LinkedList<>();
+
+    /**
+     * Map of internal queue id -> deployment instance.
+     */
+    private final Map<String, Deployment> deploymentMap = new HashMap<>();
+
+    /**
+     * <p>If the offered deployment id is unique, then insert the offered deployment at the tail of the queue.</p>
+     *
+     * <p>When an offered deployment has the same deployment id as an already-enqueued element:
+     *     - if the offered deployment meets replacement criteria, then the enqueued element is replaced by the offered
+     *       element, preserving queue order.
+     *     - otherwise, the offered deployment is ignored.</p>
+     *
+     * <p>Replacement criteria are as follows:
+     *     - offered.getDeploymentStage().getPriority() > enqueued.getDeploymentStage().getPriority(), OR
+     *     - offered.isCancelled() == true, OR
+     *     - offered.getDeploymentType().equals(SHADOW)</p>
+     *
+     * <p>When an offered deployment is a shadow deployment:
+     *     - if there is already a shadow deployment in the queue, then replace it and preserve queue order.
+     *     - otherwise, enqueue normally.</p>
+     *
+     * @param offeredDeployment the offered deployment instance.
+     * @return true if the queue was modified, otherwise false.
+     */
+    public synchronized boolean offer(Deployment offeredDeployment) {
+        if (offeredDeployment == null) {
             return false;
         }
-        return deploymentsQueue.offer(deployment);
-    }
-
-    /**
-     * Get the next deployment to be deployed.
-     *
-     * @return deployment
-     */
-    public synchronized Deployment peek() {
-        Deployment deployment = deploymentsQueue.peek();
-        // Discarding is not done at schedule time because the DeploymentService does not remove the deployments
-        // atomically DeploymentService first peeks and determine if the next deployment is actionable.
-        while (deployment != null && canDeploymentBeDiscarded(deployment)) {
-            logger.atInfo().kv("DEPLOYMENT_ID", deployment.getId())
-                    .kv("DEPLOYMENT_TYPE", deployment.getDeploymentType())
-                    .log("Discarding device deployment");
-            deploymentsQueue.remove();
-            deployment = deploymentsQueue.peek();
+        final String offeredDeploymentInternalId; // the id to employ in the internal queue
+        if (Deployment.DeploymentType.SHADOW.equals(offeredDeployment.getDeploymentType())) {
+            offeredDeploymentInternalId = SHADOW_DEPLOYMENT_QUEUE_ID;
+        } else {
+            offeredDeploymentInternalId = offeredDeployment.getId();
         }
-        return deployment;
-    }
 
-    /**
-     * Removed the deployment from the head of the queue.
-     */
-    public synchronized void remove() {
-        deploymentsQueue.remove();
-    }
-
-    public boolean isEmpty() {
-        return deploymentsQueue.isEmpty();
-    }
-
-    private boolean canDeploymentBeDiscarded(Deployment selectedDeployment) {
-        // If the selected deployment is of type shadow and there is another deployment in the queue
-        // the selected deployment can be discarded. ShadowDeploymentListener ensures that shadow deployments are
-        // queued based on the order in which they are created in the cloud.
-        if (DeploymentType.SHADOW.equals(selectedDeployment.getDeploymentType())) {
-            Iterator<Deployment> iterator = deploymentsQueue.iterator();
-            while (iterator.hasNext()) {
-                Deployment nextDeployment = iterator.next();
-                if (!selectedDeployment.equals(nextDeployment)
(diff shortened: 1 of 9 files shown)


MikeDombo:

indougnito:
fix line length for checkstyle

indougnito:
fix mock config in unit test

src/main/java/com/aws/greengrass/deployment/DeploymentService.java
```diff
@@ -272,9 +275,59 @@ private Long getPollingFrequency(Topic pollingFrequencyTopic) {
 
     @Override
     protected void shutdown() {
+        saveDeploymentQueueToConfig(); // Save any deployments in queue for next startup
         receivedShutdown.set(true);
     }
 
+    @SuppressWarnings("PMD.AvoidCatchingGenericException")
+    private void saveDeploymentQueueToConfig() {
```
rbattle, line 283: nit: use `persist` to be consistent with `persistGroupToRootComponents`, `persistAndPublishDeploymentStatus`, `persistLastSuccessfulDeployment`, etc...

src/main/java/com/aws/greengrass/deployment/DeploymentService.java
```diff
@@ -272,9 +275,59 @@ private Long getPollingFrequency(Topic pollingFrequencyTopic) {
 
     @Override
     protected void shutdown() {
+        saveDeploymentQueueToConfig(); // Save any deployments in queue for next startup
         receivedShutdown.set(true);
```
rbattle, line 279: receivedShutdown controls the loop in `startup` - should you save after it is set to `true` so that loop isn't trying to process a deployment while you are trying to save the queue?

src/main/java/com/aws/greengrass/deployment/DeploymentService.java
indougnito: agree

shaguptashaikh:
Since the job and shadow subscribers get unfinished deployments after startup are there likely to be duplicate deployments getting added to deployment queue after this change?

src/main/java/com/aws/greengrass/deployment/DeploymentService.java
indougnito: I was actually concerned that setting `receivedShutdown` to `true` would trigger `finishCurrentDeployment()` and set `currentDeploymentTaskMetadata = null` before I had a chance to save it to the config.

indougnito:
> Since the job and shadow subscribers get unfinished deployments after startup are there likely to be duplicate deployments getting added to deployment queue after this change?

Yes, but I observe that we have de-duplication logic for jobs & shadow deployments:

https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/248ca457056db5d3a8b14a9e582c6ff3a4c75774/src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java#L438

https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/248ca457056db5d3a8b14a9e582c6ff3a4c75774/src/main/java/com/aws/greengrass/deployment/IotJobsHelper.java#L659

src/main/java/com/aws/greengrass/deployment/DeploymentService.java
indougnito: So no, I don't think so.

The queue saving logic wants `currentDeploymentTaskMetadata.getDeployment()`. The deployment processing loop only modifies `currentDeploymentTaskMetadata` here, in `createNewDeployment()`:

https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/b273fa79d962fc9aedcf0abccbef01d540662b24/src/main/java/com/aws/greengrass/deployment/DeploymentService.java#L502

`createNewDeployment()` only gets invoked when `currentDeploymentTaskMetadata` is `null`, which happens when `finishCurrentDeployment()` is invoked.

indougnito:
Renaming `saveDeploymentQueueToConfig()` to `persistDeploymentQueueToConfig()`.

src/main/java/com/aws/greengrass/deployment/DeploymentService.java
rbattle: how would setting it to true trigger the finishCurrentDeployment condition?

junfuchen99:
Just thinking out loud here: currently we persist deployment document/metadata in `/deployment/ongoing`, which is not modified only until the end of `DeploymentService::finishCurrentDeployment`, so it should remain intact if deployment is interrupted by kernel shutdown.

When kernel launches, if it's in kernel activation or rollback stage, we reads from ongoing deployment directory to resume deployment
https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/main/src/main/java/com/aws/greengrass/lifecyclemanager/Kernel.java#L222-L230.

An alternative solution from saving to kernel config: we could do something similar for default stage as well at kernel launch - read metadata and offer the deployment.

indougnito:
Added a `finally` clause to the `try..catch` in `loadDeploymentQueueFromConfig()` so that the config value is always cleared after loading.

src/main/java/com/aws/greengrass/deployment/DeploymentService.java
indougnito: ah, i misread. `receivedShutdown==true` effectively just stops the loop, which is fine.

i'll swap the order of loading/setting the flag.

indougnito:
Moved the invocation of `persistDeploymentQueueToConfig()` after `receivedShutdown.set(true)`.

indougnito:
> Just thinking out loud here: currently we persist deployment document/metadata in `/deployment/ongoing`, which is not modified only until the end of `DeploymentService::finishCurrentDeployment`, so it should remain intact if deployment is interrupted by kernel shutdown.
> 
> When kernel launches, if it's in kernel activation or rollback stage, we reads from ongoing deployment directory to resume deployment https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/main/src/main/java/com/aws/greengrass/lifecyclemanager/Kernel.java#L222-L230.
> 
> An alternative solution from saving to kernel config: we could do something similar for default stage as well at kernel launch - read metadata and offer the deployment.

Wouldn't that solution fail if there were >= 2 deployments in queue?

rbattle (approved):

junfuchen99:
> > Just thinking out loud here: currently we persist deployment document/metadata in `/deployment/ongoing`, which is not modified only until the end of `DeploymentService::finishCurrentDeployment`, so it should remain intact if deployment is interrupted by kernel shutdown.
> > When kernel launches, if it's in kernel activation or rollback stage, we reads from ongoing deployment directory to resume deployment https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/main/src/main/java/com/aws/greengrass/lifecyclemanager/Kernel.java#L222-L230.
> > An alternative solution from saving to kernel config: we could do something similar for default stage as well at kernel launch - read metadata and offer the deployment.
> 
> Wouldn't that solution fail if there were >= 2 deployments in queue?

Thanks for explaining! Yea I think it there were >=2 deployments in queue, it would make sense to save all of them in config. A follow-up question would be, if there were >=2 deployments in queue, let's say {A, B}. If when executing deployment A, deployment requires a nucleus bootstrap, then after restart and kernel launch, would the queue become {A,B,A}? Since we first initializes from saved config and then reads from ongoing deployment directory. It may lead to A being executed twice when we add functionality of retry deployment.

indougnito:
> > > Just thinking out loud here: currently we persist deployment document/metadata in `/deployment/ongoing`, which is not modified only until the end of `DeploymentService::finishCurrentDeployment`, so it should remain intact if deployment is interrupted by kernel shutdown.
> > > When kernel launches, if it's in kernel activation or rollback stage, we reads from ongoing deployment directory to resume deployment https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/main/src/main/java/com/aws/greengrass/lifecyclemanager/Kernel.java#L222-L230.
> > > An alternative solution from saving to kernel config: we could do something similar for default stage as well at kernel launch - read metadata and offer the deployment.
> > 
> > 
> > Wouldn't that solution fail if there were >= 2 deployments in queue?
> 
> Thanks for explaining! Yea I think it there were >=2 deployments in queue, it would make sense to save all of them in config. A follow-up question would be, if there were >=2 deployments in queue, let's say {A, B}. If when executing deployment A, deployment requires a nucleus bootstrap, then after restart and kernel launch, would the queue become {A,B,A}? Since we first initializes from saved config and then reads from ongoing deployment directory. It may lead to A being executed twice when we add functionality of retry deployment.

As discussed offline, we need to arrange things so that any agent offering into the DeploymentQueue performs the necessary de-duplication (and in the case of bootstrap deployments, they need to replace matching OTA deployments in the queue).

I will look into consolidating our de-dupe/replacement logic into the DeploymentQueue class itself, which will allow us to delete a lot of our existing de-dupe logic (we currently have it implemented separately for IoT Jobs & Shadow deployments).

indougnito:
Re-implemented DeploymentQueue to handle de-duplication; added unit tests.

indougnito:
Shortened commit message

indougnito:
Need to look more closely at de-dupe for shadow deployments.

If multiple shadow deployments are in queue, then only the shadow deployment closest to the tail (most recently enqueued) should be actionable.  So, maybe after enqueue, if the enqueued is type==shadow, we then iterate through the queue and remove other shadow deployments.

src/main/java/com/aws/greengrass/deployment/DeploymentQueue.java
```diff
@@ -6,76 +6,162 @@
...
-     * Add a deployment to the queue.
+     * Internal queue of deployment id's. For shadow deployments, the internal queue id is not the same as the
+     * actual deployment id.
+     */
+    private final Queue<String> deploymentIdQueue = new LinkedList<>();
+
+    /**
+     * Map of internal queue id -> deployment instance.
+     */
+    private final Map<String, Deployment> deploymentMap = new HashMap<>();
```
rbattle, line 62: could you just use a LinkedHashMap here instead of maintaining two data structures?

src/main/java/com/aws/greengrass/deployment/DeploymentQueue.java
```diff
@@ -6,76 +6,162 @@
...
+     * @return true if the queue was modified, otherwise false.
      */
-    public synchronized boolean offer(Deployment deployment) {
-        //For shadow deployment when desired state is reverted, it can result in scheduling a deployment which is
-        // same as
-        if (!DeploymentType.SHADOW.equals(deployment.getDeploymentType()) && deploymentsQueue.contains(deployment)) {
+    public synchronized boolean offer(Deployment offeredDeployment) {
+        if (offeredDeployment == null) {
+            return false;
+        }
```
rbattle, line 87: nit: Queue's generally throw NPE if you offer null, or if they allow null as a value they'll process it...any reason to not throw NPE here?

Seems like we should check the deployment is not null before doing anything in the queue, rather than silently allowing usage of it (even if we report the queue is not modified)

src/main/java/com/aws/greengrass/deployment/DeploymentQueue.java
```diff
@@ -6,76 +6,162 @@
...
+        } else {
+            offeredDeploymentInternalId = offeredDeployment.getId();
+        }
+
+        // is the internal queue id already in use?
+        final String enqueuedDeploymentId = deploymentIdQueue.stream()
+                .filter(enqueuedId -> enqueuedId.equals(offeredDeploymentInternalId))
+                .findAny()
+                .orElse(null);
+        if (enqueuedDeploymentId != null) {
```
rbattle, line 100: this seems to defeat a benefit of using an Optional - avoiding nulls

It's perhaps more readable if you made this:

```
Optional<String> enqueuedDeploymentId = ...
if (enqueudDeploymentId.isPresent()) {

}
```

src/main/java/com/aws/greengrass/deployment/DeploymentQueue.java
```diff
@@ -6,76 +6,162 @@
...
+            // internal queue id is already in use
+            final Deployment enqueuedDeployment = deploymentMap.get(enqueuedDeploymentId);
+            if (enqueuedDeployment == null) {
+                logger.atError().kv(DEPLOYMENT_ID_LOG_KEY, offeredDeployment.getId())
+                        .kv("InternalQueueId", offeredDeploymentInternalId)
+                        .log("Logic error: internal queue contains id with no corresponding deployment in map");
+                return false;
+            }
```
rbattle, line 108: I think you could avoid this condition if you used a LinkedHashMap

src/main/java/com/aws/greengrass/deployment/DeploymentQueue.java
indougnito: I thought about it, but didn't like that LinkedHashMap doesn't provide a pop() mechanism like Queue. And I didn't really like the idea of iterating over all elements every time I wanted the last item.

src/main/java/com/aws/greengrass/deployment/DeploymentQueue.java
indougnito: I was actually inspired by the Queue javadoc here (https://docs.oracle.com/javase/7/docs/api/java/util/Queue.html), which states that, by convention, `offer(..)` methods return a special value on failure, whereas `add(..)` methods throw an exception.

src/main/java/com/aws/greengrass/deployment/DeploymentQueue.java
indougnito: Whoops, completely agree :)

src/main/java/com/aws/greengrass/deployment/DeploymentQueue.java
rbattle: https://docs.oracle.com/javase/7/docs/api/java/util/Queue.html#offer(E) throws NPE

the special value is intended to be used when offer would violate constraints (like if you had a fixed size)

I'd say that offering null is not a 'normal' failure case

src/main/java/com/aws/greengrass/deployment/DeploymentQueue.java
indougnito: Yeah, although this nullcheck is purely defensive.  Should never get null unless somebody introduces a bug later.

I like this better than iterating over all LinkedHashMap elements every time I want the last item...

src/main/java/com/aws/greengrass/deployment/DeploymentQueue.java
indougnito: Touche, I now agree

src/main/java/com/aws/greengrass/deployment/DeploymentQueue.java
indougnito: I am evidently mistaken about the LinkedHashMap iteration ordering - LinkedHashMap would be better here.

src/main/java/com/aws/greengrass/deployment/DeploymentQueue.java
indougnito: I was mistaken about LinkedHashMap ordering; going to swap one in

rbattle (approved):

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
```diff
@@ -462,12 +463,13 @@ public void stopAllServices(int timeoutSeconds) {
      * @param timeoutSeconds Timeout in seconds
      */
     public void softShutdown(int timeoutSeconds) {
-        logger.atDebug("system-shutdown").log("Start soft shutdown");
+        logger.atDebug(SYSTEM_SHUTDOWN_EVENT).log("Start soft shutdown");
+        stopAllServices(timeoutSeconds);
```
junfuchen99, line 467: Is it possible for switching the order of `stopAllServices` and `kernel.getContext().waitForPublishQueueToClear();`, `close(tlog);`, `kernel.writeEffectiveConfig();` break any existing workflow?

indougnito: We shouldn't close the tlog until services are stopped (since they might be writing to the config).

Does it matter if `waitForPublishQueueToClear()` happens before or after services are stopped?

indougnito: Going to move waitForPublishQueueToClear() before stopAllServices(), which restores the original relative order, just to be safe.

junfuchen99 (approved):

kucnavya (approved):

rbattle (approved):

PR #1249, opened 2022-06-29
fix: shutdown before network with systemd

### LINKED ISSUE #1248: System shutdown is delayed up to 90 seconds by greengrass service
**Describe the bug**
The greengrass systemd service needs up to 90 seconds to stop during the system shutdown. See screenshot below.
The greengrass.log file then shows several errors `com.aws.greengrass.mqttclient.MqttClient: Error unsubscribing`. See log file below.
Reason is that the system network may already have been shut down before the greengrass service is stopped.

**To Reproduce**
Shut down an Ubuntu system with the greengrass service installed by the greengrass installer.

**Expected behavior**
Fast and clean stop of the greengrass service at system shutdown.

**Actual behavior**
Delayed stop of the greengrass service at system shutdown until greengrass' MQTT/network timeouts or the systemd timeout run out.

**Environment**
 - OS: Ubuntu 20.04
 - JDK version: 11
 - Nucleus version: 2.5.5

**Additional context**

In my eyes the solution is to add the dependency `After=network.target` to the greengrass service definition. This will prevent stopping the network services before the greengrass service. See [systemd Network configuration synchronization points](https://systemd.io/NETWORK_ONLINE/).

I'll soon send a pull request for this suggested change.

The bug is not critical but unnecessarily prolongs system restarts.

Screenshot of shutdown logs:  
greengrass-shutdown-delay-screenshot

greengrass.log excerpt:
[greengrass.log](https://github.com/aws-greengrass/aws-greengrass-nucleus/files/9009356/greengrass.log).

grillbaer:
**Issue #1248**

**Description of changes:**

Added dependency `After=network.target` to greengrass service definition. 
According to the reference docs in [systemd Network configuration synchronization points](https://systemd.io/NETWORK_ONLINE/), this will ensure that the greengrass service is stopped before the network is stopped during system shutdown.

I did not add Wants and After dependencies for `network-online.target` because this would also influence the system startup and might delay starting of greengrass when the network is down for external reasons like unplugged cable etc.

To fix #1248.

=> Manually tested with greengrass installed in a Ubuntu 20.04 VM and performing several system shutdowns.

**Documentation Checklist:**

**Compatibility Checklist:**


shaguptashaikh (approved):

junfuchen99:
Hi @grillbaer, thank you so much for your input! One edge case I want to bring up - is it possible for `After=network.target` to create a risk where greengrass service is no longer able to start up in a complete offline environment?

I'm reading https://www.freedesktop.org/wiki/Software/systemd/NetworkTarget/ and I see that `network.target` only signifies if the network management software is up, so such scenario would only exist if the network management service is disabled, which I assume is very unlikely.

grillbaer:
Hi @junfuchen99, i did some testing (on Ubuntu 20.04) and investigation regarding the risk of Greengrass not starting up without network.

When the network is disconnected (unplugged) during system startup and thus the network interface does not get online (no IP address from DHCP), the startup of greengrass and a bunch of other services (like snapd) are delayed by 120 seconds. After 120 seconds systemd gives up waiting for the network to get online.

Important: this behavior is _independent_ of `After=network.target`. This is how it already works today without the change. I tested both cases.

I think the reason for the startup delay without network is that greengrass and many other service starts are triggered by `WantedBy=multi-user.target` which itself is not started before the network online status is reached, either successfully with status network online or failing after a 120 seconds timeout.

=> So i think `After=network.target` does not change anything of the current Greengrass startup behavior, but brings in the desired shutdown sequence.

junfuchen99 (approved):
Thanks for the follow-up!

PR #1258, opened 2022-07-12
feat: allow StreamManager to use PutComponentMetric always

nikkhilmuthye:
**Issue #, if available:**

**Description of changes:**
Allow list Stream Manager to publish telemetry metrics always. 

Simply adding the ACL in the Stream Manager recipe will not work if the customer already has Stream Manager running on their device and they upgrade. We need a way for publishing the telemetry metrics from Stream Manager when the customer upgrades it without having the customers to reset the configuration.

**Documentation Checklist:**

**Compatibility Checklist:**


src/main/java/com/aws/greengrass/builtin/services/telemetry/ComponentMetricIPCEventStreamAgent.java
```diff
@@ -39,6 +40,8 @@
 import static com.aws.greengrass.ipc.modules.ComponentMetricIPCService.PUT_COMPONENT_METRIC_SERVICE_NAME;
 
 public class ComponentMetricIPCEventStreamAgent {
+    private static final String STREAM_MANAGER_SERVICE_NAME = "aws.greengrass.StreamManager";
+    private static final List<String> AUTHORIZED_COMPONENTS = Collections.singletonList(STREAM_MANAGER_SERVICE_NAME);
```
aggaayus, line 44: We need to tag this line with the SIM that we have created for revisiting defaultConfiguration behavior. I like the idea of keeping this a list, but do we expect its size to increase ?

nikkhilmuthye: This is a common practice. We have done that in other places as well.

rbattle: +1 on adding a comment here

rbattle (approved):

PR #1260, opened 2022-07-13
feat: add a new API for getting the cert chain using security service

saranyailla:
**Issue #, if available:**

**Description of changes:**
- CryptoKey provider interface is updated with the addition of a new API for getting the certificate chain. 
- Adds a new API in SecurityService to get certificate chain using private key and certificate URIs by using the newly updated interface. However, this change is also compatible with older implementations of the providers as certificate chain is obtained by using key manager from the existing APIs. 

- This change is needed for enabling BYOCA feature where customers can provide private key and cert of the CA through file system or HSM. 

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/security/SecurityService.java
+++ b/src/main/java/com/aws/greengrass/security/SecurityService.java
@@ -183,22 +183,26 @@ public final class SecurityService {
         CryptoKeySpi provider = selectCryptoKeyProvider(privateKeyUri);
         KeyManager[] km = provider.getKeyManagers(privateKeyUri, certificateUri);
 
-        if (km.length == 1 && km[0] instanceof X509KeyManager) {
-            X509KeyManager x509KeyManager = (X509KeyManager) km[0];
-            KeyPair keyPair = provider.getKeyPair(privateKeyUri, certificateUri);
-            String[] aliases = x509KeyManager.getClientAliases(keyPair.getPublic().getAlgorithm(), null);
-            if (aliases == null) {
-                throw new CertificateChainLoadingException("Unable to find aliases in the key manager with the given "
-                        + "private key and certificate URIs");
-            }
-            for (String alias : aliases) {
-                if (x509KeyManager.getPrivateKey(alias).equals(keyPair.getPrivate())) {
-                    return x509KeyManager.getCertificateChain(alias);
-                }
+        if (km.length != 1 || !(km[0] instanceof X509KeyManager)) {
+            throw new CertificateChainLoadingException("Unable to find the X509 key manager instance to get the "
+                    + "certificate chain using the private key and certificate URIs");
+        }
+
+        X509KeyManager x509KeyManager = (X509KeyManager) km[0];
+        KeyPair keyPair = provider.getKeyPair(privateKeyUri, certificateUri);
+        String[] aliases = x509KeyManager.getClientAliases(keyPair.getPublic().getAlgorithm(), null);
+        if (aliases == null) {
+            throw new CertificateChainLoadingException("Unable to find aliases in the key manager with the given "
+                    + "private key and certificate URIs");
+        }
+        for (String alias : aliases) {
+            if (x509KeyManager.getPrivateKey(alias).equals(keyPair.getPrivate())) {
+                return x509KeyManager.getCertificateChain(alias);
             }
         }
-        throw new CertificateChainLoadingException("Unable to get certificate chain from the provider using X509 "
-                + "key manager");
+
+        throw new CertificateChainLoadingException("Unable to get the certificate chain using the private key and "
+                + "certificate URIs");
     }
 
 
--- a/src/main/java/com/aws/greengrass/security/SecurityService.java
+++ b/src/main/java/com/aws/greengrass/security/SecurityService.java
@@ -42,12 +42,14 @@ import java.util.concurrent.ConcurrentMap;
 import javax.inject.Inject;
 import javax.net.ssl.KeyManager;
 import javax.net.ssl.KeyManagerFactory;
+import javax.net.ssl.X509KeyManager;
 
 public final class SecurityService {
     private static final Logger logger = LogManager.getLogger(SecurityService.class);
     private static final String KEY_TYPE = "keyType";
     private static final String KEY_URI = "keyUri";
     private static final String CERT_URI = "certificateUri";
+    private static final String KEY_ALIAS = "private-key";
 
     // retry 3 times with exponential backoff, start with 200ms,
     // if service still not available, pop exception to the caller
@@ -149,11 +151,11 @@ public final class SecurityService {
     /**
      * Get JSSE KeyManagers, used for https TLS handshake.
      *
-     * @param privateKeyUri private key URI
+     * @param privateKeyUri  private key URI
      * @param certificateUri certificate URI
      * @return KeyManagers that manage the specified private key
      * @throws ServiceUnavailableException if crypto key provider service is unavailable
-     * @throws KeyLoadingException if crypto key provider service fails to load key
+     * @throws KeyLoadingException         if crypto key provider service fails to load key
      */
     public KeyManager[] getKeyManagers(URI privateKeyUri, URI certificateUri)
             throws ServiceUnavailableException, KeyLoadingException {
@@ -162,6 +164,47 @@ public final class SecurityService {
         CryptoKeySpi provider = selectCryptoKeyProvider(privateKeyUri);
         return provider.getKeyManagers(privateKeyUri, certificateUri);
     }
+
+    /**
+     * Get certificate chain using private and certificate URIs.
+     *
+     * @param privateKeyUri  private key URI
+     * @param certificateUri certificate URI
+     * @return X509Certificate list that corresponds to the given private key and certificate URIs
+     * @throws ServiceUnavailableException if crypto key provider service is unavailable
+     * @throws KeyLoadingException         if crypto key provider service fails to load key
+     */
+    public X509Certificate[] getCertificateChain(URI privateKeyUri, URI certificateUri)
+            throws ServiceUnavailableException, KeyLoadingException {
+        logger.atTrace().kv(KEY_URI, privateKeyUri).kv(CERT_URI, certificateUri)
+                .log("Getting the certificate chain using private key and certificate URIs");
+
+        X509Certificate[] certChain = new X509Certificate[0];
+        CryptoKeySpi provider = selectCryptoKeyProvider(privateKeyUri);
+        KeyManager[] km = provider.getKeyManagers(privateKeyUri, certificateUri);
+
+        if (km.length == 1 && km[0] instanceof X509KeyManager) {
+            X509KeyManager x509KeyManager = (X509KeyManager) km[0];
+            KeyPair keyPair = provider.getKeyPair(privateKeyUri, certificateUri);
+            String[] aliases = x509KeyManager.getClientAliases(keyPair.getPublic().getAlgorithm(), null);
+            logger.atTrace().kv(KEY_URI, privateKeyUri).kv(CERT_URI, certificateUri)
+                    .log("Getting the certificate chain from the provider using the key manager");
+            if (aliases == null) {
+                logger.atError().kv(KEY_URI, privateKeyUri).kv(CERT_URI, certificateUri)
+                        .log("Unable to find aliases in the key manager.");
+                return certChain;
+            }
+            for (String alias : aliases) {
+                if (x509KeyManager.getPrivateKey(alias).equals(keyPair.getPrivate())) {
+                    return x509KeyManager.getCertificateChain(alias);
+                }
+            }
+        }
+        logger.atError().kv(KEY_URI, privateKeyUri).kv(CERT_URI, certificateUri)
+                .log("Unable to get the certificate chain using private key and certificate URIs.");
+        return certChain;
+    }
+
 
     /**
      * Get JCA KeyPair, used for Secret manager.
@@ -312,7 +355,7 @@ public final class SecurityService {
 
                 KeyStore keyStore = KeyStore.getInstance("PKCS12");
                 keyStore.load(null);
-                keyStore.setKeyEntry("private-key", privateKey, null, certificateChain.toArray(new Certificate[0]));
+                keyStore.setKeyEntry(KEY_ALIAS, privateKey, null, certificateChain.toArray(new Certificate[0]));
 
                 KeyManagerFactory keyManagerFactory =
                         KeyManagerFactory.getInstance(KeyManagerFactory.getDefaultAlgorithm());
--- a/src/main/java/com/aws/greengrass/security/SecurityService.java
+++ b/src/main/java/com/aws/greengrass/security/SecurityService.java
@@ -188,10 +188,8 @@ public final class SecurityService {
             KeyPair keyPair = provider.getKeyPair(privateKeyUri, certificateUri);
             String[] aliases = x509KeyManager.getClientAliases(keyPair.getPublic().getAlgorithm(), null);
             if (aliases == null) {
-                logger.atError().kv(KEY_URI, privateKeyUri).kv(CERT_URI, certificateUri)
-                        .log("Unable to find aliases in the key manager.");
-                throw new CertificateChainLoadingException("Unable to get the certificate chain from the provider "
-                        + "using the key manager");
+                throw new CertificateChainLoadingException("Unable to find aliases in the key manager with the given "
+                        + "private key and certificate URIs");
             }
             for (String alias : aliases) {
                 if (x509KeyManager.getPrivateKey(alias).equals(keyPair.getPrivate())) {
@@ -199,9 +197,7 @@ public final class SecurityService {
                 }
             }
         }
-        logger.atError().kv(KEY_URI, privateKeyUri).kv(CERT_URI, certificateUri)
-                .log("Unable to get the certificate chain using private key and certificate URIs");
-        throw new CertificateChainLoadingException("Unable to get the certificate chain from the provider using the "
+        throw new CertificateChainLoadingException("Unable to get certificate chain from the provider using X509 "
                 + "key manager");
     }
 
--- a/src/main/java/com/aws/greengrass/security/SecurityService.java
+++ b/src/main/java/com/aws/greengrass/security/SecurityService.java
@@ -9,6 +9,7 @@ import com.aws.greengrass.config.CaseInsensitiveString;
 import com.aws.greengrass.deployment.DeviceConfiguration;
 import com.aws.greengrass.logging.api.Logger;
 import com.aws.greengrass.logging.impl.LogManager;
+import com.aws.greengrass.security.exceptions.CertificateChainLoadingException;
 import com.aws.greengrass.security.exceptions.KeyLoadingException;
 import com.aws.greengrass.security.exceptions.MqttConnectionProviderException;
 import com.aws.greengrass.security.exceptions.ServiceProviderConflictException;
@@ -171,10 +172,12 @@ public final class SecurityService {
      * @param privateKeyUri  private key URI
      * @param certificateUri certificate URI
      * @return X509Certificate list that corresponds to the given private key and certificate URIs
-     * @throws Exception if crypto key provider service is unavailable or fails to load key
+     * @throws ServiceUnavailableException      if crypto key provider service is unavailable
+     * @throws KeyLoadingException              if crypto key provider service fails to load key
+     * @throws CertificateChainLoadingException if no certificate chain is found
      */
-    @SuppressWarnings("PMD.SignatureDeclareThrowsException")
-    public X509Certificate[] getCertificateChain(URI privateKeyUri, URI certificateUri) throws Exception {
+    public X509Certificate[] getCertificateChain(URI privateKeyUri, URI certificateUri)
+            throws CertificateChainLoadingException, ServiceUnavailableException, KeyLoadingException {
         logger.atTrace().kv(KEY_URI, privateKeyUri).kv(CERT_URI, certificateUri)
                 .log("Getting the certificate chain using private key and certificate URIs");
         CryptoKeySpi provider = selectCryptoKeyProvider(privateKeyUri);
@@ -187,7 +190,8 @@ public final class SecurityService {
             if (aliases == null) {
                 logger.atError().kv(KEY_URI, privateKeyUri).kv(CERT_URI, certificateUri)
                         .log("Unable to find aliases in the key manager.");
-                throw new Exception("Unable to get the certificate chain from the provider using the key manager");
+                throw new CertificateChainLoadingException("Unable to get the certificate chain from the provider "
+                        + "using the key manager");
             }
             for (String alias : aliases) {
                 if (x509KeyManager.getPrivateKey(alias).equals(keyPair.getPrivate())) {
@@ -197,7 +201,8 @@ public final class SecurityService {
         }
         logger.atError().kv(KEY_URI, privateKeyUri).kv(CERT_URI, certificateUri)
                 .log("Unable to get the certificate chain using private key and certificate URIs");
-        throw new Exception("Unable to get the certificate chain from the provider using the key manager");
+        throw new CertificateChainLoadingException("Unable to get the certificate chain from the provider using the "
+                + "key manager");
     }
 
 
--- a/src/main/java/com/aws/greengrass/security/SecurityService.java
+++ b/src/main/java/com/aws/greengrass/security/SecurityService.java
@@ -171,15 +171,12 @@ public final class SecurityService {
      * @param privateKeyUri  private key URI
      * @param certificateUri certificate URI
      * @return X509Certificate list that corresponds to the given private key and certificate URIs
-     * @throws ServiceUnavailableException if crypto key provider service is unavailable
-     * @throws KeyLoadingException         if crypto key provider service fails to load key
+     * @throws Exception if crypto key provider service is unavailable or fails to load key
      */
-    public X509Certificate[] getCertificateChain(URI privateKeyUri, URI certificateUri)
-            throws ServiceUnavailableException, KeyLoadingException {
+    @SuppressWarnings("PMD.SignatureDeclareThrowsException")
+    public X509Certificate[] getCertificateChain(URI privateKeyUri, URI certificateUri) throws Exception {
         logger.atTrace().kv(KEY_URI, privateKeyUri).kv(CERT_URI, certificateUri)
                 .log("Getting the certificate chain using private key and certificate URIs");
-
-        X509Certificate[] certChain = new X509Certificate[0];
         CryptoKeySpi provider = selectCryptoKeyProvider(privateKeyUri);
         KeyManager[] km = provider.getKeyManagers(privateKeyUri, certificateUri);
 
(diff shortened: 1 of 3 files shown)


MikeDombo (commented):
Why is this required? In the design discussion we mentioned that the SecurityService, and particularly the implementations (PKCS11 provider) should not be changed as we don't have a well defined upgrade path.

src/main/java/com/aws/greengrass/security/CryptoKeySpi.java
```diff
@@ -20,4 +20,8 @@ KeyManager[] getKeyManagers(URI privateKeyUri, URI certificateUri)
     KeyPair getKeyPair(URI privateKeyUri, URI certificateUri) throws ServiceUnavailableException, KeyLoadingException;
 
     String supportedKeyType();
+
+    default String getKeyAlias(URI privateKeyUri, URI certificateUri) {
```
MikeDombo, line 24: How about implementing what I talked with you about last night? getCertificate.

Since I think we agreed that we don't care about the alias. All we want is the certificate.

src/main/java/com/aws/greengrass/security/SecurityService.java
```diff
@@ -163,6 +165,61 @@ public KeyManager[] getKeyManagers(URI privateKeyUri, URI certificateUri)
...
+    /**
+     * Get certificate chain using key alias.
+     *
+     * @param privateKeyUri  private key URI
+     * @param certificateUri certificate URI
+     * @return X509Certificate that corresponds to the given private key and certificate URIs
+     * @throws ServiceUnavailableException if crypto key provider service is unavailable
+     * @throws KeyLoadingException         if crypto key provider service fails to load key
+     */
+    public X509Certificate[] getCertificate(URI privateKeyUri, URI certificateUri)
```
MikeDombo, line 177: This api name is wrong. It isn't giving us the certificate, it is giving us the certificate chain. I don't believe that we do want (or even have) the chain. We only have a single certificate, so I'd just return that.

src/main/java/com/aws/greengrass/security/CryptoKeySpi.java
saranyailla: Changed.

src/main/java/com/aws/greengrass/security/CryptoKeySpi.java
```diff
@@ -20,4 +21,9 @@ KeyManager[] getKeyManagers(URI privateKeyUri, URI certificateUri)
     KeyPair getKeyPair(URI privateKeyUri, URI certificateUri) throws ServiceUnavailableException, KeyLoadingException;
 
     String supportedKeyType();
+
+    default X509Certificate getCertificate(URI privateKeyUri, URI certificateUri)
```
MikeDombo, line 25: not clear to me that we need both inputs, since we're requesting the certificate. But let me know if you think that the private key would be required for some reason.

src/main/java/com/aws/greengrass/security/SecurityService.java
```diff
@@ -163,14 +166,53 @@ public KeyManager[] getKeyManagers(URI privateKeyUri, URI certificateUri)
...
+     */
+    public X509Certificate getCertificate(URI privateKeyUri, URI certificateUri)
+            throws ServiceUnavailableException, KeyLoadingException {
+        logger.atTrace().kv(KEY_URI, privateKeyUri).kv(CERT_URI, certificateUri)
+                .log("Get certificate by private key and certificate URIs");
+        CryptoKeySpi provider = selectCryptoKeyProvider(privateKeyUri);
+        try {
+            return provider.getCertificate(privateKeyUri, certificateUri);
+        } catch (UnsupportedOperationException e) {
+            KeyManager[] km = provider.getKeyManagers(privateKeyUri, certificateUri);
```
MikeDombo, line 186: log that we're falling back to a search method.

src/main/java/com/aws/greengrass/security/SecurityService.java
```diff
@@ -163,14 +166,53 @@ public KeyManager[] getKeyManagers(URI privateKeyUri, URI certificateUri)
...
+            if (km.length != 1 || !(km[0] instanceof X509KeyManager)) {
+                return null;
```
MikeDombo, line 188: error log

src/main/java/com/aws/greengrass/security/SecurityService.java
```diff
@@ -163,14 +166,53 @@ public KeyManager[] getKeyManagers(URI privateKeyUri, URI certificateUri)
...
+            }
+
+            X509KeyManager x509KeyManager = (X509KeyManager) km[0];
+            KeyPair keyPair = provider.getKeyPair(privateKeyUri, certificateUri);
+            String[] aliases = x509KeyManager.getClientAliases(keyPair.getPublic().getAlgorithm(), null);
+            for (String alias : aliases) {
+                if (x509KeyManager.getPrivateKey(alias) != keyPair.getPrivate()) {
```
MikeDombo, line 195: use `.equals`?

src/main/java/com/aws/greengrass/security/SecurityService.java
```diff
@@ -163,14 +166,53 @@ public KeyManager[] getKeyManagers(URI privateKeyUri, URI certificateUri)
...
+                    continue;
+                }
+                X509Certificate[] certificateChain = x509KeyManager.getCertificateChain(alias);
+                if (certificateChain.length >= 1) {
```
MikeDombo, line 199: if the chain is greater, how do we know which of the multiple to return? Can you check that the certificate's public key matches the public/private key we want?

src/main/java/com/aws/greengrass/security/SecurityService.java
```diff
@@ -356,6 +398,19 @@ public String supportedKeyType() {
...
+        public X509Certificate getCertificate(URI privateKeyUri, URI certificateUri) throws KeyLoadingException {
+            KeyManager[] km = getKeyManagers(privateKeyUri, certificateUri);
+            if (km.length == 1 && km[0] instanceof X509KeyManager) {
+                X509KeyManager x509KeyManager = (X509KeyManager) km[0];
+                X509Certificate[] certChain = x509KeyManager.getCertificateChain(KEY_ALIAS);
+                if (certChain.length >= 1) {
+                    return certChain[0];
+                }
+            }
+            return null;
```
MikeDombo, line 411: this is totally unnecessary.

`EncryptionUtils.loadX509Certificates(Paths.get(certificateUri));`

src/main/java/com/aws/greengrass/security/SecurityService.java
jbutler: Noticed a discussion around cert vs cert chain.

Don't assume there's only 1 cert. Why would that be the case? W.r.t. which cert, there's convention. No need for fancy logic here. Keep it simple

src/main/java/com/aws/greengrass/security/SecurityService.java
saranyailla: Ok, that's better.

src/main/java/com/aws/greengrass/security/SecurityService.java
saranyailla: I'll change this API to get the cert chain. Any component that just needs the certificate can retrieve the first one in the array as long as we can impose the order `core CA cert -> intermediate cert (if any) -> root cert` while storing the cert chain in the key store. Any other opinions?

src/main/java/com/aws/greengrass/security/CryptoKeySpi.java
saranyailla: Yeah, private key is needed only in PKCS provider component because that's the label used to retrieve the cert from key store.

src/test/java/com/aws/greengrass/security/SecurityServiceTest.java
```diff
@@ -29,19 +29,21 @@
...
+import java.util.List;
 import javax.net.ssl.KeyManager;
 import javax.net.ssl.X509KeyManager;
 
 import static com.aws.greengrass.testcommons.testutilities.ExceptionLogProtector.ignoreExceptionOfType;
 import static org.hamcrest.MatcherAssert.assertThat;
-import static org.hamcrest.Matchers.containsString;
-import static org.hamcrest.Matchers.is;
-import static org.hamcrest.Matchers.notNullValue;
+import static org.hamcrest.Matchers.*;
```
MikeDombo, line 43: no * imports

MikeDombo (dismissed):

src/main/java/com/aws/greengrass/security/SecurityService.java
```diff
@@ -163,14 +168,62 @@ public KeyManager[] getKeyManagers(URI privateKeyUri, URI certificateUri)
...
+            return provider.getCertificateChain(privateKeyUri, certificateUri);
+        } catch (UnsupportedOperationException e) {
+            logger.atTrace().kv(KEY_URI, privateKeyUri).kv(CERT_URI, certificateUri)
+                    .log("Could not find the certificate using the provider implementation."
+                            + " Getting it from the key manager.");
+            KeyManager[] km = provider.getKeyManagers(privateKeyUri, certificateUri);
+            if (km.length != 1 || !(km[0] instanceof X509KeyManager)) {
+                logger.atError().kv(KEY_URI, privateKeyUri).kv(CERT_URI, certificateUri)
+                        .log("Unable to find the key manager with the given URIs.");
+                return null;
```
Nelsonochoam, line 197: Could we return an empty list instead? That ensures we adhere to the signature we are telling clients we adhere to

```suggestion
                return Collections.< X509Certificate>emptyList();
```

src/main/java/com/aws/greengrass/security/SecurityService.java
```diff
@@ -163,14 +168,62 @@ public KeyManager[] getKeyManagers(URI privateKeyUri, URI certificateUri)
...
+            }
+            X509KeyManager x509KeyManager = (X509KeyManager) km[0];
+            KeyPair keyPair = provider.getKeyPair(privateKeyUri, certificateUri);
+            String[] aliases = x509KeyManager.getClientAliases(keyPair.getPublic().getAlgorithm(), null);
+            if (aliases == null) {
+                logger.atError().kv(KEY_URI, privateKeyUri).kv(CERT_URI, certificateUri)
+                        .log("Unable to find aliases in the key manager.");
+                return null;
```
Nelsonochoam, line 205:
```suggestion
                return Collections.< X509Certificate>emptyList();
```

src/test/java/com/aws/greengrass/security/SecurityServiceTest.java
```diff
@@ -345,4 +353,103 @@ void GIVEN_key_and_cert_uri_WHEN_get_mqtt_builder_from_default_THEN_succeed() th
...
+    void GIVEN_key_and_cert_uri_WHEN_getCertificateChain_and_providerImplemented_THEN_getCertChainFromProvider()
+            throws Exception {
+        URI keyUri = new URI("pkcs11:object=key-label");
+        URI certificateUri = new URI("file:///path/to/certificate");
+        when(mockKeyProvider.supportedKeyType()).thenReturn("PKCS11");
+        when(mockKeyProvider.getCertificateChain(keyUri, certificateUri)).thenReturn(Collections.emptyList());
+
+        service.registerCryptoKeyProvider(mockKeyProvider);
+        List<X509Certificate> certs = service.getCertificateChain(keyUri, certificateUri);
+        assertThat(certs, notNullValue());
```
Nelsonochoam, line 367: Could we also assert that the value we get here is the cert that we expect. Just returning an List with a random `X509Certificate` or an empty array would make this test pass

src/test/java/com/aws/greengrass/security/SecurityServiceTest.java
```diff
@@ -345,4 +353,103 @@ void GIVEN_key_and_cert_uri_WHEN_get_mqtt_builder_from_default_THEN_succeed() th
...
+        when(mockKeyProvider.supportedKeyType()).thenReturn("PKCS11");
+        when(mockKeyProvider.getCertificateChain(keyUri, certificateUri))
+                .thenThrow(UnsupportedOperationException.class);
+        when(x509KeyManager.getClientAliases(any(), any())).thenReturn(aliases);
+        when(x509KeyManager.getPrivateKey(aliases[0])).thenReturn(mockPrivateKey);
+        when(x509KeyManager.getCertificateChain("test")).thenReturn(x509Certificates);
+
+        service.registerCryptoKeyProvider(mockKeyProvider);
+        List<X509Certificate> certs = service.getCertificateChain(keyUri, certificateUri);
+        assertThat(certs, notNullValue());
```
Nelsonochoam, line 394: Could we also assert that the value we get here is the cert that we expect. Just returning an List with a random X509Certificate or an empty array would make this test pass

src/test/java/com/aws/greengrass/security/SecurityServiceTest.java
```diff
@@ -345,4 +353,103 @@ void GIVEN_key_and_cert_uri_WHEN_get_mqtt_builder_from_default_THEN_succeed() th
...
+        URI keyUri = new URI("pkcs11:object=key-label");
+        URI certificateUri = new URI("file:///path/to/certificate");
+        KeyManager[] mockKeyManagers = {mock(X509KeyManager.class), mock(X509KeyManager.class)};
+        when(mockKeyProvider.getKeyManagers(keyUri, certificateUri)).thenReturn(mockKeyManagers);
+        when(mockKeyProvider.supportedKeyType()).thenReturn("PKCS11");
+        when(mockKeyProvider.getCertificateChain(keyUri, certificateUri))
+                .thenThrow(UnsupportedOperationException.class);
+        service.registerCryptoKeyProvider(mockKeyProvider);
+        List<X509Certificate> certs = service.getCertificateChain(keyUri, certificateUri);
+        assertThat(certs, nullValue());
```
Nelsonochoam, line 409: If you follow my suggestions we might have to change this to something that asserts empty certs.

src/main/java/com/aws/greengrass/security/SecurityService.java
```diff
@@ -163,14 +168,62 @@ public KeyManager[] getKeyManagers(URI privateKeyUri, URI certificateUri)
...
+                .log("Getting the certificate chain using private key and certificate URIs");
+        CryptoKeySpi provider = selectCryptoKeyProvider(privateKeyUri);
+        try {
+            return provider.getCertificateChain(privateKeyUri, certificateUri);
+        } catch (UnsupportedOperationException e) {
+            logger.atTrace().kv(KEY_URI, privateKeyUri).kv(CERT_URI, certificateUri)
+                    .log("Could not find the certificate using the provider implementation."
+                            + " Getting it from the key manager.");
+            KeyManager[] km = provider.getKeyManagers(privateKeyUri, certificateUri);
+            if (km.length != 1 || !(km[0] instanceof X509KeyManager)) {
```
Nelsonochoam, line 194: Can we guarantee that the X509KeyManager is always going to be returned in position 0?. What if for some reason the API changes and it gets returned in position 1 and the length is greater that 1?. This is forcing the client to have implicit understanding of how the provider stores the managers. I believe we should just tell the provider to give us the `x509KeyManager` like

```
X509KeyManager x509KeyManager = provider.getKeyManager(pkUri, certUri, X509)

if (x509KeyManager == null) {
    return Collections.emptyList()
}
```

saranyailla: In both file and pkcs providers, we create a keystore with only one key along with its **x509** type certs and `init` the keymanager instance with it. So, we'll only have one key manager and that will be of type `X509KeyManager`. Otherwise, it is neither guaranteed that it will be of size 1 not x509 can be found at index 0.

Nelsonochoam: Got it, but the caller basically needs to have all that information an do this type of checks every time they want to get the `X509KeyManager`.  The disadvantages of this is that

* the caller is aware of the data structure we are using
* It makes things harder to change, if we want to change how the provider stores the managers we then have to make sure we change all the callers.

jbutler: `!A || !B` logic that you want to evaluate to false? You're making me recruit way too many brain cells here. Please invert this condition logic and flip the if/else.

`km.length == 1 && km[0] instanceof X509KeyManager`

src/main/java/com/aws/greengrass/security/CryptoKeySpi.java
```diff
@@ -20,4 +24,9 @@ KeyManager[] getKeyManagers(URI privateKeyUri, URI certificateUri)
     KeyPair getKeyPair(URI privateKeyUri, URI certificateUri) throws ServiceUnavailableException, KeyLoadingException;
 
     String supportedKeyType();
+
+    default List<X509Certificate> getCertificateChain(URI privateKeyUri, URI certificateUri)
+            throws ServiceUnavailableException, KeyLoadingException, CertificateException, IOException {
+        throw new UnsupportedOperationException("This method is not supported.");
+    }
```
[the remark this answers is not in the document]
saranyailla: Newly added API reduces the number of calls involved such as the following in getting the cert chain..
 - getting the key manager (validate URIs, get key store, get key pair)
- getting the key pair (validate URIs, get key store)
- look up for aliases and find the one that matches the private key iteratively

src/main/java/com/aws/greengrass/security/SecurityService.java
```diff
@@ -163,14 +168,63 @@ public KeyManager[] getKeyManagers(URI privateKeyUri, URI certificateUri)
...
+     *
+     * @param privateKeyUri  private key URI
+     * @param certificateUri certificate URI
+     * @return X509Certificate list that corresponds to the given private key and certificate URIs
+     * @throws ServiceUnavailableException if crypto key provider service is unavailable
+     * @throws KeyLoadingException         if crypto key provider service fails to load key
+     * @throws CertificateException if a valid certificate is not found at the URI specified
+     * @throws IOException         if reading file at the specified URI fails
+     */
+    public List<X509Certificate> getCertificateChain(URI privateKeyUri, URI certificateUri)
```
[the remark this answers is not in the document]
saranyailla: Updated.

src/main/java/com/aws/greengrass/security/SecurityService.java
saranyailla: Had to code this way initially to avoid PMD violation for nested if statements. Updated it.

src/main/java/com/aws/greengrass/security/SecurityService.java
```diff
@@ -163,6 +165,47 @@ public KeyManager[] getKeyManagers(URI privateKeyUri, URI certificateUri)
...
+        if (km.length == 1 && km[0] instanceof X509KeyManager) {
+            X509KeyManager x509KeyManager = (X509KeyManager) km[0];
+            KeyPair keyPair = provider.getKeyPair(privateKeyUri, certificateUri);
+            String[] aliases = x509KeyManager.getClientAliases(keyPair.getPublic().getAlgorithm(), null);
+            logger.atTrace().kv(KEY_URI, privateKeyUri).kv(CERT_URI, certificateUri)
+                    .log("Getting the certificate chain from the provider using the key manager");
+            if (aliases == null) {
+                logger.atError().kv(KEY_URI, privateKeyUri).kv(CERT_URI, certificateUri)
+                        .log("Unable to find aliases in the key manager.");
+                return certChain;
```
MikeDombo, line 195: this is an error case, shouldn't we be throwing an exception, not an empty array?

src/main/java/com/aws/greengrass/security/SecurityService.java
```diff
@@ -163,6 +165,47 @@ public KeyManager[] getKeyManagers(URI privateKeyUri, URI certificateUri)
...
+            }
+            for (String alias : aliases) {
+                if (x509KeyManager.getPrivateKey(alias).equals(keyPair.getPrivate())) {
+                    return x509KeyManager.getCertificateChain(alias);
+                }
+            }
+        }
+        logger.atError().kv(KEY_URI, privateKeyUri).kv(CERT_URI, certificateUri)
+                .log("Unable to get the certificate chain using private key and certificate URIs.");
+        return certChain;
```
MikeDombo, line 205: likewise down here. Throw exception?

src/main/java/com/aws/greengrass/security/SecurityService.java
saranyailla: I think it's on the caller to decide what to do with the output. It might not always be an exception on the other side.

MikeDombo: With this pattern though, the caller doesn't really know that anything has gone wrong. It isn't a pattern that everyone always agrees with, but Greengrass code guidelines do have us used checked exceptions. So use a checked exception for exceptional cases (which this is).

src/main/java/com/aws/greengrass/security/SecurityService.java
```diff
@@ -163,6 +165,47 @@ public KeyManager[] getKeyManagers(URI privateKeyUri, URI certificateUri)
...
+
+        X509Certificate[] certChain = new X509Certificate[0];
+        CryptoKeySpi provider = selectCryptoKeyProvider(privateKeyUri);
+        KeyManager[] km = provider.getKeyManagers(privateKeyUri, certificateUri);
+
+        if (km.length == 1 && km[0] instanceof X509KeyManager) {
+            X509KeyManager x509KeyManager = (X509KeyManager) km[0];
+            KeyPair keyPair = provider.getKeyPair(privateKeyUri, certificateUri);
+            String[] aliases = x509KeyManager.getClientAliases(keyPair.getPublic().getAlgorithm(), null);
+            logger.atTrace().kv(KEY_URI, privateKeyUri).kv(CERT_URI, certificateUri)
```
jbutler: Redundant log. Remove

src/main/java/com/aws/greengrass/security/SecurityService.java
```diff
@@ -163,6 +165,42 @@ public KeyManager[] getKeyManagers(URI privateKeyUri, URI certificateUri)
...
+        KeyManager[] km = provider.getKeyManagers(privateKeyUri, certificateUri);
+
+        if (km.length == 1 && km[0] instanceof X509KeyManager) {
+            X509KeyManager x509KeyManager = (X509KeyManager) km[0];
+            KeyPair keyPair = provider.getKeyPair(privateKeyUri, certificateUri);
+            String[] aliases = x509KeyManager.getClientAliases(keyPair.getPublic().getAlgorithm(), null);
+            if (aliases == null) {
+                logger.atError().kv(KEY_URI, privateKeyUri).kv(CERT_URI, certificateUri)
+                        .log("Unable to find aliases in the key manager.");
+                throw new Exception("Unable to get the certificate chain from the provider using the key manager");
```
jbutler, line 190: You can add a new checked exception for this -- perhaps throw a `CertLoadingException` instead of a `KeyLoadingException`.

src/main/java/com/aws/greengrass/security/SecurityService.java
```diff
@@ -163,6 +165,42 @@ public KeyManager[] getKeyManagers(URI privateKeyUri, URI certificateUri)
...
+            }
+            for (String alias : aliases) {
+                if (x509KeyManager.getPrivateKey(alias).equals(keyPair.getPrivate())) {
+                    return x509KeyManager.getCertificateChain(alias);
+                }
+            }
+        }
+        logger.atError().kv(KEY_URI, privateKeyUri).kv(CERT_URI, certificateUri)
+                .log("Unable to get the certificate chain using private key and certificate URIs");
+        throw new Exception("Unable to get the certificate chain from the provider using the key manager");
```
MikeDombo, line 200: use a properly scoped exception. Create a new exception type if needed.

MikeDombo: `NoCertificateChainException`

src/main/java/com/aws/greengrass/security/SecurityService.java
saranyailla: Done.

src/main/java/com/aws/greengrass/security/SecurityService.java
```diff
@@ -163,6 +166,46 @@ public KeyManager[] getKeyManagers(URI privateKeyUri, URI certificateUri)
...
+
+        if (km.length == 1 && km[0] instanceof X509KeyManager) {
+            X509KeyManager x509KeyManager = (X509KeyManager) km[0];
+            KeyPair keyPair = provider.getKeyPair(privateKeyUri, certificateUri);
+            String[] aliases = x509KeyManager.getClientAliases(keyPair.getPublic().getAlgorithm(), null);
+            if (aliases == null) {
+                logger.atError().kv(KEY_URI, privateKeyUri).kv(CERT_URI, certificateUri)
+                        .log("Unable to find aliases in the key manager.");
+                throw new CertificateChainLoadingException("Unable to get the certificate chain from the provider "
+                        + "using the key manager");
```
MikeDombo, line 194: don't log _and_ throw. Only throw. The exception message should tell you what went wrong (ie. we couldn't find any aliases)

src/main/java/com/aws/greengrass/security/SecurityService.java
```diff
@@ -163,6 +166,46 @@ public KeyManager[] getKeyManagers(URI privateKeyUri, URI certificateUri)
...
+            for (String alias : aliases) {
+                if (x509KeyManager.getPrivateKey(alias).equals(keyPair.getPrivate())) {
+                    return x509KeyManager.getCertificateChain(alias);
+                }
+            }
+        }
+        logger.atError().kv(KEY_URI, privateKeyUri).kv(CERT_URI, certificateUri)
+                .log("Unable to get the certificate chain using private key and certificate URIs");
+        throw new CertificateChainLoadingException("Unable to get the certificate chain from the provider using the "
+                + "key manager");
```
MikeDombo, line 205: Only throw. And have multiple exception messages based on what specifically went wrong. We need to be able to diagnose this without access to the customer's device. 

In this place, this exception can happen for more than once reason which will be tough to figure out.

src/main/java/com/aws/greengrass/security/SecurityService.java
```diff
@@ -163,6 +166,46 @@ public KeyManager[] getKeyManagers(URI privateKeyUri, URI certificateUri)
...
+     * @throws CertificateChainLoadingException if no certificate chain is found
+     */
+    public X509Certificate[] getCertificateChain(URI privateKeyUri, URI certificateUri)
+            throws CertificateChainLoadingException, ServiceUnavailableException, KeyLoadingException {
+        logger.atTrace().kv(KEY_URI, privateKeyUri).kv(CERT_URI, certificateUri)
+                .log("Getting the certificate chain using private key and certificate URIs");
+        CryptoKeySpi provider = selectCryptoKeyProvider(privateKeyUri);
+        KeyManager[] km = provider.getKeyManagers(privateKeyUri, certificateUri);
+
+        if (km.length == 1 && km[0] instanceof X509KeyManager) {
```
MikeDombo, line 186: I'd suggest using an "early exit" style.

`if (km.length != 1 || !(km[0] instanceof X509KeyManager)) { throw <blah> }`

src/main/java/com/aws/greengrass/security/SecurityService.java
```diff
@@ -163,6 +166,42 @@ public KeyManager[] getKeyManagers(URI privateKeyUri, URI certificateUri)
...
+                throw new CertificateChainLoadingException("Unable to find aliases in the key manager with the given "
+                        + "private key and certificate URIs");
+            }
+            for (String alias : aliases) {
+                if (x509KeyManager.getPrivateKey(alias).equals(keyPair.getPrivate())) {
+                    return x509KeyManager.getCertificateChain(alias);
+                }
+            }
+        }
+        throw new CertificateChainLoadingException("Unable to get certificate chain from the provider using X509 "
```
MikeDombo, line 200: this can still happen for more than one reason (if no aliases are equal)

MikeDombo (dismissed):

MikeDombo (approved):

jbutler (approved):

PR #1263, opened 2022-07-14
test: integ tests of putComponentMetric IPC API

kucnavya:
**Issue #, if available:**

**Description of changes:**
Adding integration tests for Business metrics

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/ipc/IPCComponentMetricsTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/ipc/IPCComponentMetricsTest.java
@@ -1,0 +1,385 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.aws.greengrass.integrationtests.ipc;
+
+import com.aws.greengrass.config.Topics;
+import com.aws.greengrass.dependency.State;
+import com.aws.greengrass.deployment.DeviceConfiguration;
+import com.aws.greengrass.integrationtests.BaseITCase;
+import com.aws.greengrass.integrationtests.util.ConfigPlatformResolver;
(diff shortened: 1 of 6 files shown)


nikkhilmuthye (approved):

src/integrationtests/java/com/aws/greengrass/integrationtests/ipc/IPCComponentMetricsTest.java
```diff
@@ -0,0 +1,232 @@
...
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.atLeastOnce;
+import static org.mockito.Mockito.lenient;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+import static org.mockito.internal.verification.VerificationModeFactory.atLeast;
+
+@ExtendWith({GGExtension.class, MockitoExtension.class})
+public class IPCComponentMetricsTest extends BaseITCase {
```
aggaayus, line 82: 1. The responsibility of this API is to validate, process and create metric factory object which in turn creates a log file. Hence, the happy path for this API should validate that a log file with correct name and data is created in the right directory and stop at that point. The functionality of aggregating data in that log file, publishing through MQTT is the responsibility of TelemetryAgent for which we have separate integration tests. 
2. We should add test cases for unhappy path as well. For ex - input to API contains a metric that was not a part of ACL policy, input of invalid metric values (< 0) etc.

nikkhilmuthye: For Point 1, while I agree with this. I also see an additional logic where we merge in the two features since they are meant to work together. There is no scenario where the `PutComponentMetric` API is useful without the `TelemetryAgent`.

aggaayus (approved):
Thanks for making the tests exhaustive !!

nikkhilmuthye (approved):

PR #1268, opened 2022-07-19
fix: do not wait for mqtt client disconnect on context shutdown

junfuchen99:
**Issue #, if available:**

**Description of changes:**
During context shutdown and Mqtt client closing, do not wait for `AwsIotMqttClient::disconnect`.

Since SDK v1.9.2, once the SDK/CRT has been told to shutdown, we cannot guarantee that the SDK/CRT code will complete as expected because of our JVM access being taken away. However, since the program is shutting down, there should not be a need to wait on the code to finish since it will be terminated and all processes stopped. 

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
+++ b/src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
@@ -9,6 +9,7 @@ import com.aws.greengrass.config.Topics;
 import com.aws.greengrass.logging.api.Logger;
 import com.aws.greengrass.logging.impl.LogManager;
 import com.aws.greengrass.util.Coerce;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import lombok.AccessLevel;
 import lombok.Getter;
 import lombok.Setter;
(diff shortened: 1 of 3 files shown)


src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
```diff
@@ -284,13 +284,6 @@ int getTimeout() {
                 MqttClient.DEFAULT_MQTT_OPERATION_TIMEOUT, MqttClient.MQTT_OPERATION_TIMEOUT_KEY));
     }
 
-    int getCloseTimeout() {
-        // Use a shorter timeout for disconnection when closing client,
```
MikeDombo, line 288: why remove this? You're still using a timeout in the close

src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
```diff
@@ -406,9 +399,19 @@ public void close() {
...
             logger.atError().log("Error while disconnecting the MQTT client", e);
         }
     }
+
+    public synchronized void closeOnShutdown() {
+        if (resubscribeFuture != null && !resubscribeFuture.isDone()) {
+            logger.atTrace().log("Canceling resubscribe future");
+            resubscribeFuture.cancel(true);
+        }
+        // During Nucleus shutdown, disconnecting AwsIotMqttClient would time out because SDK/CRT backend cannot
```
MikeDombo, line 413: this comment doesn't explain what you're doing.

The important part is that you aren't waiting for the disconnect future (or returning it).

junfuchen99: will update the comment

src/main/java/com/aws/greengrass/mqttclient/AwsIotMqttClient.java
```diff
@@ -406,9 +399,19 @@ public void close() {
...
         }
         try {
-            disconnect().get(getCloseTimeout(), TimeUnit.MILLISECONDS);
+            disconnect().get(getTimeout(), TimeUnit.MILLISECONDS);
         } catch (InterruptedException | ExecutionException | TimeoutException e) {
             logger.atError().log("Error while disconnecting the MQTT client", e);
         }
     }
+
+    public synchronized void closeOnShutdown() {
```
MikeDombo, line 408: I don't think this should be synchronized. You need to sync very carefully in this class because we've had deadlocks before.

junfuchen99: will update and suppress FB warnings instead.

MikeDombo (approved):

shaguptashaikh (approved):

indougnito (approved):

PR #1275, opened 2022-08-04
feature: propagate component status codes to FSS

indougnito:
**Issue #, if available:**

**Description of changes:**

Propagates component status codes to FSS.

Implements a new lifecycle method reportState(newState, statusCode) accepting a status code parameter, which will be propagated to FSS in a fashion similar to lifecycle state.

Invoking reportState(newState) without a status code parameter causes a default status code to be propagated, e.g. ComponentStatusCode.RUN_ERRORED if the new state is ERRORED and the previous state was RUNNING, or ComponentStatusCode.NONE if the new state is healthy.

Component status updates will now be triggered on ERRORED in addition to the existing BROKEN state transitions 

Also updates the config README to call attention to some Topic naming details.

We would like to know why a component failed, i.e. its status code, in addition to the fact that it failed.

**Documentation Checklist:**

**Compatibility Checklist:**


src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
```diff
@@ -223,6 +278,12 @@ private Topic initStateTopic(final Topics topics) {
         return state;
     }
 
+    private Topic initTopic(final Topics topics, final String topicName) {
```
rbattle, line 281: nit: initStateTopic should be rewritten to call this

src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
```diff
@@ -265,11 +326,11 @@ private void setDesiredState(State... state) {
...
             // try insert to the queue, if queue full doesn't block.
-            enqueueStateEvent("DesiredStateUpdated");
+            enqueueStateEvent(new DesiredStateUpdatedEvent());
         }
     }
 
-    private void enqueueStateEvent(Object event) {
+    private void enqueueStateEvent(StateEvent event) {
         if (!stateEventQueue.offer(event)) {
             logger.error("couldn't put the new event to stateEventQueue");
```
rbattle, line 335: nit: should the event be logged here?

src/main/java/com/aws/greengrass/status/FleetStatusService.java
```diff
@@ -295,6 +295,21 @@ private void handleServiceStateChange(GreengrassService greengrassService, State
             updatedGreengrassServiceSet.add(greengrassService);
         }
 
+        // Always report status on errored state, evaluate overall status based on current state of all components
+        if (newState.equals(State.ERRORED)) {
+            synchronized (updatedGreengrassServiceSet) {
+                Instant now = Instant.now();
+                AtomicReference<OverallStatus> overAllStatus = new AtomicReference<>();
```
rbattle, line 302: we've seen some circumstances where overallStatus is null - is it worth initializing it to HEALTHY?

rbattle (approved):
looks ok - few minor comments

src/main/java/com/aws/greengrass/status/FleetStatusService.java
junfuchen99: overallStatus would be null if kernel just launches and there is no service running. Should we consider it healthy or assign a new value like "UNKNOWN"?

(I think we can address this issue in a separate pr.

src/main/java/com/aws/greengrass/lifecyclemanager/GenericExternalService.java
```diff
@@ -807,4 +810,12 @@ protected Exec addUser(Exec exec, boolean requiresPrivilege) {
         }
         return exec;
     }
+
+    @AllArgsConstructor
+    @Data
+    private static class RunResult {
+        private RunStatus runStatus;
+        private Exec exec;
+        private ComponentStatusCode statusCode;
```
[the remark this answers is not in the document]
junfuchen99: Could you also add integ test for an invalid runwith or skipif config to make sure that that path works?

junfuchen99 (dismissed):

junfuchen99 (dismissed):

junfuchen99 (approved):

kucnavya (approved):

Issue #825, opened 2021-01-22
com.aws.greengrass.mqttclient.spool: New persistent storage spooler(s) : sqlite spool / filespool


kris-sum:
**Feature Description**
Add new spoolers with persistent storage - either sqlite spool or a filesystem spool. These would be drop-in replacements for the InMemorySpool.

**Use Case**
Where greengrass devices operate in conditions where they may be offline without network access for long periods of time (days, weeks) the **InMemorySpool** may not be sufficient for storing proxied IoT MQTT messages. Additionally, large MQTT messages will quickly fill up the default size of the InMemoryPool even when offline for short periods of time.

**Proposed Solution**
A basic **filesystem** spool would be the simplest to write, with each message stored as a file. But persisting many small messages could potentially fill up available inodes and have poor throughput performance.
An **sqlite** database spool would allow persistent storage to a single `datastore.sqlite` file, but with more CPU overhead.

_The sqlite spool would be my preferred implementation._

Some limits should still be enforced on the persistent spoolers - potentially
* maxNumberOfMessages
* maxSizeOnDisk (harder to evalute for sqlite?)

**Other**
Would this be something that the AWS team would work on internally / would you accept a PR if we implemented the feature?

* [x] :wave: I may be able to implement this feature request

MikeDombo:
Hi Kris,
This feature is definitely on our roadmap. I don't have an ETA for you right now, but we're certainly aware of the case and we'll work to address it soon.

kris-sum:
That's great :) 

We'll be needing this for production in the next 6 months so we might start coding this and raise a PR, unless you know it'll be addressed before then.

kris-sum:
@MikeDombo Sorry to chase this up, but is there a rough roadmap ETA? (and would you accept a PR for it?)

MikeDombo:
Yes we would consider a PR for this feature. I cannot share any ETA at this time, sorry.

plafer:
We also need this feature. Would you now be able to share an ETA, by any chance? :)

MikeDombo:
@plafer No, sorry I do not have any ETA to share at this time.

TamoshaytisV:
[...]

PR #1277, opened 2022-08-12
feat: integrate persistence spooler plugin with MqttClient

allanli524:
**Description of changes:**

Changes were made in the MqttClient and Kernel components of the Nucleus. Code was added to allow the integration of the new "persistence spool plugin", which will introduce persistence to the spooler in MqttClient. The kernel lifecycle is slightly tweaked so that the classpath scans for plugins occurs before startables starts to allow for MqttClient to access the new plugin's class as it initializes.

UPDATE CHANGES:

1. Persistent Queue sync is best effort mode: In case of Disk Spool Storage type, we try to sync message IDs from Disk spooler to the main `queueOfMessageId`. If we fail while doing this sync, we continue with Disk Spooler storage rather than defaulting back to InMemorySpooler (as "Disk" is what the customer configured). 
2. Always create InMemory spooler, fallback to adding a message into InMemory spooler in case adding into Disk spooler fails. Always Read from InMemory spooler first as there might be messages due to fallback.
3. Added Unit test for QueueSync and fallback behavior.

This change is necessary to integrate with the new plugin component "persistence spool", which will allow users to adopt the persistence mode of the spooler by deploying it.

The unit tests and integration tests in the nucleus was slightly changed to reflect the constructor changes this commit made. No new tests were added in this commit. Running "mvn verify" locally on this branch yielded the same result as the main branch.

The plugin component is still being tested, unit tests and integration tests for the plugin will be available with the plugin's codebase.

DO NOT MERGE THIS BRANCH.


src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -213,6 +238,61 @@ public long getCurrentSpoolerSize() {
...
+                SpoolMessage message = spooler.getMessageById(currentId);
+                PublishRequest request = message.getRequest();
+                try {
+                    queueCapacityCheck(request);
+
+                    queueOfMessageId.putLast(currentId);
+                    if (currentId > highestId) {
+                        highestId = currentId;
+                    }
+                } catch (InterruptedException | SpoolerStoreException e) {
```
MikeDombo, line 262: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

allanli524: Thanks, I have added Thread.currentThread().interrupt() in a new commit.

src/main/java/com/aws/greengrass/mqttclient/spool/InMemorySpool.java
```diff
@@ -27,4 +29,9 @@ public void add(long id, SpoolMessage message) {
         messages.put(id, message);
     }
 
+    @Override
+    public List<Long> getAllSpoolMessageIds() {
+        return new ArrayList<>();
```
vaibhavmurkute, line 34: Shouldn't it be returning `messages.keySet()` with `Set<Long>` return type?

allanli524: This interface method was written for the plugin to return the existing messageIds in the persistent database with order, which is why the return type must be List<Long> to keep the order. So that when the spooler starts the persistent queue can be synced with the queue in Spool.java.

Since InMemorySpool is not persistent, there will be no messageIds to sync when the spooler starts, so this implementation is just a placeholder that returns an empty list.

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -213,6 +238,64 @@ public long getCurrentSpoolerSize() {
...
+     * ids into queueOfMessageId
+     */
+    private void persistentQueueSync(List<Long> diskQueueOfIds) {
+        if (!diskQueueOfIds.isEmpty()) {
+            long highestId = -1;
+            int i;
+            for (i = 0; i < diskQueueOfIds.size(); i += 1) {
+                //Check for queue space and remove if necessary
+                long currentId = diskQueueOfIds.get(i);
+                SpoolMessage message = spooler.getMessageById(currentId);
```
vaibhavmurkute, line 253: Possible NPE here, `spooler` may not have been instantiated (setup).

allanli524: Thank you for catching this, I will fix it for the next commit

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -213,6 +238,67 @@ public long getCurrentSpoolerSize() {
...
+                try {
+                    queueCapacityCheck(request);
+
+                    queueOfMessageId.putLast(currentId);
+                    if (currentId > highestId) {
+                        highestId = currentId;
+                    }
+                } catch (SpoolerStoreException e) {
+                    logger.atError().log(e);
+                    throw new RuntimeException(e);
```
vaibhavmurkute, line 266: Can we directly throw `SpoolerStoreException` ? We usually avoid log _and_ throw: either log or throw.

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -213,6 +238,67 @@ public long getCurrentSpoolerSize() {
...
+                } catch (InterruptedException e) {
+                    Thread.currentThread().interrupt();
+                    logger.atError().log(e);
+                    throw new RuntimeException(e);
```
vaibhavmurkute, line 270: No need to log, just rethrow `InterruptedException`.

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -213,6 +238,67 @@ public long getCurrentSpoolerSize() {
...
+                }
+                logger.atInfo()
+                        .kv("Number of MQTT messages", diskQueueOfIds.size())
+                        .log("Messages added to spool runtime queue");
```
vaibhavmurkute, line 274: Guess this log will be printed for every spooled message, might want to move it out of the loop. 
Also a key-value nit: `"Number of MQTT messages"` --> `"numSpoolerMessages"`

src/main/java/com/aws/greengrass/mqttclient/spool/InMemorySpool.java
vaibhavmurkute: This seems wrong. Irrespective of persistent or inMemory spool, what happens if higher level abstractions call this method, e.g. `CloudMessageSpool.getAllSpoolMessageIds()`

allanli524: Changing for next commit

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
allanli524: I will throw the SpoolerStoreException as a ServiceLoadException so that it will be handled by the setUpSpooler() method.

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
allanli524: thanks, revised

src/main/java/com/aws/greengrass/mqttclient/spool/CloudMessageSpool.java
```diff
@@ -5,11 +5,15 @@
...
 
     SpoolMessage getMessageById(long id);
 
     void removeMessageById(long id);
 
     void add(long id, SpoolMessage message);
-}
+
+    List<Long> getAllSpoolMessageIds();
+}
```
MikeDombo, line 19: line at end of file

src/main/java/com/aws/greengrass/mqttclient/spool/CloudMessageSpool.java
```diff
@@ -5,11 +5,15 @@
...
 public interface CloudMessageSpool {
 
     SpoolMessage getMessageById(long id);
 
     void removeMessageById(long id);
 
     void add(long id, SpoolMessage message);
-}
+
+    List<Long> getAllSpoolMessageIds();
```
MikeDombo, line 18: do we really need a `List` specifically? Can we get away with a generic `Collection`?

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -81,11 +90,43 @@ private void setSpoolerConfigFromDeviceConfig(Topics topics) {
...
         if (config.getStorageType() == SpoolerStorageType.Memory) {
             return new InMemorySpool();
+        } else if (config.getStorageType() == SpoolerStorageType.FileSystem) {
+            try {
+                return getPersistenceSpoolGGService();
+            } catch (ServiceLoadException e) {
+                //log and use InMemorySpool
+                logger.atWarn()
+                        .kv(GG_PERSISTENCE_SPOOL_SERVICE_NAME_KEY, config.getPersistenceSpoolServiceName())
+                        .kv("ServiceLoadException", e.getMessage())
```
MikeDombo, line 100: don't log the message of the error. Log the full cause using `.cause()`

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -81,11 +90,43 @@ private void setSpoolerConfigFromDeviceConfig(Topics topics) {
...
+                        .log("Persistence Spool set up failure, "
+                                + "defaulting to In-Memory Spool, retry when spooler accessed again.");
+                return new InMemorySpool();
+            } catch (InterruptedException e) {
+                throw new RuntimeException(e);
```
MikeDombo, line 105: this is not the way to handle an interruption.

Do:

`Thread.currentThread().interrupt();`

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -81,11 +90,43 @@ private void setSpoolerConfigFromDeviceConfig(Topics topics) {
     private CloudMessageSpool setupSpooler() {
         if (config.getStorageType() == SpoolerStorageType.Memory) {
             return new InMemorySpool();
+        } else if (config.getStorageType() == SpoolerStorageType.FileSystem) {
```
MikeDombo, line 93: I don't think this storage type should be called filesystem. This is actually more of a generic solution, since it will find any plugin and can thus be implemented in any way. Suggest changing FileSystem -> Plugin.

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -213,6 +240,66 @@ public long getCurrentSpoolerSize() {
...
 
+    /**
+     * Extract message ids from the persistenceSpool plugin's on disk database and insert the message \
+     * ids into queueOfMessageId, this function is only used in FileSystem storage mode.
+     * @param diskQueueOfIds list of messageIds to sync
+     * @param persistenceSpool instance of CloudMessageSpool
+     */
+    private void persistentQueueSync(List<Long> diskQueueOfIds, CloudMessageSpool persistenceSpool)
+            throws InterruptedException, ServiceLoadException {
+        if (!diskQueueOfIds.isEmpty()) {
```
MikeDombo, line 252: use an error-first style. 

`if (diskQueueOfIds.isEmpty()) { return; }`

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -213,6 +240,66 @@ public long getCurrentSpoolerSize() {
...
+            long highestId = -1;
+            int i;
+            for (i = 0; i < diskQueueOfIds.size(); i += 1) {
```
MikeDombo, line 255: use an enhanced (for-each) for loop.

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -213,6 +240,66 @@ public long getCurrentSpoolerSize() {
...
+                    queueCapacityCheck(request);
+
+                    queueOfMessageId.putLast(currentId);
+                    if (currentId > highestId) {
+                        highestId = currentId;
+                    }
+                } catch (SpoolerStoreException e) {
+                    throw new ServiceLoadException(e);
+                } catch (InterruptedException e) {
+                    Thread.currentThread().interrupt();
```
MikeDombo, line 270: don't both interrupt and rethrow. Choose 1. In this case, just don't catch the exception and let it bubble up.

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -213,6 +240,66 @@ public long getCurrentSpoolerSize() {
...
+                PublishRequest request = message.getRequest();
+                try {
+                    queueCapacityCheck(request);
+
+                    queueOfMessageId.putLast(currentId);
+                    if (currentId > highestId) {
+                        highestId = currentId;
+                    }
+                } catch (SpoolerStoreException e) {
+                    throw new ServiceLoadException(e);
```
MikeDombo, line 268: why are you wrapping the real exception in a service load exception? Sounds like this method should just throw `SpoolerStoreException`

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -213,6 +240,66 @@ public long getCurrentSpoolerSize() {
...
+        if (curMessageQueueSizeInBytes.get() > getSpoolConfig().getSpoolSizeInBytes()) {
+            removeOldestMessage();
+        }
 
+        if (curMessageQueueSizeInBytes.get() > getSpoolConfig().getSpoolSizeInBytes()) {
+            curMessageQueueSizeInBytes.getAndAdd(-1L * messageSizeInBytes);
+            throw new SpoolerStoreException("Message spool is full. Message could not be added.");
+        }
+    }
+}
```
MikeDombo, line 305: EOF

src/test/java/com/aws/greengrass/mqttclient/InMemorySpoolTest.java
```diff
@@ -43,12 +44,13 @@ class InMemorySpoolTest {
     private Spool spool;
     Configuration config = new Configuration(new Context());
     private static final String GG_SPOOL_MAX_SIZE_IN_BYTES_KEY = "maxSizeInBytes";
+    private final Kernel kernel = new Kernel();
```
MikeDombo, line 47: if you create a kernel you must shut it down. 

Call shutdown in `AfterEach`

MikeDombo: You can also just use a mock kernel.

src/test/java/com/aws/greengrass/mqttclient/MqttClientTest.java
```diff
@@ -98,6 +99,7 @@ class MqttClientTest {
     @Mock
     Spool spool;
 
+    Kernel kernel = new Kernel();
```
MikeDombo, line 102: likewise.

src/main/java/com/aws/greengrass/mqttclient/spool/CloudMessageSpool.java
allanli524: I think it is better to use a List because the collection needs to retain order. It is mainly used to sync the existing disk queue with the runtime queue. 

Unless Collection is able to retain the order of the messageIds that we get directly from SQLite. 

What do you think?

MikeDombo: Could use `Iterable` instead, which would be ordered.

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
allanli524: I think the naming of this was done before me, but I can change it

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
allanli524: Okay I will implement an enhanced for loop

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -213,6 +238,62 @@ public long getCurrentSpoolerSize() {
...
+     */
+    private void persistentQueueSync(Iterable<Long> diskQueueOfIds, CloudMessageSpool persistenceSpool)
+            throws InterruptedException, SpoolerStoreException {
+        if (!diskQueueOfIds.iterator().hasNext()) {
+            return;
+        }
+        long highestId = -1;
+        int numMessages = 0;
+        int queueOfMessageIdInitSize = queueOfMessageId.size();
+        for (long currentId: diskQueueOfIds) {
```
MikeDombo, line 256: reformat this file. 

Space before `:`

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -81,11 +89,42 @@ private void setSpoolerConfigFromDeviceConfig(Topics topics) {
...
+     * @return CloudMessageSpool instance
+     * @throws ServiceLoadException thrown if the service cannot be located
+     */
+    private CloudMessageSpool getPersistenceSpoolGGService()
+            throws ServiceLoadException, InterruptedException, SpoolerStoreException {
+        GreengrassService locatedService = kernel.locate(config.getPersistenceSpoolServiceName());
+        if (locatedService instanceof CloudMessageSpool) {
+            CloudMessageSpool persistenceSpool = (CloudMessageSpool) locatedService;
+            persistentQueueSync(persistenceSpool.getAllSpoolMessageIds(), persistenceSpool);
+            logger.atInfo().log("Persistent Spooler has been set up.");
```
MikeDombo, line 119: no period at the end of log messages

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -81,11 +89,42 @@ private void setSpoolerConfigFromDeviceConfig(Topics topics) {
...
         if (config.getStorageType() == SpoolerStorageType.Memory) {
             return new InMemorySpool();
+        } else if (config.getStorageType() == SpoolerStorageType.Plugin) {
+            try {
+                return getPersistenceSpoolGGService();
+            } catch (ServiceLoadException | SpoolerStoreException e) {
+                //log and use InMemorySpool
+                logger.atWarn()
+                        .kv(GG_PERSISTENCE_SPOOL_SERVICE_NAME_KEY, config.getPersistenceSpoolServiceName())
+                        .cause(e).log();
```
MikeDombo, line 99: add a message to the log

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -483,13 +485,13 @@ public synchronized void unsubscribe(UnsubscribeRequest request)
...
-                                    subscriptions.put(e.getKey(), subscriberForTopic.get().getValue());
-                                }
-                            });
+                                        // recalculate and replace the client
+                                        Optional<Map.Entry<MqttTopic, AwsIotMqttClient>> subscriberForTopic =
+                                                findExistingSubscriberForTopic(e.getKey().getTopic());
+                                        if (subscriberForTopic.isPresent()) {
+                                            subscriptions.put(e.getKey(), subscriberForTopic.get().getValue());
+                                        }
+                                    });
```
vaibhavmurkute, line 494: Does this file also need reformat?
Rest of the changes look good to me.

dflatow:
Can someone explain to me if this PR will solve this issue: https://github.com/aws-greengrass/aws-greengrass-nucleus/issues/825? Thank you

jcosentino11:
> Can someone explain to me if this PR will solve this issue: #825? Thank you

Partially--this PR handles the integration between a new plugin and the nucleus.  The plugin itself will also need to be released in order for the issue to be fully resolved.

dflatow:
>>  Can someone explain to me if this PR will solve this issue: https://github.com/aws-greengrass/aws-greengrass-nucleus/issues/825? Thank you
>
> Partially--this PR handles the integration between a new plugin and the nucleus. The plugin itself will also need to be released in order for the issue to be fully resolved.

Cool well I'm looking forward to seeing this realeased!

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -56,37 +61,83 @@ public Spool(DeviceConfiguration deviceConfiguration) {
...
             return new InMemorySpool();
+        } else if (config.getStorageType() == SpoolerStorageType.Disk) {
+            try {
+                return getPersistenceSpoolGGService();
+            } catch (ServiceLoadException | IOException e) {
+                //log and use InMemorySpool
+                logger.atWarn()
+                        .kv(PERSISTENCE_SPOOL_SERVICE_NAME_KEY, config.getPersistenceSpoolServiceName())
+                        .cause(e).log("Persistence spool set up failed, defaulting to in-memory mode");
+                return new InMemorySpool();
```
MikeDombo, line 103: always create an in memory spooler so that we can fallback to it if anything fails.
if we cannot write to the disk spool, we need to fallback to in memory so that we are able to send messages

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -132,8 +169,12 @@ public synchronized SpoolMessage addMessage(Publish request) throws InterruptedE
         return message;
     }
 
-    private void addMessageToSpooler(long id, SpoolMessage message) {
-        spooler.add(id, message);
+    private void addMessageToSpooler(long id, SpoolMessage message) throws SpoolerStoreException {
+        try {
+            spooler.add(id, message);
+        } catch (IOException e) {
+            throw new SpoolerStoreException(e);
```
MikeDombo, line 176: must fallback

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
vabhasin: Sounds right! Will add that.

src/main/java/com/aws/greengrass/mqttclient/spool/CloudMessageSpool.java
```diff
@@ -5,11 +5,15 @@
...
 public interface CloudMessageSpool {
 
     SpoolMessage getMessageById(long id);
 
     void removeMessageById(long id);
 
-    void add(long id, SpoolMessage message);
+    void add(long id, SpoolMessage message) throws IOException;
+
+    Iterable<Long> getAllSpoolMessageIds() throws IOException;
```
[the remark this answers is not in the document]
vabhasin: Oh no, it's actually getting all Message IDs...not the message itself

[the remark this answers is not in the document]
jcosentino11: whoops, long day... `getAllMessageIds()`, main thought is that `spool` is redundant

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
MikeDombo: We don't necessarily want to switch the spooler, we just want to fallback. So when writing, try writing to disk then fallback if needed.
When reading, read from memory first, then read from disk.

MikeDombo: Yes, we need to check to see if there are any messages which were put their as the fallback, otherwise they'd never be sent which is very bad.

MikeDombo (dismissed):

jcosentino11 (dismissed):

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -133,7 +169,12 @@ public synchronized SpoolMessage addMessage(Publish request) throws InterruptedE
     }
 
     private void addMessageToSpooler(long id, SpoolMessage message) {
-        spooler.add(id, message);
+        try {
+            spooler.add(id, message);
+        } catch (IOException e) {
+            logger.atWarn().log("Disk Spooler failed to add Message, adding message to InMemory Spooler", e);
```
jcosentino11, line 175: technically `spooler` could be in memory spooler---in that case this would be a retry which we probably don't want or may want to log differently

jcosentino11: same in getMessageById

vabhasin: Yea, spooler could be memory spooler. But Exception in add is only thrown by DiskSpooler, so I'm relying on that.

jcosentino11: can we add a comment for that then since it's not obvious, at least to me :)

jcosentino11: what about `getMessageById`?   in worst case we'd do two lookups

vabhasin: Yea, I can change that. Thanks!

MikeDombo (dismissed):

jcosentino11 (dismissed):

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
```diff
@@ -140,18 +140,17 @@ public void launch() {
...
         // This guarantees that IPC, for example, is running before any user code
         for (Class<? extends Startable> c : startables) {
             kernel.getContext().get(c).startup();
         }
 
-        final List<DeviceIdentityInterface> provisioningPlugins = findProvisioningPlugins();
-        // Must be called before everything else so that these are available to be
-        // referenced by main/dependencies of main
-        final Queue<String> autostart = findBuiltInServicesAndPlugins(); //NOPMD
-        loadPlugins();
```
junfuchen99, line 154: Why did we need to reorder this event?

jcosentino11 (commented):
Tested manually that spooler plugin loads after reverting kernel change?

MikeDombo (approved):

jcosentino11 (approved):

PR #1287, opened 2022-08-24
feat: deployment error code setup

junfuchen99:
**Issue #, if available:**

**Description of changes:**
This pr sets up necessary data types and utils for deployment error code. Specifically:
1. Defines deployment error code and error type enum. Assigns each error code with an error type.
2. Modifies `DeploymentException` to maintain a list of error codes and a map of error context. Updated its child exceptions.
3. Implements utility functions to traverse the chain of exceptions and generate the error code stack.
4. Added unit tests.

To help service gain better visibility of deployment failures on device.
These changes are part of our ongoing improvements to FSS and device health notifications.

**Documentation Checklist:**

**Compatibility Checklist:**


src/main/java/com/aws/greengrass/deployment/errorcode/DeploymentErrorCodeUtils.java
```diff
@@ -0,0 +1,225 @@
...
+    /**
+     * Helper function to inline add an error code to the head of list.
+     *
+     * @param errorCodeList the destination list
+     * @param errorCode     error code to be added to the head
+     * @return the updated list
+     */
+    public static List<DeploymentErrorCode> inlineAddToFront(List<DeploymentErrorCode> errorCodeList,
+                                                             DeploymentErrorCode errorCode) {
+        errorCodeList.add(0, errorCode);
```
MikeDombo, line 201: warning that you cannot use this for `Arrays.asList` as they're unmodifiable. You may way to handle that better here, and return a list which you can modify.

src/main/java/com/aws/greengrass/deployment/errorcode/DeploymentErrorCodeUtils.java
```diff
@@ -0,0 +1,225 @@
...
+        if (arn.isEmpty()) {
+            throw new IllegalArgumentException("Empty component arn");
+        }
+
+        final Matcher matcher = COMPONENT_ARN_PATTERN.matcher(arn);
+        if (matcher.matches()) {
+            String accountId = matcher.group(5);
+            return "aws".equals(accountId);
+        }
+        throw new IllegalArgumentException("Component arn is not valid");
```
MikeDombo, line 222: we don't like unchecked exceptions.

src/test/java/com/aws/greengrass/deployment/errorcode/DeploymentErrorCodeUtilsTest.java
```diff
@@ -0,0 +1,195 @@
...
+        assertEquals(first.size(), second.size());
+        for (int i = 0; i < first.size(); i++) {
+            assertEquals(first.get(i), second.get(i));
+        }
+    }
+
+    private static void assertListEqualsWithoutOrder(List<String> first, List<String> second) {
+        assertTrue(first.size() == second.size() && first.containsAll(second) && second.containsAll(first));
+    }
+}
```
MikeDombo, line 195: EOF

junfuchen99: will fix

src/main/java/com/aws/greengrass/deployment/errorcode/DeploymentErrorCodeUtils.java
junfuchen99: will change to another exception type

src/main/java/com/aws/greengrass/deployment/errorcode/DeploymentErrorCodeUtils.java
junfuchen99: Yea I was planning to wrap a `new ArrayList<>` around `Arrays.asList` before it passes into this function, but I agree maybe I should just return a new list here instead.

MikeDombo: It is fine if you're modifying a modifiable list, just maybe add a check for an unmodifyable list too and handle that case.

junfuchen99: Updated to always use a modifiable list. Also added a check for unsupported operation.

MikeDombo (approved):

src/main/java/com/aws/greengrass/componentmanager/plugins/docker/exceptions/DockerPullException.java
```diff
@@ -0,0 +1,22 @@
...
+package com.aws.greengrass.componentmanager.plugins.docker.exceptions;
+
+import static com.aws.greengrass.deployment.errorcode.DeploymentErrorCode.DOCKER_PULL_ERROR;
+
+public class DockerPullException extends DockerImageDownloadException {
+    static final long serialVersionUID = -3387516993124229948L;
+
+    public DockerPullException(String message) {
+        super(message);
+        super.getErrorCodes().add(DOCKER_PULL_ERROR);
```
rbattle, line 15: nit - why not just have a super class method: addErrorCode that does this. or even super constructor that takes a the error code as a param and adds it along with the initial one

src/main/java/com/aws/greengrass/deployment/errorcode/DeploymentErrorCode.java
```diff
@@ -0,0 +1,122 @@
...
+    S3_HEAD_OBJECT_ACCESS_DENIED(DeploymentErrorType.PERMISSION_ERROR),
+    S3_HEAD_OBJECT_RESOURCE_NOT_FOUND(DeploymentErrorType.REQUEST_ERROR),
+    S3_GET_BUCKET_LOCATION_ACCESS_DENIED(DeploymentErrorType.PERMISSION_ERROR),
+    S3_GET_BUCKET_LOCATION_RESOURCE_NOT_FOUND(DeploymentErrorType.REQUEST_ERROR),
+    S3_GET_BUCKET_ACCESS_DENIED(DeploymentErrorType.PERMISSION_ERROR),
+    S3_GET_BUCKET_RESOURCE_NOT_FOUND(DeploymentErrorType.REQUEST_ERROR),
+
+    // Nucleus errors
+    EMPTY_DEPLOYMENT_REQUEST(DeploymentErrorType.NUCLEUS_ERROR),
+    DEPLOYMENT_DOCUMENT_PARSE_ERROR(DeploymentErrorType.NUCLEUS_ERROR),
```
rbattle, line 96: are these nucleus errors or should we categorize them as some type of cloud service response error? I suppose for local deployments that doesn't make sense though.

src/main/java/com/aws/greengrass/deployment/errorcode/DeploymentErrorCode.java
```diff
@@ -0,0 +1,122 @@
...
+    DEPLOYMENT_TYPE_NOT_VALID(DeploymentErrorType.NUCLEUS_ERROR),
+    RESOLVE_COMPONENT_CANDIDATES_BAD_RESPONSE(DeploymentErrorType.NUCLEUS_ERROR),
```
rbattle, line 98: also seems like it should be tied to a cloud error rather than Nucleus

src/main/java/com/aws/greengrass/deployment/errorcode/DeploymentErrorCode.java
```diff
@@ -0,0 +1,122 @@
...
+    COMPONENT_METADATA_NOT_VALID(DeploymentErrorType.NUCLEUS_ERROR),
```
rbattle, line 99: does that come from component recipe or is this after downloading we check something and determine it is invalid?

rbattle: also should this be a COMPONENT_ERROR?

src/main/java/com/aws/greengrass/componentmanager/plugins/docker/exceptions/DockerPullException.java
junfuchen99: We discussed offline. Super constructor would not be efficient. Will wrap with an `addErrorCode`.

src/main/java/com/aws/greengrass/deployment/errorcode/DeploymentErrorCodeUtils.java
```diff
@@ -0,0 +1,213 @@
...
+
+        List<String> errorStack = errorCodeSet.stream().map(Enum::toString).collect(Collectors.toList());
+        // remove duplicate types
+        List<String> errorTypes = errorCodeSet.stream().map(DeploymentErrorCode::getErrorType).distinct()
+                .filter(type -> !type.equals(DeploymentErrorType.NONE)).map(Enum::toString)
+                .collect(Collectors.toList());
+        return new Pair<>(errorStack, errorTypes);
+    }
+
+    private static void translateExceptionToErrorCode(Set<DeploymentErrorCode> errorCodeSet, Throwable e,
```
rbattle, line 106: should you log when it is an unhandled exception type?

junfuchen99: We synced offline. Since we are always logging the original exception when deployment fails, there is no need to add additional log here.

src/main/java/com/aws/greengrass/deployment/errorcode/DeploymentErrorCode.java
junfuchen99: this is where an invalid component metadata is received in deployment doc. As others, it will be nucleus error if local deployment; cloud service error if cloud deployments

rbattle (approved):

MikeDombo (approved):

PR #1297, opened 2022-09-07
fix(config): race condition when subscribing to changing configuration

MikeDombo:
**Issue #, if available:**

**Description of changes:**
Fixes a race condition which causes this failure: https://github.com/aws-greengrass/aws-greengrass-nucleus/runs/8234655490?check_suite_focus=true.

```
software.amazon.awssdk.eventstreamrpc.UnmappedDataException: Found model-type {aws.greengrass#ConfigurationUpdateEvents} which does not map into Java class: software.amazon.awssdk.aws.greengrass.model.SubscribeToConfigurationUpdateResponse
```

Fixes it by using an atomicboolean which is set to true once the IPC subscription response has been sent to the client. Only when this flag is `true` will a configuration update cause the data to actually send.

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/builtin/services/configstore/ConfigStoreIPCEventStreamAgent.java
+++ b/src/main/java/com/aws/greengrass/builtin/services/configstore/ConfigStoreIPCEventStreamAgent.java
@@ -58,6 +58,7 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.BiConsumer;
 import java.util.function.Consumer;
 import javax.inject.Inject;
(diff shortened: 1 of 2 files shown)


junfuchen99 (dismissed):

src/main/java/com/aws/greengrass/builtin/services/configstore/ConfigStoreIPCEventStreamAgent.java
```diff
@@ -402,10 +404,14 @@ public SubscribeToConfigurationUpdateResponse handleRequest(SubscribeToConfigura
             });
         }
 
+        @Override
+        public void afterHandleRequest() {
+            active.set(true);
```
rbattle, line 409: should it be set to false in on `onStreamClosed` or is that a scenario that doesn't matter?

MikeDombo: Shouldn't matter. When the stream is closed, the _send_ method won't do anything anyway.

rbattle (dismissed):

rbattle (approved):

junfuchen99 (approved):

PR #1308, opened 2022-09-15
feat: allow component configuration changes to move out of BROKEN

MikeDombo:
**Issue #, if available:**

**Description of changes:**
This change calls `requestReinstall` when configuration changes the lifecycle of a component which is currently BROKEN. Today, broken components can only get un-broken by changing the install portion of lifecycle, or by changing the version of a component. This will allow more components to get unbroken more easily and in a more understandable way.

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/lifecyclemanager/GenericExternalServiceIntegTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/lifecyclemanager/GenericExternalServiceIntegTest.java
@@ -591,13 +591,15 @@ class GenericExternalServiceIntegTest extends BaseITCase {
         CountDownLatch serviceErroredLatch = new CountDownLatch(2);
         CountDownLatch serviceBrokenLatch = new CountDownLatch(1);
         List<Pair<ComponentStatusDetails, Long>> componentStatus = new ArrayList<>();
+
+        String serviceName = "ServiceA";
         kernel.getContext().addGlobalStateChangeListener((service, oldState, newState) -> {
-            if ("ServiceA".equals(service.getName()) && State.ERRORED.equals(newState)) {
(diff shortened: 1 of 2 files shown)


src/integrationtests/java/com/aws/greengrass/integrationtests/lifecyclemanager/GenericExternalServiceIntegTest.java
```diff
@@ -591,13 +591,16 @@ void GIVEN_service_starts_up_WHEN_service_breaks_THEN_status_details_persisted_f
         CountDownLatch serviceErroredLatch = new CountDownLatch(2);
         CountDownLatch serviceBrokenLatch = new CountDownLatch(1);
         List<Pair<ComponentStatusDetails, Long>> componentStatus = new ArrayList<>();
+
+        String serviceName = "ServiceA";
+        CountDownLatch finalServiceErroredLatch = serviceErroredLatch;
```
shaguptashaikh, line 596: Nit- why do you have these assignments? Could just use two separate latches or reinitialize the original one for your next set of assertions

shaguptashaikh (approved):

src/integrationtests/java/com/aws/greengrass/integrationtests/lifecyclemanager/GenericExternalServiceIntegTest.java
MikeDombo: This is required as lambdas must have "effectively final" variables in their captures.

I could create a new variable too, that also works. Re-assigning the value is what breaks the effectively final.

shaguptashaikh (approved):

junfuchen99 (approved):

indougnito (approved):

PR #1311, opened 2022-09-20
fix: FSS robustness improvements

junfuchen99:
**Description of changes:**
1. Always initialize overall status to healthy instead of null;
2. Avoid spamming FSS updates when mqtt client disconnects and reconnects without component state change by adding a 60 second time window;
3. Send nucleus launch message at the end of `KernelLifecycle::launch` instead of when setting up iot job communication.

Improve the robustness of FSS.

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
+++ b/src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
@@ -180,6 +180,15 @@ public class KernelLifecycle {
         logger.atInfo().setEventType("system-start").addKeyValue("main", kernel.getMain()).log();
         startupAllServices();
 
+        try {
+            GreengrassService fleetStatusService = kernel.locate(FleetStatusService.FLEET_STATUS_SERVICE_TOPICS);
+            if (fleetStatusService instanceof FleetStatusService) {
+                ((FleetStatusService) fleetStatusService).triggerFleetStatusUpdateAtKernelLaunch();
+            }
+        } catch (ServiceLoadException e) {
+            logger.atError().setCause(e).log("Failed to send status update at kernel launch because kernel was "
+                    + "unable to locate FleetStatusService");
+        }
     }
 
     @SuppressWarnings("PMD.AvoidCatchingGenericException")
--- a/src/main/java/com/aws/greengrass/status/FleetStatusService.java
+++ b/src/main/java/com/aws/greengrass/status/FleetStatusService.java
@@ -78,6 +78,7 @@ public class FleetStatusService extends GreengrassService {
             "$aws/things/{thingName}/greengrassv2/health/json";
     public static final String FLEET_STATUS_TEST_PERIODIC_UPDATE_INTERVAL_SEC = "fssPeriodicUpdateIntervalSec";
     public static final int DEFAULT_PERIODIC_PUBLISH_INTERVAL_SEC = 86_400;
+    public static final int MINIMAL_RECONNECT_PUBLISH_INTERVAL_SEC = 60;
     public static final String FLEET_STATUS_PERIODIC_PUBLISH_INTERVAL_SEC = "periodicStatusPublishIntervalSeconds";
     static final String FLEET_STATUS_SEQUENCE_NUMBER_TOPIC = "sequenceNumber";
     static final String FLEET_STATUS_LAST_PERIODIC_UPDATE_TIME_TOPIC = "lastPeriodicUpdateTime";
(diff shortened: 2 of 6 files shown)


src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
```diff
@@ -180,6 +180,15 @@ public void launch() {
...
 
+        try {
+            GreengrassService fleetStatusService = kernel.locate(FleetStatusService.FLEET_STATUS_SERVICE_TOPICS);
+            if (fleetStatusService instanceof FleetStatusService) {
+                ((FleetStatusService) fleetStatusService).triggerFleetStatusUpdateAtKernelLaunch();
+            }
+        } catch (ServiceLoadException e) {
+            logger.atError().setCause(e).log("Failed to send status update at kernel launch because kernel was "
+                    + "unable to locate FleetStatusService");
+        }
```
shaguptashaikh, line 191: You've only requested startup for services at this point but they may not be running. Is that going to be problematic? does it make sense to wait for main to reach desired state instead?

junfuchen99: As we discussed offline, the nucleus launch message should be the first FSS message a device sends after launch.  If any component goes wrong while starting up, they would error and trigger their own FSS updates. 

Also FSS only needs to be initialized and postinject to send an update; doesn't have to be running.

Hence, it's preferable to not wait for main to reach desired state

src/main/java/com/aws/greengrass/status/FleetStatusService.java
```diff
@@ -337,21 +337,24 @@ private void updatePeriodicFleetStatusData() {
...
-     * @param isConfigurationUpdate true if the update is triggered by device configuration changes
-     *                              false if the update is triggered at kernel launch IoTJobsHelper post-inject
+     * Trigger a Fleet Status update at kernel launch.
      */
-    public void updateFleetStatusUpdateForAllComponents(Boolean isConfigurationUpdate) {
-        if (isConfigurationUpdate) {
-            updateFleetStatusUpdateForAllComponents(Trigger.NETWORK_RECONFIGURE);
+    public void triggerFleetStatusUpdateAtKernelLaunch() {
+        if (!deviceConfiguration.isDeviceConfiguredToTalkToCloud()) {
+            logger.atWarn().log("Failed to send status update at kernel launch because device is offline");
```
rbattle, line 344: should we be warning about this? Seems like it would be expected behavior to not make network calls when device is not configured with cloud endpoints

I suggest dropping to DEBUG and saying something like: "Will not send status update for Nucleus Launch as device is not configured for online access"

src/main/java/com/aws/greengrass/status/FleetStatusService.java
```diff
@@ -419,6 +422,13 @@ private void updateEventTriggeredFleetStatusData(DeploymentInformation deploymen
             removedDependenciesSet.forEach(serviceFssTracksMap::remove);
             removedDependenciesSet.clear();
 
+            // Do not send empty FSS update when network reconnects in order to avoid spamming messages on flaky
+            // network.
+            // TODO: throttling mechanism for FSS updates
+            if (updatedGreengrassServiceSet.isEmpty() && Trigger.RECONNECT.equals(trigger)) {
```
rbattle, line 428: this optimizes for the worst case condition of a flaky network (or two devices with the same name booting each other off repeatedly). However there is a nice thing that customers lose here - when there is a network outage, they could see when GG reconnects. 

Rather than dropping all, did you consider keeping track of the last time a message was sent and not sending if it was within a time window, like 60s or 5m?

That would eliminate the spam, but still provide some more value.

If reconnect trigger is intended only to be used when something is removed while the device is offline, then update the comments to add more detail about that.

src/main/java/com/aws/greengrass/status/FleetStatusService.java
junfuchen99: As we discussed offline, will keep it at WARN level to keep parity with other device offline message.

src/main/java/com/aws/greengrass/status/FleetStatusService.java
junfuchen99: Added a 60 seconds time window

src/main/java/com/aws/greengrass/status/FleetStatusService.java
```diff
@@ -545,6 +567,11 @@ private Topic getPeriodicUpdateTimeTopic() {
         return config.lookup(FLEET_STATUS_LAST_PERIODIC_UPDATE_TIME_TOPIC).dflt(Instant.now().toEpochMilli());
     }
 
+    private Topic getReconnectUpdateTimeTopic() {
+        // default to zero so that first Reconnect update would go thru
+        return config.lookup(FLEET_STATUS_LAST_RECONNECT_UPDATE_TIME_TOPIC).dflt(0);
+    }
```
shaguptashaikh, line 573: Why do you need to maintain this in config?

junfuchen99: Updated to only store in memory

shaguptashaikh (approved):

indougnito (approved):

PR #1319, opened 2022-09-23
fix: createUser, createGroup and addUserToGroup do not support Busybox

### LINKED ISSUE #911: createUser and  createGroup do not support Busybox
The createUser function uses a nonportable useradd to create the user which is not available in busybox by default:

runCmd("useradd -r -m " + user, o -> {

https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/87d89a61342f20e2e84f505af38bd89f68514c2f/src/main/java/com/aws/greengrass/util/platforms/unix/UnixPlatform.java#L355

Busyboxy supports a adduser command the equivalent of which would be "adduser -S " + user (the -S is for system).

Similarly, the createGroup function uses 

 runCmd("groupadd -r " + group, o -> {

https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/87d89a61342f20e2e84f505af38bd89f68514c2f/src/main/java/com/aws/greengrass/util/platforms/unix/UnixPlatform.java#L361

the equivalent Busybox command is

"addgroup -S " + group

It should be noted that on Debian based systems the adduser and addgroup commands are deemed more "user friendly"  though they both use --system instead of -S for their runtime flags.

### LINKED ISSUE #912: addUserToGroup does not support Busybox
The function addUserToGroup uses usermod under the hood to add a user to a group

 runCmd("usermod -a -G " + group + " " + user, o -> {

https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/87d89a61342f20e2e84f505af38bd89f68514c2f/src/main/java/com/aws/greengrass/util/platforms/unix/UnixPlatform.java#L367

Busybox based Linux use the command equivalent 

"addgroup " + user + " " + group

https://busybox.net/downloads/BusyBox.html

This option also works on Debian based systems if both the user and the group are pre-existing cf. man 8 adduser

Mirrorxxi:
**Issue #, if available:**
Closes #911
Closes #912

**Description of changes:**
Attempts to use `adduser` when `useradd` fails. Attempts to use `addgroup` when `groupadd` fails. Attempts to use `addgroup` when `usermod` fails.

Support for systems that do not have `useradd`, `groupadd`, and `usermod`

**Documentation Checklist:**

**Compatibility Checklist:**


src/main/java/com/aws/greengrass/util/platforms/unix/UnixPlatform.java
```diff
@@ -355,20 +355,36 @@ public UnixRunWithGenerator getRunWithGenerator() {
...
     public void createUser(String user) throws IOException {
-        runCmd("useradd -r -m " + user, o -> {
-        }, "Failed to create user");
+        try {
+            runCmd("useradd -r -m " + user, o -> {
+            }, "Failed to create user");
+        } catch (IOException e) {
+            runCmd("adduser -S  " + user, l -> {
+            }, "Failed to create user");
+        }
```
MikeDombo, line 364: Instead of just trying both commands, I would prefer if we can identify which command will work before calling it.

src/main/java/com/aws/greengrass/util/platforms/unix/UnixPlatform.java
```diff
@@ -355,20 +356,58 @@ public UnixRunWithGenerator getRunWithGenerator() {
...
+     * Check if command exists.
+     * @param command command
+     * @return boolean
+     * @throws IOException throws when io exception occurs
+     */
+    protected boolean checkIfCommandExists(String command) throws IOException {
+        AtomicBoolean flag = new AtomicBoolean(false);
+
+        // Determine whether this command exists
+        runCmd("command -v " + command + "|wc -l", o -> {
```
MikeDombo, line 406: You cannot rely on the command 'command' and 'wc' existing on the target platform. This has to work with no additional commands. You must determine what exists from entirely within Java.

src/main/java/com/aws/greengrass/util/platforms/unix/UnixPlatform.java
```diff
@@ -355,20 +356,51 @@ public UnixRunWithGenerator getRunWithGenerator() {
...
+    }
+
+    /**
+     * Check if command exists.
+     *
+     * @param command command
+     * @return boolean
+     */
+    protected boolean checkIfCommandExists(String command) {
+        return UnixPlatform.getInstance().createNewProcessRunner().which(command) != null;
```
MikeDombo, line 403: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

**Problem**
Analysis of this package determined that this line of code might contain a resource leak. A resource leak can slow down or crash your system.

**Fix**
Consider closing the resource returned by the *createNewProcessRunner* method in a try-finally block. Execution paths that do not contain closure statements were detected.

**More info**
[View resource management guidelines at oracle.com](https://www.oracle.com/technetwork/articles/java/trywithresources-401775.html) (external link).

MikeDombo:
@Mirrorxxi 
Please rebase your change, then we should be able to merge this.

@shaguptashaikh please review.

MikeDombo (dismissed):

src/test/java/com/aws/greengrass/util/platforms/unix/UnixPlatformTest.java
```diff
@@ -5,28 +5,100 @@
...
+    private final static String group = "ggc_group01";
+
+    @Test
+    public void GIVEN_busybox_environment_WHEN_add_user_THEN_add_user_success() {
+        try {
+            spyPlatform.runCmd("useradd -r -m " + user, o -> {
+            }, "Failed to create user with useradd");
+        } catch (IOException e) {
+            logger.info(e.getMessage());
+        }
```
shaguptashaikh, line 53: What are you verifying with this? Seems like you're actually running these commands, so the test will behave differently depending on where it runs, and since you're catching and logging the exception, the test will never really fail but not accurately verify anything either, what purpose does this serve?

src/test/java/com/aws/greengrass/util/platforms/unix/UnixPlatformTest.java
```diff
@@ -5,28 +5,100 @@
...
+
+        try {
+            spyPlatform.createUser(user);
```
shaguptashaikh, line 56: Same for this, seems like everything you're testing is in the catch block below?

shaguptashaikh (commented):
Same questions for all tests

src/test/java/com/aws/greengrass/util/platforms/unix/UnixPlatformTest.java
yiwenTS: Since the pull requests are to run on Ubuntu os (our automation test platform), but not the busybox env, then the way of thinking for the test case is as below 5 steps:
1. create a user named "ggc_user01" (the purpose is to let createUser method to run the busybox command "adduser")
2. call createUser method to create ggc_user01 (run "useradd" command), then IOException generated
3. Entered outter catch block, and run the busybox command "adduser"
4. Second IOException generated since busybox command "adduser" is running on Ubuntu os

createUserMethod
testcase

MikeDombo (approved):

shaguptashaikh:
Can you provide some details here about what manual testing you performed?

yiwenTS:
Hi @shaguptashaikh, here are some records for manual test.
1. ggc_user and ggc_group can be created & ggc_user can be added to ggc_group in Alpine os by nucleus now:
https://user-images.githubusercontent.com/111258711/196684073-8f91b3ba-d811-4500-b499-0e2a85596562.mp4

2. ggc_user and ggc_group can be created & ggc_user can be added to ggc_group in Ubuntu os by nucleus now:
https://user-images.githubusercontent.com/111258711/196684124-83194337-c54e-491e-99f5-2f7fd5491786.mp4

Below is record to show above commands are not supported in Alpine in the past:
https://user-images.githubusercontent.com/111258711/196684188-c637763b-e770-485a-b58d-768092e20d1a.mp4

Use below command to start Alpine os and then install nucleus:
docker pull alpine
docker images 
docker run -dit  -v /gginstall:/gginstall  --name alpine-test alpine
docker exec -it alpine-test /bin/sh

shaguptashaikh (approved):

PR #1321, opened 2022-09-26
fix: update ggc version env on version update

shaguptashaikh:
**Issue #, if available:**
-

**Description of changes:**
Version env should be set up on launch but with preloaded config.tlog config value does not get updated due to the use of `lookup` then `dflt`

In order to correctly update the ggc_version env var for component processes when they're restarted after version update

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
+++ b/src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
@@ -328,7 +328,7 @@ public class DeviceConfiguration {
     void initializeNucleusVersion(String nucleusComponentName, String nucleusComponentVersion) {
         kernel.getConfig().lookup(SERVICES_NAMESPACE_TOPIC, nucleusComponentName,
                 VERSION_CONFIG_KEY).dflt(nucleusComponentVersion);
-        kernel.getConfig().lookup(SETENV_CONFIG_NAMESPACE, GGC_VERSION_ENV).dflt(nucleusComponentVersion);
+        kernel.getConfig().lookup(SETENV_CONFIG_NAMESPACE, GGC_VERSION_ENV).overrideValue(nucleusComponentVersion);
     }
 
     void initializeComponentStore(KernelAlternatives kernelAlts, String nucleusComponentName,


src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
```diff
@@ -328,7 +328,7 @@ void initializeNucleusLifecycleConfig(String nucleusComponentName, ComponentReci
     void initializeNucleusVersion(String nucleusComponentName, String nucleusComponentVersion) {
         kernel.getConfig().lookup(SERVICES_NAMESPACE_TOPIC, nucleusComponentName,
                 VERSION_CONFIG_KEY).dflt(nucleusComponentVersion);
-        kernel.getConfig().lookup(SETENV_CONFIG_NAMESPACE, GGC_VERSION_ENV).dflt(nucleusComponentVersion);
+        kernel.getConfig().lookup(SETENV_CONFIG_NAMESPACE, GGC_VERSION_ENV).withValue(nucleusComponentVersion);
```
MikeDombo, line 331: use `overrideValue` so that the timestamp is ignored and it always uses the value you give it.

MikeDombo, line 331:
```suggestion
        kernel.getConfig().lookup(SETENV_CONFIG_NAMESPACE, GGC_VERSION_ENV).overrideValue(nucleusComponentVersion);
```

shaguptashaikh: Why would we want to retain the timestamp though? that's what `overrideValue` does. `withValue` uses the current timestamp which I think is more accurate

MikeDombo: Because the timestamp isn't important. If the system clock is wrong, then this value may not be updated which is why I added the override method.

MikeDombo (approved):

kucnavya (approved):

PR #1325, opened 2022-09-28
fix: increase timeout for executor service shutdown when sigterm is received

junfuchen99:
**Description of changes:**
1. Add a 5-second timeout for executor service stopping during kernel shutdown on external signal.

Currently, when sigterm is received, executor shutdown and context shutdown happens instantly (within 1ms). Sometimes   threads under executor service are not interrupted properly because of such short timing.

Adding a 5-second timeout to interrupt threads properly.

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
+++ b/src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
@@ -82,6 +82,7 @@ import static com.aws.greengrass.util.Utils.deepToString;
 public class KernelLifecycle {
     private static final Logger logger = LogManager.getLogger(KernelLifecycle.class);
     private static final int EXECUTOR_SHUTDOWN_TIMEOUT_SECONDS = 30;
+    private static final int EXECUTOR_SERVICE_SHUTDOWN_TIMEOUT_SECONDS = 5;
     // Enum for provision policy will exist in common library package
     // This will be done as part of re-provisioning
     // TODO:  Use the enum from common library when available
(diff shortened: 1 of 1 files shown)


MikeDombo (approved):

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
```diff
@@ -523,10 +524,14 @@ public void shutdown(int timeoutSeconds) {
...
-            boolean scheduledExecutorTerminated = scheduledExecutorService.awaitTermination(timeoutSeconds,
-                    TimeUnit.SECONDS);
+            // when kernel shuts down due to external signal, give some time for executor service to stop so that
+            // threads are interrupted correctly
+            int executorServiceShutdownTimeoutSecond =
+                    timeoutSeconds == -1 ? EXECUTOR_SERVICE_SHUTDOWN_TIMEOUT_SECONDS : timeoutSeconds;
+            boolean executorTerminated =
+                    executorService.awaitTermination(executorServiceShutdownTimeoutSecond, TimeUnit.SECONDS);
+            boolean scheduledExecutorTerminated =
+                    scheduledExecutorService.awaitTermination(executorServiceShutdownTimeoutSecond, TimeUnit.SECONDS);
```
jbutler, line 534: Do you want to wait this timeout for each executor? Or should you be computing a remaining time for the second await?

jbutler (approved):
Approved based on conversation with Junfu.

There is a existing UAT which asserts on executor service shutdown logs which is failing as a result of fast Nucleus shutdown. While there is no critical need that we shutdown this service, we're doing it to improve odds that tasks actually get interrupted in case they need to do any cleanup as a result.

Issue #1111, opened 2021-10-08
(nucleus): Remove container images downloaded via `docker:` artifacts when the component is removed


massi-ang:
**Feature Description**
When a component that has specified artifacts using the `docker:` key is removed, the DockerApplicationManager plugin should remove the image if no other deployed component requires it

**Use Case**
Edge devices are limited in resources, and removing unused artifacts is a best practice

**Proposed Solution**
The only workaround is to write a custom component that monitor the images and removes those for which there are no running containers. 

**Other**

* [ ] :wave: I may be able to implement this feature request
* [ ] :warning: This feature might incur a breaking change

shaguptashaikh:
Hi @massi-ang, thanks for this reminder, we are currently thinking about providing a generic cleanup/uninstall mechanism requested by customers which can help clean up artifacts/installations like images. We will update here once we're able to work on that.

PR #1326, opened 2022-09-29
feat: remove docker image when component deleted (#1111)

leoSRM:
**Issue #1111:**

**Description of changes:**

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/componentmanager/ComponentStore.java
+++ b/src/main/java/com/aws/greengrass/componentmanager/ComponentStore.java
@@ -16,13 +16,13 @@ import com.aws.greengrass.componentmanager.models.RecipeMetadata;
 import com.aws.greengrass.config.PlatformResolver;
 import com.aws.greengrass.constants.FileSuffix;
 import com.aws.greengrass.deployment.errorcode.DeploymentErrorCode;
-import com.aws.greengrass.lifecyclemanager.GreengrassService;
 import com.aws.greengrass.logging.api.Logger;
 import com.aws.greengrass.logging.impl.LogManager;
 import com.aws.greengrass.util.Coerce;
 import com.aws.greengrass.util.Digest;
 import com.aws.greengrass.util.NucleusPaths;
 import com.aws.greengrass.util.SerializerFactory;
+import com.aws.greengrass.util.Utils;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.vdurmont.semver4j.Requirement;
 import com.vdurmont.semver4j.Semver;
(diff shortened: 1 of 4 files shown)


src/main/java/com/aws/greengrass/componentmanager/ComponentManager.java
```diff
@@ -508,6 +513,9 @@ public void cleanupStaleVersions() {
             for (String compVersion : removeVersions) {
                 try {
                     ComponentIdentifier identifier = new ComponentIdentifier(compName, new Semver(compVersion));
+                    // issues #1111
+                    // delete docker image before removing the recipe file
+                    removeDockerImage(identifier);
```
shaguptashaikh, line 518: This should be done inside `componentStore.deleteComponent`. And ideally, since there's a new type of artifact to cleanup now, it will make sense to refactor [artifact deletion code here](https://github.com/leoSRM/aws-greengrass-nucleus/blob/7978cba7dc6894abe74ece967e4cca02d95c044b/src/main/java/com/aws/greengrass/componentmanager/ComponentStore.java#L250,L271) into the respective artifact downloaders (might rename downloader classes too since they would handle cleanup in addition to downloads)

src/main/java/com/aws/greengrass/componentmanager/ComponentManager.java
```diff
@@ -667,4 +675,49 @@ private String getFileName(File f) {
...
+        logger.atDebug("removeDockerImage-start").log();
+        List<String> allRecipeContentList = componentStore.getAllComponentRecipeContent();
+        Optional<String> componentRecipeContent = componentStore.findComponentRecipeContent(componentIdentifier);
+        List<String> toBeRemovedList = new ArrayList<String>();
+        if (componentRecipeContent.isPresent()) {
+            RecipeLoader.RecipeFormat format = RecipeLoader.RecipeFormat.YAML;
+            // the recipe is JSON or YML
+            if (componentRecipeContent.get().startsWith("{")) {
+                format = RecipeLoader.RecipeFormat.JSON;
+            }
```
shaguptashaikh, line 696: Why do you need all of this? Are recipes ever saved as JSON?

src/main/java/com/aws/greengrass/componentmanager/ComponentManager.java
```diff
@@ -667,4 +675,49 @@ private String getFileName(File f) {
...
+            // get recipe
+            com.amazon.aws.iot.greengrass.component.common.ComponentRecipe componentRecipe =
+                    RecipeLoader.parseRecipe(componentRecipeContent.get(), format);
+            List<String> currentUriList =
+                    componentRecipe.getManifests().stream()
+                            .map(PlatformSpecificManifest::getArtifacts)
+                            .flatMap(Collection::stream)
+                            .map(item -> item.getUri().toString())
```
shaguptashaikh, line 704: You can filter directly here on the URI scheme, see [here](https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/main/src/main/java/com/aws/greengrass/componentmanager/builtins/ArtifactDownloaderFactory.java#L81)

src/main/java/com/aws/greengrass/componentmanager/ComponentManager.java
```diff
@@ -667,4 +675,49 @@ private String getFileName(File f) {
...
+            for (String uriString : currentUriList) {
+                // this docker image only used by itself
+                if (allRecipeContentList.stream().filter(item -> item.contains(uriString)).count() < 2) {
+                    toBeRemovedList.add(uriString);
+                }
+            }
+            // run docker command to remove image
+            Exec exec = Platform.getInstance().createNewProcessRunner();
+            logger.info("removeDockerImage-start" + toBeRemovedList.toString());
+            toBeRemovedList.forEach(uriString -> {
```
shaguptashaikh, line 718: Please implement this [here](https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/main/src/main/java/com/aws/greengrass/componentmanager/plugins/docker/DefaultDockerClient.java) along with the rest of the docker cli usage  
And use the [parser](https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/main/src/main/java/com/aws/greengrass/componentmanager/plugins/docker/DockerImageArtifactParser.java) to get the image information from the artifact URI

src/main/java/com/aws/greengrass/componentmanager/plugins/docker/DefaultDockerClient.java
```diff
@@ -166,6 +167,16 @@ private Optional<UserNotAuthorizedForDockerException> checkUserAuthorizationErro
         return Optional.ofNullable(error);
     }
 
+    public void deleteImage(Image image) throws DockerImageDeleteException {
+        CliResponse response = runDockerCmd(String.format("docker rmi -f %s", image.getImageFullName()));
+        if (response.exit.isPresent() && response.exit.get() == 0) {
+            return;
+        } else {
+            throw new DockerImageDeleteException("Unexpected error while trying to perform docker rmi",
```
MikeDombo, line 175: if it is failing, we need to log the reason why it failed so that we and the customer can debug the issue. Log the standard out and standard error response from the command execution.

src/main/java/com/aws/greengrass/componentmanager/ComponentStore.java
```diff
@@ -230,6 +244,9 @@ ComponentRecipe getPackageRecipe(@NonNull ComponentIdentifier pkgId) throws Pack
     void deleteComponent(@NonNull ComponentIdentifier compId) throws PackageLoadingException {
         logger.atDebug("delete-component-start").kv("componentIdentifier", compId).log();
         IOException exception = null;
+        // issues #1111
+        // delete docker image before removing the recipe file
+        removeDockerImage(compId);
```
MikeDombo, line 249: I think we can do this in a much better way which is properly extensible.

I'd want to add a new method to the `ArtifactDownloader` interface which is for cleaning up artifacts. Then, here you will use `getArtifactDownloader` to get the proper downloader for each artifact of the component, then call the cleanup method.

src/main/java/com/aws/greengrass/componentmanager/ComponentStore.java
```diff
@@ -555,4 +572,77 @@ private String getHashOfComponentName(String componentName) throws HashingAlgori
...
         }
     }
+
+    /**
+     * issues #1111
+     * Get all locally available component recipe content
+     *
+     * @return list of all component recipe content
+     */
+    public List<String> getAllComponentRecipeContent() {
```
MikeDombo, line 582: we should not be reading all the component recipes into memory, this will use too much memory.

Let's target only what we actually need. Get a list of the components, then for each component we want to read the artifacts and determine which docker image(s) would be safe to delete.

PR #1327, opened 2022-09-30
fix: shutdown external services with the old version of the shutdown command

MikeDombo:
**Issue #, if available:**

**Description of changes:**
Currently, a deployment simply updates the configuration of a component, and the component itself is responsible for updating itself appropriately in reaction to that change. What this means is that if a component is updated from v1 to v2, the component will have v2's lifecycle configuration immediately. The component reacts by reinstalling since the version changed. As part of reinstallation, the component will first shutdown. However, the shutdown command that it will use is the shutdown command for v2 because the configuration has already been changed. This means that the process of v1 will try to be stopped by the v2 shutdown script; this may or may not work.

This change fixes this problem by caching the correct shutdown script at `startup` time. The shutdown will then use this cached value instead of reading from the updated config value.

Added a new case into an existing test to verify that the correct shutdown script is used when the version changes.

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/lifecyclemanager/GenericExternalServiceIntegTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/lifecyclemanager/GenericExternalServiceIntegTest.java
@@ -345,7 +345,8 @@ class GenericExternalServiceIntegTest extends BaseITCase {
     }
 
     @Test
-    void GIVEN_running_service_WHEN_version_config_changes_THEN_service_reinstalls() throws Exception {
+    void GIVEN_running_service_WHEN_version_config_changes_THEN_service_reinstalls_and_prev_version_shutdown_is_used()
+            throws Exception {
         ConfigPlatformResolver.initKernelWithMultiPlatformConfig(kernel,
                 getClass().getResource("service_with_dynamic_config.yaml"));
         CountDownLatch mainRunning = new CountDownLatch(1);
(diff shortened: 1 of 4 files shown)


src/main/java/com/aws/greengrass/lifecyclemanager/GenericExternalService.java
```diff
@@ -698,19 +730,28 @@ protected RunResult run(String name, Topic t, String cmd, IntConsumer background
...
-            trackingList.add(exec);
+        Exec finalExec = exec;
+        CrashableSupplier<RunStatus, InterruptedException> doRun = () -> {
+            logger.atDebug().setEventType("generic-service-run").log();
+            // Track all running processes that we fork
+            if (finalExec.isRunning()) {
+                trackingList.add(finalExec);
+            }
+            return shellRunner.successful(finalExec, t.getFullName(),
+                    background, this) ? RunStatus.OK : RunStatus.Errored;
```
shaguptashaikh, line 742: When a cached exec runs and fails, looks like it won't update the status code in `RunResult`?

src/main/java/com/aws/greengrass/lifecyclemanager/GenericExternalService.java
```diff
@@ -584,7 +613,8 @@ public void handleError() throws InterruptedException {
                 Lifecycle.TIMEOUT_NAMESPACE_TOPIC));
 
         CountDownLatch handlerExecutionCdl = new CountDownLatch(1);
-        run(Lifecycle.LIFECYCLE_RECOVER_NAMESPACE_TOPIC, c -> handlerExecutionCdl.countDown(), lifecycleProcesses);
+        run(Lifecycle.LIFECYCLE_RECOVER_NAMESPACE_TOPIC, c -> handlerExecutionCdl.countDown(), lifecycleProcesses,
```
shaguptashaikh, line 616: nit - would've been nice to leave the original run method and have it call the new method with `runImmediately` as `true` so that you don't see booleans passed around in a lot of places

src/main/java/com/aws/greengrass/lifecyclemanager/GenericExternalService.java
MikeDombo: The runresult isn't used by anyone, so not, it isn't updated, but it doesn't matter.

shaguptashaikh: Why is it not used? if shutdown fails we need to set the status code correctly no?

MikeDombo: also the `RunStatus` is still returned by the cached method, so you can check the result if you wanted.

MikeDombo: This is our current implementation: https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/main/src/main/java/com/aws/greengrass/lifecyclemanager/GenericExternalService.java#L536-L540

shaguptashaikh: RunStatus doesn't have enough info, but since it's not used right now and component doesn't go to errored state this is probably okay

shaguptashaikh (dismissed):

src/integrationtests/java/com/aws/greengrass/integrationtests/lifecyclemanager/GenericExternalServiceIntegTest.java
```diff
@@ -363,13 +364,50 @@ void GIVEN_running_service_WHEN_version_config_changes_THEN_service_reinstalls()
...
+                service.getServiceConfig().find(SERVICE_LIFECYCLE_NAMESPACE_TOPIC, Lifecycle.LIFECYCLE_SHUTDOWN_NAMESPACE_TOPIC)
+                        .withValue("echo shutdown v1.0.1");
+            });
+
+            assertTrue(serviceReinstalled.await(60, TimeUnit.SECONDS));
+            componentShutdown.get(0, TimeUnit.SECONDS);
+        }
+
+        // Now when we shutdown it should use the latest version which is 1.0.1
+        kernel.locate("service_with_dynamic_config").requestStop();
```
shaguptashaikh, line 396: Can it shut down too soon and your log listener won't be invoked?

shaguptashaikh (approved):

shaguptashaikh (approved):

rbattle (approved):

PR #1329, opened 2022-10-10
feat: Linux Control Group version 2 API support cgroup v2

ChangxinDong:
**Issue #, if available:**

**Description of changes:**

**Documentation Checklist:**

**Compatibility Checklist:**


shaguptashaikh (changes requested):
Too much duplicate code both between the existing Cgroup enum and the new CgroupV2 class, and also between the existing linux system resource controller and the new v2 class you're adding. The only difference I see is the cgroup file names to write to and the content to write which needs to be derived from the system resource config passed. Please refactor and move everything that's common into base classes. Might not even need a new v2 resource controller class if you simply refactor the cgroup enum/classes to appropriately override methods that provide file names, and also add new methods that will format the cgroup file contents for each cgroup/resource limit

src/main/java/com/aws/greengrass/util/platforms/unix/linux/Cgroup.java
```diff
@@ -63,8 +84,18 @@ public Path getSubsystemComponentPath(String componentName) {
...
 
+    /**
+     * get component memory limit path.
+     *
+     * @param componentName componentName
+     * @return memory limit Path
+     */
     public Path getComponentMemoryLimitPath(String componentName) {
-        return getSubsystemComponentPath(componentName).resolve(CGROUP_MEMORY_LIMITS);
+        if (StringUtils.isEmpty(osString)) {
```
MikeDombo, line 94: use `Utils.isEmpty` not StringUtils.

MikeDombo: What does it mean when the osString is empty? What is the criteria for using MEMORY_MAX instead of CGROUP_MEMORY_LIMITS?

src/main/java/com/aws/greengrass/util/platforms/unix/linux/CgroupSubSystemV2.java
```diff
@@ -0,0 +1,37 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.aws.greengrass.util.platforms.unix.linux;
+
+public enum CgroupSubSystemV2 {
+    Memory("", ""), CPU("", ""), Freezer("", ""),
+    Unified("","");
```
MikeDombo, line 10: I think we should have a better way to represent this than just a bunch of empty strings.

src/main/java/com/aws/greengrass/util/platforms/unix/linux/CgroupSubSystemV2.java
```diff
@@ -0,0 +1,37 @@
...
+
+    private String osString;
+    private String mountSrc;
+
+    CgroupSubSystemV2(String osString, String mountSrc) {
```
MikeDombo, line 15: getComponentMemoryLimitPath and those similar methods seem like they should be implemented here instead of checking for an empty string in the Cgroup class.

MikeDombo (changes requested):

src/main/java/com/aws/greengrass/util/platforms/unix/linux/Cgroup.java
ChangxinDong: When the osString is empty means using the CgroupSubSystemV2,because all the osString in the CgroupSubSystemV2 is empty. We use MEMORY_MAX for v2 and CGROUP_MEMORY_LIMITS for v1.

src/integrationtests/java/com/aws/greengrass/integrationtests/lifecyclemanager/GenericExternalServiceIntegTest.java
```diff
@@ -620,6 +640,84 @@ void GIVEN_linux_resource_limits_WHEN_it_changes_THEN_component_runs_with_new_re
...
+        systemResourceController = linuxPlatform.getSystemResourceController();
+        LinuxSystemResourceControllerV2 controllerV2 = (LinuxSystemResourceControllerV2) systemResourceController;
+        Field memoryCgroupField = LinuxSystemResourceControllerV2.class.getSuperclass().getDeclaredField("memoryCgroup");
+        memoryCgroupField.setAccessible(true);
+        Cgroup memoryCgroup = (Cgroup) memoryCgroupField.get(controllerV2);
+        Field subsystem = memoryCgroup.getClass().getDeclaredField("subSystem");
+        subsystem.setAccessible(true);
+        CgroupSubSystemV2 cg = (CgroupSubSystemV2) subsystem.get(memoryCgroup);
+        Field f = cg.getClass().getInterfaces()[0].getDeclaredField("CGROUP_ROOT");
+        setFinalStatic(f, Paths.get(ROOT_PATH_STRING));
```
MikeDombo, line 678: why are you using reflection?

Even in testing, reflection should not be used.

src/main/java/com/aws/greengrass/util/platforms/unix/linux/LinuxSystemResourceControllerV2.java
```diff
@@ -0,0 +1,132 @@
...
+import java.nio.file.Files;
+import java.nio.file.StandardOpenOption;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * https://www.kernel.org/doc/Documentation/cgroup-v2.txt
+ */
+public class LinuxSystemResourceControllerV2 extends LinuxSystemResourceController {
```
MikeDombo, line 25: Does this really need to be a different class? I was hoping that by properly abstracting the cgroup v1 and v2 using `CGroupSubSystemPath` interface, that we would not need the system resource controller to have 2 variants as well.

src/integrationtests/java/com/aws/greengrass/integrationtests/lifecyclemanager/GenericExternalServiceIntegTest.java
ChangxinDong: we are confused about how to do it without reflection. Since the automation test platform has Cgroup V1 being activated, thus we have to simulate Cgroup V2 is active. And because the folders of Cgroup V1 are read-only, therefore we have to change CGROUP_ROOT path to a test path so that the folder/file structures of V2 can be created under that test root path. But due to CGROUP_ROOT is a final field in interface “CGroupSubSystemPath.java”, so without reflection, this path is not able to be reset to the test path. We also tried mock/stub with @InjectionMocks, @Spy, @Mock, but as kernel.lanuch() is the core method called in integration test, it is too deep for stubbing to get effective. Could you please also have a look at this issue, if there is any other approach to change the CGROUP_ROOT path?

src/main/java/com/aws/greengrass/util/platforms/unix/linux/LinuxSystemResourceControllerV2.java
ChangxinDong: We removed the “LinuxSystemResourceControllerV2.java” file according to last comments.

src/integrationtests/java/com/aws/greengrass/integrationtests/lifecyclemanager/GenericExternalServiceIntegTest.java
MikeDombo: What I'd want in this case is that the test is skipped on OSes which do not have Cgroup v2. When running on a device which does have cgroup v2 enabled, then I'd want the v2 test to run and not the v1 test.

src/main/java/com/aws/greengrass/util/platforms/unix/linux/CGroupSubSystemPath.java
```diff
@@ -0,0 +1,120 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+@SuppressFBWarnings(value = "DMI_HARDCODED_ABSOLUTE_FILENAME",
+        justification = "CGroupSubSystemPath virtual filesystem path cannot be relative")
+public interface CGroupSubSystemPath {
```
MikeDombo, line 21: `CGroupSubSystemPaths`

src/main/java/com/aws/greengrass/util/platforms/unix/linux/Cgroup.java
```diff
@@ -5,81 +5,117 @@
...
 /**
- * Represents Linux cgroup v1 subsystems.
+ * Represents Linux cgroup subsystems.
  */
-@SuppressFBWarnings(value = "DMI_HARDCODED_ABSOLUTE_FILENAME", justification = "Cgroup virtual filesystem path "
-        + "cannot be relative")
-public enum Cgroup {
-    Memory("memory"), CPU("cpu,cpuacct"), Freezer("freezer", "freezer");
+public class Cgroup {
+    private final CGroupSubSystemPath subSystem;
```
MikeDombo, line 18: `subSystem` doesn't seem appropriate here. `subSystemPaths` instead?

src/main/java/com/aws/greengrass/util/platforms/unix/linux/Cgroup.java
```diff
@@ -5,81 +5,117 @@
...
 
 /**
- * Represents Linux cgroup v1 subsystems.
+ * Represents Linux cgroup subsystems.
  */
-@SuppressFBWarnings(value = "DMI_HARDCODED_ABSOLUTE_FILENAME", justification = "Cgroup virtual filesystem path "
-        + "cannot be relative")
-public enum Cgroup {
-    Memory("memory"), CPU("cpu,cpuacct"), Freezer("freezer", "freezer");
+public class Cgroup {
```
MikeDombo, line 17: Why does this class exist at all?

All of its methods are just calling into the same methods in `CGroupSubSystemPath`, so why does this class exist at all? Just use `CGroupSubSystemPath` directly if that's what we need.

src/main/java/com/aws/greengrass/util/platforms/unix/linux/CgroupSubSystem.java
```diff
@@ -0,0 +1,169 @@
...
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.StandardOpenOption;
+import java.util.List;
+import java.util.Set;
+
+@SuppressFBWarnings(value = "DMI_HARDCODED_ABSOLUTE_FILENAME",
+        justification = "CgroupSubSystem virtual filesystem path cannot be relative")
+public enum CgroupSubSystem implements CGroupSubSystemPath {
```
MikeDombo, line 22: `CGroupV1`

src/main/java/com/aws/greengrass/util/platforms/unix/linux/CgroupSubSystemV2.java
```diff
@@ -0,0 +1,145 @@
...
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.StandardOpenOption;
+import java.util.List;
+import java.util.Set;
+
+@SuppressFBWarnings(value = "DMI_HARDCODED_ABSOLUTE_FILENAME",
+        justification = "CgroupSubSystemV2 virtual filesystem path cannot be relative")
+public enum CgroupSubSystemV2 implements CGroupSubSystemPath {
```
MikeDombo, line 22: `CGroupV2`

src/main/java/com/aws/greengrass/util/platforms/unix/linux/CgroupSubSystemV2.java
```diff
@@ -0,0 +1,145 @@
...
+
+    @Override
+    public Path getSubsystemGGPath() {
+        return getSubsystemRootPath().resolve(GG_NAMESPACE);
+    }
+
+    @Override
+    public Path getSubsystemComponentPath(String componentName) {
+        return getSubsystemGGPath().resolve(componentName);
+    }
```
MikeDombo, line 44: These implementations are all the same in both V1 and V2. These should be defaults in the interface.

Everything which is _not_ different between V1 and V2 should be in the interface or a common parent class.

src/main/java/com/aws/greengrass/util/platforms/unix/linux/CGroupSubSystemPaths.java
```diff
@@ -0,0 +1,126 @@
...
+        return null;
+    }
+
+    default Path getComponentCpuMaxPath(String componentName) {
+        return null;
+    }
+
+    default Path getCgroupFreezePath(String componentName) {
+        return null;
+    }
```
MikeDombo, line 86: Why are these defaults `null`? If these must be implemented, don't provide a `default`, so then the must be implemented by the subclass.

MikeDombo (commented):
Please comment on this PR with test results for the cgroup tests from a system using cgroup v1 and a system using cgroup v2. Show that the unit and integration tests for this feature are passing on both types of cgroup.

Also post a different comment that explains the manual testing performed to ensure that the feature is working properly for v1 and v2.

src/integrationtests/java/com/aws/greengrass/integrationtests/lifecyclemanager/GenericExternalServiceIntegTest.java
```diff
@@ -751,18 +799,34 @@ void GIVEN_service_starts_up_WHEN_startup_times_out_THEN_timeout_error_code_pers
...
+        byte[] buf2 = Files.readAllBytes(CGroupV1.CPU.getComponentCpuQuotaPath(componentName));
+        byte[] buf3 = Files.readAllBytes(CGroupV1.CPU.getComponentCpuPeriodPath(componentName));
 
         int quota = Integer.parseInt(new String(buf2, StandardCharsets.UTF_8).trim());
         int period = Integer.parseInt(new String(buf3, StandardCharsets.UTF_8).trim());
         int expectedQuota = (int) (cpus * period);
         assertThat(expectedQuota, equalTo(quota));
     }
 
+    private void assertResourceLimits_V2(long memory, double cpus) throws Exception {
```
MikeDombo, line 814: `assertResourceLimitsCgroupV2`

src/main/java/com/aws/greengrass/util/platforms/unix/linux/CGroupSubSystemPaths.java
MikeDombo: Do not just make a change, please explain what you are thinking.

src/main/java/com/aws/greengrass/util/platforms/unix/linux/CGroupV2.java
```diff
@@ -0,0 +1,129 @@
...
+        return getSubsystemComponentPath(componentName).resolve(CPU_MAX);
+    }
+
+    @Override
+    public Path getCgroupFreezePath(String componentName) {
+        return getSubsystemComponentPath(componentName).resolve(CGROUP_FREEZE);
+    }
+
+    @Override
+    public void initializeCgroup(GreengrassService component, LinuxPlatform platform) throws IOException {
```
MikeDombo, line 66: how does this method differ from what is in CgroupV1? Can they share the same implementation? This code looks like it is just duplicated.

src/main/java/com/aws/greengrass/util/platforms/unix/linux/CGroupV2.java
```diff
@@ -0,0 +1,129 @@
...
+        //Enable controllers for root group
+        Files.write(getRootSubTreeControlPath(),
+                CGROUP_SUBTREE_CONTROL_CONTENT.getBytes(StandardCharsets.UTF_8));
+        //Enable controllers for gg group
+        Files.write(getGGSubTreeControlPath(),
+                CGROUP_SUBTREE_CONTROL_CONTENT.getBytes(StandardCharsets.UTF_8));
+    }
+
+    @Override
+    public void handleCpuLimits(GreengrassService component, double cpu) throws IOException {
```
MikeDombo, line 91: Same question here, this looks mostly the same as the V1 implementation, can they share the same implementation?

ChangxinDong: Method handleCpuLimits use different file paths and data format in the  V1 and V2. I have no idea how to make them into one method. Do you have any suggestion?

src/main/java/com/aws/greengrass/util/platforms/unix/linux/CGroupSubSystemPaths.java
ChangxinDong: Oh,they are not necessary any more, I will remove them from the interface.At the beginning,I want to the controller share the same method.

yiwenTS:
Test Case 1: Limitation for memory (V2)
image
image
Test Case 2: Limitation for CPU (V2)
image
Test Case 3: Limitation for Freezer (V2)
image
image

yiwenTS:
Test Case 4: Limitation for memory (V1)
image
image
Test Case 5: Limitation for CPU (V1)
image
Test Case 6: Limitation for Freezer (V1)
image
image

yiwenTS:
Integration test result for Cgroup V2:
image

yiwenTS:
Unit test result for Cgroup V2:
image

junfuchen99:
Feature released in https://github.com/aws-greengrass/aws-greengrass-nucleus/pull/1733

PR #1330, opened 2022-10-10
fix: parse status codes from greengrass cloud exceptions

junfuchen99:
**Description of changes:**
1. Parse status codes from greengrass exceptions instead of relying on its type.

Sometimes when a cloud api call fails due to missing permissions, a generic `GreengrassV2DataException` is thrown instead of a specific `AccessDeniedException`, which causes the error code logic to handle it incorrectly.

**Documentation Checklist:**

**Compatibility Checklist:**


MikeDombo (approved):

MikeDombo:

MikeDombo (approved):

MikeDombo:
We also need to really audit our retry logic in general, and with regards to this change too. As you found, we aren't always retrying when we should, and we also know that the exception may not be what we expect. We should use the http status codes.

shaguptashaikh (approved):

indougnito (approved):

PR #1338, opened 2022-10-17
fix: protect generateFailureMessage from null messages

MikeDombo:
**Issue #, if available:**

**Description of changes:**
`Throwable.getMessage()` can return null, specifically it always returns null for NullPointerException. Ironically, this causes a NPE to be thrown from the StringBuilder's constructor. This change protects this method from null messages and will add the exception classname instead of the message when there is no message.

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/util/Utils.java
+++ b/src/main/java/com/aws/greengrass/util/Utils.java
@@ -193,13 +193,24 @@ public final class Utils {
      * @return String chain of exceptions and messages.
      */
     public static String generateFailureMessage(Throwable t) {
-        StringBuilder failureMessage = new StringBuilder(t.getMessage());
+        if (t == null) {
+            return "No error, null throwable";
+        }
+        StringBuilder failureMessage = new StringBuilder(getMessageFromThrowable(t));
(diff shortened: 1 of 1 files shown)


src/main/java/com/aws/greengrass/util/Utils.java
```diff
@@ -193,15 +193,28 @@ public static String getUltimateMessage(Throwable t) {
...
         while (temp.getCause() != null) {
             temp = temp.getCause();
-            failureMessage.append(". ").append(temp.getMessage());
+            failureMessage.append(". ");
+            getMessageFromThrowable(temp, failureMessage);
         }
         return failureMessage.toString();
     }
 
+    private static void getMessageFromThrowable(Throwable t, StringBuilder builder) {
```
rbattle, line 210: nit: why not return String so you don't need to pass in the builder?

MikeDombo: Yeah thanks, good call. Done.

kucnavya (approved):

PR #1345, opened 2022-10-19
feat(reject deployment): reject stale deployments

devnaruka:
**Issue #, if available:**

**Description of changes:**
1. Enforce deployments to be processed based on creation time sequence. All stale deployments would be rejected.
2. Fleet status update for rejection status would have detailed status.
3. Update unit tests for group to last deployment topic lookup in deployment config.
4. Add integration tests for subgroup deployments out-of-order deployment scenarios.

With the new sub-group deployment feature, a device could receive deployments which override components for one group but these could arrive in different order. The mechanism added in the PR enforce only new deployments to be processed and all stale deployments would get rejected.

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/deployment/SubGroupDeploymentIntegrationTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/deployment/SubGroupDeploymentIntegrationTest.java
@@ -13,9 +13,13 @@ import com.aws.greengrass.config.Topic;
 import com.aws.greengrass.config.Topics;
 import com.aws.greengrass.dependency.State;
 import com.aws.greengrass.deployment.DeploymentQueue;
+import com.aws.greengrass.deployment.DeploymentService;
 import com.aws.greengrass.deployment.DeploymentStatusKeeper;
 import com.aws.greengrass.deployment.DeviceConfiguration;
+import com.aws.greengrass.deployment.IotJobsClientWrapper;
+import com.aws.greengrass.deployment.IotJobsHelper;
 import com.aws.greengrass.deployment.ThingGroupHelper;
+import com.aws.greengrass.deployment.exceptions.DeploymentRejectedException;
 import com.aws.greengrass.deployment.model.Deployment;
 import com.aws.greengrass.helper.PreloadComponentStoreHelper;
 import com.aws.greengrass.integrationtests.BaseITCase;
@@ -23,24 +27,36 @@ import com.aws.greengrass.integrationtests.util.ConfigPlatformResolver;
 import com.aws.greengrass.lifecyclemanager.GreengrassService;
 import com.aws.greengrass.lifecyclemanager.Kernel;
 import com.aws.greengrass.mqttclient.MqttClient;
+import com.aws.greengrass.mqttclient.PublishRequest;
 import com.aws.greengrass.status.FleetStatusService;
+import com.aws.greengrass.status.model.DeploymentInformation;
+import com.aws.greengrass.status.model.FleetStatusDetails;
+import com.aws.greengrass.status.model.Trigger;
 import com.aws.greengrass.testcommons.testutilities.NoOpPathOwnershipHandler;
 import com.aws.greengrass.util.Coerce;
 import com.aws.greengrass.util.Utils;
 import com.fasterxml.jackson.databind.DeserializationFeature;
+import com.fasterxml.jackson.databind.JsonMappingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtensionContext;
 import org.mockito.Mock;
+import software.amazon.awssdk.arns.Arn;
 import software.amazon.awssdk.core.exception.SdkClientException;
+import software.amazon.awssdk.crt.mqtt.QualityOfService;
+import software.amazon.awssdk.iot.iotjobs.model.JobStatus;
+import software.amazon.awssdk.iot.iotjobs.model.UpdateJobExecutionRequest;
+import software.amazon.awssdk.iot.iotjobs.model.UpdateJobExecutionResponse;
+import software.amazon.awssdk.iot.iotjobs.model.UpdateJobExecutionSubscriptionRequest;
 
 import java.io.File;
 import java.io.IOException;
 import java.net.URI;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -49,9 +65,12 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Consumer;
 import java.util.stream.Collectors;
 import javax.annotation.Nonnull;
 
@@ -63,11 +82,21 @@ import static com.aws.greengrass.lifecyclemanager.GreengrassService.SERVICES_NAM
 import static com.aws.greengrass.status.FleetStatusService.FLEET_STATUS_SERVICE_TOPICS;
 import static com.aws.greengrass.testcommons.testutilities.ExceptionLogProtector.ignoreExceptionOfType;
 import static com.aws.greengrass.util.Utils.copyFolderRecursively;
+import static com.github.grantwest.eventually.EventuallyLambdaMatcher.eventuallyEval;
 import static java.nio.file.StandardCopyOption.REPLACE_EXISTING;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.contains;
+import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.notNullValue;
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.anyInt;
+import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.doNothing;
+import static org.mockito.Mockito.spy;
 import static org.mockito.Mockito.when;
 
 
@@ -83,20 +112,27 @@ public class SubGroupDeploymentIntegrationTest extends BaseITCase {
             Arrays.asList(DeploymentCapability.SUBGROUP_DEPLOYMENTS.toString());
 
     private Kernel kernel;
+    private DeviceConfiguration deviceConfiguration;
     private DeploymentQueue deploymentQueue;
     private Path localStoreContentPath;
     @Mock
     private ThingGroupHelper thingGroupHelper;
     @Mock
     private MqttClient mqttClient;
-
+    private IotJobsClientWrapper iotJobsClientWrapper;
     private Map<String, CountDownLatch> groupLatchMap;
+    private Map<String, Map<String, Object>> groupDeploymentStatus;
+    private Map<String, AtomicReference<FleetStatusDetails>> fleetStatusUpdates;
+    private Map<String, UpdateJobExecutionRequest> jobStatusUpdates;
 
     @BeforeEach
     void before(ExtensionContext context) throws Exception {
         ignoreExceptionOfType(context, PackageDownloadException.class);
         ignoreExceptionOfType(context, SdkClientException.class);
 
+        groupDeploymentStatus = new ConcurrentHashMap<>();
+        fleetStatusUpdates = new ConcurrentHashMap<>();
+        jobStatusUpdates = new ConcurrentHashMap<>();
         groupLatchMap = new ConcurrentHashMap<>();
 
         kernel = new Kernel();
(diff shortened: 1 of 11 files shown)


src/integrationtests/java/com/aws/greengrass/integrationtests/BaseITCase.java
```diff
@@ -148,4 +150,14 @@ public static void deleteWindowsTestUser(String username) throws IOException, In
         // Delete test user's home directory
         FileUtils.deleteQuietly(new File("C:\\Users\\" + username));
     }
+
+    @SuppressWarnings({"PMD.SignatureDeclareThrowsException"})
+    public static void overrideInstanceField(Object target, String field, Object value) throws Exception {
```
MikeDombo, line 155: remove now?

src/integrationtests/java/com/aws/greengrass/integrationtests/deployment/SubGroupDeploymentIntegrationTest.java
```diff
@@ -91,12 +109,17 @@ public class SubGroupDeploymentIntegrationTest extends BaseITCase {
...
 
     private Map<String, CountDownLatch> groupLatchMap;
+    private Map<String, Map<String, Object>> groupDeploymentStatus;
+    private Map<String, AtomicReference<FleetStatusDetails>> fleetStatusUpdates;
 
     @BeforeEach
     void before(ExtensionContext context) throws Exception {
         ignoreExceptionOfType(context, PackageDownloadException.class);
         ignoreExceptionOfType(context, SdkClientException.class);
+        ignoreExceptionOfType(context, DeploymentRejectedException.class);
```
MikeDombo, line 119: recommend only doing this ignore in the tests which need it, not in the beforeeach

src/main/java/com/aws/greengrass/deployment/DefaultDeploymentTask.java
```diff
@@ -111,6 +114,20 @@ public DeploymentResult call() throws InterruptedException {
                     .kv("Deployment service config", deploymentServiceConfig.toPOJO().toString())
                     .log("Starting deployment task");
 
+            // Enforce deployment sequence validation using timestamps:
+            // If we have already processed a more recent deployment for a thing group, then we will drop
+            // this deployment and only process it if it is newer (or same) deployment for that thing group.
+            if (isStaleDeployment(deploymentDocument)) {
+                logger.atInfo().setEventType(DEPLOYMENT_TASK_EVENT_TYPE)
+                        .kv("Deployment service config", deploymentServiceConfig.toPOJO().toString())
+                        .log("Nucleus has a newer deployment for '{}' target. Rejecting the deployment.",
```
MikeDombo, line 123: logs should not end with a period, we already append a period to all logs.

src/main/java/com/aws/greengrass/deployment/DefaultDeploymentTask.java
```diff
@@ -111,6 +114,20 @@ public DeploymentResult call() throws InterruptedException {
                     .kv("Deployment service config", deploymentServiceConfig.toPOJO().toString())
                     .log("Starting deployment task");
 
+            // Enforce deployment sequence validation using timestamps:
+            // If we have already processed a more recent deployment for a thing group, then we will drop
+            // this deployment and only process it if it is newer (or same) deployment for that thing group.
+            if (isStaleDeployment(deploymentDocument)) {
+                logger.atInfo().setEventType(DEPLOYMENT_TASK_EVENT_TYPE)
+                        .kv("Deployment service config", deploymentServiceConfig.toPOJO().toString())
```
MikeDombo, line 122: I don't think we should be dumping the entire config, particularly at info level.

src/main/java/com/aws/greengrass/deployment/DefaultDeploymentTask.java
MikeDombo: What do we know about the newer deployment? We should say what we do know, such as the timestamp or the configuration arn of the newer version that's currently applied.

src/main/java/com/aws/greengrass/deployment/DeploymentService.java
```diff
@@ -414,7 +428,21 @@ private void persistGroupToRootComponents(DeploymentDocument deploymentDocument)
                 groupTopics.remove();
             }
         });
+
+        Topics groupLastDeploymentTopics = config.lookupTopics(GROUP_TO_LAST_DEPLOYMENT_TOPICS);
+        groupLastDeploymentTopics.forEach(node -> {
+            Topics groupTopics = (Topics) node;
```
MikeDombo, line 434: don't blindly cast, check using instanceof first

src/main/java/com/aws/greengrass/deployment/DefaultDeploymentTask.java
```diff
@@ -182,6 +199,33 @@ public DeploymentResult call() throws InterruptedException {
...
+        long timestamp = Coerce.toLong(lastDeployment.find(GROUP_TO_LAST_DEPLOYMENT_TIMESTAMP_KEY));
+
+        // if don't have last deployment detail, then its a new deployment
+        if (timestamp == 0 || deploymentDocument.getTimestamp() == null) {
+            return false;
+        }
+
+        // return stale only if new deployment's timestamp is older
+        // than last deployment
+        return deploymentDocument.getTimestamp().compareTo(timestamp) < 0;
```
MikeDombo, line 226: since timestamp is a long, just using < seems more clear to me than using compareTo.

src/integrationtests/java/com/aws/greengrass/integrationtests/BaseITCase.java
```diff
@@ -148,4 +150,14 @@ public static void deleteWindowsTestUser(String username) throws IOException, In
...
+        Field f = target.getClass().getDeclaredField(field);
+        f.setAccessible(true);
+        Field modifiersField = Field.class.getDeclaredField("modifiers");
+        modifiersField.setAccessible(true);
+        modifiersField.setInt(f, f.getModifiers() & ~Modifier.FINAL);
+        f.set(target, value);
```
shaguptashaikh, line 161: Why do you need reflection?

src/integrationtests/java/com/aws/greengrass/integrationtests/BaseITCase.java
devnaruka: I removed it! The branch seems to be stale somehow. Pushed it again.

src/main/java/com/aws/greengrass/deployment/DefaultDeploymentTask.java
```diff
@@ -111,6 +114,20 @@ public DeploymentResult call() throws InterruptedException {
                     .kv("Deployment service config", deploymentServiceConfig.toPOJO().toString())
                     .log("Starting deployment task");
 
+            // Enforce deployment sequence validation using timestamps:
+            // If we have already processed a more recent deployment for a thing group, then we will drop
+            // this deployment and only process it if it is newer (or same) deployment for that thing group.
+            if (isStaleDeployment(deploymentDocument)) {
```
shaguptashaikh, line 120: Wonder if this belongs to the Jobs/Shadow helpers before even enqueuing the deployment, there's some dedupe/replace logic and this seems like another such use case

MikeDombo: I talked to @devnaruka about that, I don't think they have the capability to fail/reject a deployment, all they're able to do is to drop it. In this case, we want to signal the rejection back to the cloud.

src/main/java/com/aws/greengrass/deployment/DefaultDeploymentTask.java
devnaruka: What should we log here for the customers? DeploymentID and configurationArn would that be ok? Do we need anything else?

src/integrationtests/java/com/aws/greengrass/integrationtests/deployment/SubGroupDeploymentIntegrationTest.java
devnaruka: I would need to get the ExtensionContext somehow in the test to set this up .. not sure if that's possible. I will look for it.

MikeDombo: You just add it as a parameter in the test method, it will be given to you.

src/main/java/com/aws/greengrass/deployment/DefaultDeploymentTask.java
MikeDombo: Config arn for the deployment we're rejecting and the config arn for the deployment that we're keeping, I guess?

src/main/java/com/aws/greengrass/deployment/DeploymentService.java
devnaruka: I am using the same pattern as line 423.

MikeDombo: That doesn't mean that it is good :P

Definitely please do check for the type before casting. This protects us in case we change the format, or if the user breaks our config somehow.

src/integrationtests/java/com/aws/greengrass/integrationtests/BaseITCase.java
devnaruka: this is old change and no longer part of the commit, not sure how you are seeing it.

src/main/java/com/aws/greengrass/deployment/DefaultDeploymentTask.java
devnaruka: We would need to parse the deployment configuration to get the groups. We don't parse the configuration in the jobs helper.

To keep the change footprint minimal I kept it here.

src/main/java/com/aws/greengrass/deployment/DefaultDeploymentTask.java
devnaruka: I am removing that, as the config is already dumped right above this check.

src/main/java/com/aws/greengrass/deployment/DefaultDeploymentTask.java
```diff
@@ -111,6 +115,27 @@ public DeploymentResult call() throws InterruptedException {
...
+                logger.atInfo().setEventType(DEPLOYMENT_TASK_EVENT_TYPE)
+                        .log("Nucleus has a newer deployment for '{}' target. Rejecting the deployment",
+                                deploymentDocument.getGroupName());
+
+                Topics lastDeployment =
+                        deploymentServiceConfig.lookupTopics(DeploymentService.GROUP_TO_LAST_DEPLOYMENT_TOPICS)
+                                .lookupTopics(deploymentDocument.getGroupName());
```
MikeDombo, line 128: you don't need 2 calls to lookup topics, just use `lookupTopics(GROUP, getGroupName())`

devnaruka: I didn't know that :)

src/main/java/com/aws/greengrass/deployment/DeploymentService.java
```diff
@@ -402,19 +418,36 @@ private void finishCurrentDeployment() throws InterruptedException {
...
         Topics groupMembershipTopics = config.lookupTopics(GROUP_MEMBERSHIP_TOPICS);
         deploymentGroupTopics.forEach(node -> {
-            Topics groupTopics = (Topics) node;
-            if (groupMembershipTopics.find(groupTopics.getName()) == null
-                    && !groupTopics.getName().startsWith(DEVICE_DEPLOYMENT_GROUP_NAME_PREFIX)
-                    && !groupTopics.getName().equals(LOCAL_DEPLOYMENT_GROUP_NAME)) {
-                logger.info("Removing mapping for thing group " + groupTopics.getName());
-                groupTopics.remove();
+            if (node instanceof Topics) {
+                Topics groupTopics = (Topics) node;
```
MikeDombo, line 425: thanks

src/main/java/com/aws/greengrass/deployment/DeploymentService.java
```diff
@@ -402,19 +418,36 @@ private void finishCurrentDeployment() throws InterruptedException {
...
+        });
+
+        Topics groupLastDeploymentTopics = config.lookupTopics(GROUP_TO_LAST_DEPLOYMENT_TOPICS);
+        groupLastDeploymentTopics.forEach(node -> {
+            if (node instanceof Topics) {
+                Topics groupTopics = (Topics) node;
+                if (groupMembershipTopics.find(groupTopics.getName()) == null && !groupTopics.getName()
+                        .startsWith(DEVICE_DEPLOYMENT_GROUP_NAME_PREFIX) && !groupTopics.getName()
+                        .equals(LOCAL_DEPLOYMENT_GROUP_NAME)) {
+                    logger.info("Removing last deployment information for thing group " + groupTopics.getName());
```
MikeDombo, line 442: I don't think this needs to be info.

devnaruka: At what level this should be logged then?

src/main/java/com/aws/greengrass/deployment/DeploymentService.java
```diff
@@ -402,19 +418,36 @@ private void finishCurrentDeployment() throws InterruptedException {
...
+                    groupTopics.remove();
```
MikeDombo, line 443: What's going on here exactly? Why are we removing the group to last deployment info?

Relatedly, when we find out that we're no longer a member of a group, we should be cleaning up the persisted information about the last deployment from that group.

src/main/java/com/aws/greengrass/deployment/DeploymentService.java
MikeDombo: No, lesser than that. Debug

src/main/java/com/aws/greengrass/deployment/DeploymentService.java
MikeDombo: To make this more readable, let's break it out (and the one above which is doing the same logic) into a method whose name says what we're doing. Because what we're doing is cleaning up old information about groups that this device is no longer a member of.

src/main/java/com/aws/greengrass/deployment/DeploymentService.java
```diff
@@ -427,11 +438,51 @@ private void persistGroupToRootComponents(DeploymentDocument deploymentDocument)
...
+                .replaceAndWait(lastDeploymentDetails);
+
+        // persist group to root packages mapping
         deploymentGroupTopics.lookupTopics(deploymentDocument.getGroupName())
                 .replaceAndWait(deploymentGroupToRootPackages);
         setComponentsToGroupsMapping(deploymentGroupTopics);
     }
 
+    /**
+     * Group memberships for a device can change. If the device is no longer part of a group, then perform cleanup.
```
MikeDombo, line 456: yeah, perfect. Way more clear, thanks!

MikeDombo (dismissed):

src/integrationtests/java/com/aws/greengrass/integrationtests/deployment/SubGroupDeploymentIntegrationTest.java
```diff
@@ -251,6 +329,24 @@ private void verifyServicesRemoved(String... services) {
         }
     }
 
+    private void verifyRejectionStatus(String... groups) {
```
shaguptashaikh, line 332: Is there a way to verify that the Rejected status is reported to Jobs, since that also goes through the MQTT client similar to FSS updates? Not blocking this PR for that but if it should be tested somewhere, in a UAT would be fine too if that's easier

devnaruka: Capturing jobs status was little tricky, but I have added it.

src/main/java/com/aws/greengrass/deployment/DefaultDeploymentTask.java
```diff
@@ -111,6 +115,27 @@ public DeploymentResult call() throws InterruptedException {
...
+
+                String lastDeploymentConfigArn =
+                        Coerce.toString(lastDeployment.find(GROUP_TO_LAST_DEPLOYMENT_CONFIG_ARN_KEY));
+                return new DeploymentResult(DeploymentResult.DeploymentStatus.REJECTED, new DeploymentRejectedException(
+                        String.format("Nucleus has a newer deployment for '%s' "
+                                        + "target deployed by '%s'. Rejecting the deployment from '%s'",
+                                deploymentDocument.getGroupName(), lastDeploymentConfigArn,
+                                deploymentDocument.getConfigurationArn()),
+                        DeploymentErrorCode.REJECTED_STALE_DEPLOYMENT));
+            }
```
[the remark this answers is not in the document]
devnaruka: Done

src/main/java/com/aws/greengrass/deployment/DefaultDeploymentTask.java
```diff
@@ -182,6 +207,31 @@ public DeploymentResult call() throws InterruptedException {
         }
     }
 
+    private boolean isStaleDeployment(DeploymentDocument deploymentDocument) {
```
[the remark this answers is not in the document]
devnaruka: Done

shaguptashaikh (approved):

MikeDombo (approved):

PR #1350, opened 2022-10-26
fix: handle empty recipe content correctly when calculating digest

junfuchen99:
**Description of changes:**
1. Handle empty(corrupted) recipe file correctly when calculating recipe digest while loading plugin components at kernel launch.

Before this change, if a deployed plugin component's locally stored recipe is corrupted and emptied, then kernel launch would fail due to an IllegalArgumentException thrown at `Digest::calculate` and stuck in a restart loop.

After this change, a ServiceLoadException would be thrown and kernel would create a new instance of UnloadableService.

Manually tested the fix.

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/componentmanager/ComponentStore.java
+++ b/src/main/java/com/aws/greengrass/componentmanager/ComponentStore.java
@@ -59,6 +59,8 @@ public class ComponentStore {
     private static final String LOG_KEY_RECIPE_METADATA_FILE_PATH = "RecipeMetadataFilePath";
     private static final String RECIPE_SUFFIX = ".recipe";
 
+    private static final String PLUGIN_LOAD_ERROR_EVENT = "plugin-load-error";
+
     private final NucleusPaths nucleusPaths;
     private final PlatformResolver platformResolver;
     private final RecipeLoader recipeLoader;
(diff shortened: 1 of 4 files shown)


src/main/java/com/aws/greengrass/componentmanager/ComponentStore.java
```diff
@@ -192,7 +193,8 @@ Optional<String> findComponentRecipeContent(@NonNull ComponentIdentifier compone
         }
 
         try {
-            return Optional.of(new String(Files.readAllBytes(recipePath), StandardCharsets.UTF_8));
+            String recipeContent = new String(Files.readAllBytes(recipePath), StandardCharsets.UTF_8);
+            return Utils.isEmpty(recipeContent) ? Optional.empty() : Optional.of(recipeContent);
```
shaguptashaikh, line 197: This will result in the plugin being treated as unloadable but the error message will say that the plugin has been modified - which implies modified by someone intentionally, rather than saying the recipe is corrupt/invalid. Should we have an exception thrown from the validateDigest method that can be used to communicate the problem better?

shaguptashaikh (commented):
How do we plan to test this?

src/main/java/com/aws/greengrass/componentmanager/ComponentStore.java
junfuchen99: updated the pr to clarify the error message

junfuchen99:
> How do we plan to test this?

I updated the unit test to verify the code path is correct. If we think necessary, we can add a UAT to test it end-to-end.

shaguptashaikh (approved):

MikeDombo:

shaguptashaikh (approved):

indougnito (approved):

Issue #1348, opened 2022-10-25
TokenExchange Service


skulljoi:
**Describe the bug**
When using node 18 and the device has support to ipv6.
the token exchange server only start on 127.0.0.1 not in both 127.0.0.1 and ::1.
So then the AWS-SDK-JAVASCRIPT try to use the address from the environment variable that comes with localhost, some times it goes to 127.0.0.1 and some times ::1, when the ::1 is is the one it is not able to get the TOKENS.

**To Reproduce**
Install greengrass in a device that has both protocols ipv4 and ipv6, create a simple component that need to use AWS Tokens, and you will see that it is not able to use.
Please see: https://github.com/aws/aws-sdk-js-v3/issues/4048

**Expected behavior**
GG token exchange should work on both protocols or it could be configurable.

**Actual behavior**
GG Token exchange server only start on 127.0.0.1 when localhost is resolved to ::1 the AWS-SDK could not connect to the container server to grab the credentials to be use.

**Environment**
 - OS: rapibian / armv7l GNU/Linux
 - JDK version: openjdk 11.0.16 2022-07-19
 - Nucleus version: 2.8.1
 - Node version: v18.10.0
 - AWS SDK Javascript version: "@aws-sdk/*": "^3.192.0",

**Additional context**
If you bypass the address like the code bellow it works
```
// FIX FOR GG problem with ipv6
if (process.env.AWS_CONTAINER_CREDENTIALS_FULL_URI) {
	let localhostIpv4 = process.env.AWS_CONTAINER_CREDENTIALS_FULL_URI;
	localhostIpv4 = localhostIpv4.replace('localhost', '127.0.0.1');
	process.env.AWS_CONTAINER_CREDENTIALS_FULL_URI = localhostIpv4;
}
```

E.g. what is the impact of the bug?
It make the component that need to access aws resources with the TOKENS not work.

jcosentino11:
Hi @skulljoi, thanks for reporting the issue! We're aware of it and are currently working on a fix to have TES listen on ipv4 and ipv6. Will update this thread with progress, thanks!

jcosentino11:
[...]

PR #1351, opened 2022-10-31
feat(tes): support dual-stack

jcosentino11:
**Issue #, if available:**
resolves https://github.com/aws-greengrass/aws-greengrass-nucleus/issues/1348

**Description of changes:**
Modifies TES web server to listen on `127.0.0.1` and `::1`, allowing it to handle IPv4 and IPv6 requests.

Presently, if a customer component attempts to call TES over IPv6, it will fail because TES is listening on just IPv4 (see issue linked at top).

Additionally, verified manually in context of containerized lambda function that there was connectivity to TES from ipv4 and ipv6 addresses. 

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/tes/HttpServerImpl.java
+++ b/src/main/java/com/aws/greengrass/tes/HttpServerImpl.java
@@ -34,20 +34,14 @@ public class HttpServerImpl implements Server {
 
     /**
      * HTTP server port, provided by configuration.
+     *
+     * <p>The actual HTTP server port may be different from this,
+     * if configured port is system-assigned (0).
      */
     private final int configuredPort;
 
-    /**
-     * The actual HTTP server port, since the
-     * configured port may be system-assigned (0).
-     */
-    private final int resolvedPort;
-
-    /**
-     * Reference to servers, we attempt to have one
-     * for IPV4 and IPV6, if those stacks are available.
-     */
-    private final List<HttpServer> servers;
+    private HttpServer serverIPv4;
+    private HttpServer serverIPv6;
 
     @Inject
     private final HttpHandler credentialRequestHandler;
@@ -66,20 +60,17 @@ public class HttpServerImpl implements Server {
         this.credentialRequestHandler = credentialRequestHandler;
         this.executorService = executorService;
         this.configuredPort = port;
-        this.servers = createServers(configuredPort);
-        this.resolvedPort = this.servers.get(0).getAddress().getPort();
+        createServers(configuredPort);
     }
 
     /**
      * Create ipv4 and ipv6 servers, both using the specified port.
      *
      * @param port chosen port for all created servers
-     * @return list of http servers
      * @throws IOException if an I/O error occurs during server creation
      * @throws BindException if the chosen port is unavailable
      */
-    private List<HttpServer> createServers(int port) throws IOException {
-        HttpServer serverIPv4 = null;
+    private void createServers(int port) throws IOException {
         InetSocketAddress addrIPv4 = new InetSocketAddress(ADDR_IPV4, port);
         try {
             serverIPv4 = HttpServer.create(addrIPv4, 0);
@@ -92,18 +83,15 @@ public class HttpServerImpl implements Server {
                             + "Proceeding anyway as IPv4 is likely unavailable on the host system");
         }
 
-        HttpServer serverIPv6 = null;
         InetSocketAddress addrIPv6 =
                 new InetSocketAddress(ADDR_IPV6,
-                // use the same port from IPv4 server
-                serverIPv4 == null ? port : serverIPv4.getAddress().getPort());
+                        // use the same port from IPv4 server
+                        serverIPv4 == null ? port : serverIPv4.getAddress().getPort());
         try {
             serverIPv6 = HttpServer.create(addrIPv6, 0);
         } catch (SocketException e) {
             if (canBindToAddress(ADDR_IPV6)) {
-                if (serverIPv4 != null) { // cleanup
-                    stop(serverIPv4);
-                }
+                stop();
                 throw e;
             }
             if (serverIPv4 == null) {
@@ -115,8 +103,6 @@ public class HttpServerImpl implements Server {
                     .log("Unable to bind HTTP server. "
                             + "Proceeding anyway as IPv6 is likely unavailable on the host system");
         }
-
-        return Stream.of(serverIPv4, serverIPv6).filter(Objects::nonNull).collect(Collectors.toList());
     }
 
     private boolean canBindToAddress(String hostname) {
@@ -130,7 +116,7 @@ public class HttpServerImpl implements Server {
 
     @Override
     public void start() {
-        servers.forEach(this::start);
+        getServers().forEach(this::start);
     }
 
     private void start(HttpServer server) {
@@ -141,7 +127,7 @@ public class HttpServerImpl implements Server {
 
     @Override
     public void stop() {
-        servers.forEach(this::stop);
+        getServers().forEach(this::stop);
     }
 
     private void stop(HttpServer server) {
@@ -149,12 +135,22 @@ public class HttpServerImpl implements Server {
     }
 
     List<InetSocketAddress> getServerAddresses() {
-        return servers.stream()
+        return getServers().stream()
                 .map(HttpServer::getAddress)
                 .collect(Collectors.toList());
     }
 
     int getServerPort() {
-        return resolvedPort;
+        return getServers().stream()
+                .findFirst()
+                .map(HttpServer::getAddress)
+                .map(InetSocketAddress::getPort)
+                .orElseThrow(() -> new IllegalStateException("No servers created"));
+    }
+
+    private List<HttpServer> getServers() {
+        return Stream.of(serverIPv4, serverIPv6)
+                .filter(Objects::nonNull)
+                .collect(Collectors.toList());
     }
 }
--- a/src/main/java/com/aws/greengrass/tes/HttpServerImpl.java
+++ b/src/main/java/com/aws/greengrass/tes/HttpServerImpl.java
@@ -87,7 +87,7 @@ public class HttpServerImpl implements Server {
             if (canBindToAddress(ADDR_IPV4)) {
                 throw e;
             }
-            logger.atDebug().cause(e).kv("address", addrIPv4)
+            logger.atInfo().cause(e).kv("address", addrIPv4)
                     .log("Unable to bind HTTP server. "
                             + "Proceeding anyway as IPv4 is likely unavailable on the host system");
         }
@@ -111,7 +111,7 @@ public class HttpServerImpl implements Server {
                 // such as no ports being available.
                 throw e;
             }
-            logger.atDebug().cause(e).kv("address", addrIPv4)
+            logger.atInfo().cause(e).kv("address", addrIPv6)
                     .log("Unable to bind HTTP server. "
                             + "Proceeding anyway as IPv6 is likely unavailable on the host system");
         }
--- a/src/main/java/com/aws/greengrass/tes/HttpServerImpl.java
+++ b/src/main/java/com/aws/greengrass/tes/HttpServerImpl.java
@@ -195,7 +195,7 @@ public class HttpServerImpl implements Server {
                 .findFirst()
                 .map(HttpServer::getAddress)
                 .map(InetSocketAddress::getPort)
-                .orElseThrow(() -> new IllegalStateException("No servers created"));
+                .orElse(configuredPort);
     }
 
     private List<HttpServer> getServers() {
--- a/src/main/java/com/aws/greengrass/tes/HttpServerImpl.java
+++ b/src/main/java/com/aws/greengrass/tes/HttpServerImpl.java
@@ -9,35 +9,121 @@ import com.sun.net.httpserver.HttpHandler;
 import com.sun.net.httpserver.HttpServer;
 
 import java.io.IOException;
+import java.net.BindException;
 import java.net.InetSocketAddress;
+import java.net.Socket;
+import java.util.List;
+import java.util.Objects;
 import java.util.concurrent.ExecutorService;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import javax.inject.Inject;
 
 public class HttpServerImpl implements Server {
-    public static final String URL = "/2016-11-01/credentialprovider/";
-    private static final int STOP_TIMEOUT_SECONDS = 1;
-    private static final String ANY_INTERFACE = "::";
-    private HttpServer server;
+    static final String URL = "/2016-11-01/credentialprovider/";
+    private static final int TIME_TO_WAIT_BEFORE_SHUTDOWN_IN_SECONDS = 1;
+    @SuppressWarnings("PMD.AvoidUsingHardCodedIP")
+    private static final String ADDR_IPV4 = "127.0.0.1";
+    @SuppressWarnings("PMD.AvoidUsingHardCodedIP")
+    private static final String ADDR_IPV6 = "::1";
+
+    /**
+     * HTTP server port, provided by configuration.
+     */
     private final int configuredPort;
+
+    /**
+     * The actual HTTP server port, since the
+     * configured port may be system-assigned (0).
+     */
+    private final int resolvedPort;
+
+    /**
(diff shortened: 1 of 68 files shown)


src/main/java/com/aws/greengrass/tes/HttpServerImpl.java
```diff
@@ -31,25 +28,32 @@ public class HttpServerImpl implements Server {
...
         this.executorService = executorService;
     }
 
     @Override
-    public void start() {
-        httpImpl.createContext(URL, credentialRequestHandler);
-        httpImpl.setExecutor(executorService);
-        httpImpl.start();
+    public void start() throws IOException {
+        server = HttpServer.create(new InetSocketAddress(ANY_INTERFACE, configuredPort), 0);
```
MikeDombo, line 40: so this makes us listen on any interface for both ipv4 and ipv6?

Doesn't this also mean that TES is available on the LAN (we don't want that)?

MikeDombo: Why are we starting a new server within `start()`, that makes start non-idempotent. We could end up leaking server instances it seems.

jcosentino11: > so this makes us listen on any interface for both ipv4 and ipv6?
> 
> Doesn't this also mean that TES is available on the LAN (we don't want that)?

Yes and yes.  So even though we check Authorization header, reason is we don't want to increase attack surface?  
Will revise

> Why are we starting a new server within `start()`, that makes start non-idempotent. We could end up leaking server instances it seems.

No reason, was just removing network binding from the HttpServerImpl constructor, will revert

src/main/java/com/aws/greengrass/tes/HttpServerImpl.java
```diff
@@ -26,30 +57,104 @@ public class HttpServerImpl implements Server {
...
+     * @throws BindException if the chosen port is unavailable
+     */
+    private List<HttpServer> createServers(int port) throws IOException {
+        HttpServer serverIPv4 = null;
+        InetSocketAddress addrIPv4 = new InetSocketAddress(ADDR_IPV4, port);
+        try {
+            serverIPv4 = HttpServer.create(addrIPv4, 0);
+        } catch (SocketException e) {
+            if (canBindToAddress(ADDR_IPV4)) {
+                throw e;
```
MikeDombo, line 88: what happens if we throw here? Will it retry? How do we recover?

src/main/java/com/aws/greengrass/tes/HttpServerImpl.java
```diff
@@ -26,30 +57,104 @@ public class HttpServerImpl implements Server {
...
+            }
+            logger.atDebug().cause(e).kv("address", addrIPv4)
```
MikeDombo, line 90: this sounds more like a warning, or info

src/main/java/com/aws/greengrass/tes/HttpServerImpl.java
```diff
@@ -26,30 +57,104 @@ public class HttpServerImpl implements Server {
...
+                new InetSocketAddress(ADDR_IPV6,
+                // use the same port from IPv4 server
+                serverIPv4 == null ? port : serverIPv4.getAddress().getPort());
+        try {
+            serverIPv6 = HttpServer.create(addrIPv6, 0);
+        } catch (SocketException e) {
+            if (canBindToAddress(ADDR_IPV6)) {
+                if (serverIPv4 != null) { // cleanup
+                    stop(serverIPv4);
+                }
```
MikeDombo, line 106: can we simplify this cleanup logic by just calling `stop()`? and letting stop figure out which server(s) need to stop

src/main/java/com/aws/greengrass/tes/HttpServerImpl.java
jcosentino11: will bump to info

src/main/java/com/aws/greengrass/tes/HttpServerImpl.java
jcosentino11: The thought was to let this bubble up as component ERROR and have the component retry.  In the case where port is 0, it's a good chance that the second time would pass.  (Side question--is there ever a time configured port is not the default of 0?  I know we have a `configuration.port` option defined in TokenExchangeService, but it doesn't appear to be set anywhere?)

If error persists and we get into a BROKEN state, then I imagine recovery would be a redeployment, the customer would need to clear up ports on their device

src/main/java/com/aws/greengrass/tes/HttpServerImpl.java
jcosentino11: yep will rework

src/main/java/com/aws/greengrass/tes/HttpServerImpl.java
MikeDombo: The port may be set by the customer.

I think we'd want to retry indefinitely if the port is 0 because we should always be able to find some port to use.

jcosentino11: Sorry the above applies to the similar throw below for ipv6.  When this line throws it should only be possible if configured port != 0 and is already bound, so something is wrong with the configuration or the port needs to be unblocked (so i guess same question still stands if that port configuration is ever expected to change)

jcosentino11: "should only" since there's a bit of a race---in theory `HttpServer.create(addrIPv4, 0);` could throw when port == 0 when all ports are taken, then `canBindToAddress(ADDR_IPV4)` might return true if a port was freed between those two calls 🤷 

regardless, will add in explicit retrying when port is system-picked and bubble up exception when port is user configured 👍

src/main/java/com/aws/greengrass/tes/HttpServerImpl.java
```diff
@@ -5,51 +5,202 @@
...
+            RetryUtils.runWithRetry(retryConfig,
+                    () -> {
+                        doCreateServers(port);
+                        return null;
+                    },
+                    "create-http-servers",
+                    logger);
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+        } catch (IOException e) {
```
MikeDombo, line 109: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

**Problem**: An exception is caught and re-thrown within the catch clause without any other operations being performed on the exception, making the catch clause unnecessary.

**Fix**: Perform additional operations on the exception that has been caught. You can wrap a standard exception into a custom exception to add additional information or special handling. When you do that, make sure to set the original exception as the cause. Otherwise, you will lose the stack trace and message of the original exception, which will make it difficult to analyze the event that caused the exception.  
[Learn more](https://www.ibm.com/support/pages/best-practice-catching-and-re-throwing-java-exceptions)

src/main/java/com/aws/greengrass/tes/HttpServerImpl.java
```diff
@@ -5,51 +5,202 @@
...
+
+    HttpServerImpl(int port, HttpHandler credentialRequestHandler,
+                   ExecutorService executorService, HttpServerProvider httpServerProvider, SocketFactory socketFactory)
+            throws IOException {
         this.credentialRequestHandler = credentialRequestHandler;
         this.executorService = executorService;
+        this.configuredPort = port;
+        this.httpServerProvider = httpServerProvider;
+        this.socketFactory = socketFactory;
+        createServers(configuredPort);
```
MikeDombo, line 75: having this is the constructor scares me. Creating and starting the server should be moved into `start`.

src/main/java/com/aws/greengrass/tes/HttpServerImpl.java
```diff
@@ -5,51 +5,202 @@
...
+        try {
+            RetryUtils.runWithRetry(retryConfig,
+                    () -> {
+                        doCreateServers(port);
+                        return null;
+                    },
+                    "create-http-servers",
+                    logger);
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
```
MikeDombo, line 108: let this method throw the interrupt up.

src/main/java/com/aws/greengrass/tes/HttpServerImpl.java
```diff
@@ -5,51 +5,202 @@
...
+                .collect(Collectors.toList());
     }
 
     int getServerPort() {
-        return httpImpl.getAddress().getPort();
+        return getServers().stream()
+                .findFirst()
+                .map(HttpServer::getAddress)
+                .map(InetSocketAddress::getPort)
+                .orElseThrow(() -> new IllegalStateException("No servers created"));
```
MikeDombo, line 198: illegal state is a bad exception. Use checked exceptions or return a reasonable default value.

src/test/java/com/aws/greengrass/tes/HttpServerImplTest.java
```diff
@@ -6,90 +6,340 @@
...
 class HttpServerImplTest {
-    private static final String mockResponse = "Hello World";
 
-    private final ExecutorService executorService = Executors.newFixedThreadPool(1);
+    private static final String MOCK_CREDENTIAL_RESPONSE = "Hello World";
+    private static final String[] hosts = {"localhost", "127.0.0.1", "::1"};
 
-    @Mock
-    private HttpHandler mockHttpHandler;
+    ExecutorService executorService = TestUtils.synchronousExecutorService();
```
MikeDombo, line 59: shutdown this executor

jcosentino11: it's a no-op though? will add for completeness though

src/main/java/com/aws/greengrass/tes/HttpServerImpl.java
jcosentino11: I did at one point 😅 
will revise and make sure start is idempotent

jcosentino11:
just confirmed passing on UATs

src/main/java/com/aws/greengrass/tes/TokenExchangeService.java
```diff
@@ -85,14 +85,14 @@ public void postInject() {
...
-    protected void startup() {
+    protected void startup() throws InterruptedException {
         logger.atInfo().addKeyValue(PORT_TOPIC, port).addKeyValue(IOT_ROLE_ALIAS_TOPIC, iotRoleAlias)
                 .log("Attempting to start server at configured port {}", port);
         try {
             validateConfig();
-            server = new HttpServerImpl(port, credentialRequestHandler, executor);
-            server.start();
-            logger.atInfo().log("Started server at port {}", server.getServerPort());
+            server.start(port, credentialRequestHandler);
```
MikeDombo, line 93: Why did you change the signature of `start()`? Keep it just `start()`. Everything which was set in the constructor is fine, leave that alone. Just don't create and start the server in the constructor.

MikeDombo (approved):

src/main/java/com/aws/greengrass/tes/HttpServerImpl.java
```diff
@@ -5,51 +5,203 @@
...
+    }
+
+    @SuppressWarnings({"PMD.AvoidCatchingGenericException", "PMD.AvoidRethrowingException"})
+    private void createServersWithRetry(int port) throws IOException, InterruptedException {
+        RetryUtils.RetryConfig retryConfig = RetryUtils.RetryConfig.builder()
+                .maxAttempt(Integer.MAX_VALUE)
+                .initialRetryInterval(Duration.ofSeconds(1L))
+                .maxRetryInterval(Duration.ofMinutes(1L))
+                .retryableExceptions(Collections.singletonList(SocketException.class))
+                .build();
```
shaguptashaikh, line 140: This doesn't need to be here, you can make this a static constant

src/main/java/com/aws/greengrass/tes/HttpServerImpl.java
```diff
@@ -5,51 +5,203 @@
...
+     * Create ipv4 and ipv6 servers, both using the specified port.
+     *
+     * @param port chosen port for all created servers
+     * @throws IOException if an I/O error occurs during server creation
+     * @throws BindException if the chosen port is unavailable
+     */
+    private void doCreateServers(int port) throws IOException {
+        InetSocketAddress addrIPv4 = new InetSocketAddress(ADDR_IPV4, port);
+        try {
+            serverIPv4 = httpServerProvider.createHttpServer(addrIPv4, 0);
```
shaguptashaikh, line 167: Is it okay to block on this with infinite retries and never even attempt to create the ipv6 server or is the idea that this can only happen when the port is unavailable so the ipv6 server will also fail anyway? Would we instead benefit from configurable retries the default for which could be infinite retries?

src/main/java/com/aws/greengrass/tes/HttpServerImpl.java
```diff
@@ -5,51 +5,203 @@
...
+        return Stream.of(serverIPv4, serverIPv6)
+                .filter(Objects::nonNull)
+                .collect(Collectors.toList());
+    }
+
+    private void createServers(int port) throws IOException, InterruptedException {
+        if (!getServers().isEmpty()) {
+            logger.atInfo().kv("port", port).log("Skipping server creation, servers already running");
+            return;
+        }
```
shaguptashaikh, line 121: How does this work when we need to restart if some configuration for TES has changed?

jcosentino11: new `HttpServerImpl` is created on each `startup`, https://github.com/aws-greengrass/aws-greengrass-nucleus/pull/1351/files#diff-53c817ca085df370edb8025c1ae9981c0e0d62c164ade69a91290289b3e7d573R95, so it'll respond to config changes.  Technically this path won't ever get triggered today, more of a protection against if `HttpServerImpl#start` were called twice with the same instance.

shaguptashaikh: I see, but looks like that happens on service startup for TES, is config expected to change and take effect without service restart? (I know this is not part of this change and how it existed, just curious how this currently works and if have scope for improvement)

src/main/java/com/aws/greengrass/tes/HttpServerImpl.java
jcosentino11: I think we should avoid having TES in a false half-state, where v4/v6 are both supported, but we only create one because the other fails due to port being unavailable.  So if v4 fails when we reasonably believe ipv4 is supported on the customer device, then it wouldn't make sense to try v6.  And inversely where v4 succeeds but v6 fails, we'd want to retry until we get both.

Configurable retries could be beneficial but I'm not sure if we'd be retrying much anyway.  I imagine the chance of a system picking a port that gets swiped before the v6 server is created is pretty low.  Maybe it would be better to fail more quickly to signal to the customer that there's some larger issue going on? (cc: @MikeDombo)

src/main/java/com/aws/greengrass/tes/HttpServerImpl.java
jcosentino11: Ahh, yeah I was just focused on keeping the behavior similar to before, port change taking effect after restart.  I don't see why we couldn't react to port changes in the future.  In the original PR for TES there's a [TODO](https://github.com/aws-greengrass/aws-greengrass-nucleus/pull/240/files#r424732402) for it. Although there's a comment in the code rn saying that [port change shouldn't be allowed](https://github.com/aws-greengrass/aws-greengrass-nucleus/commit/fa3cf396d0001602e291070c970f568f4534eec4#diff-c2b86c6fe185344327806e5cb614804c7e81c6abd7ee63b41f42384fb28d4238R58), curious what that's about

src/main/java/com/aws/greengrass/tes/HttpServerImpl.java
MikeDombo: Would we be able to do a lookup for `localhost` and get the address(es) for it? Instead of always trying to bind to ::1, we should only bind to ::1 and 127.0.01 if both are resolved for localhost address. This way, if a system only has ipv4 we should work fine, and the same if the system only has ipv6.

PR #1353, opened 2022-11-01
feat: [Lifecycle] fix blocking install step

Mirrorxxi:
**Issue #, if available:**
The blocking behavior of handleCurrentStateNew prevents the lifecycle from reacting to lifecycle changes such as requestStop.

**Description of changes:**

This is the manual test result: 
[testResult.zip](https://github.com/aws-greengrass/aws-greengrass-nucleus/files/10000570/testResult.zip)

**Documentation Checklist:**

**Compatibility Checklist:**


src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
```diff
@@ -680,14 +680,13 @@ private void handleCurrentStateErrored(Optional<State> desiredState, State prevS
...
+            // For new, we can interrupt it and cancel the installation when it is requested to shutdown nucleus
             // For both starting and running, make sure we stop first before retrying
+            case NEW:
             case STARTING:
             case RUNNING:
                 internalReportState(State.STOPPING);
                 break;
-            case NEW: // error in installing.
-                internalReportState(State.NEW);
-                break;
```
MikeDombo, line 690: Install should not go through the stopping phase. Stop only runs after `run` or `startup`. Install should be interruptable and using the `close` method of `Exec`.

MikeDombo (changes requested):
This pull request does not solve the issue.

Your change must include a modification in https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/main/src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java#L489.

When `close` is called on the GreengrassService instance and the service is currently executing the install script, the script needs to be interrupted and killed using `Exec.close()`

src/test/resources/com/aws/greengrass/lifecyclemanager/InstallSleepServiceConfig.json
```diff
@@ -0,0 +1,18 @@
+{
```
MikeDombo, line 1: remove all these JSON and YAML files which aren't being used.

src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
```diff
@@ -519,9 +521,23 @@ private void handleCurrentStateNew(Optional<State> desiredState) throws Interrup
             greengrassService.serviceErrored(ee);
         } catch (TimeoutException te) {
             greengrassService.serviceErrored(ComponentStatusCode.INSTALL_TIMEOUT, "Timeout in install");
-        } finally {
-            stopBackingTask();
         }
+
+        asyncFinishAction.set((Object stateEvent) -> {
+            // if a state is reported
```
MikeDombo, line 527: why is this the right async finish action?

For `handleCurrentStateInstalledAsync`, it just uses `stopBackingTask()` which seems more correct to me.

Can you explain why all this code is here?

src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
```diff
@@ -519,9 +521,23 @@ private void handleCurrentStateNew(Optional<State> desiredState) throws Interrup
             greengrassService.serviceErrored(ee);
         } catch (TimeoutException te) {
             greengrassService.serviceErrored(ComponentStatusCode.INSTALL_TIMEOUT, "Timeout in install");
-        } finally {
-            stopBackingTask();
         }
+
```
MikeDombo, line 525: to make this change work, we cannot be blocking on L516

src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
Mirrorxxi: We want to set installation phase as background process, so we make some changes in GenericExternalService. After getting the new State from it's return infomation , then we can do some judgement here. But you pointed out that no changes should be made in GenericExternalService, so the change here doesn't seem to work either.

src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
```diff
@@ -486,41 +486,63 @@ private void handleCurrentStateBroken(Optional<State> desiredState, State previo
...
+                    // Bail out if we're not in the expected state
+                    return;
+                }
+                try {
+                    greengrassService.install();
+                    if (!State.ERRORED.equals(lastReportedState.get()) && !schedule.isDone()) {
+                        internalReportState(State.INSTALLED);
+                        schedule.cancel(true);
+                    }
+                } catch (InterruptedException t) {
```
MikeDombo, line 525: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
```diff
@@ -486,41 +486,64 @@ private void handleCurrentStateBroken(Optional<State> desiredState, State previo
     }
 
     @SuppressWarnings("PMD.AvoidCatchingThrowable")
-    private void handleCurrentStateNew(Optional<State> desiredState) throws InterruptedException {
+    private void handleCurrentStateNew(Optional<State> desiredState,
```
MikeDombo, line 489: rename to `handleCurrentStateNewAsync` because it should now be asynchronous.

src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
```diff
@@ -486,41 +486,64 @@ private void handleCurrentStateBroken(Optional<State> desiredState, State previo
...
+                if (nextDesiredState.isPresent() && nextDesiredState.get().equals(State.INSTALLED)) {
+                    schedule.cancel(false);
+                } else {
+                    schedule.cancel(true);
+                }
+                stopBackingTask();
+                return true;
+            });
+        } else {
+            internalReportState(State.STOPPING);
```
MikeDombo, line 546: I don't think that reporting stopping is what we want. After install is cancelled, it should still be `NEW` since it was never installed.

src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
```diff
@@ -486,41 +486,64 @@ private void handleCurrentStateBroken(Optional<State> desiredState, State previo
...
+                    if (!State.ERRORED.equals(lastReportedState.get()) && !schedule.isDone()) {
+                        internalReportState(State.INSTALLED);
+                        schedule.cancel(true);
+                    }
+                } catch (InterruptedException t) {
+                    logger.atWarn("service-install-interrupted").log("Service interrupted while running install");
+                    Thread.currentThread().interrupt();
+                } catch (Throwable t) {
+                    greengrassService.serviceErrored(t);
+                }
```
MikeDombo, line 530: All the branches after `greengrassService.install()` should be canceling the scheduled task

src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
```diff
@@ -486,41 +486,64 @@ private void handleCurrentStateBroken(Optional<State> desiredState, State previo
...
+            }, LIFECYCLE_INSTALL_NAMESPACE_TOPIC);
+
+            asyncFinishAction.set((stateEvent) -> {
+                // If desiredState is updated
+                Optional<State> nextDesiredState = peekOrRemoveFirstDesiredState(State.NEW);
+                // Don't finish the state handling if the new desiredState is still INSTALLED
+                if (nextDesiredState.isPresent() && nextDesiredState.get().equals(State.INSTALLED)) {
+                    schedule.cancel(false);
+                } else {
+                    schedule.cancel(true);
```
MikeDombo, line 540: why are there 2 branches here?

Using `cancel(true)` should be fine for both. The scheduled task isn't really interruptible anyway, so in practice there shouldn't be any difference between true and false.

src/test/java/com/aws/greengrass/lifecyclemanager/LifecycleTest.java
```diff
@@ -180,12 +180,38 @@ void GIVEN_state_new_WHEN_requestStart_called_THEN_install_invoked() throws Inte
...
+
+        lifecycle.initLifecycleThread();
+        lifecycle.requestStart();
+
+        verify(greengrassService,timeout(1000).atLeastOnce()).install();
+
+        lifecycle.requestStop();
+
+        assertThat(installInterrupted.await(1000, TimeUnit.MILLISECONDS), is(true));
+        verify(greengrassService,timeout(1000)).shutdown();
```
MikeDombo, line 212: I don't want shutdown to be called when we're killing `install`. Shutdown is for stopping `startup`/`run` only.

src/test/java/com/aws/greengrass/lifecyclemanager/LifecycleTest.java
```diff
@@ -180,12 +180,38 @@ void GIVEN_state_new_WHEN_requestStart_called_THEN_install_invoked() throws Inte
         lifecycle.initLifecycleThread();
         lifecycle.requestStart();
 
-        verify(greengrassService, timeout(1000)).install();
+        verify(greengrassService, timeout(1000).atLeastOnce()).install();
```
MikeDombo, line 183: why is this changed? Did you change break this test?

This test should only be calling install one time.

Mirrorxxi: After install(), we will judge whether to report the INSTALLED status according to this condition :
if (!State.ERRORED.equals(lastReportedState.get()) && !schedule.isDone())

In asynchronous operation, we directly set schedule. cancel (true):

Therefore, the execution order of the install thread and schedule thread is unstable. Sometimes the test is called install() only once, sometimes it is called install twice (because INSTALLED was not reported for the first time) . So we made such changes.

MikeDombo: That doesn't sound acceptable at all. The `install` step should only be getting called once. This is an incredibly simple test, there must not be any race conditions in it.

src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
```diff
@@ -93,7 +94,8 @@ public class Lifecycle {
     private final AtomicLong stateGeneration = new AtomicLong();
     private final GreengrassService greengrassService;
 
-
+    // number of times asynchronous methods are called during installation
+    private final AtomicInteger newStateUpdateCount = new AtomicInteger(0);
```
MikeDombo, line 98: None of our other states need this, yet they are able to work asynchronously.

src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
```diff
@@ -486,42 +488,61 @@ private void handleCurrentStateBroken(Optional<State> desiredState, State previo
...
             try {
                 greengrassService.install();
+                if (!State.ERRORED.equals(lastReportedState.get()) && !schedule.isDone()) {
+                    internalReportState(State.INSTALLED);
+                    schedule.cancel(true);
+                }
             } catch (InterruptedException t) {
+                schedule.cancel(true);
                 logger.atWarn("service-install-interrupted").log("Service interrupted while running install");
+                Thread.currentThread().interrupt();
```
junfuchen99, line 529: you don't need this

src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
```diff
@@ -486,42 +488,61 @@ private void handleCurrentStateBroken(Optional<State> desiredState, State previo
...
+                }, timeout, TimeUnit.SECONDS);
+
         replaceBackingTask(() -> {
             if (!State.NEW.equals(getState()) || getStateGeneration().get() != currentStateGeneration) {
                 // Bail out if we're not in the expected state
                 return;
             }
             try {
                 greengrassService.install();
+                if (!State.ERRORED.equals(lastReportedState.get()) && !schedule.isDone()) {
```
junfuchen99, line 522: you don't need to check `!schedule.isDone()` because state will be ERRORED if schedule timed out

src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
```diff
@@ -486,42 +488,61 @@ private void handleCurrentStateBroken(Optional<State> desiredState, State previo
...
-                LIFECYCLE_INSTALL_NAMESPACE_TOPIC, DEFAULT_INSTALL_STAGE_TIMEOUT_IN_SEC);
-
-        try {
-            backingTask.get().get(installTimeOut, TimeUnit.SECONDS);
-            if (!State.ERRORED.equals(lastReportedState.get())) {
-                internalReportState(State.INSTALLED);
+        asyncFinishAction.set((stateEvent) -> {
+            // For the first request, if the stateEvent is not a StateTransitionEvent,
+            // wait again to avoid execution twice in the installation phase.
+            if (0 == newStateUpdateCount.getAndIncrement() && stateEvent instanceof DesiredStateUpdatedEvent) {
```
junfuchen99, line 539: I think you should mirror `handleStateTransitionStartingToRunningAsync`:
1. if a state is reported, then it means the install() has finished, so cancel the schedule and return true;
2. if a desired state is updated, then check if the new desired state is still INSTALLED,
a. if it is, then return false
b. if it isn't, then stopBackingTask and return true

src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
Mirrorxxi: The unit tests related to the installation timeout cannot pass if we don't check it.

src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
Mirrorxxi: Even though making changes like this, it still cannot solve the problem that greengrassService. install() sometimes is called twice (Has an impact on this unit test "LifecycleTest.GIVEN_state_new_WHEN_requestStart_called_THEN_install_invoked"). Is this acceptable ?

MikeDombo: No, this is absolutely not acceptable. Install cannot be called twice when it should only run once.

PR #1361, opened 2022-11-15
test(windows): deflake windows testing, extending the timeout

Mirrorxxi:
**Issue #, if available:**
About "DeploymentTaskIntegrationTest.GIVEN_kernel_with_no_deployment_history_WHEN_new_service_install_breaks_THEN_services_are_rolled_back" , The timeout is 60 in Failure3RollbackDeployment.json . The condition for stopping deployment （the state changes to BROKEN) is to report ERRORED state at least 3 times. The ERRORED state is only reported twice within 60 seconds sometimes, so the deployment is not stopped and rolled back immediately. The next test will be executed after the timeout period expires, so the problem arises when the first deployment reports the third error status. 

**Description of changes:**

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,39 @@
+name: "CodeQL"
+
+on:
+  pull_request:
+    branches: [ "*" ]
+  schedule:
+    - cron: "34 1 * * 5"
+
+jobs:
+  analyze:
+    name: Analyze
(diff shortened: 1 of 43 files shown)


junfuchen99 (commented):
The change looks correct. I'm rerunning the Windows workflow a few times to confirm if this fix the issue.

MikeDombo (approved):

PR #1362, opened 2022-11-17
fix: restart nucleus if deployment removes a plugin

shaguptashaikh:
**Issue #, if available:**
-
**Description of changes:**
If a deployment is about to remove a plugin component, force nucleus restart to unload the plugin's code
Draft PR for discussion and early feedback, fix and test are a little silly, will improve if we decide to take this route

Without this, subsequent deployment of the same plugin can cause unexpected behavior when nucleus hasn't restarted between the the two deployments

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/deployment/bootstrap/BootstrapManager.java
+++ b/src/main/java/com/aws/greengrass/deployment/bootstrap/BootstrapManager.java
@@ -15,6 +15,7 @@ import com.aws.greengrass.deployment.exceptions.DeviceConfigurationException;
 import com.aws.greengrass.deployment.exceptions.ServiceUpdateException;
 import com.aws.greengrass.lifecyclemanager.GreengrassService;
 import com.aws.greengrass.lifecyclemanager.Kernel;
+import com.aws.greengrass.lifecyclemanager.PluginService;
 import com.aws.greengrass.lifecyclemanager.exceptions.InputValidationException;
 import com.aws.greengrass.lifecyclemanager.exceptions.ServiceLoadException;
 import com.aws.greengrass.logging.api.Logger;
@@ -46,6 +47,7 @@ import java.util.NoSuchElementException;
 import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.TimeoutException;
+import java.util.stream.Collectors;
 import javax.annotation.concurrent.NotThreadSafe;
 import javax.inject.Inject;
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/aws/greengrass/deployment/bootstrap/BootstrapManager.java
```diff
@@ -153,6 +158,19 @@ public boolean isBootstrapRequired(Map<String, Object> newConfig)
...
     }
 
+    private boolean willRemovePlugins(Map<String, Object> serviceConfig) {
+        Set<String> pluginsToBeRemoved = kernel.orderedDependencies().stream()
+                .filter(s -> s instanceof PluginService)
+                .filter(s -> !serviceConfig.containsKey(s.getName()))
+                .map(GreengrassService::getName)
+                .collect(Collectors.toSet());
+        if (!pluginsToBeRemoved.isEmpty()) {
+            logger.atDebug().kv("plugins-to-remove", pluginsToBeRemoved)
```
MikeDombo, line 168: info log and put "bootstrap" in it somewhere

MikeDombo (approved):
Need to verify with UAT of course, but this LGTM

src/main/java/com/aws/greengrass/deployment/bootstrap/BootstrapManager.java
```diff
@@ -136,7 +138,10 @@ public boolean isBootstrapRequired(Map<String, Object> newConfig)
             }
         });
         if (componentsRequiresBootstrapTask.isEmpty()) {
-            return nucleusConfigValidAndNeedsRestart;
+            // Force restart if
+            // 1. any nucleus config change requires restart or
+            // 2. if any plugin will be removed in the deployment to ensure plugin cleanup
+            return nucleusConfigValidAndNeedsRestart || willRemovePlugins(serviceConfig);
```
MikeDombo, line 144: so this will return that bootstrap is needed, but will it actually cause it to restart? Don't we need to tell it somewhere that the bootstrap that we want is a restart?

MikeDombo: This definitely could be right, I'm just not confident without real testing.

shaguptashaikh:
> PluginComponentTest.GIVEN_plugin_added_and_removed_WHEN_plugin_added_again_THEN_plugin_is_loaded_into_JVM:255 expected: <null> but was: <com.aws.greengrass.dependency.Context$Value@67301e27>

I think this test needs to be removed and replaced with a UAT because now there's going to be a restart

src/main/java/com/aws/greengrass/deployment/bootstrap/BootstrapManager.java
shaguptashaikh: It'll by default be a restart, we already do this for things like when runWith config changes, which is a similar in that it is not a real bootstrap task as such

MikeDombo (approved):

junfuchen99 (approved):

PR #1368, opened 2022-11-23
fix(sdk-clients): use http client builder to avoid leaks

junfuchen99:
**Description of changes:**
Address same memory leak situations as in https://github.com/aws-greengrass/aws-greengrass-client-device-auth/pull/201.
- Searched for all usages of `.httpClient`and switched to `.httpClientBuilder`
- Deprecated `ProxyUtils::getSdkHttpClient`

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/componentmanager/builtins/GreengrassRepositoryDownloader.java
+++ b/src/main/java/com/aws/greengrass/componentmanager/builtins/GreengrassRepositoryDownloader.java
@@ -234,7 +234,7 @@ public class GreengrassRepositoryDownloader extends ArtifactDownloader {
     }
 
     SdkHttpClient getSdkHttpClient() {
-        return ProxyUtils.getSdkHttpClient();
+        return ProxyUtils.getSdkHttpClientBuilder().build();
     }
 
     private long getContentLengthLong(SdkHttpResponse sdkHttpResponse) {
--- a/src/main/java/com/aws/greengrass/componentmanager/plugins/docker/EcrAccessor.java
+++ b/src/main/java/com/aws/greengrass/componentmanager/plugins/docker/EcrAccessor.java
@@ -57,7 +57,8 @@ public class EcrAccessor {
             region = Coerce.toString(deviceConfiguration.getAWSRegion());
         }
 
-        return EcrClient.builder().httpClient(ProxyUtils.getSdkHttpClient())
+        return EcrClient.builder()
+                .httpClientBuilder(ProxyUtils.getSdkHttpClientBuilder())
                 .region(Region.of(region))
                 .credentialsProvider(lazyCredentialProvider).build();
     }
--- a/src/main/java/com/aws/greengrass/util/ProxyUtils.java
+++ b/src/main/java/com/aws/greengrass/util/ProxyUtils.java
@@ -229,8 +229,9 @@ public final class ProxyUtils {
      * @return httpClient built with a ProxyConfiguration, if a proxy is configured, otherwise
      *         a default httpClient
      *
-     * @deprecated This method creates potential memory leak situations when wrapped by an SDK client.
-     *         Recommend to use <code>ProxyUtils.getSdkHttpClientBuilder</code> instead.
+     * @deprecated Using this method in an SDK client builder would create a non-managed HTTP client, which does not
+     *         close when the SDK client is closed. Recommend to use <code>ProxyUtils.getSdkHttpClientBuilder</code>
+     *         instead.
      *
      * @see <a href="https://github.com/aws-greengrass/aws-greengrass-nucleus/pull/1368">depreacted reason</a>
      *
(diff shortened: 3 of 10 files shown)


src/main/java/com/aws/greengrass/componentmanager/plugins/docker/EcrAccessor.java
```diff
@@ -57,7 +57,8 @@ public EcrClient getClient(String region) {
             region = Coerce.toString(deviceConfiguration.getAWSRegion());
         }
 
-        return EcrClient.builder().httpClient(ProxyUtils.getSdkHttpClient())
+        return EcrClient.builder()
+                .httpClientBuilder(ProxyUtils.getSdkHttpClientBuilder().useIdleConnectionReaper(false))
```
MikeDombo, line 61: instead of setting `idleconnectionreaper` here, let's just set it when we make the builder, set it by default.

src/main/java/com/aws/greengrass/componentmanager/plugins/docker/EcrAccessor.java
```diff
@@ -57,7 +57,8 @@ public EcrClient getClient(String region) {
             region = Coerce.toString(deviceConfiguration.getAWSRegion());
         }
 
-        return EcrClient.builder().httpClient(ProxyUtils.getSdkHttpClient())
```
MikeDombo, line 60: Let's mark `getSdkHttpClient` as deprecated and any public method that returns a client and not a builder.

src/main/java/com/aws/greengrass/componentmanager/plugins/docker/EcrAccessor.java
junfuchen99: updated

src/main/java/com/aws/greengrass/componentmanager/plugins/docker/EcrAccessor.java
junfuchen99: updated

src/main/java/com/aws/greengrass/util/ProxyUtils.java
```diff
@@ -228,6 +228,12 @@ public static void setDeviceConfiguration(DeviceConfiguration deviceConfiguratio
      *
      * @return httpClient built with a ProxyConfiguration, if a proxy is configured, otherwise
      *         a default httpClient
+     *
+     * @deprecated This method creates potential memory leak situations when wrapped by an SDK client.
```
shaguptashaikh, line 232: Instead of saying 'potential memory leak situation' we could call out the exact behavior that leads to the leak

shaguptashaikh (dismissed):

src/main/java/com/aws/greengrass/util/ProxyUtils.java
junfuchen99: updated

shaguptashaikh (dismissed):

src/main/java/com/aws/greengrass/componentmanager/builtins/GreengrassRepositoryDownloader.java
```diff
@@ -234,7 +234,7 @@ private String getArtifactDownloadURL(ComponentIdentifier componentIdentifier, S
     }
 
     SdkHttpClient getSdkHttpClient() {
```
MikeDombo, line 236: this is still returning a _built_ client, not a builder. So is the user of `getSdkHttpClient` handling that fact correctly?

src/main/java/com/aws/greengrass/util/ProxyUtils.java
```diff
@@ -228,6 +228,13 @@ public static void setDeviceConfiguration(DeviceConfiguration deviceConfiguratio
...
      *         a default httpClient
+     *
+     * @deprecated Using this method in an SDK client builder would create a non-managed HTTP client, which does not
+     *         close when the SDK client is closed. Recommend to use <code>ProxyUtils.getSdkHttpClientBuilder</code>
+     *         instead.
+     *
+     * @see <a href="https://github.com/aws-greengrass/aws-greengrass-nucleus/pull/1368">depreacted reason</a>
+     *
      */
     public static SdkHttpClient getSdkHttpClient() {
```
MikeDombo, line 239: `@Deprecated` is an annotation, use it.

src/main/java/com/aws/greengrass/componentmanager/builtins/GreengrassRepositoryDownloader.java
junfuchen99: All users of `getSdkHttpClient` call it in a try with resource directly and not in a SDK builder. I verified that `client.close()` is called after try with resource block finishes.

src/main/java/com/aws/greengrass/util/ProxyUtils.java
junfuchen99: updated

MikeDombo (dismissed):

shaguptashaikh (dismissed):

MikeDombo (approved):

shaguptashaikh (approved):

PR #1372, opened 2022-11-30
fix: cache http client for ggv2 data client properly

junfuchen99:
**Issue #, if available:**

**Description of changes:**
- always keep a cache of http client for ggv2 data client;
- reconfigure http client only when private key or cert or root ca path changes.

We made a recent change in https://github.com/aws-greengrass/aws-greengrass-nucleus/pull/1368 to cache http client for ggv2 data client. Before it, we never closed v2 data client properly. 

However, it exposed a new issue - if other plugins are still using the data client while we are trying to close the client, then their api calls would fail with `java.lang.IllegalStateException: Connection pool shut down`. 

Specifically, when we fetch data client for the first time, we update data plane port with the default port number (if not previously configured), which is a device config change and triggers a client cleanup and causes api call failure.
`thing group helper calls fetchGreengrassV2DataClient -> device config change triggers clean client -> thing group helper fails to use the client and call listThingGroupsForCoreDevice`

Failing E2E tests are now passing.

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/util/GreengrassServiceClientFactory.java
+++ b/src/main/java/com/aws/greengrass/util/GreengrassServiceClientFactory.java
@@ -61,6 +61,12 @@ public class GreengrassServiceClientFactory {
             if (WhatHappened.interiorAdded.equals(what) || WhatHappened.timestampUpdated.equals(what)) {
                 return;
             }
+            if (validString(node, DEVICE_PARAM_ROOT_CA_PATH) || validString(node, DEVICE_PARAM_CERTIFICATE_FILE_PATH)
+                    || validString(node, DEVICE_PARAM_PRIVATE_KEY_PATH)) {
+                logger.atInfo().kv("node", node.getFullName()).log("Closing cached http client for Greengrass v2 "
+                        + "data client due to device config change");
+                cleanHttpClient();
+            }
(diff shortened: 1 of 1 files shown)


src/main/java/com/aws/greengrass/util/GreengrassServiceClientFactory.java
```diff
@@ -184,5 +204,6 @@ private void configureClient(DeviceConfiguration deviceConfiguration) {
             }
         }
         this.greengrassV2DataClient = clientBuilder.build();
+        this.greengrassV2DataClient.close();
```
MikeDombo, line 207: why are you building the client and instantly closing it?

junfuchen99: typo - forgot to remove this line when doing some other testing

MikeDombo (approved):

shaguptashaikh:
Can you link the E2E test that caught this if you have time? I'm just worried that we're accidentally catching these problems in OTF UATs and E2E tests and want us to spend a minute on checking that we have good coverage around reconfiguring, using across plugins etc

shaguptashaikh:
And, did we run the failing E2E test locally to verify this?

PR #1388, opened 2023-01-05
feat(mqtt5): add MQTT 5 publish/subscribe models

MikeDombo:
**Issue #, if available:**

**Description of changes:**
This PR adds classes for publish/subscribe/unsubscribe into the Nucleus so that we will be able to update the MQTT client to support MQTT5 and utilize these models for the public interface. I chose not to extend the existing interfaces and instead create new interfaces because the currently existing interfaces depend on objects that are in the CRT's MQTT 3 implementation such as `QualityOfService` and `MqttMessage`. In order to maintain compatibility with MQTT 3, the MQTT client in Nucleus can be updated to support both `Publish` (v5) and `PublishRequest` (v3) using translation code to copy the common fields. The same will also apply for subscriptions and message receipt events.

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/mqttclient/v5/Publish.java
+++ b/src/main/java/com/aws/greengrass/mqttclient/v5/Publish.java
@@ -1,0 +1,87 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.aws.greengrass.mqttclient.v5;
+
+import lombok.Builder;
+import lombok.NonNull;
+import lombok.Value;
+
+import java.util.List;
+
+@SuppressWarnings({"PMD.ClassWithOnlyPrivateConstructorsShouldBeFinal", "PMD.ExcessiveParameterList"})
+@Value
+public class Publish {
+    @NonNull String topic;
+    @NonNull QOS qos;
+    /**
+     * Retain the message in the cloud MQTT broker (only last message with retain is actually kept).
+     * Subscribers will immediately receive the last retained message when they first subscribe.
+     */
+    boolean retain;
+    byte[] payload;
+    PayloadFormatIndicator payloadFormat;
+    Long messageExpiryIntervalSeconds;
+    String responseTopic;
+    byte[] correlationData;
+    String contentType;
+    List<UserProperty> userProperties;
+
+    // Readonly - Not used for Publish, only set for received messages
+    List<Long> subscriptionIdentifiers;
+
+    @Builder
+    protected Publish(String topic, QOS qos, boolean retain, byte[] payload, PayloadFormatIndicator payloadFormat,
+                      Long messageExpiryIntervalSeconds, String responseTopic, byte[] correlationData,
+                      String contentType, List<UserProperty> userProperties, List<Long> subscriptionIdentifiers) {
+        // Intern the string to deduplicate topic strings in memory
+        this.topic = topic.intern();
+        if (qos == null) {
+            qos = QOS.AT_LEAST_ONCE;
+        }
+        this.qos = qos;
+        this.retain = retain;
+        this.payload = payload;
+        this.payloadFormat = payloadFormat;
+        this.messageExpiryIntervalSeconds = messageExpiryIntervalSeconds;
+        if (responseTopic != null) {
+            responseTopic = responseTopic.intern();
+        }
+        this.responseTopic = responseTopic;
+        this.correlationData = correlationData;
+        if (contentType != null) {
+            contentType = contentType.intern();
+        }
+        this.contentType = contentType;
+        this.userProperties = userProperties;
+        this.subscriptionIdentifiers = subscriptionIdentifiers;
+    }
+
+    public enum PayloadFormatIndicator {
+        /**
+         * The payload is arbitrary binary data.
+         */
+        BYTES(0),
+
+        /**
+         * The payload is a well-formed utf-8 string value.
+         */
+        UTF8(1);
+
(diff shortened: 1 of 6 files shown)


src/main/java/com/aws/greengrass/mqttclient/v5/Unsubscribe.java
```diff
@@ -0,0 +1,19 @@
...
+import lombok.NonNull;
+import lombok.Value;
+
+import java.util.function.Consumer;
+
+@Builder
+@Value
+public class Unsubscribe {
+    @NonNull String topic;
+    @NonNull Consumer<Publish> callback;
```
jbutler, line 18: Should the callback accept a `Publish`? Or do we need an unsuback?

src/main/java/com/aws/greengrass/mqttclient/v5/Subscribe.java
```diff
@@ -0,0 +1,55 @@
...
+@Builder
+@Value
+public class Subscribe {
+    @NonNull String topic;
+    @Builder.Default
+    @NonNull QOS qos = QOS.AT_LEAST_ONCE;
+
+    Boolean noLocal;
+    Boolean retainAsPublished;
+    RetainHandlingType retainHandlingType;
```
jbutler, line 24: Explicit defaults?

src/main/java/com/aws/greengrass/mqttclient/v5/Unsubscribe.java
MikeDombo: this callback isn't being called, rather it is being used to remove the correct callback from the client's internal mapping. I don't think we really care about the unsuback as a user of the Nucleus mqtt client. The nucleus can and should abstract any failures of that.

src/main/java/com/aws/greengrass/mqttclient/v5/Subscribe.java
MikeDombo: defaults are null which is defined in the spec. Want me to set them null here?

src/main/java/com/aws/greengrass/mqttclient/v5/UserProperty.java
```diff
@@ -0,0 +1,20 @@
...
+import lombok.Value;
+
+import java.util.Map;
+
+@Value
+public class UserProperty {
+    String key;
+    String value;
+
+    public static UserProperty fromEntry(Map.Entry<String, String> e) {
```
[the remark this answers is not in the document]
MikeDombo: Sure that's fine.

src/main/java/com/aws/greengrass/mqttclient/v5/Subscribe.java
MikeDombo: updated.

src/main/java/com/aws/greengrass/mqttclient/v5/Unsubscribe.java
MikeDombo: Updated field name and commented.

src/main/java/com/aws/greengrass/mqttclient/v5/Publish.java
```diff
@@ -0,0 +1,87 @@
...
+
+    // Readonly - Not used for Publish, only set for received messages
+    List<Long> subscriptionIdentifiers;
+
+    @Builder
+    protected Publish(String topic, QOS qos, boolean retain, byte[] payload, PayloadFormatIndicator payloadFormat,
+                      Long messageExpiryIntervalSeconds, String responseTopic, byte[] correlationData,
+                      String contentType, List<UserProperty> userProperties, List<Long> subscriptionIdentifiers) {
+        // Intern the string to deduplicate topic strings in memory
+        this.topic = topic.intern();
```
jbutler, line 40: Do we need this here? Or should we intern later at the spooler?

I'm assuming the builder methods will call this constructor, but it seems if we got rid of the string interning that we could just let lombok do all of this for us

MikeDombo: This is the same that we do in the existing publish request for mqtt 3. I think that encapsulating it here like this simplifies things. No one really needs to look at the constructor implementation, just using the builder will all work fine. Keeps the spooler logic somewhat simpler.

MikeDombo: This would apply not only to IPC, but to any usage of publish internally, so doing the interning here is still appropriate IMO.

jbutler (approved):

jcosentino11 (approved):

src/main/java/com/aws/greengrass/mqttclient/v5/UserProperty.java
```diff
@@ -0,0 +1,14 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.aws.greengrass.mqttclient.v5;
+
+import lombok.Value;
+
+@Value
+public class UserProperty {
+    String key;
```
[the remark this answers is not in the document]
MikeDombo: > Followed by a UTF-8 String Pair. The User Property is allowed to appear multiple times to represent multiple name, value pairs. The same name is allowed to appear more than once.
> The Server MUST send all User Properties unaltered in a PUBLISH packet when forwarding the Application Message to a Client [MQTT-3.3.2-17]. The Server MUST maintain the order of User Properties when forwarding the Application Message [MQTT-3.3.2-18].

 
Spec doesn't say non-null. Certainly would be a bit weird, but I don't think it is disallowed.

PR #1389, opened 2023-01-13
fix(mqtt): ensure new clients are created with unused client IDs

MikeDombo:
**Issue #, if available:**

**Description of changes:**
Currently, our MQTT client spawns new connections as needed when each individual connection reaches 50 subscriptions. As subscriptions are then removed, the connections will be closed. Connections will not be closed in a deterministic order. Since connections are not necessarily closed in the reverse order, when we try to spawn a new connection, it determines the next client ID to use by looking only the size of the connection list. This may result in us choosing a client ID which we're already using, thus making the old and new connection unusable due to duplicate client IDs.

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
+++ b/src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
@@ -37,6 +37,7 @@ import software.amazon.awssdk.iot.AwsIotMqttConnectionBuilder;
 
 import java.io.Closeable;
 import java.time.Duration;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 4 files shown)


src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -118,6 +119,7 @@ public class MqttClient implements Closeable {
     private final List<AwsIotMqttClient> connections = new CopyOnWriteArrayList<>();
     private final Map<SubscribeRequest, AwsIotMqttClient> subscriptions = new ConcurrentHashMap<>();
     private final Map<MqttTopic, AwsIotMqttClient> subscriptionTopics = new ConcurrentHashMap<>();
+    private final Set<Integer> activeClientIds = new HashSet<>();
```
shaguptashaikh, line 122: What would we lose if we just tracked a counter an simply increased it, not worrying about if a connection that had a smaller number in the client id got closed?

MikeDombo: lose? nothing really. Just having large client IDs potentially.

shaguptashaikh: Not a strong opinion but I don't think we can have any guarantees on how large the client ids can get anyway, so I'd personally prefer to simplify this and not maintain a set.

MikeDombo: In this case, I prefer keeping the numbers smaller, especially for IoT Policies which may require customers to specify the exact client IDs that will be used. This method would let them specify a finite number of client IDs, say 1-10.

shaguptashaikh (approved):

junfuchen99 (approved):

PR #1395, opened 2023-01-31
fix: increase file download buffer size

MikeDombo:
**Issue #, if available:**

**Description of changes:**
Increase download/write buffer size to 64K from 1K. 1K is too small to be efficient. Particularly since we're writing to disk with SYNC, we want each write to be as efficient as possible and using 64K blocks should be closer to ideal.

Writing 1GB on mac with 1K buffer size takes 28s. Writing using 8KB buffer takes 4s. Using 64KB takes us to ~1.5s.

Writing 1GB on mac without SYNC using a 1K buffer size takes 2.6s.

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/componentmanager/Unarchiver.java
+++ b/src/main/java/com/aws/greengrass/componentmanager/Unarchiver.java
@@ -38,6 +38,9 @@ public class Unarchiver {
     static void unzip(File zipFile, File destDir) throws IOException {
         try (ZipFile zf = new ZipFile(zipFile)) {
             Enumeration<? extends ZipEntry> entries = zf.entries();
+            // IOUtils uses a 4K buffer by default. Using 64K will make things go faster.
+            byte[] buffer = new byte[1024 * 64];
+
             while (entries.hasMoreElements()) {
                 ZipEntry zipEntry = entries.nextElement();
                 File newFile = safeNewZipFile(destDir, zipEntry);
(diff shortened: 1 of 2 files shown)


shaguptashaikh (approved):

src/main/java/com/aws/greengrass/componentmanager/Unarchiver.java
```diff
@@ -38,6 +38,9 @@ public void unarchive(Unarchive method, File toUnarchive, Path unarchiveInto) th
     static void unzip(File zipFile, File destDir) throws IOException {
         try (ZipFile zf = new ZipFile(zipFile)) {
             Enumeration<? extends ZipEntry> entries = zf.entries();
+            // IOUtils uses a 4K buffer by default. Using 64K will make things go faster.
+            byte[] buffer = new byte[1024 * 64];
```
shaguptashaikh, line 42: Nit- could we use the same constant here so if we change it again later you don't have to remember to change here too

shaguptashaikh (approved):

src/main/java/com/aws/greengrass/componentmanager/Unarchiver.java
MikeDombo: Yes, though I don't know that these actually should be coupled. Unarchiving is entirely local, while downloading is over the internet.

shaguptashaikh (approved):

kucnavya (approved):

PR #1399, opened 2023-02-06
fix: truncate job status detail values if over limit

shaguptashaikh:
**Issue #, if available:**
-

**Description of changes:**
Job status detail values have a limit of 1024 chars https://docs.aws.amazon.com/general/latest/gr/iot_device_management.html#iot_device_management_quotas which is currently not respected, can lead to failures in publishing deployment status back. Only truncating these values for Jobs because full messages can still be helpful, and hence are retained for other deployment types and FSS

To avoid failures publishing job status due to large failure messages that exceed the limit.

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/deployment/IotJobsHelper.java
+++ b/src/main/java/com/aws/greengrass/deployment/IotJobsHelper.java
@@ -22,6 +22,7 @@ import com.aws.greengrass.status.FleetStatusService;
 import com.aws.greengrass.status.model.Trigger;
 import com.aws.greengrass.util.Coerce;
 import com.aws.greengrass.util.SerializerFactory;
+import com.aws.greengrass.util.Utils;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import lombok.AccessLevel;
 import lombok.Getter;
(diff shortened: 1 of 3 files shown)


src/main/java/com/aws/greengrass/deployment/IotJobsHelper.java
```diff
@@ -429,6 +429,11 @@ public void updateJobStatus(String jobId, JobStatus status, HashMap<String, Stri
                             .log("Job status updated rejected");
                     gotResponse.completeExceptionally(new Exception(response.message));
                 });
+
+        // Truncate status detail map values longer than the 1024 characters limit
+        statusDetailsMap.entrySet().forEach(e -> statusDetailsMap.put(e.getKey(),
+                e.getValue().substring(0, Math.min(e.getValue().length() - 1, 1023))));
```
MikeDombo, line 435: is the min really needed? If you just use 1023 does substring complain if the string is shorter?

Can you add a Util for truncating to a specific length and including ellipses so that we know it was truncated?

MikeDombo: also, this needs to be length, not length-1.

kucnavya (approved):

PR #1410, opened 2023-02-16
fix: improvements to config loading logic

junfuchen99:
**Description of changes:**
1. Refactor config loading logic.
2. Call tlog recovery logic when `config.tlog` file does not exist.
3. Check the existence of `config.tlog.old` before loading `config.tlog`. If the old tlog exists, then last tlog truncation was interrupted and we should overwrite the `config.tlog`.
4. Add `tlog~` in config read supported extension.
5. Improvements to logging.

To fix a bug in current config recovery logic. Today, if `config.tlog` file does not at nucleus start, no backup files are read into the config and kernel launches with an empty config.

Furthermore, the condition of `config.tlog` missing may be triggered if nucleus is terminated during tlog truncation. Specifically, `config.tlog` is moved to `config.tlog.old` and new configs are written to `config.tlog+` and moved to `config.tlog` only after write is complete.

The `initConfigAndTlog` is also huge. Attempted to refactor it by breaking down to multiple methods .

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
+++ b/src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
@@ -266,13 +266,14 @@ public class KernelLifecycle {
     void initConfigAndTlog() {
         try {
             Path transactionLogPath = nucleusPaths.configPath().resolve(Kernel.DEFAULT_CONFIG_TLOG_FILE);
-            boolean readFromTlog = false;
+            boolean readFromTlog = true;
 
             if (Objects.nonNull(kernelCommandLine.getProvidedConfigPathName())) {
                 // If a config file is provided, kernel will use the provided file as a new base
                 // and ignore existing config and tlog files.
                 // This is used by the nucleus bootstrap workflow
                 kernel.getConfig().read(kernelCommandLine.getProvidedConfigPathName());
+                readFromTlog = false;
             } else {
                 Path bootstrapTlogPath = nucleusPaths.configPath().resolve(Kernel.DEFAULT_BOOTSTRAP_CONFIG_TLOG_FILE);
 
@@ -285,10 +286,10 @@ public class KernelLifecycle {
                 // if config.tlog is valid, read the tlog first because the yaml config file may not be up to date
                 if (transactionTlogValid) {
                     kernel.getConfig().read(transactionLogPath);
-                    readFromTlog = true;
                 } else {
                     // if config.tlog is not valid, try to read config from backup tlogs
                     readConfigFromBackUpTLog(transactionLogPath, bootstrapTlogPath);
+                    readFromTlog = false;
                 }
 
                 // read from external configs
@@ -354,7 +355,7 @@ public class KernelLifecycle {
                 return false;
             }
         }
-        // also delete the new file as part of undoing the effect of incomplete truncation
+        // also delete the new file (config.tlog+) as part of undoing the effect of incomplete truncation
         Path newTlogPath = CommitableFile.getNewFile(transactionLogPath);
         try {
             Files.deleteIfExists(newTlogPath);
--- a/src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
+++ b/src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
@@ -51,6 +51,7 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.nio.file.StandardCopyOption;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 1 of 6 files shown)


src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
```diff
@@ -265,94 +266,55 @@ void initConfigAndTlog(String configFilePath) {
...
+                boolean externalConfigFromCmd = Utils.isNotEmpty(kernelCommandLine.getProvidedInitialConfigPath());
+                if (externalConfigFromCmd) {
+                    externalConfig = Paths.get(kernelCommandLine.getProvidedInitialConfigPath());
                 }
-
+                // not validating its content since the file could be in non-tlog format
                 boolean externalConfigExists = Files.exists(externalConfig);
                 // If there is no tlog, or the path was provided via commandline, read in that file
-                if ((externalConfigFromCmd || !tlogExists) && externalConfigExists) {
+                if ((externalConfigFromCmd || !transactionTlogValid) && externalConfigExists) {
```
kucnavya, line 303: Do we really need `externalConfigFromCmd` (as long as `externalConfigExists` is true, we should read from `externalConfig`) ?

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
```diff
@@ -265,94 +266,55 @@ void initConfigAndTlog(String configFilePath) {
     void initConfigAndTlog() {
         try {
             Path transactionLogPath = nucleusPaths.configPath().resolve(Kernel.DEFAULT_CONFIG_TLOG_FILE);
-            boolean readFromNonTlog = false;
+            boolean readFromTlog = false;
```
kucnavya, line 269: Should the default be true, now that this boolean is the opposite of `readFromNonTlog` ?

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
```diff
@@ -366,6 +328,68 @@ void initConfigAndTlog() {
...
+                    oldTlogPath, transactionLogPath);
+            try {
+                Files.move(oldTlogPath, transactionLogPath, StandardCopyOption.REPLACE_EXISTING);
+            } catch (IOException e) {
+                logger.atError().setCause(e).log("An IO error occurred while moving the old tlog file. Will "
+                        + "attempt to load from backup configs");
+                return false;
+            }
+        }
+        // also delete the new file as part of undoing the effect of incomplete truncation
```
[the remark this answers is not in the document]
junfuchen99: will update

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
junfuchen99: It's not as long as `externalConfigExists` is true

Only when 1) there is no valid `conifg.tlog` or 2) the commandline provides a path using the `--init-config` option

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
```diff
@@ -366,6 +328,68 @@ void initConfigAndTlog() {
...
+     * @param transactionLogPath path to config.tlog
+     * @return true if last tlog truncation was complete or if we are able to undo its effect;
+     *         false only if there was an IO error while undoing its effect (renaming the old tlog file)
+     */
+    private boolean handleIncompleteTlogTruncation(Path transactionLogPath) {
+        Path oldTlogPath = ConfigurationWriter.getOldTlogPath(transactionLogPath);
+        // At the beginning of tlog truncation, the original config.tlog file is moved to config.tlog.old
+        // If .old file exists, then the last truncation was incomplete, so we need to undo its effect by moving it
+        // back to the original location.
+        if (Files.exists(oldTlogPath)) {
```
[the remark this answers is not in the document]
junfuchen99: My understanding is that, if the .old file exists, then the truncation must have failed.

In this case I trust the .old file more because it's directly moved from last valid config. I don't trust the config.tlog as much even if it exists because it must have been created sometime during truncation and more likely to be incomplete.

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
junfuchen99: will update

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
kucnavya: Talked offline. Seems like we read both from `config.tlog` and `externalConfig` when `transactionTlogValid` is true. If that is the intended behavior then this check makes sense to me.

kucnavya (approved):

PR #1413, opened 2023-02-22
chore: exception and logging for incompatible platform claim

alter-mage:
**Issue #, if available:**

**Description of changes:**
Use substring matching to handle the incompatible platform claim and unavailable component exceptions differently, added a new exception class and message to be emitted to console in case of an incompatible platform claim exception

Cloud throws an ItemNotFoundException when resolveComponentCandidates API request is called with an incompatible platform claim. The message which needs to be emitted to the console should different from NoAvailableComponentVersionException in this case

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/componentmanager/ComponentServiceHelper.java
+++ b/src/main/java/com/aws/greengrass/componentmanager/ComponentServiceHelper.java
@@ -5,6 +5,7 @@
 
 package com.aws.greengrass.componentmanager;
 
+import com.aws.greengrass.componentmanager.exceptions.IncompatiblePlatformClaimByComponentException;
 import com.aws.greengrass.componentmanager.exceptions.NoAvailableComponentVersionException;
 import com.aws.greengrass.componentmanager.exceptions.PackagingException;
 import com.aws.greengrass.config.PlatformResolver;
@@ -57,7 +58,7 @@ public class ComponentServiceHelper {
      * @param versionRequirements       component dependents version requirement map
      * @return resolved component version and recipe
      * @throws NoAvailableComponentVersionException if no applicable version available in cloud service
-     * @throws Exception when not able to retrieve greengrasV2DataClient
+     * @throws Exception when not able to retrieve greengrassV2DataClient
      */
     @SuppressWarnings({"PMD.PreserveStackTrace", "PMD.SignatureDeclareThrowsException"})
     ResolvedComponentVersion resolveComponentVersion(String componentName, Semver localCandidateVersion,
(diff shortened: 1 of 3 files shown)


src/main/java/com/aws/greengrass/componentmanager/ComponentServiceHelper.java
```diff
@@ -84,14 +85,21 @@ ResolvedComponentVersion resolveComponentVersion(String componentName, Semver lo
...
+        try (GreengrassV2DataClient greengrassV2DataClient = RetryUtils.runWithRetry(clientExceptionRetryConfig,
                 clientFactory::fetchGreengrassV2DataClient, "get-greengrass-v2-data-client", logger)) {
-            result = greengrasV2DataClient.resolveComponentCandidates(request);
+            result = greengrassV2DataClient.resolveComponentCandidates(request);
         } catch (ResourceNotFoundException e) {
-            logger.atDebug().kv("componentName", componentName).kv("versionRequirements", versionRequirements)
-                    .log("No applicable version found in cloud registry", e);
-            throw new NoAvailableComponentVersionException("No cloud component version satisfies the requirements.",
-                    componentName, versionRequirements);
+            if (e.getMessage().contains("claim platform")) {
```
junfuchen99, line 92: This is not reliable way of checking platform incompatible error. If cloud log message ever changes, this would break.

Instead of log parsing and creating a new error code, I would recommend directly logging the message we get from cloud's `ResourceNotFoundException` into `NoAvailableComponentVersionException`

alter-mage: Yes, I agree but I discussed this case with Michael. The message from cloud is only dumped into the logs and the message which is displayed on the console or through cli is what we build in makeMessage(). Having the same version conflict message for incompatible platform claim is not a clear way to explain what really went wrong.

junfuchen99: The `makeMessage` function also includes the initial message right? so if you throw something like `NoAvailableComponentVersionException("No cloud component version satisfies the requirements due to " + e.getMessage(),
                    componentName, versionRequirements);`, customer can still see it in the console.

alter-mage: Yes, but after the initial message, the customer will see `Check whether the version constraints conflict and that the component exists in your AWS account with a version that matches the version constraints.  If the version constraints conflict, revise deployments to resolve the conflict.` That might give away the wrong information about version conflict and not incompatible platform.

src/main/java/com/aws/greengrass/componentmanager/ComponentServiceHelper.java
```diff
@@ -84,14 +85,25 @@ ResolvedComponentVersion resolveComponentVersion(String componentName, Semver lo
...
```
MikeDombo, line 92: the message is "claim platform"? The grammar should be "claims platform" probably. Are you sure this is accurate?

MikeDombo, line 92: message is allowed to be null, so check for that.
`String message = e.getMessage()`, `if (message == null) <something>`

src/main/java/com/aws/greengrass/componentmanager/ComponentServiceHelper.java
```diff
@@ -84,14 +85,25 @@ ResolvedComponentVersion resolveComponentVersion(String componentName, Semver lo
...
+                logger.atInfo().kv("componentName", componentName).kv("versionRequirements", versionRequirements)
```
MikeDombo, line 93: we don't need to log at info here; this will be logged by the caller

src/main/java/com/aws/greengrass/componentmanager/ComponentServiceHelper.java
```diff
@@ -84,14 +85,25 @@ ResolvedComponentVersion resolveComponentVersion(String componentName, Semver lo
...
+                        .log("The platform claimed by the component is incompatible", e);
+                throw new IncompatiblePlatformClaimByComponentException("The platform claimed by the component is incompatible.",
+                        componentName, platform.name());
```
MikeDombo, line 96: not sure what the platform name even is, we never set that field at all. The platform is a map of attributes, that's what you need to pass here.

MikeDombo: `platformResolver.getCurrentPlatform()`

alter-mage: I got the "name" member from ComponentPlatform class of the Greengrassv2 data plane library, but that's incorrect

src/main/java/com/aws/greengrass/componentmanager/exceptions/IncompatiblePlatformClaimByComponentException.java
```diff
@@ -0,0 +1,38 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.aws.greengrass.componentmanager.exceptions;
+
+import com.vdurmont.semver4j.Requirement;
+
+import java.util.Map;
+
+import static com.aws.greengrass.deployment.errorcode.DeploymentErrorCode.*;
```
junfuchen99, line 12: avoid star imports

src/main/java/com/aws/greengrass/componentmanager/ComponentServiceHelper.java
junfuchen99: How do you plan to test this change? You can either do an integ test and mock the exception returned by cloud. Or you will need to add a UAT.

MikeDombo (approved):

src/main/java/com/aws/greengrass/componentmanager/ComponentServiceHelper.java
```diff
@@ -84,14 +85,32 @@ ResolvedComponentVersion resolveComponentVersion(String componentName, Semver lo
...
+            result = greengrassV2DataClient.resolveComponentCandidates(request);
         } catch (ResourceNotFoundException e) {
-            logger.atDebug().kv("componentName", componentName).kv("versionRequirements", versionRequirements)
-                    .log("No applicable version found in cloud registry", e);
-            throw new NoAvailableComponentVersionException("No cloud component version satisfies the requirements.",
-                    componentName, versionRequirements);
+            if (e.getMessage() == null) {
+                throw new PackagingException(
+                    "Component service returns invalid exception message",
+                    DeploymentErrorCode.RESOLVE_COMPONENT_CANDIDATES_BAD_RESPONSE);
```
junfuchen99, line 95: This is a cloud service error, which we count as a system error. I don't think we have any commitment from cloud to never return a null message, and in this case we know that the exception type is ResourceNotFoundException.

I would recommend to just throw a NoAvailableComponentVersionException with default message if the exception message is null.

junfuchen99 (approved):

MikeDombo (approved):

PR #1421, opened 2023-03-10
fix: re-negotiate with cloud if local recipe metadata contains a mismatched region

junfuchen99:
**Description of changes:**
- validate the region of component version arn stored in recipe metadata of local candidate; if it doesn't match, then renegotiate with cloud.
- added unit test

We are seeing on rare occasions when customer launches nucleus in region A whereas the local recipe metadata of a 1p component is pointing to a different region B for local deployment. This changes allows deployment to renegotiate with cloud and get the correct recipe metadata.

**Documentation Checklist:**

**Compatibility Checklist:**


indougnito (approved):

MikeDombo (commented):
This will fail early, but is there any way we can just _not_ fail? ie. Delete/pretend the metadata doesn't exist and have us download it again from the right region?

MikeDombo (approved):

indougnito (approved):

PR #1425, opened 2023-03-22
fix: use FileChannel and force instead of SYNC for downloads

kucnavya:
**Issue #, if available:**

**Description of changes:**
Writing to disk with SYNC causes a slowdown, ultimately causing the TCP connection reset frequently.
This change uses FileChannel instead and force syncs the underlying FileChannel only at the end.

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/componentmanager/Unarchiver.java
+++ b/src/main/java/com/aws/greengrass/componentmanager/Unarchiver.java
@@ -13,7 +13,8 @@ import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
-import java.nio.file.Files;
+import java.nio.channels.Channels;
+import java.nio.channels.FileChannel;
 import java.nio.file.Path;
 import java.nio.file.StandardOpenOption;
 import java.util.Enumeration;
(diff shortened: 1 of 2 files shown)


src/main/java/com/aws/greengrass/componentmanager/Unarchiver.java
```diff
@@ -50,12 +51,13 @@ static void unzip(File zipFile, File destDir) throws IOException {
                     Utils.createPaths(newFile.getParentFile().toPath());
                     // Only unarchive when the destination file doesn't exist or the file sizes don't match
                     if (!newFile.exists() || zipEntry.getSize() != newFile.length()) {
-                        try (OutputStream fos = Files.newOutputStream(newFile.toPath(),
-                                StandardOpenOption.CREATE,
-                                StandardOpenOption.TRUNCATE_EXISTING,
-                                StandardOpenOption.SYNC);
-                             InputStream is = zf.getInputStream(zipEntry)) {
+                        try (FileChannel fc = FileChannel.open(newFile.toPath(), StandardOpenOption.CREATE,
```
MikeDombo, line 54: need to add StandardOpenOption.WRITE here I think. newOutputStream adds it by default, but FileChannel doesn't.

MikeDombo: fixed it.

MikeDombo (approved):

PR #1437, opened 2023-04-17
fix: remove extra config during deployment rollback

shaguptashaikh:
**Issue #, if available:**
-
**Description of changes:**
If deployment added configuration that is expected to follow the 'Replace' behavior, upon rollback it should be removed, but since rollback is done by performing a merge of the old snapshot, nothing gets removed. This change removes everything from the current Configuration object that is not in the rollback snapshot but its parent node has 'Replace' behavior 

To ensure rollbacks don't leave behind nodes that were newly added by the deployment and prevent scenarios like a faulty lifecycle script left behind still executes after rollback and produces erroneous results 

Tested with unit and user acceptance tests

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/config/ConfigurationReader.java
+++ b/src/main/java/com/aws/greengrass/config/ConfigurationReader.java
@@ -38,6 +38,22 @@ public final class ConfigurationReader {
      */
     public static void mergeTLogInto(Configuration config, Reader reader, boolean forceTimestamp,
                                      Predicate<Node> mergeCondition) throws IOException {
+        mergeTLogInto(config, reader, forceTimestamp, mergeCondition, ConfigurationMode.WITH_VALUES);
+    }
+
+    /**
+     * Merge the given transaction log into the given configuration.
+     *
+     * @param config            configuration to merge into
+     * @param reader            reader of the transaction log to read from
+     * @param forceTimestamp    should ignore if the proposed timestamp is older than current
+     * @param mergeCondition    Predicate that returns true if the provided Topic should be merged and false if not
+     * @param configurationMode Configuration mode
+     * @throws IOException if reading fails
+     */
+    private static void mergeTLogInto(Configuration config, Reader reader, boolean forceTimestamp,
+                                      Predicate<Node> mergeCondition, ConfigurationMode configurationMode)
+            throws IOException {
         try (BufferedReader in = reader instanceof BufferedReader ? (BufferedReader) reader
                 : new BufferedReader(reader)) {
             String l;
@@ -51,7 +67,9 @@ public final class ConfigurationReader {
                         if (mergeCondition != null && !mergeCondition.test(targetTopic)) {
                             continue;
                         }
-                        targetTopic.withNewerValue(tlogline.timestamp, tlogline.value, forceTimestamp);
+                        if (ConfigurationMode.WITH_VALUES.equals(configurationMode)) {
+                            targetTopic.withNewerValue(tlogline.timestamp, tlogline.value, forceTimestamp);
+                        }
                     } else if (WhatHappened.removed.equals(tlogline.action)) {
                         Node n = config.findNode(tlogline.topicPath);
                         if (n == null) {
@@ -97,9 +115,47 @@ public final class ConfigurationReader {
         }
     }
 
-    private static void mergeTLogInto(Configuration c, Path p) throws IOException {
+    private static void mergeTLogInto(Configuration c, Path p, ConfigurationMode configurationMode) throws IOException {
         try (BufferedReader bufferedReader = Files.newBufferedReader(p)) {
-            mergeTLogInto(c, bufferedReader, false, null);
+            mergeTLogInto(c, bufferedReader, false, null, configurationMode);
+        }
+    }
+
+    /**
+     * In place update for the given config object from the given transaction log, adhering to the given update behavior
+     * tree and without losing listeners. Config listeners fire asynchronously as nodes update so if you need the
+     * listeners to wait before all nodes are updated, run this on the context thread to synchronize.
+     *
+     * @param config         configuration to merge into
+     * @param tlogPath       path of the tlog file to read to-be-replaced config from
+     * @param forceTimestamp should ignore if the proposed timestamp is older than current
+     * @param mergeCondition Predicate that returns true if the provided Topic should be merged and false if not
+     * @param tree           Merge behavior hierarchy for the update
+     * @throws IOException if update fails
+     */
+    public static void updateFromTLog(Configuration config, Path tlogPath, boolean forceTimestamp,
+                                      Predicate<Node> mergeCondition, UpdateBehaviorTree tree) throws IOException {
+
+        // Merge tlog into configuration so that nodes to retain get replaced
+        ConfigurationReader.mergeTLogInto(config, tlogPath, forceTimestamp, mergeCondition);
+
+        // Remove nodes from configuration that are not present in the provided tlog and whose
+        // update behavior in the provided tree is 'Replace'
+        ConfigurationReader.discardNodesNotInTLog(config, tlogPath, tree);
+    }
+
+    private static void discardNodesNotInTLog(Configuration config, Path tlogPath, UpdateBehaviorTree tree)
+            throws IOException {
+        try (Context stubContext = new Context()) {
+            Configuration tlogMirror = createFromTLog(stubContext, tlogPath, ConfigurationMode.SKELETON_ONLY);
+
+            config.deepForEach((n, b) -> {
+                if (UpdateBehaviorTree.UpdateBehavior.REPLACE.equals(b) && tlogMirror.findNode(n.path()) == null) {
+                    logger.atTrace().kv("node-to-remove", n.getFullName())
+                            .log("Removing config node not in source tlog");
+                    n.remove();
+                }
+            }, tree);
         }
     }
 
@@ -166,7 +222,35 @@ public final class ConfigurationReader {
      */
     public static Configuration createFromTLog(Context context, Path p) throws IOException {
         Configuration c = new Configuration(context);
-        ConfigurationReader.mergeTLogInto(c, p);
+        mergeTLogInto(c, p, ConfigurationMode.WITH_VALUES);
         return c;
+    }
+
+    /**
+     * Create a Configuration based on a transaction log's path.
+     *
+     * @param context           root context for the configuration
+     * @param p                 path to the transaction log
+     * @param configurationMode Configuration mode
+     * @return Configuration from the transaction log
+     * @throws IOException if reading the transaction log fails
+     */
+    static Configuration createFromTLog(Context context, Path p, ConfigurationMode configurationMode)
+            throws IOException {
+        Configuration c = new Configuration(context);
+        mergeTLogInto(c, p, configurationMode);
+        return c;
+    }
+
+    enum ConfigurationMode {
+        /**
+         * Use when only operating on / traversing key path hierarchy.
+         */
+        SKELETON_ONLY,
+
+        /**
+         * Use when you want a full configuration instance.
+         */
+        WITH_VALUES
     }
 }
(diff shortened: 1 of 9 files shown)


src/main/java/com/aws/greengrass/config/ConfigurationReader.java
```diff
@@ -103,6 +103,38 @@ private static void mergeTLogInto(Configuration c, Path p) throws IOException {
...
+        // Remove extraneous nodes from configuration
+        ConfigurationReader.discardNodesNotInTLog(config, tlogPath, tree);
+    }
+
+    private static void discardNodesNotInTLog(Configuration config, Path tlogPath, UpdateBehaviorTree tree)
+            throws IOException {
+        Configuration tlogMirror = createFromTLog(new Context(), tlogPath);
+
+        config.deepForEach((n, b) -> {
+            if (UpdateBehaviorTree.UpdateBehavior.REPLACE.equals(b) && tlogMirror.findNode(n.path()) == null) {
```
MikeDombo, line 132: technically you should be looking up the behavior for this node in particular right?

src/main/java/com/aws/greengrass/config/ConfigurationReader.java
```diff
@@ -103,6 +103,38 @@ private static void mergeTLogInto(Configuration c, Path p) throws IOException {
...
+        // Merge tlog into configuration so that nodes to retain get replaced
+        ConfigurationReader.mergeTLogInto(config, tlogPath, forceTimestamp, mergeCondition);
+
+        // Remove extraneous nodes from configuration
+        ConfigurationReader.discardNodesNotInTLog(config, tlogPath, tree);
+    }
+
+    private static void discardNodesNotInTLog(Configuration config, Path tlogPath, UpdateBehaviorTree tree)
+            throws IOException {
+        Configuration tlogMirror = createFromTLog(new Context(), tlogPath);
```
MikeDombo, line 129: this requires double the amount of config memory to store this.

MikeDombo: context not closed

src/main/java/com/aws/greengrass/config/ConfigurationReader.java
shaguptashaikh: `b'` is the node's parent's behavior, and parent's behavior determines if the node should be removed so the parent gets 'replaced' as a result.

src/main/java/com/aws/greengrass/config/ConfigurationReader.java
shaguptashaikh: Yeah I don't like it either

src/main/java/com/aws/greengrass/config/ConfigurationReader.java
shaguptashaikh: But I don't know how it will move past the root node this way, I'll know when I test it

src/main/java/com/aws/greengrass/config/ConfigurationReader.java
```diff
@@ -38,6 +38,21 @@ private ConfigurationReader() {
...
+     * Merge the given transaction log into the given configuration.
+     *
+     * @param config            configuration to merge into
+     * @param reader            reader of the transaction log to read from
+     * @param forceTimestamp    should ignore if the proposed timestamp is older than current
+     * @param mergeCondition    Predicate that returns true if the provided Topic should be merged and false if not
+     * @param configurationMode Configuration mode
+     * @throws IOException if reading fails
+     */
+    private static void mergeTLogInto(Configuration config, Reader reader, boolean forceTimestamp,
```
MikeDombo, line 54: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

The cyclomatic complexity of this method is 16. By comparison, 99% of the methods in the CodeGuru reference dataset have a lower cyclomatic complexity. This indicates the method has a high number of decisions and it can make the logic difficult to understand and test. We recommend that you simplify this method or break it into multiple methods.

src/main/java/com/aws/greengrass/config/ConfigurationReader.java
```diff
@@ -51,7 +67,9 @@ public static void mergeTLogInto(Configuration config, Reader reader, boolean fo
                         if (mergeCondition != null && !mergeCondition.test(targetTopic)) {
                             continue;
                         }
-                        targetTopic.withNewerValue(tlogline.timestamp, tlogline.value, forceTimestamp);
+                        if (ConfigurationMode.WITH_VALUES.equals(configurationMode)) {
+                            targetTopic.withNewerValue(tlogline.timestamp, tlogline.value, forceTimestamp);
+                        }
```
MikeDombo, line 72: in all config modes it is still modifying the config, removing nodes on L82 for example. Is that desirable?

src/main/java/com/aws/greengrass/config/ConfigurationReader.java
```diff
@@ -166,7 +219,30 @@ public static boolean validateTlog(Path tlogPath) {
...
+     * @throws IOException if reading the transaction log fails
+     */
+    public static Configuration createFromTLog(Context context, Path p, ConfigurationMode configurationMode)
+            throws IOException {
+        Configuration c = new Configuration(context);
+        mergeTLogInto(c, p, configurationMode);
+        return c;
+    }
+
+    enum ConfigurationMode {
```
MikeDombo, line 242: this is needed to call a public API, so it should be public.

src/main/java/com/aws/greengrass/deployment/activator/DeploymentActivator.java
```diff
@@ -138,4 +138,42 @@ protected UpdateBehaviorTree createDeploymentMergeBehavior(long deploymentTimest
 
         return rootMergeBehavior;
     }
+
+    private UpdateBehaviorTree createRollbackMergeBehavior() {
+        // root: MERGE
+        //   services: MERGE
```
MikeDombo, line 144: we should be removing any services that were removed right?

MikeDombo: Default should be replace when we rollback. What are the few (if any) instances where we actually do want to keep updated/added values?

shaguptashaikh: That should be taken care of through the service dependency tree and `service.remove()` that happens as part of `removeObsoleteServices` part of roll-forward/rollback deployment.

src/main/java/com/aws/greengrass/config/ConfigurationReader.java
shaguptashaikh: It is intentional, but I don't know if I can change this to make it clearer

src/main/java/com/aws/greengrass/config/ConfigurationReader.java
shaguptashaikh: I made the API private instead because for now I don't see the need to make it public

src/main/java/com/aws/greengrass/deployment/activator/DeploymentActivator.java
shaguptashaikh: I could make it replace for services, because I also agree that most things for rollbacks should be replace and only a few things like the private and runtime namespaces should be overridden to merge. But I've left it as 'merge' behavior to be careful

src/main/java/com/aws/greengrass/config/ConfigurationReader.java
shaguptashaikh: I take this back, it needs to be the node, not its parent

src/main/java/com/aws/greengrass/config/ConfigurationReader.java
```diff
@@ -97,9 +115,46 @@ public static void mergeTLogInto(Configuration config, Path tlogPath, boolean fo
...
+     * tree and without losing listeners.
+     *
+     * @param config         configuration to merge into
+     * @param tlogPath       path of the tlog file to read to-be-replaced config from
+     * @param forceTimestamp should ignore if the proposed timestamp is older than current
+     * @param mergeCondition Predicate that returns true if the provided Topic should be merged and false if not
+     * @param tree           Merge behavior hierarchy for the update
+     * @throws IOException if update fails
+     */
+    public static void updateFromTLog(Configuration config, Path tlogPath, boolean forceTimestamp,
```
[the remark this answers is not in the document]
shaguptashaikh: Deployment task thread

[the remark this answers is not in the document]
MikeDombo: I think we need to run it on the context thread so that we don't have partial config changes start firing subscriptions.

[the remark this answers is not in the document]
shaguptashaikh: Rollbacks never had it, but I agree. I've changed rollback config change to run on the publish thread

MikeDombo (approved):

src/main/java/com/aws/greengrass/config/ConfigurationReader.java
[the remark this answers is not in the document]
rbattle: If there are specific threads that should or should not call this, can you add that to the javadoc?

src/main/java/com/aws/greengrass/config/ConfigurationReader.java
```diff
@@ -166,7 +220,30 @@ public static boolean validateTlog(Path tlogPath) {
...
+     */
+    static Configuration createFromTLog(Context context, Path p, ConfigurationMode configurationMode)
+            throws IOException {
+        Configuration c = new Configuration(context);
+        mergeTLogInto(c, p, configurationMode);
+        return c;
+    }
+
+    enum ConfigurationMode {
+        // Use when only operating on / traversing key path hierarchy
```
rbattle, line 244: nit: make this JavaDoc comment so IDEs will show it

src/main/java/com/aws/greengrass/deployment/activator/DeploymentActivator.java
rbattle: question: Do we lack test coverage to make you more confident in making it replace? What is better about merge makes you more confident in that behavior?

src/main/java/com/aws/greengrass/config/ConfigurationReader.java
```diff
@@ -97,9 +115,45 @@ public static void mergeTLogInto(Configuration config, Path tlogPath, boolean fo
...
+     * @param tree           Merge behavior hierarchy for the update
+     * @throws IOException if update fails
+     */
+    public static void updateFromTLog(Configuration config, Path tlogPath, boolean forceTimestamp,
+                                      Predicate<Node> mergeCondition, UpdateBehaviorTree tree) throws IOException {
+
+        // Merge tlog into configuration so that nodes to retain get replaced
+        ConfigurationReader.mergeTLogInto(config, tlogPath, forceTimestamp, mergeCondition);
+
+        // Remove extraneous nodes from configuration
```
rbattle, line 141: can you add context here about what type of nodes could remain in configuration? This comment doesn't tell me anything more than what the "discardNodesNotInTlog" does

src/main/java/com/aws/greengrass/deployment/activator/DeploymentActivator.java
shaguptashaikh: At first it was mostly about causing least disruption and because I know that leaving this as merge doesn't mean we don't remove any service instances that must be removed. Another reason that I just realized - the order for rollback steps where you merge config then shut down and close obsolete services, and then remove the configuration for those services that we currently already follow is more appropriate. If we made everything 'replace' here then we remove config for services before closing them and it could lead to situations like no longer having the shutdown script for a service and not cleaning it up properly.

src/main/java/com/aws/greengrass/config/ConfigurationReader.java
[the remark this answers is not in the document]
shaguptashaikh: It will depend on what you use this for. For instance if you were to use this method on nucleus startup before any services have been started up, then there's no need to synchronize. But I added some notes to the javadoc

MikeDombo (dismissed):

MikeDombo (approved):

src/main/java/com/aws/greengrass/deployment/activator/DeploymentActivator.java
rbattle: thanks for diving into that. That makes sense

rbattle (approved):

PR #1441, opened 2023-04-19
fix: allow new services to wait for nucleus restart before install

alter-mage:
**Issue #, if available:**

**Description of changes:**
While soft shutdown of Nucleus, create new services added to the deployment config and close them before nucleus restart. This prevents nucleus restart (in case of ota deployment/bootstrap steps) to interrupt service lifecycles. As Nucleus resumes default deployment (after all pending bootstrap steps are completed), all services in deployment config will request re-installation.

Nucleus restart was causing component lifecycles to interrupt

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/deployment/activator/KernelUpdateActivator.java
+++ b/src/main/java/com/aws/greengrass/deployment/activator/KernelUpdateActivator.java
@@ -21,8 +21,10 @@ import com.aws.greengrass.util.Utils;
 
 import java.io.IOException;
 import java.nio.file.Path;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import javax.inject.Inject;
 
@@ -31,6 +33,7 @@ import static com.aws.greengrass.deployment.DeploymentConfigMerger.MERGE_CONFIG_
 import static com.aws.greengrass.deployment.bootstrap.BootstrapSuccessCode.REQUEST_REBOOT;
 import static com.aws.greengrass.deployment.bootstrap.BootstrapSuccessCode.REQUEST_RESTART;
 import static com.aws.greengrass.deployment.model.Deployment.DeploymentStage.KERNEL_ROLLBACK;
+import static com.aws.greengrass.lifecyclemanager.GreengrassService.SERVICES_NAMESPACE_TOPIC;
 
 /**
  * Activation and rollback of Kernel update deployments.
@@ -70,9 +73,12 @@ public class KernelUpdateActivator extends DeploymentActivator {
 
         DeploymentDocument deploymentDocument = deployment.getDeploymentDocumentObj();
         KernelLifecycle lifecycle = kernel.getContext().get(KernelLifecycle.class);
+
+        Set<String> servicesConfig = newConfig.get(SERVICES_NAMESPACE_TOPIC) == null ? Collections.emptySet() :
+                ((Map<String, Object>)newConfig.get(SERVICES_NAMESPACE_TOPIC)).keySet();
         // Preserve tlog state before launch directory is updated to reflect ongoing deployment.
         // Wait for all services to close.
-        lifecycle.softShutdown(30);
+        lifecycle.softShutdown(30, servicesConfig);
 
         updateConfiguration(deploymentDocument.getTimestamp(), newConfig);
 
--- a/src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
+++ b/src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
@@ -457,11 +457,24 @@ public class KernelLifecycle {
      * Shutdown all services in dependency order.
      *
      * @param timeoutSeconds timeout seconds for waiting all services to shutdown. Use -1 to wait infinitely.
+     * @param serviceNames Configuration of services yet to be merged
      */
     @SuppressWarnings("PMD.AvoidCatchingThrowable")
-    public void stopAllServices(int timeoutSeconds) {
-        GreengrassService[] d = kernel.orderedDependencies().toArray(new GreengrassService[0]);
+    public void stopAllServices(int timeoutSeconds, Set<String> serviceNames) {
+        HashSet<String> runningServices = kernel.orderedDependencies().stream().map(GreengrassService::getName)
+                .collect(Collectors.toCollection(HashSet::new));
+        ArrayList<GreengrassService> newServices = serviceNames.stream()
+                .filter(service -> !runningServices.contains(service))
+                .collect(Collectors.toCollection(ArrayList::new))
+                .stream()
+                .map(kernel::locateIgnoreError)
+                .collect(Collectors.toCollection(ArrayList::new));
+        logger.atDebug()
+                .kv("new-services", newServices.stream().map(GreengrassService::getName)
+                        .collect(Collectors.toCollection(ArrayList::new))).log();
+        kernel.clearODcache();
 
+        GreengrassService[] d = kernel.orderedDependencies().toArray(new GreengrassService[0]);
         CompletableFuture<?>[] arr = new CompletableFuture[d.length];
         for (int i = d.length - 1; i >= 0; --i) { // shutdown in reverse order
             String serviceName = d[i].getName();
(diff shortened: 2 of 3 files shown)


src/main/java/com/aws/greengrass/deployment/activator/KernelUpdateActivator.java
```diff
@@ -70,9 +73,12 @@ public void activate(Map<String, Object> newConfig, Deployment deployment,
 
         DeploymentDocument deploymentDocument = deployment.getDeploymentDocumentObj();
         KernelLifecycle lifecycle = kernel.getContext().get(KernelLifecycle.class);
+
+        Set<String> servicesConfig = newConfig.get(SERVICES_NAMESPACE_TOPIC) == null ? Collections.emptySet() :
```
shaguptashaikh, line 77: Cal it `services` or `serviceNames` or something that doesn't give an impression that this has all service config like lifecycle, component config, etc.

src/main/java/com/aws/greengrass/deployment/activator/KernelUpdateActivator.java
```diff
@@ -70,9 +73,12 @@ public void activate(Map<String, Object> newConfig, Deployment deployment,
...
+                ((Map<String, Object>)newConfig.get(SERVICES_NAMESPACE_TOPIC)).keySet();
```
shaguptashaikh, line 78: Spacing, format the code with our formatting guidelines for all PRs

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
```diff
@@ -457,11 +457,24 @@ public void startupAllServices() {
      * Shutdown all services in dependency order.
      *
      * @param timeoutSeconds timeout seconds for waiting all services to shutdown. Use -1 to wait infinitely.
+     * @param serviceNames Configuration of services yet to be merged
      */
     @SuppressWarnings("PMD.AvoidCatchingThrowable")
-    public void stopAllServices(int timeoutSeconds) {
-        GreengrassService[] d = kernel.orderedDependencies().toArray(new GreengrassService[0]);
+    public void stopAllServices(int timeoutSeconds, Set<String> serviceNames) {
+        HashSet<String> runningServices = kernel.orderedDependencies().stream().map(GreengrassService::getName)
```
rbattle, line 464: nit: just use `Set<String> runningServices` and `List<GreengrassService> newServices`

also, is this all the services in GG, not just the running services?

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
```diff
@@ -457,11 +457,24 @@ public void startupAllServices() {
...
+                .collect(Collectors.toCollection(HashSet::new));
```
shaguptashaikh, line 465: Use the interface in variable declaration not the implementation, i.e. `Set<String> runningServices =`. Also applies to the next line

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
```diff
@@ -457,11 +457,24 @@ public void startupAllServices() {
...
+        ArrayList<GreengrassService> newServices = serviceNames.stream()
```
shaguptashaikh, line 466: Why does this need to be a list and not a set?

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
```diff
@@ -457,11 +457,24 @@ public void startupAllServices() {
...
+                .filter(service -> !runningServices.contains(service))
+                .collect(Collectors.toCollection(ArrayList::new))
+                .stream()
+                .map(kernel::locateIgnoreError)
```
shaguptashaikh, line 470: You should be able to map then filter and not need to collect and recreate another stream in between. `locateIgnoreError` will simply return the running instance of a service if the service exists IIRC.

rbattle, line 470: nit: why do you even need to filter for "running services"
couldn't you just do something like

```
.map(kernel::locateIgnoreError)
.filter(s -> s.getClass().instanceOf(UnloadableService.class))
.collect(...)
```

you shouldn't have to collect and then stream either, you can chain the filter and map together

shaguptashaikh: Also, why does this only need to be done for new services and not running services, could running services not have a lifecycle script change in the config merge that could trigger its lifecycle listener, why won't we need to wait for that?

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
```diff
@@ -457,11 +457,24 @@ public void startupAllServices() {
...
+                .collect(Collectors.toCollection(ArrayList::new));
+        logger.atDebug()
+                .kv("new-services", newServices.stream().map(GreengrassService::getName)
+                        .collect(Collectors.toCollection(ArrayList::new))).log();
```
shaguptashaikh, line 474: You don't seem to be using the newServices list for anything?

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
indougnito, line 464: Does this new section of code need a comment?

What's actually being done with `newServices`, other than passing it to logger.atDebug()?

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
alter-mage: Yes, tbh, I just need the debug log to avoid PMD violation
locateIgnoreError should create the new services which do not exist and the original loop to close all services should close the newly created ones. This is able to resolve the scenario where a deployment consists of
1. Nucleus OTA update
2. Component with long install lifecycle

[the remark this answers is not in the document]
alter-mage: Tbh, I am logging the new services only to avoid PMD violations
The idea is that, `locateIgnoreError` will create the newServices if they are already not running and the close all services loop afterwards should close all the services (including the newly created). This should suppress the execution of lifecycle steps of new services until an re-installation is triggered after all bootstrap steps are completed.

shaguptashaikh: Then I don't think you need to do this locate stuff in the `stopAllServices` method, you could instead make another method called `initServices` or something with the newConfig object and call it before `stopAllServices`. If you plan to keep it this way then definitely add comments explaining what's being done.

shaguptashaikh: Secondly, in that case you only need to call `locate` on the service list from newConfig because like I said, locate doesn't do anything but return the existing instance for a running service. And you can use a simple forEach loop for that

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
alter-mage: I did not investigate this, will do that now

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
alter-mage: Right, but `locate` will throw an error on services which do not exist, `localIgnoreError` will create the service if it does not exist. I think creating the new service and closing that should silence the dependency init in the lifecycle listener and new services will  get re-installed after the bootstrap steps are completed.

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
shaguptashaikh: I have the answer to my second question, since the running services are in the orderedDependencies list, nucleus shutdown will shut them down and wait as it is.

alter-mage: I was thinking something else, when `newConfig` is merged after `softShutdown`, if an existing service has one of its lifecycle steps modified then that should trigger a `dependencyInit` for the lifecycle listener. This is not something that I investigated.

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
shaguptashaikh: I mean, both `locate` and `locateIgnoreError` will return the existing instance when called on a running service, so you don't need to filter running services out from the list you get from `newConfig`, that's all

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
alter-mage: Oh but the service is closed

shaguptashaikh:
Can you explain why installation of the new services is being triggered when the service isn't even initialized? Doesn't lifecycle state machine only go from new -> installing when you requestInstall or requestStart, and that must be done explicitly for a new service only after it's been initialized through locate/locateIgnoreError?

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
alter-mage: I see, yeah that makes sense

shaguptashaikh (commented):
Also, I think we need an integ test that tries to add a service and ensure it closes before nucleus is shut down. Since all of this happens before the first shutdown for the bootstrap deployment, it should be possible to test in an integ test

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
shaguptashaikh: Maybe what you're saying is possible too, we should check whether that does or doesn't happen

MikeDombo:
Resolved by #1455.

PR #1445, opened 2023-04-20
fix: Allow environment variables to override defaults in the loader

CosmicPenguin:
**Issue #, if available:**

**Description of changes:**

Use := to allow environment variables to override the defaults for GG_ROOT, LOADER_DIR and CONFIG_FILE. This allows the caller to set a GG_ROOT somewhere other than ../../../../ but still use the loader, config and jar. This is useful to allow devices that use signed read-only root filesystems to ship a loader and bootstrap the nucleus on a different persistent filesystem on target.

Targets that use a signed or secure boot process for the root filesystem can ship the loader and use it to bootstrap to a nucleus install somewhere else on target using environment variables set in the systemd service or elsewhere during boot.

I didn't see any loader specific tests but can confirm that the desired behavior is seen both with bash and a posix compatible shell.

**Documentation Checklist:**

**Compatibility Checklist:**


junfuchen99:
Hi @CosmicPenguin , thanks for the contribution! Looks like your commit message is failing the commitlint check because "Allow" is upper-case.

MikeDombo (approved):

indougnito (approved):

MikeDombo:
Reverting in #1458. This change broke the Nucleus loader.

PR #1451, opened 2023-04-27
fix(windows): persist spaces in windows jvm options

junfuchen99:
**Description of changes:**
- Allow persisting spaces in JVM options on Windows. e.g. `-DPath="C:\Program Files\"`

If not, nucleus would fail to restart/set up as system service if there are spaces in provided jvm options.

This change essentially wraps all JVM options with double quotes when persisting to `launch.param`.

Testing will be done in UAT.

We decided to only implement this fix for Windows, because Linux kernel always wraps ${JVM_OPTIONS} with single quotes, so there is no way to escape the spaces correctly by only modifying the `launch.param`. Modifying the loader script would break older nucleus versions if downgrading nucleus.

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
+++ b/src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
@@ -16,6 +16,7 @@ import com.aws.greengrass.componentmanager.models.ComponentRecipe;
 import com.aws.greengrass.config.CaseInsensitiveString;
 import com.aws.greengrass.config.ChildChanged;
 import com.aws.greengrass.config.Node;
+import com.aws.greengrass.config.PlatformResolver;
 import com.aws.greengrass.config.Topic;
 import com.aws.greengrass.config.Topics;
 import com.aws.greengrass.config.Validator;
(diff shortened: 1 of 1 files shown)


MikeDombo (commented):
How does this work with the bootstrap and loader script? When we bootstrap, we do an echo into the launch param file.

junfuchen99:
> How does this work with the bootstrap and loader script? When we bootstrap, we do an echo into the launch param file.

Just verified with a UAT. For jvm option like `-Dpath="c a"`, it will persist in launch.param as `"-Dpath=c a"`, which works with loader.cmd. 

It also persists in nucleus config so when nucleus bootstraps, it echoes the correct launch.param to the new directory.
```
{"TS":2,"TP":["services","aws.greengrass.Nucleus","configuration","jvmOptions"],"W":"changed","V":"\"-Dlog.fmt=TEXT\" \"-Dlog.level=INFO\" \"-Dlog.store=FILE\" \"-Dpath=c a\""}
```

Xuqing888 (approved):

indougnito (approved):

PR #1455, opened 2023-05-01
fix: remove deps topic listener during service close

alter-mage:
**Issue #, if available:**

**Description of changes:**
Remove the service dependency topic watcher when closing the service. On a high level, while Nucleus is performing bootstrap and requesting restarts/reboots when required, removing the watcher should prevent new services from requesting installation, causing their lifecycle steps to be executed and possibly interrupted by restarts/reboots.

When device receives a deployment with bootstrap steps, the deployment is activated via a KernelUpdateActivator which stops all services, applies bootstrap, requests nucleus restart/device reboot, requests new services to be started.
Nucleus collects the new config to be merged as a map of config keys and objects and closes all services before applying the merge. The config merge invokes various subscribers to react to the change in the config.
We let inbuilt java methods take care of the ordering in which service config needs to be updated
https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/e447acd4e0946a1c188d5bf4876de626a727cf28/src/main/java/com/aws/greengrass/config/Topics.java#L316-L320

The service `main` is dependent on all the new services for each deployment and registers a dependency topic listener whenever the new config for `main` is being merged
https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/e447acd4e0946a1c188d5bf4876de626a727cf28/src/main/java/com/aws/greengrass/lifecyclemanager/GreengrassService.java#L199-L210

Any change in the config path `services/<service-name>/dependencies` triggers a subscription which loads all of `service-name` (mostly `main`) dependency services.
If a dependency service config is updated after `main` loads that service then a lifecycle change of that service will trigger a re-installation and cause its lifecycle steps to start executing
https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/e447acd4e0946a1c188d5bf4876de626a727cf28/src/main/java/com/aws/greengrass/lifecyclemanager/GenericExternalService.java#L129-L136C3

This is not a desirable experience as services should only be started once all bootstrap steps have been executed and Nucleus explicitly asks for the services to be started.

**Documentation Checklist:**

**Compatibility Checklist:**


src/main/java/com/aws/greengrass/lifecyclemanager/GreengrassService.java
```diff
@@ -449,6 +449,7 @@ protected CompletableFuture<Void> close(boolean waitForDependers) {
                         logger.error("Interrupted waiting for dependers to exit");
                     }
                 }
+                externalDependenciesTopic.remove();
```
junfuchen99, line 452: As communicated offline, directly calling `externalDependenciesTopic.remove()` will remove the entire node. You should just remove the watcher

src/main/java/com/aws/greengrass/config/Topic.java
```diff
@@ -53,6 +54,27 @@ public Topic subscribe(Subscriber s) {
...
+     * This is the preferred way to get a value from a configuration. Instead of {@code setValue(configValue.getOnce())}
+     * use {@code configValue.get((nv,ov)->setValue(nv)) }
+     * This way, every change to the config file will get forwarded to the object.
+     *</p>
+     *
+     * @param s subscriber
+     * @param service service subscribing to dependency topic
+     * @return this topic
+     */
+    public Topic subscribe(Subscriber s, GreengrassService service) {
```
MikeDombo, line 69: why are you adding this? This is highly specific logic for services which should not exist in the config.

shaguptashaikh: This is oddly placed in the Topic class, should be in GreengrassService

src/integrationtests/java/com/aws/greengrass/integrationtests/deployment/DeploymentConfigMergingTest.java
```diff
@@ -17,14 +19,12 @@
 import com.aws.greengrass.integrationtests.BaseITCase;
 import com.aws.greengrass.integrationtests.ipc.IPCTestUtils;
 import com.aws.greengrass.integrationtests.util.ConfigPlatformResolver;
-import com.aws.greengrass.lifecyclemanager.GenericExternalService;
-import com.aws.greengrass.lifecyclemanager.GlobalStateChangeListener;
-import com.aws.greengrass.lifecyclemanager.GreengrassService;
-import com.aws.greengrass.lifecyclemanager.Kernel;
+import com.aws.greengrass.lifecyclemanager.*;
```
shaguptashaikh, line 22: No wildcard imports, undo this change and import specific additional things you need

src/integrationtests/java/com/aws/greengrass/integrationtests/deployment/DeploymentConfigMergingTest.java
```diff
@@ -217,6 +234,45 @@ void GIVEN_kernel_running_single_service_WHEN_merge_changes_service_THEN_service
         }
     }
 
+    @Test
+    void GIVEN_kernel_running_single_service_WHEN_merge_config_with_bootstrap_THEN_dependency_watcher_is_removed()
```
shaguptashaikh, line 238: Why were we not able to make a test for the kernel update workflow and have to try the regular deployment merge?

src/integrationtests/java/com/aws/greengrass/integrationtests/deployment/DeploymentConfigMergingTest.java
alter-mage: yep!

src/integrationtests/java/com/aws/greengrass/integrationtests/deployment/DeploymentConfigMergingTest.java
shaguptashaikh: Ok, looks like you are trying to merge something with bootstrap so it should take the bootstrap route, but what is this test doing? I don't see any verifications after the merge is performed

src/main/java/com/aws/greengrass/config/Topic.java
alter-mage: You mean the new subscribe method or the topic watchers?

src/main/java/com/aws/greengrass/lifecyclemanager/GreengrassService.java
```diff
@@ -84,6 +86,8 @@ public class GreengrassService implements InjectionActions {
 
     // dependencies that are explicitly declared by customer in config store.
     private final Topic externalDependenciesTopic;
+    @Getter @Setter
+    private Watcher externalDependenciesTopicWatcher;
```
shaguptashaikh, line 90: Will there always be just one watcher?

src/integrationtests/java/com/aws/greengrass/integrationtests/deployment/DeploymentConfigMergingTest.java
alter-mage: I was using this test to investigate the root cause of this problem. Since KernelUpdateActivator eventually calls System.exit, I dont think I can use this integ test for any verification and planning to remove this. The plan is to instead test the E2E with a UAT. But I have kept the test here for now as a draft.

src/main/java/com/aws/greengrass/lifecyclemanager/GreengrassService.java
alter-mage: There could be more, but removing just the one watcher which is registered when the service is initialized should be sufficient

src/main/java/com/aws/greengrass/config/Topic.java
shaguptashaikh: Both

shaguptashaikh: Why does this need to be done this way? Why can this code not live in the service class?

src/main/java/com/aws/greengrass/lifecyclemanager/GreengrassService.java
shaguptashaikh: There could be more watchers for dependencies? and if so, why is removing just one sufficient?

src/test/java/com/aws/greengrass/lifecyclemanager/GreengrassServiceTest.java
```diff
@@ -116,4 +131,35 @@ void GIVEN_service_WHEN_dependencies_change_THEN_service_dependencies_updates()
...
+        serviceConfiguration.read(configPath);
+        Topics root = serviceConfiguration.getRoot();
+        Topics serviceTopics = spy(root.findTopics(SERVICES_NAMESPACE_TOPIC, "B"));
+        Topic externalDependencyTopic = mock(Topic.class);
+
+        doReturn(context).when(serviceTopics).getContext();
+        doReturn(externalDependencyTopic).when(serviceTopics).createLeafChild(eq(SERVICE_DEPENDENCIES_NAMESPACE_TOPIC));
+        doReturn(externalDependencyTopic).when(externalDependencyTopic).dflt(any());
+        doReturn(Collections.emptyList()).when(externalDependencyTopic).getOnce();
+        lenient().doReturn(externalDependencyTopic).when(externalDependencyTopic).subscribe(any());
```
shaguptashaikh, line 149: Why lenient? if this is specifically being mocked for this test, then it either gets invoked or it doesn't

alter-mage: I always run into a `Strict stubbing argument mismatch` if I do not use lenient

alter-mage: Okay, I dont think this is required at all

src/main/java/com/aws/greengrass/lifecyclemanager/GreengrassService.java
alter-mage: I have not come across a situation where there were more than one watchers for a service dependency topic, I did not exhaustively look at situations on this. However, it seemed like the watcher which was registered during the initial watcher registration still being alive when services were closed was causing all the problems. I also wanted to make minimal changes on this.

alter-mage: But if there is a recommendation that the entire Watchers set should be emptied when a service is being closed then that can be done as well

src/main/java/com/aws/greengrass/config/Topic.java
alter-mage: I wanted to maintain the original definition of `.subscribe` while also allowing `GreengrassService` to store a reference of the dependency watcher

src/test/java/com/aws/greengrass/lifecyclemanager/GreengrassServiceTest.java
```diff
@@ -116,4 +131,35 @@ void GIVEN_service_WHEN_dependencies_change_THEN_service_dependencies_updates()
...
+        serviceConfiguration.read(configPath);
+        Topics root = serviceConfiguration.getRoot();
+        Topics serviceTopics = spy(root.findTopics(SERVICES_NAMESPACE_TOPIC, "B"));
+        Topic externalDependencyTopic = mock(Topic.class);
+
+        doReturn(context).when(serviceTopics).getContext();
+        doReturn(externalDependencyTopic).when(serviceTopics).createLeafChild(eq(SERVICE_DEPENDENCIES_NAMESPACE_TOPIC));
+        doReturn(externalDependencyTopic).when(externalDependencyTopic).dflt(any());
+        doReturn(Collections.emptyList()).when(externalDependencyTopic).getOnce();
+//        doReturn(externalDependencyTopic).when(externalDependencyTopic).subscribe(any());
```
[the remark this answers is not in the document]
alter-mage: I meant to delete that :D

indougnito (dismissed):

src/test/java/com/aws/greengrass/lifecyclemanager/GreengrassServiceTest.java
```diff
@@ -57,6 +70,8 @@ void beforeEach() throws IOException, URISyntaxException, ServiceLoadException {
         Path configPath = Paths.get(this.getClass().getResource("services.yaml").toURI());
         context = spy(new Context());
         context.put(Kernel.class, kernel);
+        ExecutorService executorService = Executors.newCachedThreadPool();
```
MikeDombo, line 73: do not create a new pool.

If you must, then make sure it gets shutdown.

src/test/java/com/aws/greengrass/lifecyclemanager/GreengrassServiceTest.java
```diff
@@ -74,6 +90,7 @@ void beforeEach() throws IOException, URISyntaxException, ServiceLoadException {
     @AfterEach
     void afterEach() throws IOException {
         context.close();
+        executorService.shutdown();
```
MikeDombo, line 93: shutdownnow

alter-mage: Got it, I used shutdownNow before but modified to shutdown
Is there a reason to prefer shutdownNow over shutdown?

MikeDombo: shutdown now kills anything that's currently running, we don't want anything stray to keep running.

indougnito (approved):

src/test/java/com/aws/greengrass/lifecyclemanager/GreengrassServiceTest.java
```diff
@@ -116,4 +133,34 @@ void GIVEN_service_WHEN_dependencies_change_THEN_service_dependencies_updates()
...
+        AtomicBoolean watcherRemoved = new AtomicBoolean(false);
+        doAnswer((Answer<Void>) invocationOnMock -> {
+            watcherRemoved.set(true);
+            return null;
+        }).when(externalDependencyTopic).remove(any());
+
+        service.close().get(30, TimeUnit.SECONDS);
+
+        //THEN
+        assertTrue(watcherRemoved.get());
```
shaguptashaikh, line 164: Does `service.close()` ever get undone expecting the service lifecycle for the same service instance to progress as usual? Or does it always only happen during nucleus shutdown or service removal?

shaguptashaikh: The reason I'm asking is to see whether something needs to be tested for that

shaguptashaikh (approved):

junfuchen99 (approved):
Nice work getting this across the finish line

PR #1465, opened 2023-05-10
feat: allow local deployment cancellation

junfuchen99:
**Description of changes:**
- necessary nucleus changes to support local deployment cancellation.
- update device sdk snapshot

Will be covered by tests in CLI repo. Nucleus changes here are minimal and do not require additional testings.

CLI will offer a deployment object with `isCancelled = true`. When DeploymentService processes the deployment it calls `cancelCurrentDeployment` which handles deployment cancellation regardless of deployment type. The only change required is making sure the deployment status is persisted after cancellation takes effect.

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/software/amazon/awssdk/aws/greengrass/GeneratedAbstractCancelLocalDeploymentOperationHandler.java
+++ b/src/main/java/software/amazon/awssdk/aws/greengrass/GeneratedAbstractCancelLocalDeploymentOperationHandler.java
@@ -1,0 +1,27 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package software.amazon.awssdk.aws.greengrass;
+
+import software.amazon.awssdk.aws.greengrass.model.CancelLocalDeploymentRequest;
+import software.amazon.awssdk.aws.greengrass.model.CancelLocalDeploymentResponse;
+import software.amazon.awssdk.eventstreamrpc.OperationContinuationHandler;
+import software.amazon.awssdk.eventstreamrpc.OperationContinuationHandlerContext;
+import software.amazon.awssdk.eventstreamrpc.OperationModelContext;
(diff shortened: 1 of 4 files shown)


src/main/java/software/amazon/awssdk/aws/greengrass/GeneratedAbstractCancelLocalDeploymentOperationHandler.java
```diff
@@ -0,0 +1,27 @@
...
+package software.amazon.awssdk.aws.greengrass;
+
+import software.amazon.awssdk.aws.greengrass.model.CancelLocalDeploymentRequest;
+import software.amazon.awssdk.aws.greengrass.model.CancelLocalDeploymentResponse;
+import software.amazon.awssdk.eventstreamrpc.OperationContinuationHandler;
+import software.amazon.awssdk.eventstreamrpc.OperationContinuationHandlerContext;
+import software.amazon.awssdk.eventstreamrpc.OperationModelContext;
+import software.amazon.awssdk.eventstreamrpc.model.EventStreamJsonMessage;
+
+public abstract class GeneratedAbstractCancelLocalDeploymentOperationHandler extends
```
MikeDombo, line 15: do we want/need this? Cloud deployments do not have an equivalent. Cloud deployments are canceled when a new deployment comes in.

MikeDombo (approved):

src/main/java/software/amazon/awssdk/aws/greengrass/GeneratedAbstractCancelLocalDeploymentOperationHandler.java
junfuchen99: Discussed offline.

Today, when customer cancels a cloud deployment, they call`aws greengrassv2 cancel-deployment --deployment-id deploymentId` and we are trying to bring parity to that CLI experience, which will require a new CLI command.

shaguptashaikh (approved):

PR #1477, opened 2023-05-26
feat: split startable services

vabhasin:
**Issue #, if available:**
MQTT Spool fails to locate the specified DiskSpooler Implementation as a Plugin Service, as the logic to find and load Plugins is after we start all the startable services. Startables include the MqttProxyIPCService which needs the Spooler implementation plugin. This behavior is only observed in Integration testing as we look for plugin service in Class path.
**Description of changes:**
Split the startable services and start everything except MqttProxyIPCService before plugins and start MQTT service after the plugins.

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
+++ b/src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
@@ -111,7 +111,9 @@ public class KernelLifecycle {
     @Setter(AccessLevel.PACKAGE)
     private List<Class<? extends Startable>> startables = Arrays.asList(IPCEventStreamService.class,
             AuthorizationService.class, ConfigStoreIPCService.class, LifecycleIPCService.class,
-            PubSubIPCService.class, MqttProxyIPCService.class, ComponentMetricIPCService.class);
+            PubSubIPCService.class, ComponentMetricIPCService.class);
+    @Setter(AccessLevel.PACKAGE)
+    private List<Class<? extends Startable>> mqttStartable = Collections.singletonList(MqttProxyIPCService.class);
     @Getter
     private ConfigurationWriter tlog;
     private GreengrassService mainService;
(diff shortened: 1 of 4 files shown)


src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
```diff
@@ -152,6 +154,11 @@ public void launch() {
         // referenced by main/dependencies of main
         final Queue<String> autostart = findBuiltInServicesAndPlugins(); //NOPMD
         loadPlugins();
+
+        for (Class<? extends Startable> c : mqttStartable) {
+            kernel.getContext().get(c).startup();
+        }
+
```
MikeDombo, line 161: you aren't loading mqttstartable

MikeDombo (commented):
Does this fully address the initial problem you were trying to resolve?

shaguptashaikh:
> Does this fully address the initial problem you were trying to resolve?

What is the initial problem we're trying to solve, I'm curious but I don't see any details. @vabhasin could you fill out the PR template?

vabhasin:
> > Does this fully address the initial problem you were trying to resolve?
> 
> What is the initial problem we're trying to solve, I'm curious but I don't see any details. @vabhasin could you fill out the PR template?

Updated the Description with the details

src/test/java/com/aws/greengrass/lifecyclemanager/KernelTest.java
```diff
@@ -282,6 +285,21 @@ void GIVEN_kernel_WHEN_locate_finds_class_definition_in_config_THEN_create_servi
         assertEquals("testImpl", service2.getName());
     }
 
+    @SuppressWarnings("PMD.AvoidCatchingGenericException")
```
MikeDombo, line 288: Without splitting the startables, does this fail?

MikeDombo (approved):

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
```diff
@@ -152,6 +155,14 @@ public void launch() {
         // referenced by main/dependencies of main
         final Queue<String> autostart = findBuiltInServicesAndPlugins(); //NOPMD
         loadPlugins();
+
+        // Start MqttProxyIPCService after plugins are loaded, as it requires
+        // DiskSpooler Implementation Plugin. This behavior is only needed in testing
+        // as we scan our own classpath to find the @ImplementsService
```
junfuchen99, line 161: Do you know if we have any other plugin components that may require publishing to iot core? If any, would they break since now they are loaded before MqttProxyIPCService?

vabhasin: Not that I know of. I believe none of them would do any publishing before their startup phase and we do a startupAllServices at a later point.

MikeDombo (approved):

junfuchen99 (approved):

PR #1479, opened 2023-05-30
feat: execute sync when called by disk spooler implementation

vabhasin:
**Issue #, if available:**

**Description of changes:**
Modify Spool to only perform a sync of message IDs when called by the Disk Spooler implementation. This is because we don't know if the database for the Disk Spooler implementation would be ready by the time we start with the sync in our initialization. 
Corresponding change needs to be made in aws-greengrass-DiskSpooler to call `executeSync` in it's `startup()`.

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
+++ b/src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
@@ -22,6 +22,7 @@ import java.util.concurrent.BlockingDeque;
 import java.util.concurrent.LinkedBlockingDeque;
 import java.util.concurrent.atomic.AtomicLong;
 import javax.annotation.Nullable;
+import javax.inject.Inject;
 
 public class Spool {
     private static final Logger logger = LogManager.getLogger(Spool.class);
@@ -34,7 +35,7 @@ public class Spool {
     private static final SpoolerStorageType DEFAULT_SPOOL_STORAGE_TYPE = SpoolerStorageType.Memory;
     private static final int DEFAULT_SPOOL_MAX_MESSAGE_QUEUE_SIZE_IN_BYTES = (int) (2.5 * 1024 * 1024); // 2.5MB
     private final DeviceConfiguration deviceConfiguration;
-    private final CloudMessageSpool spooler;
+    private CloudMessageSpool spooler;
     private final InMemorySpool inMemorySpooler;
     private final Kernel kernel;
     private final AtomicLong nextId = new AtomicLong(0);
@@ -48,6 +49,7 @@ public class Spool {
      * @param deviceConfiguration the device configuration
      * @param kernel              a kernel instance
      */
+    @Inject
     public Spool(DeviceConfiguration deviceConfiguration, Kernel kernel) {
         this.deviceConfiguration = deviceConfiguration;
         Topics topics = this.deviceConfiguration.getSpoolerNamespace();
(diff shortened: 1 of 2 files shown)


src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -312,6 +291,38 @@ public void persistentQueueSync(Iterable<Long> diskQueueOfIds, CloudMessageSpool
...
+     * Extract message ids from the persistenceSpool plugin's disk database and insert the message
+     * ids into queueOfMessageId, this function is only used in Disk storage mode. If Sync fails midway,
+     * we continue anyway with that DiskSpooler. If we fail to get all Message IDs from Disk Spooler Database,
+     * we default to InMemory spooler.
+     *
+     * @param diskSpooler   the disk spooler instance for syncing messages
+     */
+    public void executeQueueSync(CloudMessageSpool diskSpooler) {
+        try {
+            persistentQueueSync(diskSpooler.getAllMessageIds(), diskSpooler);
```
MikeDombo, line 304: why are we doing this? Isn't this effectively just copying from diskSpooler into diskSpooler? What is this accomplishing

vabhasin: We maintain a Queue which has all message IDs that we mean to send to Cloud. When publishing or adding messages to spool, we use this queue to figure out what to publish next or if we can add more messages to spool.
We are performing this sync to basically refill that queue, with the messages that we see in persistent storage

vabhasin: Thought of a case just now, where we only restart the DiskSpooler component(which calls ExecuteSync) but not Nucleus, the queue would already contain these message IDs in that case. Adding a check for that so that we don't add duplicate message IDs to main queue while syncing.

MikeDombo: instead of adding checks, is it possible to dedupe? use an ordered set for example/map

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -273,6 +273,10 @@ private void persistentQueueSync(Iterable<Long> diskQueueOfIds, CloudMessageSpoo
         int numMessages = 0;
         int queueOfMessageIdInitSize = queueOfMessageId.size();
         for (long currentId : diskQueueOfIds) {
+            // Check if Queue of message IDs already contains this ID
+            if (queueOfMessageId.contains(currentId)) {
```
MikeDombo, line 277: this doesn't seem threadsafe when there are other threads adding and removing messages at the same time.

vabhasin: I did not find something particular to contains method, but looks like from java Docs all implementations of BlockingDequeue are threadsafe for all actions
https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/BlockingDeque.html

MikeDombo: just because it is safe doesn't mean it is correct.

MikeDombo: Think about what happens if a message is either added or removed before or after your `contains` check. Does that make the check and subsequent actions incorrect?

vabhasin: No, they won't matter as any add operation around that time would result in a new incremented message ID that we don't care about while performing this sync. As for the case that this code change covers, it doesn't matter.
There could be some weird edge cases resulting from a series of actions. I can go over them separately...

MikeDombo: I'd like you to think about a way that we can potentially delete the queue. It will be taking up a lot of memory and we probably can find a better way

MikeDombo (approved):

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
vabhasin: Got it, yes agreed

jcosentino11 (approved):

PR #1484, opened 2023-06-02
test: speed up tests

MikeDombo:
**Issue #, if available:**

**Description of changes:**
- Close handle on windows after updating IPC socket permissions - this forced the IPC server to stay alive forever
- Set retry interval to 0 during tests which are testing retries
- Remove unnecessary lifecycle scripts from `main` in several tests
- Remove the default true value for scan self classpath which means that plugins will need to enable that in their own testing, but it speeds up Nucleus testing
- Set deployment polling frequency to 0 instead of 1 in some tests

- Decrease test time by 7 minutes on linux!
- Decrease test time by 7-15 minutes on windows!

**Documentation Checklist:**

**Compatibility Checklist:**


src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -288,76 +288,80 @@ protected MqttClient(DeviceConfiguration deviceConfiguration,
...
-                        try {
-                            connection.reconnect(getMqttOperationTimeoutMillis());
-                            brokenConnections.remove(connection);
-                        } catch (InterruptedException | ExecutionException | TimeoutException e) {
-                            logger.atError().setCause(e).kv(CLIENT_ID_KEY, connection.getClientId())
-                                    .log("Error while reconnecting MQTT client");
+                            try {
+                                connection.reconnect(getMqttOperationTimeoutMillis());
+                                brokenConnections.remove(connection);
+                            } catch (InterruptedException | ExecutionException | TimeoutException e) {
```
MikeDombo, line 351: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

src/integrationtests/resources/com/aws/greengrass/integrationtests/ipc/ipc.yaml
```diff
@@ -2,9 +2,8 @@
 system:
   certificateFilePath: "/greengrass/v2/device.pem.crt"
   privateKeyPath: "/greengrass/v2/private.pem.key"
-  rootCaPath: "/greengrass/v2/AmazonRootCA1.pem"
-  rootpath: "/greengrass/v2"
   ipcSocketPath: "/ipcFolder/ipc.socket"
+  rootpath: "/greengrass/v2"
```
junfuchen99, line 6: why do we need this change?

src/main/java/com/aws/greengrass/lifecyclemanager/GreengrassService.java
```diff
@@ -495,6 +495,8 @@ public void addOrUpdateDependency(GreengrassService dependencyService, Dependenc
                 context.get(Kernel.class).clearODcache();
                 return new DependencyInfo(dependencyType, isDefault, listener);
             });
+            // Clear cache after updating dependency list
+            context.get(Kernel.class).clearODcache();
```
junfuchen99, line 499: Why do we need to clear again? I thought line 495 already clear it

src/integrationtests/java/com/aws/greengrass/integrationtests/BaseITCase.java
```diff
@@ -66,7 +66,6 @@ public class BaseITCase {
 
     @BeforeEach
     void setRootDir() {
-        System.setProperty("aws.greengrass.scanSelfClasspath", "true");
```
junfuchen99, line 69: Would this break other plugin integ tests? like log manager. I think they extend BaseITCase. https://github.com/aws-greengrass/aws-greengrass-log-manager/blob/main/src/integrationtests/java/com/aws/greengrass/integrationtests/logmanager/LogManagerTest.java#L97

src/integrationtests/resources/com/aws/greengrass/integrationtests/ipc/ipc.yaml
MikeDombo: It isn't needed and removing it ensures we are running from our own temp dir instead of a shared space that may have anything in it.

MikeDombo: Removing root ca path prevents exceptions in MQTT client because the ca doesn't exist.

src/main/java/com/aws/greengrass/lifecyclemanager/GreengrassService.java
MikeDombo: I saw a test fail once which could have been due to a race, so clearing again here to try and avoid that, won't hurt performance since there's such a small likelihood that anyone is querying the list

src/integrationtests/java/com/aws/greengrass/integrationtests/BaseITCase.java
MikeDombo: Yes this is mentioned in the description. Plugins will need to set it for their integration tests. It speeds us up here so that's worthwhile imo

junfuchen99 (approved):

src/integrationtests/java/com/aws/greengrass/integrationtests/lifecyclemanager/KernelTest.java
```diff
@@ -245,7 +245,7 @@ void GIVEN_service_startup_always_fail_WHEN_kernel_launches_THEN_service_go_brok
                 serviceBroken.countDown();
             }
         });
-        assertTrue(serviceBroken.await(TIMEOUT, TimeUnit.SECONDS));
+        assertTrue(serviceBroken.await(60, TimeUnit.SECONDS));
```
indougnito, line 248: why increase 45s -> 60s?  was the test flaky?

indougnito (approved):

src/integrationtests/java/com/aws/greengrass/integrationtests/lifecyclemanager/KernelTest.java
MikeDombo: Yes, timing out on windows.

PR #1486, opened 2023-06-09
fix: sync spool message queue during initialization

vabhasin:
**Issue #, if available:**
1. Having the sync queue run from Persistent spooler can cause some edge cases which would result in some unwanted scenarios like adding 2 same message IDs into the queue, or failure to add message in DB. These are difficult to debug and would cause unnecessary complexity. Already saw a couple instances while running UATs, like the FSS message is added to spooler when kernel is launched, and this happens even before we complete the sync. This results in failure to add a message with message ID 0, as there's already a message in DB with that ID which was not synced. We can remove this complexity by doing the sync during initialization before any spool message is added/removed from queue. 
2. We need to restart nucleus if Spooler configuration changes include a change in `SpoolStorageType`. Changing storage from Disk to Memory will result in losing messages and could break Nucleus and our ability to do deployments, if we don't restart Nucleus.

**Description of changes:**
1. Basically revert this commit (https://github.com/aws-greengrass/aws-greengrass-nucleus/commit/9d0d18e9ff0f929a5108e5f93d5df13ba30fc134) and keep some of the changes that are required. 
2. Add check for `SpoolStorageType` change in `BootstrapManager`

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/deployment/bootstrap/BootstrapManager.java
+++ b/src/main/java/com/aws/greengrass/deployment/bootstrap/BootstrapManager.java
@@ -20,7 +20,6 @@ import com.aws.greengrass.lifecyclemanager.exceptions.InputValidationException;
 import com.aws.greengrass.lifecyclemanager.exceptions.ServiceLoadException;
 import com.aws.greengrass.logging.api.Logger;
 import com.aws.greengrass.logging.impl.LogManager;
-import com.aws.greengrass.mqttclient.spool.SpoolerStorageType;
 import com.aws.greengrass.util.Coerce;
 import com.aws.greengrass.util.CommitableReader;
 import com.aws.greengrass.util.CommitableWriter;
@@ -202,9 +201,7 @@ public class BootstrapManager implements Iterator<BootstrapTaskStatus>  {
         if (newStorageType == null
                 && !(DEFAULT_SPOOL_STORAGE_TYPE.toString().equalsIgnoreCase(currentSpoolerStorageType))
                 || newStorageType instanceof String
-                && !currentSpoolerStorageType.equalsIgnoreCase((String) newStorageType)
-                || newStorageType instanceof SpoolerStorageType
-                && !currentSpoolerStorageType.equalsIgnoreCase(newStorageType.toString())) {
+                && !currentSpoolerStorageType.equalsIgnoreCase((String) newStorageType)) {
             logger.atInfo().kv(DEVICE_SPOOLER_NAMESPACE, newSpooler).log(RESTART_REQUIRED_MESSAGE);
             return true;
         }
--- a/src/main/java/com/aws/greengrass/deployment/bootstrap/BootstrapManager.java
+++ b/src/main/java/com/aws/greengrass/deployment/bootstrap/BootstrapManager.java
@@ -20,6 +20,7 @@ import com.aws.greengrass.lifecyclemanager.exceptions.InputValidationException;
 import com.aws.greengrass.lifecyclemanager.exceptions.ServiceLoadException;
 import com.aws.greengrass.logging.api.Logger;
 import com.aws.greengrass.logging.impl.LogManager;
+import com.aws.greengrass.mqttclient.spool.SpoolerStorageType;
 import com.aws.greengrass.util.Coerce;
 import com.aws.greengrass.util.CommitableReader;
 import com.aws.greengrass.util.CommitableWriter;
@@ -58,6 +59,7 @@ import static com.aws.greengrass.deployment.DeviceConfiguration.DEVICE_PARAM_PRO
 import static com.aws.greengrass.deployment.DeviceConfiguration.DEVICE_PARAM_PROXY_URL;
 import static com.aws.greengrass.deployment.DeviceConfiguration.DEVICE_PARAM_PROXY_USERNAME;
 import static com.aws.greengrass.deployment.DeviceConfiguration.DEVICE_PROXY_NAMESPACE;
+import static com.aws.greengrass.deployment.DeviceConfiguration.DEVICE_SPOOLER_NAMESPACE;
 import static com.aws.greengrass.deployment.DeviceConfiguration.RUN_WITH_DEFAULT_POSIX_SHELL;
 import static com.aws.greengrass.deployment.DeviceConfiguration.RUN_WITH_DEFAULT_POSIX_SHELL_VALUE;
 import static com.aws.greengrass.deployment.DeviceConfiguration.RUN_WITH_DEFAULT_POSIX_USER;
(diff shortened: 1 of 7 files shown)


src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -61,6 +59,7 @@ public Spool(DeviceConfiguration deviceConfiguration, Kernel kernel) {
         topics.subscribe((what, node) -> {
             if (WhatHappened.childChanged.equals(what) && node != null) {
                 setSpoolerConfigFromDeviceConfig(topics);
+                spooler = setupSpooler();
```
MikeDombo, line 62: Thread safety? What's going on with the old spooler you're replacing, won't we lose messages?

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -114,10 +112,24 @@ private CloudMessageSpool setupSpooler() {
...
+                logger.atWarn()
+                        .kv(PERSISTENCE_SPOOL_SERVICE_NAME_KEY, config.getPersistenceSpoolServiceName())
+                        .cause(e).log("Persistence spool queue sync was not completed");
+            } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+                logger.atWarn()
+                        .kv(PERSISTENCE_SPOOL_SERVICE_NAME_KEY, config.getPersistenceSpoolServiceName())
+                        .cause(e).log("Persistence spool queue sync was not completed");
+            }
+            logger.atInfo().log("Persistent Spooler has been set up");
```
MikeDombo, line 131: don't log this here because it isn't true. Log this on L121.

src/main/java/com/aws/greengrass/deployment/bootstrap/BootstrapManager.java
```diff
@@ -188,6 +192,25 @@ private boolean mqttVersionHasChanged(Map<String, Object> newNucleusParameters,
         return false;
     }
 
+    private boolean spoolerStorageTypeHasChanged(Map<String, Object> newNucleusParameters,
+                                          DeviceConfiguration currentDeviceConfiguration) {
+        String currentSpoolerStorageType = Coerce.toString(
+                currentDeviceConfiguration.getSpoolerNamespace().findOrDefault(DEFAULT_SPOOL_STORAGE_TYPE,
+                        SPOOL_STORAGE_TYPE_KEY));
+        Map<String, Object> newSpooler = (Map<String, Object>) newNucleusParameters.get(DEVICE_SPOOLER_NAMESPACE);
```
MikeDombo, line 200: do we know that `newNucleusParameters` is non-null at this point?

src/main/java/com/aws/greengrass/deployment/bootstrap/BootstrapManager.java
```diff
@@ -188,6 +192,25 @@ private boolean mqttVersionHasChanged(Map<String, Object> newNucleusParameters,
...
+        Object newStorageType = newSpooler == null ? null : newSpooler.get(SPOOL_STORAGE_TYPE_KEY);
+        if (newStorageType == null
+                && !(DEFAULT_SPOOL_STORAGE_TYPE.toString().equalsIgnoreCase(currentSpoolerStorageType))
+                || newStorageType instanceof String
+                && !currentSpoolerStorageType.equalsIgnoreCase((String) newStorageType)
+                || newStorageType instanceof SpoolerStorageType
```
MikeDombo, line 206: it will never be `SpoolerStorageType`

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
vabhasin: I was hoping we should be okay if QueueSync is stopped due to storage limitation, and still return Persistent Spooler as a spooler. Also, the case for QueueSync being stopped due to some interrupt. The function will only throw IOException if in case `getAllMessageIds` fail and in that case we default to InMemory spooler

MikeDombo: Let's say that you cannot setup the database. Is the persistent spooler correctly setup?

vabhasin: If DB setup fails, locate would give ServiceLoadException and we will default to InMemorySpooler.

MikeDombo: initializing in the constructor isn't a good idea. Constructors in java have limitations compared to regular methods.

src/main/java/com/aws/greengrass/deployment/bootstrap/BootstrapManager.java
vabhasin: Yes

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -114,10 +112,25 @@ private CloudMessageSpool setupSpooler() {
      * @throws ServiceLoadException thrown if the service cannot be located
      */
     private CloudMessageSpool getPersistenceSpoolGGService()
-            throws ServiceLoadException {
+            throws ServiceLoadException, IOException {
         GreengrassService locatedService = kernel.locate(config.getPersistenceSpoolServiceName());
         if (locatedService instanceof CloudMessageSpool) {
-            return (CloudMessageSpool) locatedService;
+            CloudMessageSpool persistenceSpool = (CloudMessageSpool) locatedService;
+            persistenceSpool.initializeSpooler();
```
MikeDombo, line 119: initialize needs to be able to throw and you need to handle that case when it does throw.

vabhasin: It will be covered under `IOException` above.

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
jcosentino11: it feels confusing to log both success and error at the same time, can we keep them separate?  For error scenario, i think we can provide more context to the customer. e.g. will this drop data? what will happen moving forward (still using persistence?)

jcosentino11 (approved):

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
vabhasin: Sure, can I make a note of this and add it separately. I'm thinking something like "continuing with Persistent Spooler anyways" in both the catch cases so that we have some context.

PR #1488, opened 2023-06-15
fix: do not bootloop when bootstrap fails and metadata is missing/corrupted

alter-mage:
**Issue #, if available:**

**Description of changes:**
1. If a bootstrap task fails, prepare rollback irrespective if deployment metadata is readable/writable or not
2. If rollback preparation fails and
    a. deployment metadata was updated, simply restart the kernel
    b. deployment metadata was updated due to corrupted, launch kernel as default
3. Safe delete deployment metadata from file system before modification

After a restart, if
1. deployment stage is determined to be bootstrap,
4. a pending bootstrap task fails, and
5. persisted deployment metadata file is either missing or corrupted

then Nucleus will end up being stuck in a bootloop, effectively bricking the core device.
Note that simply taking out the rollback preparation step from try block at
https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/main/src/main/java/com/aws/greengrass/lifecyclemanager/Kernel.java#L222
is not enough, as rollback preparation may fail causing a similar bootloop as explained above. Here, the deployment should proceed as default and complete. This way the core device can receive deployments later.

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -234,7 +234,7 @@
         <dependency>
             <groupId>software.amazon.awssdk.iotdevicesdk</groupId>
             <artifactId>aws-iot-device-sdk</artifactId>
-            <version>1.12.2-CLI-SNAPSHOT</version>
+            <version>1.13.2-CLI-SNAPSHOT</version>
         </dependency>
         <dependency>
             <groupId>org.bouncycastle</groupId>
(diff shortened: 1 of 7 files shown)


src/main/java/com/aws/greengrass/lifecyclemanager/Kernel.java
```diff
@@ -233,6 +245,7 @@ public Kernel launch() {
                 context.put(DeploymentQueue.class, deploymentQueue);
                 try {
                     Deployment deployment = deploymentDirectoryManager.readDeploymentMetadata();
+                    logger.atInfo().kv("deployment-metadata", deployment).log("After reading deployment metadata");
```
indougnito, line 248: Is this log line intentional?

alter-mage: Oops! This is not supposed to be here.

PR #1490, opened 2023-06-19
fix: always remove message from memory spooler

vabhasin:
**Issue #, if available:**
Given spooler storageType "Disk", and `addMessage` fails for some reason, we add the message to Memory spooler as a fallback. When we call `removeMessageById` it will try to remove message from DiskSpooler with the given ID but message was never added to persistent spooler and instead added to Memory. Due to this, the message will actually never be removed from the memory spooler , which is bad. 
**Description of changes:**
- Always remove message from Memory Spooler.
- Removed a check in `persistentQueueSync` which is never possible now that we moved `persistentQueueSync` back to spooler initialization. Earlier the check was added as we were calling queue sync from Disk Spooler implementation, in that case the condition would have occurred when we restarted the disk Spooler component.

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
+++ b/src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
@@ -230,7 +230,11 @@ public class Spool {
     public void removeMessageById(long messageId) {
         SpoolMessage toBeRemovedMessage = getMessageById(messageId);
         if (toBeRemovedMessage != null) {
-            spooler.removeMessageById(messageId);
+            // Always remove from InMemory Spooler in case message was added into Memory spooler due to fallback
+            inMemorySpooler.removeMessageById(messageId);
+            if (config.getStorageType() == SpoolerStorageType.Disk) {
+                spooler.removeMessageById(messageId);
+            }
             int messageSize = toBeRemovedMessage.getRequest().getPayload().length;
             curMessageQueueSizeInBytes.getAndAdd(-1L * messageSize);
         }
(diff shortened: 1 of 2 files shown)


src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -300,10 +304,6 @@ public void persistentQueueSync(Iterable<Long> diskQueueOfIds, CloudMessageSpool
         int numMessages = 0;
         int queueOfMessageIdInitSize = queueOfMessageId.size();
         for (long currentId : diskQueueOfIds) {
-            // Check if Queue of message IDs already contains this ID
-            if (queueOfMessageId.contains(currentId)) {
```
MikeDombo, line 304: why is this removed?

MikeDombo (commented):
Where is the NPE? That needs to be fixed and these changes don't really look like they're fixing the root cause.

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
vabhasin: I have it in description.

MikeDombo: What happens if it ID does exist in the queue and you're wrong? Is that a problem?

MikeDombo:
Fix your PR title. It must conform to conventional commit format.

vabhasin:
> Where is the NPE? That needs to be fixed and these changes don't really look like they're fixing the root cause.

To be clear, there would be SQLException due to NPE would be from DiskSpooler Remove method, we catch that SQLException. The `removeMessageById` from spool wouldn't throw anything. And we should be ok with this scenario, as its just the case where we fallback to adding to Memory.

MikeDombo:
Why is there an NPE in disk spooler? That needs to be fixed

vabhasin:
> Why is there an NPE in disk spooler? That needs to be fixed

Never mind, my mistake. I was using a modified exampleSpooler component in my UAT run from which I had the error. The modified disk spooler had some extra logs and was trying to print message from DiskSpooler from within the remove method. Long story short, no NPE.

MikeDombo (approved):

jcosentino11 (approved):

PR #1496, opened 2023-07-07
fix: check if qos0 message check is required

vabhasin:
**Issue #, if available:**
When GG is offline, or when spool storage is full, we check through all messages to see if i we can remove any QOS 0 message to accommodate the incoming message. This loop reads through all messages in spooler and is heavy as it reads through database, in case its a disk spooler. This was discovered as we saw high CPU usage in the Disk spooler performance tests which ran spooler tests with QOS 1 messages and at 100 TPS in an unstable network (online <-> offline every 5 mins)

**Description of changes:**
1. Have a flag to see if we need to check for QOS 0 messages or not, when the removal method is called. We set the flag to true every time a new message is added to spooler queue, and set it to false at the end of removal method. Basically, a flag saying "do we need to check for qos 0 message because the queue has been added to"

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
+++ b/src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
@@ -20,6 +20,7 @@ import java.io.IOException;
 import java.util.Iterator;
 import java.util.concurrent.BlockingDeque;
 import java.util.concurrent.LinkedBlockingDeque;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
 import javax.annotation.Nullable;
 
@@ -46,7 +47,7 @@ public class Spool {
      * when removeMessagesWithQosZero is called.
      * Set it back to false at the end of the removeMessagesWithQosZero method.
      */
-    private boolean qos0MessageCheckRequired = false;
+    private final AtomicBoolean qos0MessageCheckRequired = new AtomicBoolean(false);
     private final AtomicLong curMessageQueueSizeInBytes = new AtomicLong(0);
     private SpoolerConfig config;
 
@@ -176,7 +177,7 @@ public class Spool {
         SpoolMessage message = SpoolMessage.builder().id(id).request(request).build();
         addMessageToSpooler(id, message);
         queueOfMessageId.putLast(id);
-        qos0MessageCheckRequired = true;
+        qos0MessageCheckRequired.set(true);
         return message;
     }
 
@@ -257,7 +258,7 @@ public class Spool {
     }
 
     private void removeMessagesWithQosZero(boolean needToCheckCurSpoolerSize) {
-        if (!qos0MessageCheckRequired) {
+        if (!qos0MessageCheckRequired.get()) {
             return;
         }
         Iterator<Long> messageIdIterator = queueOfMessageId.iterator();
@@ -274,7 +275,7 @@ public class Spool {
                 }
             }
         }
-        qos0MessageCheckRequired = false;
+        qos0MessageCheckRequired.set(false);
     }
 
     private boolean addJudgementWithCurrentSpoolerSize(boolean needToCheckCurSpoolerSize) {
@@ -323,7 +324,7 @@ public class Spool {
             queueCapacityCheck(request, false);
 
             queueOfMessageId.putLast(currentId);
-            qos0MessageCheckRequired = true;
+            qos0MessageCheckRequired.set(true);
             if (currentId > highestId) {
                 highestId = currentId;
             }
--- a/src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
+++ b/src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
@@ -14,6 +14,7 @@ import com.aws.greengrass.lifecyclemanager.exceptions.ServiceLoadException;
 import com.aws.greengrass.logging.api.Logger;
 import com.aws.greengrass.logging.impl.LogManager;
 import com.aws.greengrass.mqttclient.v5.Publish;
+import com.aws.greengrass.mqttclient.v5.QOS;
 import com.aws.greengrass.util.Coerce;
 
 import java.io.IOException;
@@ -324,7 +325,9 @@ public class Spool {
             queueCapacityCheck(request, false);
 
             queueOfMessageId.putLast(currentId);
-            qos0MessageCheckRequired.set(true);
+            if (request.getQos().equals(QOS.AT_MOST_ONCE)) {
+                qos0MessageCheckRequired.set(true);
+            }
             if (currentId > highestId) {
                 highestId = currentId;
             }
--- a/src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
+++ b/src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
@@ -263,8 +263,10 @@ public class Spool {
         removeMessagesWithQosZero(false);
     }
 
-    private synchronized void removeMessagesWithQosZero(boolean needToCheckCurSpoolerSize) {
-        if (!qos0MessageCheckRequired.get()) {
+    private void removeMessagesWithQosZero(boolean needToCheckCurSpoolerSize) {
+        // Don't proceed if we are doing removal to make space (not due to being offline) and qos0 message check
+        // is not required
+        if (needToCheckCurSpoolerSize && !qos0MessageCheckRequired.get()) {
             return;
         }
         Iterator<Long> messageIdIterator = queueOfMessageId.iterator();
--- a/src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
+++ b/src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
@@ -42,11 +42,16 @@ public class Spool {
     private final AtomicLong nextId = new AtomicLong(0);
     private final BlockingDeque<Long> queueOfMessageId = new LinkedBlockingDeque<>();
     /**
-     * Flag to see if we need to check for QOS0 messages or not.
-     * It is set to true everytime a new message has been added to spooler queue.
-     * If the flag is true, we will attempt to remove QOS 0 messages from queue
-     * when removeMessagesWithQosZero is called.
-     * Set it back to false at the end of the removeMessagesWithQosZero method.
+     * Flag to see if we need to check for QOS0 messages or not, when we attempt to remove QOS0 messages
+     * with removeMessagesWithQosZeromethod.
+     * removeMessagesWithQosZero is called to remove QOS0 messages from Queue either when we are offline
+     * or when we want to make space to accommodate a new incoming message.
+     * - It is set to true everytime a new message has been added to spooler queue.
+     * - If the flag is true, we will check the queue to look for QOS 0 messages
+     *  when removeMessagesWithQosZero is called.
+     * - It is set back to false at the end of the removeMessagesWithQosZero method.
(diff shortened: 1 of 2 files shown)


src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -250,7 +259,8 @@ public void popOutMessagesWithQosZero() {
 
     private void removeMessagesWithQosZero(boolean needToCheckCurSpoolerSize) {
         Iterator<Long> messageIdIterator = queueOfMessageId.iterator();
-        while (messageIdIterator.hasNext() && addJudgementWithCurrentSpoolerSize(needToCheckCurSpoolerSize)) {
+        while (messageIdIterator.hasNext() && addJudgementWithCurrentSpoolerSize(needToCheckCurSpoolerSize)
+                && qos0MessagesInSpooler.get() > 0) {
```
jcosentino11, line 263: Super super nit: we can check qos first and avoid function call

jcosentino11 (dismissed):

MikeDombo (commented):
Rather than counting, how about a simpler (and less error prone) boolean flag. ex. set to true when _any_ message is added to the queue. If the flag is true, then we will attempt to remove qos 0 messages from the queue when the removal method is called. After calling the removal method, set it back to true. Basically just a flag which means "do we need to check for qos 0 because the queue has been added to"

vabhasin:
> 

Thanks, that's definitely a better way.

vabhasin:
> Rather than counting, how about a simpler (and less error prone) boolean flag. ex. set to true when _any_ message is added to the queue. If the flag is true, then we will attempt to remove qos 0 messages from the queue when the removal method is called. After calling the removal method, set it back to true. Basically just a flag which means "do we need to check for qos 0 because the queue has been added to"


Thanks, that's definitely a better way to do it.

jcosentino11 (dismissed):
Lgtm, unit test?

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -39,6 +39,14 @@ public class Spool {
...
     private final AtomicLong nextId = new AtomicLong(0);
     private final BlockingDeque<Long> queueOfMessageId = new LinkedBlockingDeque<>();
+    /**
+     * Flag to see if we need to check for QOS0 messages or not.
+     * It is set to true everytime a new message has been added to spooler queue.
+     * If the flag is true, we will attempt to remove QOS 0 messages from queue
+     * when removeMessagesWithQosZero is called.
+     * Set it back to false at the end of the removeMessagesWithQosZero method.
+     */
+    private boolean qos0MessageCheckRequired = false;
```
MikeDombo, line 49: this must be an atomicboolean

vabhasin:
> Lgtm, unit test?

Existing unit tests should verify that the overall behavior is not changed, testing just this flag part is not possible unless we change a couple of method return types. Went through the logic once again to confirm that when we do set the flag to false, we do it for a valid reason (check removed all qos0 messages or there weren't any in the first place)

jcosentino11:
> > Lgtm, unit test?
> 
> Existing unit tests should verify that the overall behavior is not changed, testing just this flag part is not possible unless we change a couple of method return types. Went through the logic once again to confirm that when we do set the flag to false, we do it for a valid reason:
> 
>     1. Check removed all qos0 messages OR
> 
>     2. There weren't any in the first place OR
> 
>     3. We removed some qos0 messages and stopped midway because size issues were solved, in that case we set it back to true immediately after adding the original message (for which we did this removal process)

Can force disconnect then verify number of times getMessageById is called?

jcosentino11 (dismissed):

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -263,6 +274,7 @@ private void removeMessagesWithQosZero(boolean needToCheckCurSpoolerSize) {
                 }
             }
         }
+        qos0MessageCheckRequired = false;
```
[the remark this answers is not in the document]
vabhasin: When not removing all messages (For example, to make space for incoming message), it will make sure to set the flag to true after it adds that new message. In case it removes some QOS0 messages and still fails to make space, we are ok with keeping the flag to false (as no need to check for QOS0 messages, as we can't make space even after deleting them)

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -168,7 +178,7 @@ public synchronized SpoolMessage addMessage(Publish request) throws InterruptedE
         SpoolMessage message = SpoolMessage.builder().id(id).request(request).build();
         addMessageToSpooler(id, message);
         queueOfMessageId.putLast(id);
-
+        qos0MessageCheckRequired.set(true);
```
jcosentino11, line 181: check for qos0 here too?

jcosentino11: do we also need to set in `addId()` too?

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -249,6 +259,9 @@ public void popOutMessagesWithQosZero() {
     }
 
     private void removeMessagesWithQosZero(boolean needToCheckCurSpoolerSize) {
+        if (!qos0MessageCheckRequired.get()) {
+            return;
+        }
```
jcosentino11, line 264: side note: i think there's a concurrency issue. `removeMessagesWithQosZero` is called 1) when MqttClient connection is interrupted and 2) during spool addMessage.   The latter is synchronized, but the former is not.   Technically a different issue so can be addressed separately, just wanted to call it out

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
vabhasin: We don't need the check here, as in this case we want to have it set to true for any new message added. 
Look at case number 3 above in my comment. In that condition, even after doing a check once, it's not guaranteed that there are no more QOS0 messages left in queue, we just stopped because storage issue is solved. Now, _any_ new message addition would change those storage conditions and that's why should set the flag to true again for any new message not just a QOS0 message.

The reason we can have that check in `persistentQueueSync` function is because we don't replace QOS0 messages using `removeMessagesWithQosZero`(The only place where flag could go back to false) while doing the queue sync. 

For the addId(), we don't need it there as well. As that is only called by mqttclient when we try to publish a message (do a `popId` first), but actually fail to publish it and add it to retries (do a `addId`). That's not a new message being added but an old message that was popped from spooler queue (and never actually removed from spooler) but never published and added back to queue.

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
vabhasin: Thanks...Nice catch, I can make a note and add it to next PR.

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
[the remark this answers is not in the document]
jcosentino11: What about case for mqtt disconnect and reconnect?

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
jcosentino11: I might be missing something, but always setting the flag on addMessage wouldn't give performance benefits?  subsequent calls to addMessage would always perform qos0 checks

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -311,6 +325,9 @@ public void persistentQueueSync(Iterable<Long> diskQueueOfIds, CloudMessageSpool
             queueCapacityCheck(request, false);
 
             queueOfMessageId.putLast(currentId);
+            if (request.getQos().equals(QOS.AT_MOST_ONCE)) {
```
MikeDombo, line 328: switch the order. `QOS.AT_MOST_ONCE.equals(request.getQos())`

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
vabhasin: At queue capacity, it will only be triggered at every addMessage if the previous Removal Procedure actually resulted in adding a new message. Otherwise, if it didn’t, the flag will remain false as we exit `addMessage` with SpoolerStorageException. Subsequent addMessage checks are not performed then.

MikeDombo (approved):

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
MikeDombo: still a problem?

vabhasin: Umm..yes, we don't know if some add/remove message could change conditions midway. I'm not sure, will it hurt adding synchronized though?

MikeDombo: If `needToCheckCurSize` is false, maybe just don't check the flag. Always do the cleanup.

vabhasin: We are looping through the whole database (even if we don't have QOS0 messages), if we don't check the flag in that case.

MikeDombo: Yes, but that's only called on disconnect, not on every single message. So probably not a big deal.

vabhasin: To confirm we are concerned about the additional delay it will cause due to being synchronized right? In that case, wanted to mention...`addMessage` anyways is synchronized (the other place which calls `removeMessagesWithQosZero`).

MikeDombo: I'm just concerned about a potential deadlock, shouldn't happen, but adding synchronized where it wasn't before does have potential problems.

MikeDombo: Given that we are using an atomic boolean, this should still be correct

MikeDombo: If we exit early and don't remove 1 qos 0 message that doesn't seem too bad anyway.

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -263,8 +263,10 @@ public void popOutMessagesWithQosZero() {
         removeMessagesWithQosZero(false);
     }
 
-    private synchronized void removeMessagesWithQosZero(boolean needToCheckCurSpoolerSize) {
-        if (!qos0MessageCheckRequired.get()) {
+    private void removeMessagesWithQosZero(boolean needToCheckCurSpoolerSize) {
+        // Don't proceed if we are doing removal to make space (not due to being offline) and qos0 message check
+        // is not required
+        if (needToCheckCurSpoolerSize && !qos0MessageCheckRequired.get()) {
```
MikeDombo, line 269: Undo this change.

MikeDombo (approved):

jcosentino11 (approved):

MikeDombo (approved):

PR #1515, opened 2023-08-11
fix: do not move terminated service to desired state on kernel shutdown

junfuchen99:
**Description of changes:**
- Fixes a race condition between component error and kernel shutdown
- Adds integ test

If component error and kernel shutdown happen at the same time (e.g. SIGKILL is sent to both component process and nucleus process), there is a chance that component restarts during kernel shutdown.

Specifically, if the component has a hard depender, it waits for the depender to close first during kernel shutdown and allows other state transitions.

After this change, while it still waits for the depender to close, it does not restart.

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/lifecyclemanager/GenericExternalService.java
+++ b/src/main/java/com/aws/greengrass/lifecyclemanager/GenericExternalService.java
@@ -578,7 +578,8 @@ public class GenericExternalService extends GreengrassService {
         resetRunWith(); // reset runWith - a deployment can change user info
     }
 
-    private synchronized void stopAllLifecycleProcesses() {
+    // public for integ test use only
+    public synchronized void stopAllLifecycleProcesses() {
         stopProcesses(lifecycleProcesses);
     }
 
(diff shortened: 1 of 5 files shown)


src/main/java/com/aws/greengrass/lifecyclemanager/GenericExternalService.java
```diff
@@ -578,7 +578,8 @@ protected synchronized void shutdown() {
         resetRunWith(); // reset runWith - a deployment can change user info
     }
 
-    private synchronized void stopAllLifecycleProcesses() {
```
rbattle, line 581: could this be package private instead of public?

i.e. 
```
synchronized void stopAllLifecycleProcesses()
```

rbattle: oh nevermind...that integ test is in a separate package

MikeDombo (commented):
This may work, IDK, but it wasn't what I was thinking of. The fix I was thinking is in the greengrass service close method to mark the service as shutdown immediately upon closing and not after waiting for all dependent components to stop.
Another thing to consider is not waiting for dependent components to stop at all.

junfuchen99:
As Michael and I discussed offline, we will adopt a different approach that addresses this issue.

PR #1520, opened 2023-08-22
fix(log): bump ALPN old Java log up to warning

MikeDombo:
**Issue #, if available:**

**Description of changes:**

**Documentation Checklist:**

**Compatibility Checklist:**


alter-mage (approved):

rbattle (approved):
wow - that should never have been a debug level message

MikeDombo:
> wow - that should never have been a debug level message

Unfortunately it will cause some spam if people are using old java versions. We do not know at this point in the code if ALPN would be necessary (ie. we're connecting to an IoT dataplane endpoint and using port 443), so this will log even if the HTTP call will work because you're using port 8443 (or aren't calling iot dataplane).

indougnito (approved):

PR #1537, opened 2023-10-03
chore: send component update policy in deployment status update

alter-mage:
**Issue #, if available:**

**Description of changes:**
Send component update policy action as part of the deployment status update.

Components should be able to consume the deployment status updates from Nucleus but only when deployment component update policy action is set to `NOTIFY`. 

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/deployment/DeploymentService.java
+++ b/src/main/java/com/aws/greengrass/deployment/DeploymentService.java
@@ -50,6 +50,7 @@ import lombok.Getter;
 import lombok.Setter;
 import org.apache.commons.io.FileUtils;
 import software.amazon.awssdk.iot.iotjobs.model.JobStatus;
+import software.amazon.awssdk.services.greengrassv2.model.DeploymentComponentUpdatePolicyAction;
 
 import java.io.IOException;
 import java.nio.file.Files;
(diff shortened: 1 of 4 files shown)


src/main/java/com/aws/greengrass/deployment/DeploymentService.java
```diff
@@ -323,6 +324,9 @@ private void finishCurrentDeployment() throws InterruptedException {
         String configurationArn = currentDeploymentTaskMetadata.getConfigurationArn();
         DeploymentType type = currentDeploymentTaskMetadata.getDeploymentType();
         List<String> rootPackages = currentDeploymentTaskMetadata.getRootPackages();
+        DeploymentComponentUpdatePolicyAction componentUpdatePolicy =
```
junfuchen99, line 327: do you need to check if `getComponentUpdatePolicy` returns null in these three places?

alter-mage: Yeah, but I don't see how could be set to null in any of the deployment types. Thing and thing group deployments force the customer to select either notify or skip and we hard code for local deployments.
If I do catch a null update policy, what should be the default?

shaguptashaikh (commented):
It feels wrong to include the component update policy in deployment status info. 1. It's  configuration set by a user, not status that needs to be sent back to the user for observability. 2. The information also gets sent to cloud in status updates, will that break anything in the cloud or be wasteful?

PR #1540, opened 2023-10-10
feat: bootstrap-on-rollback

indougnito:
**Issue #, if available:**

**Description of changes:**

Support the execution of bootstrap steps during rollback.

Implemented is an "opt-in" feature; bootstrap-on-rollback will run for given component if, and only if, in the context of the rollback deployment, a specific boolean flag is set inside that component's bootstrap lifecycle configuration.

Example configuration snippet:

```
...
  - Platform:
      os: linux
    Lifecycle:
      bootstrap:
        script: |-
          echo "Bootstrapping"
          exit 100
        bootstrapOnRollback: true
...
```

Bootstrapping is an important feature that allows developers to restart the Nucleus, or the device, as part of the deployment workflow. Providing optional access to bootstrapping, in the context of a rollback deployment, gives developers more design flexibility. For example, a rollback might necessitate a firmware downgrade, which might require a device restart in order to take effect.

**Documentation Checklist:**

**Compatibility Checklist:**


src/main/java/com/aws/greengrass/lifecyclemanager/Kernel.java
```diff
@@ -196,24 +199,19 @@ public Kernel launch() {
...
+                        boolean bootstrapOnRollbackRequired = kernelAlts.prepareBootstrapOnRollbackIfNeeded(
+                                this.context, deploymentDirectoryManager, bootstrapManager);
+                        // Save deployment error information
                         Deployment deployment = deploymentDirectoryManager.readDeploymentMetadata();
-                        deployment.setDeploymentStage(DeploymentStage.KERNEL_ROLLBACK);
+                        deployment.setDeploymentStage(
+                                bootstrapOnRollbackRequired ? ROLLBACK_BOOTSTRAP : KERNEL_ROLLBACK);
                         Pair<List<String>, List<String>> errorReport =
                                 DeploymentErrorCodeUtils.generateErrorReportFromExceptionStack(e);
                         deployment.setErrorStack(errorReport.getLeft());
```
alter-mage, line 217: I think we should modify the log in the catch section for this try block. Is it possible for the try block to throw IOException besides not being able to read deployment metadata?

indougnito: `kernelAlts.prepareBootstrapOnRollbackIfNeeded()` doesn't throw IOException.

Only sources of IOException are `deploymentDirectoryManager.readDeploymentMetadata()` and `deploymentDirectoryManager.writeDeploymentMetadata()`.

We could change the log text from "Could not read" to "Could not read/write"

src/main/java/com/aws/greengrass/lifecyclemanager/KernelAlternatives.java
```diff
@@ -392,6 +417,91 @@ public void cleanupLaunchDirectoryLinks() {
...
+            final Set<String> componentsToExclude = bootstrapManager.getPendingTasks();
+            logger.atDebug().kv("components", componentsToExclude)
+                    .log("These components did not bootstrap during the target deployment. "
+                            + "They will be excluded from bootstrap-on-rollback.");
+            // Exclude components that are not explicitly configured to bootstrap-on-rollback
+            Set<String> unconfiguredComponents = getComponentsNotConfiguredToBootstrapOnRollback(rollbackConfig);
+            logger.atDebug().kv("components", unconfiguredComponents)
+                    .log("These components are not configured to execute bootstrap steps during rollback. "
+                            + "They will be excluded from bootstrap-on-rollback.");
+            componentsToExclude.addAll(unconfiguredComponents);
```
junfuchen99, line 451: move these out to a separate function?

src/main/java/com/aws/greengrass/lifecyclemanager/KernelAlternatives.java
```diff
@@ -392,6 +417,91 @@ public void cleanupLaunchDirectoryLinks() {
...
+                return false;
+            }
+        } else {
+            // Bootstrap-on-rollback is not required, so ensure that the task file is deleted.
+            try {
+                bootstrapManager.deleteBootstrapTaskList(rollbackBootstrapTaskFilePath);
+            } catch (IOException exc) {
+                logger.atError().log("Bootstrap-on-rollback task file could not be cleaned up", exc);
+                return false;
+            }
```
junfuchen99, line 480: would this scenario happen in reality?

src/main/java/com/aws/greengrass/lifecyclemanager/KernelAlternatives.java
```diff
@@ -392,6 +417,91 @@ public void cleanupLaunchDirectoryLinks() {
...
+            return Collections.emptySet();
+        }
+        Set<String> componentsNotConfiguredToBootstrapOnRollback = new HashSet<>();
+        services.forEach((service) -> {
+            String serviceName = service.getName();
+            if (service instanceof Topics) {
+                boolean bootstrapOnRollback = Coerce.toBoolean(((Topics) service).findOrDefault(false,
+                        SERVICE_LIFECYCLE_NAMESPACE_TOPIC, LIFECYCLE_BOOTSTRAP_NAMESPACE_TOPIC,
+                        BOOTSTRAP_ON_ROLLBACK_CONFIG_KEY));
+                if (! bootstrapOnRollback) {
```
junfuchen99, line 497: nit: white space

src/main/java/com/aws/greengrass/lifecyclemanager/Kernel.java
```diff
@@ -196,24 +199,19 @@ public Kernel launch() {
...
-                            + " requested to complete bootstrap task");
-
-                    shutdown(30, exitCode == REQUEST_REBOOT ? REQUEST_REBOOT : REQUEST_RESTART);
+                    Path bootstrapTaskFilePath = deploymentDirectoryManager.getBootstrapTaskFilePath();
+                    executeBootstrapTasksAndShutdown(bootstrapManager, bootstrapTaskFilePath);
                 } catch (ServiceUpdateException | IOException e) {
                     logger.atError().log("Deployment bootstrap failed", e);
                     try {
+                        // Bootstrapping for target deployment failed, so check if bootstrap-on-rollback is needed
+                        boolean bootstrapOnRollbackRequired = kernelAlts.prepareBootstrapOnRollbackIfNeeded(
```
junfuchen99, line 209: if a component bootstrap times out in `executeOneBootstrapTask`, it throws a `ServiceUpdateException` directly without persisting bootstrap list. 

I think this case is not handled because you rely on updated bootstrap list to check which component needs roll back.

junfuchen99: Can we also add a test case for this?

src/main/java/com/aws/greengrass/lifecyclemanager/KernelAlternatives.java
```diff
@@ -392,6 +417,91 @@ public void cleanupLaunchDirectoryLinks() {
...
+        try {
+            rollbackBootstrapTaskFilePath = deploymentDirectoryManager.getRollbackBootstrapTaskFilePath();
+        } catch (IOException exc) {
+            logger.atError().log("Bootstrap-on-rollback task file path could not be resolved", exc);
+            return false;
+        }
+        if (bootstrapOnRollbackRequired) {
+            // Bootstrap-on-rollback is required, so write the task file.
+            try {
+                bootstrapManager.persistBootstrapTaskList(rollbackBootstrapTaskFilePath);
```
junfuchen99, line 468: How do you ensure bootstrap order?

src/main/java/com/aws/greengrass/lifecyclemanager/KernelAlternatives.java
indougnito: I am trying to preclude possible corner cases.

During rollback, Nucleus uses the existence of the `rollback_bootstrap_tasks.json` file to determine whether to launch in deployment stage `ROLLBACK_BOOTSTRAP` vs. `KERNEL_ROLLBACK`.

I'm imagining a scenario where a rogue `rollback_bootstrap_tasks.json` gets saved into the file system, as a result of some improper Nucleus upgrade/downgrade. In that case, the best thing to do is to pro-actively delete the file (if it exists) when we know that `ROLLBACK_BOOTSTRAP` is not the desired next deployment stage.

src/main/java/com/aws/greengrass/lifecyclemanager/KernelAlternatives.java
indougnito: Oh, my style is to do this deliberately... it draws the eye to the NOT operator.

src/main/java/com/aws/greengrass/lifecyclemanager/KernelAlternatives.java
indougnito: Agree

src/main/java/com/aws/greengrass/lifecyclemanager/Kernel.java
indougnito: I assume you are referring to the possibility that `executeBootstrapTasksAndShutdown` throws a ServiceUpdateException during `executeOneBootstrapTask`, leading into the current context.

Actually, when `executeOneBootstrapTask` fails, the bootstrap task list is never updated!

I am resolving this problem by adding a private member `activeTask` into `BootstrapManager`, which allows me to flag this component for bootstrap-on-rollback. See usage of `BootstrapManager.getPendingTasks()`.

src/main/java/com/aws/greengrass/lifecyclemanager/KernelAlternatives.java
indougnito: But for you, I will remove the space :)

src/main/java/com/aws/greengrass/lifecyclemanager/Kernel.java
indougnito: I'm going to rename `BootstrapManager.getPendingTasks()` -> `BootstrapManager.getUnstartedTasks()` and add some testing into `BootstrapManagerTest`.

src/main/java/com/aws/greengrass/deployment/activator/KernelUpdateActivator.java
```diff
@@ -119,7 +120,10 @@ void rollback(Deployment deployment, Throwable failureCause) {
         deployment.setErrorTypes(errorReport.getRight());
         deployment.setStageDetails(Utils.generateFailureMessage(failureCause));
 
-        deployment.setDeploymentStage(KERNEL_ROLLBACK);
+        final boolean bootstrapOnRollbackRequired = kernelAlternatives.prepareBootstrapOnRollbackIfNeeded(
+                kernel.getContext(), deploymentDirectoryManager, bootstrapManager);
+
+        deployment.setDeploymentStage(bootstrapOnRollbackRequired ? ROLLBACK_BOOTSTRAP : KERNEL_ROLLBACK);
```
MikeDombo, line 126: how does this work if we're rolling back to a nucleus which does not understand "ROLLBACK_BOOSTRAP"?
If you're upgrading to 2.12.0 and that fails, then it will rollback to 2.11.3, won't that fail to read this deployment stage?

indougnito: In the scenario you describe, at kernel launch, the 2.11.3 version of Nucleus will attempt to determine the deployment stage by invoking the method `kernelAlts.determineDeploymentStage`, which will return `KERNEL_ROLLBACK`. Bootstrap-on-rollback logic will not execute (and isn't present, even if one wanted to it to execute).

Counter-intuitively, the value `ROLLBACK_BOOTSTRAP` that we provide to `setDeploymentStage` here is only useful when bootstrap-on-rollback fails. Then, when the kernel subsequently launches with deploymentStage==DEFAULT, `KernelUpdateDeploymentTask` can examine the value and understand that bootstrap-on-rollback failed during one of the bootstrap steps, and then provide the appropriate deployment status (`FAILED_UNABLE_TO_ROLLBACK`).

src/main/java/com/aws/greengrass/deployment/bootstrap/BootstrapManager.java
```diff
@@ -163,6 +197,10 @@ public boolean isBootstrapRequired(Map<String, Object> newConfig)
         return nucleusConfigValidAndNeedsRestart || !bootstrapTaskStatusList.isEmpty();
     }
 
+    private boolean hasNotBeenCompleted(BootstrapTaskStatus task) {
```
MikeDombo, line 200: Here it was completed, just not necessarily successfully right? Maybe "incompleteOrErrored"?

src/main/java/com/aws/greengrass/deployment/bootstrap/BootstrapManager.java
```diff
@@ -429,6 +467,18 @@ public void persistBootstrapTaskList(Path persistedTaskFilePath) throws IOExcept
...
     }
 
+    /**
+     * Delete the bootstrap task list file, if it exists.
+     *
+     * @param persistedTaskFilePath Path to the persisted file of bootstrap tasks
+     * @throws IOException on I/O error
+     */
+    public void deleteBootstrapTaskList(Path persistedTaskFilePath) throws IOException {
+        Objects.requireNonNull(persistedTaskFilePath);
```
MikeDombo, line 477: Require throws an exception, is that desired? Should this just log and then return perhaps?

src/main/java/com/aws/greengrass/deployment/bootstrap/BootstrapManager.java
```diff
@@ -526,6 +576,7 @@ public BootstrapTaskStatus next() {
             throw new NoSuchElementException();
         }
         cursor++;
-        return bootstrapTaskStatusList.get(cursor - 1);
+        this.activeTask = bootstrapTaskStatusList.get(cursor - 1);
```
MikeDombo, line 579: Modifying the active task in here seems sketchy. Why do you need to do this? Caller of next can set it if desired.

src/main/java/com/aws/greengrass/lifecyclemanager/KernelAlternatives.java
```diff
@@ -297,6 +311,17 @@ public Deployment.DeploymentStage determineDeploymentStage(BootstrapManager boot
...
+            try {
+                Path rollbackBootstrapTasks = deploymentDirectoryManager.getRollbackBootstrapTaskFilePath();
+                if (rollbackBootstrapTasks.toFile().exists()) {
+                    bootstrapManager.loadBootstrapTaskList(rollbackBootstrapTasks);
+                    if (bootstrapManager.hasNext()) {
+                        return ROLLBACK_BOOTSTRAP;
+                    }
+                }
+            } catch (IOException e) {
+                logger.atError().setCause(e).log("Bootstrap-on-rollback task list not found or not readable");
```
MikeDombo, line 323: If it wasn't found you should not have any exception, so just saying unreadable is fine.

src/main/java/com/aws/greengrass/deployment/bootstrap/BootstrapManager.java
indougnito: NullPointerException is not desirable, I will change this

src/main/java/com/aws/greengrass/deployment/bootstrap/BootstrapManager.java
indougnito: I've implemented a method `getUnstartedTasks()` which helps determine which components need to execute bootstrap-on-rollback.  `getUnstartedTasks()` needs to exclude the bootstrap task which just failed (which I call the "active task"), so I created a member named `activeTask` and set it whenever we fetch a task from the iterator.

src/main/java/com/aws/greengrass/lifecyclemanager/KernelAlternatives.java
indougnito: That's true, I'll just log that it was unreadable.

MikeDombo (approved):

src/main/java/com/aws/greengrass/deployment/KernelUpdateDeploymentTask.java
```diff
@@ -72,6 +73,9 @@ public DeploymentResult call() {
             } else if (KERNEL_ROLLBACK.equals(stage)) {
                 result = new DeploymentResult(DeploymentResult.DeploymentStatus.FAILED_ROLLBACK_COMPLETE,
                         getDeploymentStatusDetails());
+            } else if (ROLLBACK_BOOTSTRAP.equals(stage)) {
+                result = new DeploymentResult(DeploymentResult.DeploymentStatus.FAILED_UNABLE_TO_ROLLBACK,
```
shaguptashaikh, line 77: Why is the status `FAILED_UNABLE_TO_ROLLBACK` in this case?

src/main/java/com/aws/greengrass/lifecyclemanager/KernelAlternatives.java
```diff
@@ -392,6 +417,98 @@ public void cleanupLaunchDirectoryLinks() {
...
+        } catch (IOException exc) {
+            logger.atError().log("Failed to read rollback snapshot config", exc);
+            return false;
+        }
+        boolean bootstrapOnRollbackRequired = false;
+        try {
+            // Check if we need to execute component bootstrap steps during the rollback deployment.
+            final Set<String> componentsToExclude =
+                    getComponentsToExcludeFromBootstrapOnRollback(bootstrapManager, rollbackConfig);
+            bootstrapOnRollbackRequired = bootstrapManager.isBootstrapRequired(rollbackConfig.toPOJO(),
```
junfuchen99, line 443: Inside `isBootstrapRequired`, if a geneic component can be loaded, it will check if the bootstrap has already been executed in the past - https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/4db3bfb5999db9b9bca09b59db475e292ae54610/src/main/java/com/aws/greengrass/lifecyclemanager/GenericExternalService.java#L272-L303.

We won't load any service here because no service is running. We just rely on the rollback config.

For components A, B with bootstraps, is it possible if 
- First deployment deploys A v1 successfully;
- Second deployment deploys B v1 successfully;
- Third deployment deploys B v2 which fails and attempts to roll back;
  - rollback config finds A v1 and B v1
  - isBootstrapRequired determines that both A and B bootstraps need to execute on rollback, but only B v2 should be rolling back

src/main/java/com/aws/greengrass/deployment/KernelUpdateDeploymentTask.java
indougnito: When a bootstrap-on-rollback step fails, we do the following (from Kernel.java::241-253):

1. set the deployment object's error information
2. set the deployment object's DeploymentStage enum value to ROLLBACK_BOOTSTRAP
3. offer the deployment object into the DeploymentQueue
4. launch the kernel in DEFAULT stage

Thus, the conditional logic you see here (`if (ROLLBACK_BOOTSTRAP.equals(stage)) { result = new DeploymentResult(FAILED_UNABLE_TO_ROLLBACK); ...`.

If, on the other hand, bootstrap-on-rollback had succeeded, the kernel would have progressed into the rollback activation stage (Kernel.java::257-271) and done the following:

1. set the deployment object's DeploymentStage enum value to KERNEL_ROLLBACK
2. offer the deployment object into the DeploymentQueue
3. launch the kernel in DEFAULT stage

Hence the preceding conditional logic (`if (KERNEL_ROLLBACK.equals(stage)) { result = new DeploymentResult(FAILED_ROLLBACK_COMPLETE); ...`.

src/main/java/com/aws/greengrass/lifecyclemanager/KernelAlternatives.java
indougnito: On testing this scenario, during rollback from the 3rd deployment which fails, I found `isBootstrapRequired` determines that only the bootstrap for B v1 is needed.

For component A, the debug log message is `Bootstrap is not required: bootstrap step unchanged`.  This means that, in the context of the rollback,

1) Nucleus is able to locate the bootstrap config for component A in the failed target deployment
2) Nucleus is able to locate the bootstrap config for component A in the rollback deployment
3) these two bootstrap configs are equal (according to `GenericExternalService.serviceLifecycleBootstrapEquals()`)

This makes sense to me, because the 3rd deployment made no change to the config for component A.

junfuchen99 (approved):

shaguptashaikh (approved):

PR #1543, opened 2023-10-10
fix: s3 regional endpoint cover version upgrade scenario

yitingb:
**Issue #, if available:**

**Description of changes:**
Add check system property in both constructors to cover version upgrade and Nucleus restart scenario

Manually tested install 2.10.3 and upgrade to 2.11.123 then the s3EndpointType config appeared in config.tlog, and a new private deployment also using the regional endpoint

**Documentation Checklist:**

**Compatibility Checklist:**


MikeDombo (dismissed):

junfuchen99 (dismissed):

indougnito (dismissed):

src/main/java/com/aws/greengrass/util/S3SdkClientFactory.java
```diff
@@ -109,11 +109,15 @@ private void setS3EndpointType(String type) {
...
             System.clearProperty(S3_ENDPOINT_PROP_NAME);
-            refreshClientCache();
+            if (!clientCache.isEmpty()) {
+                refreshClientCache();
+            }
             logger.atDebug().log("s3 endpoint set to global");
         } else if (!isGlobal && !S3_REGIONAL_ENDPOINT_VALUE.equals(s3EndpointSystemProp)) {
             System.setProperty(S3_ENDPOINT_PROP_NAME, S3_REGIONAL_ENDPOINT_VALUE);
-            refreshClientCache();
+            if (!clientCache.isEmpty()) {
```
MikeDombo, line 118: this is not right.

MikeDombo: The problem is that you were trying to remove `null` from the map. The _region_ was null. So fix that problem in refreshClientCache.

yitingb: I realized, should I check the key region not null?

MikeDombo (approved):

junfuchen99 (approved):

PR #1544, opened 2023-10-11
fix: log message in registerMqttConnectionProvider

gregbreen:
**Issue #, if available:**

None

**Description of changes:**

registerMqttConnectionProvider() log message seems to be a block copy gone wrong.


src/main/java/com/aws/greengrass/security/SecurityService.java
```diff
@@ -110,7 +110,7 @@ public void registerCryptoKeyProvider(CryptoKeySpi keyProvider) throws ServicePr
      */
     public void registerMqttConnectionProvider(MqttConnectionSpi mqttProvider) throws ServiceProviderConflictException {
         CaseInsensitiveString keyType = new CaseInsensitiveString(mqttProvider.supportedKeyType());
-        logger.atInfo().kv(KEY_TYPE, keyType).log("Register crypto key service provider");
+        logger.atInfo().kv(KEY_TYPE, keyType).log("Register MQTT connection provider");
```
MikeDombo, line 113:
```suggestion
        logger.atInfo().kv(KEY_TYPE, keyType).log("Register MQTT connection security provider");
```

MikeDombo:
Thank you Greg, just one update I'd request to make it clear that this is from the security service.

MikeDombo:
And your git commit message needs to comply with conventional commits. So: `fix: log message in registerMqttConnectionProvider`

src/main/java/com/aws/greengrass/security/SecurityService.java
gregbreen: @MikeDombo , happy to do so, but crypto key uses "service" rather than "security". So yeah, definitely "security"?

MikeDombo: Yes, in this case though it isn't providing an mqtt connection, rather it is providing the pluggable security for the mqtt connection. Not that it matters all that much... but I think my suggestion is still clear and accurate.

gregbreen: OK done

MikeDombo (approved):

shaguptashaikh (approved):

PR #1548, opened 2023-10-17
fix(mqtt): backoff and retry spooler connection failure

MikeDombo:
**Issue #, if available:**

**Description of changes:**
Spooler currently is immediately retrying to connect if the connection fails. This leads to log spam when PKCS11 provider is not ready yet, for example. This change backs off 2 minutes + 10 second random jitter which matches our other backoff strategies for MQTT.

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
+++ b/src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
@@ -89,6 +89,7 @@ import static com.aws.greengrass.deployment.DeviceConfiguration.DEVICE_PARAM_PRI
 import static com.aws.greengrass.deployment.DeviceConfiguration.DEVICE_PARAM_ROOT_CA_PATH;
 import static com.aws.greengrass.deployment.DeviceConfiguration.DEVICE_PARAM_THING_NAME;
 import static com.aws.greengrass.mqttclient.AwsIotMqttClient.TOPIC_KEY;
+import static com.aws.greengrass.util.RetryUtils.RANDOM;
 
 @SuppressWarnings({"PMD.AvoidDuplicateLiterals"})
 public class MqttClient implements Closeable {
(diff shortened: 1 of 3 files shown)


junfuchen99 (dismissed):

indougnito (dismissed):

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -813,7 +814,21 @@ protected void runSpooler() {
...
+                        logger.atWarn().log("Shutting down spooler task");
+                        Thread.currentThread().interrupt();
+                        break;
+                    } else {
+                        logger.atError().log("Error when publishing from spooler", e.getCause());
+                        // Do not retry connecting immediately, most likely it would fail. Backoff and retry later.
+                        // It fails when PKCS11 is not available, for example, so retrying that immediately
+                        // will just spam the logs.
+                        Thread.sleep(Duration.ofMinutes(2).toMillis() + RANDOM.nextInt(10_000));
+                    }
```
rbattle, line 830: 2 minutes seems extremely conservative. Is that how long it takes to resolve with PKCS11?

Is it possible to do exponential back-off with jitter just so a transient issue doesn't cause a 2 minute delay?

Otherwise could using a smaller baseline duration - say 30s with 30s of jitter?

MikeDombo: PKCS can take a while, but your point is of course valid. I used 2 minutes because it is what we use everywhere else, when subscribing to jobs and shadow topics for example.

rbattle: ok - maybe worth a revisit as separate issue then

junfuchen99 (approved):

indougnito (approved):

PR #1555, opened 2023-10-30
fix: bootstrap on rollback when component breaks

alter-mage:
**Issue #, if available:**

**Description of changes:**
Determine if bootstrap on rollback is required when a component errors 3 times and transitions to broken state when activating kernel.
Tested using a UAT which makes 2 deployments
1. Deploys 3 components, all with working bootstrap steps. This deployment completes with success.
2. Deploys 2 components, one with a bootstrap step and the other with a broken startup lifecycle. The second component breaks during the kernel activation stage and triggers a rollback. Before rolling back, `KernelAlternatives` determines if bootstrap rollback is needed and persists rollback bootstrap file if necessary.

With this test setup, Nucleus enters rollback bootstrap deployment stage after preparing rollback and executes bootstrap steps of the components deployment in the first deployment.

It looks like we are not bootstrapping when rollback is triggered due to a broken component when activating kernel.

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/deployment/KernelUpdateDeploymentTask.java
+++ b/src/main/java/com/aws/greengrass/deployment/KernelUpdateDeploymentTask.java
@@ -6,6 +6,7 @@
 package com.aws.greengrass.deployment;
 
 import com.aws.greengrass.componentmanager.ComponentManager;
+import com.aws.greengrass.deployment.bootstrap.BootstrapManager;
 import com.aws.greengrass.deployment.errorcode.DeploymentErrorCode;
 import com.aws.greengrass.deployment.errorcode.DeploymentErrorCodeUtils;
 import com.aws.greengrass.deployment.errorcode.DeploymentErrorType;
@@ -94,10 +95,14 @@ public class KernelUpdateDeploymentTask implements DeploymentTask {
             logger.atError("deployment-errored", e).log();
             if (KERNEL_ACTIVATION.equals(stage)) {
                 try {
-                    deployment.setDeploymentStage(KERNEL_ROLLBACK);
+                    KernelAlternatives kernelAlternatives = kernel.getContext().get(KernelAlternatives.class);
+                    final boolean bootstrapOnRollbackRequired = kernelAlternatives.prepareBootstrapOnRollbackIfNeeded(
+                            kernel.getContext(), kernel.getContext().get(DeploymentDirectoryManager.class),
+                            kernel.getContext().get(BootstrapManager.class));
+                    deployment.setDeploymentStage(bootstrapOnRollbackRequired ? ROLLBACK_BOOTSTRAP : KERNEL_ROLLBACK);
                     saveDeploymentStatusDetails(e);
                     // Rollback workflow. Flip symlinks and restart kernel
-                    kernel.getContext().get(KernelAlternatives.class).prepareRollback();
+                    kernelAlternatives.prepareRollback();
                     kernel.shutdown(30, REQUEST_RESTART);
                 } catch (IOException ioException) {
                     logger.atError().log("Failed to set up Nucleus rollback directory", ioException);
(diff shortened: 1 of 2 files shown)


src/main/java/com/aws/greengrass/deployment/KernelUpdateDeploymentTask.java
```diff
@@ -49,6 +53,9 @@ public class KernelUpdateDeploymentTask implements DeploymentTask {
     public KernelUpdateDeploymentTask(Kernel kernel, Logger logger, Deployment deployment,
                                       ComponentManager componentManager) {
         this.kernel = kernel;
+        this.kernelAlternatives = kernel.getContext().get(KernelAlternatives.class);
+        this.deploymentDirectoryManager = kernel.getContext().get(DeploymentDirectoryManager.class);
+        this.bootstrapManager = kernel.getContext().get(BootstrapManager.class);
```
MikeDombo, line 58: these are only used in `call` within an `if`, so pull them from the context when they're actually needed, not in the constructor.

src/main/java/com/aws/greengrass/deployment/KernelUpdateDeploymentTask.java
```diff
@@ -94,10 +101,12 @@ public DeploymentResult call() {
             logger.atError("deployment-errored", e).log();
             if (KERNEL_ACTIVATION.equals(stage)) {
                 try {
-                    deployment.setDeploymentStage(KERNEL_ROLLBACK);
+                    final boolean bootstrapOnRollbackRequired = kernelAlternatives.prepareBootstrapOnRollbackIfNeeded(
```
MikeDombo, line 104: customer's deployment directory did not have any `rollback_boostrap_task.txt` file, so will this even trigger in their case?

MikeDombo: Please update the description with how this change was tested to validate it works.

src/main/java/com/aws/greengrass/deployment/KernelUpdateDeploymentTask.java
alter-mage: I initially was pulling this from context when required, however, I do not want to pull, say `KernelAlternatives`, from context in the same function twice.

MikeDombo: So save it in a local variable within the method then.

src/main/java/com/aws/greengrass/deployment/KernelUpdateDeploymentTask.java
alter-mage: `prepareBootstrapOnRollbackIfNeeded` determines if bootstrap is required from the rollback config snapshot and not the `rollback_boostrap_task.txt` file. Rollback bootstrap task list is persisted in this method which is later used by the kernel (after reboot) to determine deployment stage. And that should be accurate.

alter-mage: Updated description

src/main/java/com/aws/greengrass/deployment/KernelUpdateDeploymentTask.java
alter-mage: updated

src/main/java/com/aws/greengrass/deployment/KernelUpdateDeploymentTask.java
MikeDombo: The UAT does not reboot the device, please test manually using the exact customer scenario including reboot.

junfuchen99 (approved):

MikeDombo (approved):

src/main/java/com/aws/greengrass/deployment/KernelUpdateDeploymentTask.java
alter-mage: Just validated the behavior with manual testing. Deploying components from the exact customer scenario goes through the rollback bootstrap deployment stage when rollback is triggered after component startup lifecycle errors.

PR #1556, opened 2023-11-02
fix: overlapping shutdown and startup lifecycles

jcosentino11:
Work in progress.  This is a first pass that will likely need some cleaning up; tests need to be added

**Issue #, if available:**

**Description of changes:**
It's possible for shutdown and startup phases of a service to overlap, as the shutdown lifecycle is executed asynchronously upon state transition, and it is possible for a previous backing task to be executing (even if it was cancelled).

The fix proposed here is to "join" before executing shutdowns.  Not completely trivial because we cancel futures, which will return isDone() immediately or throw CancellationException immediately on a get.

We recently saw a customer case for bridge where MQTTClient is stuck in a connect/disconnect loop. This can only happen if there are two instances of MQTTClient running at the same time (both kick each other off the connection), which I was able to trigger by reinstalling during startup (which can happen irl due to config changes):
```
AtomicBoolean onlyOnce = new AtomicBoolean();
@Override
public void startup() {
    if (onlyOnce.compareAndSet(false, true)) {
        requestReinstall();
    }
}
```

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
+++ b/src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
@@ -19,9 +19,13 @@ import lombok.AllArgsConstructor;
 import lombok.Builder;
 import lombok.Data;
 import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+import lombok.Setter;
 
 import java.time.Clock;
 import java.time.Duration;
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -33,8 +37,8 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.BlockingQueue;
-import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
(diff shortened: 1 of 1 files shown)


src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
```diff
@@ -496,25 +499,23 @@ private void handleCurrentStateNew(Optional<State> desiredState) throws Interrup
...
         try {
-            backingTask.get().get(installTimeOut, TimeUnit.SECONDS);
+            BackingTask task = replaceBackingTask(() -> {
+                if (!State.NEW.equals(getState()) || getStateGeneration().get() != currentStateGeneration) {
+                    // Bail out if we're not in the expected state
+                    return;
+                }
+                try {
+                    greengrassService.install();
+                } catch (InterruptedException t) {
```
MikeDombo, line 512: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
```diff
@@ -629,22 +630,27 @@
...
+            Instant waitStart = Instant.now();
+            waitForBackingTask(Duration.ofSeconds(timeout));
+            long timeWaited = ChronoUnit.SECONDS.between(waitStart, Instant.now());
+
+            // does not handle desiredState in STOPPING because we must stop first.
+            // does not use setBackingTask because it will cancel the existing task.
+            shutdownFuture = greengrassService.getContext().get(ExecutorService.class).submit(() -> {
+                try {
+                    greengrassService.shutdown();
+                } catch (InterruptedException i) {
```
MikeDombo, line 646: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

MikeDombo:
This seems like something that maybe the implementing classes need to resolve. Shutdown is allowed to execute while startup is also executing, that's how you'd stop a process which is running by startup; the shutdown lifecycle would stop it.

jcosentino11:
Ah that makes sense, will take another look

PR #1569, opened 2024-01-04
chore(mqtt): adjust topic limits

jcosentino11:
**Issue #, if available:**

**Description of changes:**

There is (at least one) special case where an IoT reserved topic has > 7 slashes, [`$aws/iotwireless/events/{eventName}/{eventType}/{resourceType}/{resourceIdentifierType}/{resourceID}/{id}`](https://docs.aws.amazon.com/iot/latest/developerguide/iot-wireless-event-messages.html).

> ⚠️ The following changes applies to MQTT5 only

To account for this, validation for _some_ reserved topics (`$`) has been relaxed:
1) no client-side validation of forward slashes
2) enforce a larger size limit (512 bytes), since we can't assume we know the prefix in this case.  rely on server-side validation instead and just prevent sending extremely large payloads to core.

However, we still validate 'known' types of reserved topics, to maximize customer experience (erroring early and with clear exception messaging), which include:
1) basic ingest topics (which we were already validating)
2) shared subscriptions (new validation that has been added here)

#### Implementation Notes
* Moved topic validation logic from MqttClient to its own class
* Refactored topic validation tests to be parameterized

To allow special case reserved topics with >7 forward slashes `$aws/iotwireless/events/{eventName}/{eventType}/{resourceType}/{resourceIdentifierType}/{resourceID}/{id}` to be considered valid.

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/mqttclient/IotCoreTopicValidator.java
+++ b/src/main/java/com/aws/greengrass/mqttclient/IotCoreTopicValidator.java
@@ -33,6 +33,7 @@ public final class IotCoreTopicValidator {
     private static final char RESERVED_TOPIC_PREFIX = '$';
     private static final char MULTI_LEVEL_WILDCARD = '#';
     private static final char SINGLE_LEVEL_WILDCARD = '+';
+    private static final char FORWARD_SLASH = '/';
 
     private static final String ERROR_PUBLISH_TOPIC_TOO_LONG = String.format(
             "The topic size of request must be no "
@@ -82,8 +83,6 @@ public final class IotCoreTopicValidator {
             throw new MqttRequestException(ERROR_WILDCARD_IN_PUBLISH_TOPIC);
         }
 
-        topic = topic.toLowerCase().trim();
-
         if (topic.charAt(0) != RESERVED_TOPIC_PREFIX) {
             validateEffectiveTopic(topic, operation);
             return;
@@ -128,7 +127,7 @@ public final class IotCoreTopicValidator {
     }
 
     private static void validateEffectiveTopic(String effectiveTopic, Operation operation) throws MqttRequestException {
-        if (effectiveTopic.chars().filter(num -> num == '/').count() > TOPIC_MAX_NUMBER_OF_FORWARD_SLASHES) {
+        if (effectiveTopic.chars().filter(num -> num == FORWARD_SLASH).count() > TOPIC_MAX_NUMBER_OF_FORWARD_SLASHES) {
             throw new MqttRequestException(ERROR_TOPIC_HAS_TOO_MANY_SLASHES);
         }
         if (effectiveTopic.length() > MAX_LENGTH_OF_TOPIC) {
@@ -148,7 +147,8 @@ public final class IotCoreTopicValidator {
     private static Optional<String> removePrefix(String topic, String prefixRegex) {
         String[] firstAndRest = topic.split(prefixRegex, 2);
         if (firstAndRest.length == 2) {
-            return Optional.of(firstAndRest[1]);
+            // intern to deduplicate topic in memory
+            return Optional.of(firstAndRest[1].intern());
         }
         return Optional.empty();
     }
--- a/src/main/java/com/aws/greengrass/mqttclient/IotCoreTopicValidator.java
+++ b/src/main/java/com/aws/greengrass/mqttclient/IotCoreTopicValidator.java
@@ -5,14 +5,18 @@
 
 package com.aws.greengrass.mqttclient;
 
-import com.aws.greengrass.deployment.DeviceConfiguration;
 import com.aws.greengrass.util.Utils;
+import lombok.NonNull;
 
+import java.util.Arrays;
+import java.util.Optional;
 import java.util.regex.Pattern;
-import javax.inject.Inject;
+import java.util.stream.Collectors;
+
+import static com.aws.greengrass.mqttclient.MqttClient.MQTT_VERSION_5;
 
 
-public class IotCoreTopicValidator {
+public final class IotCoreTopicValidator {
 
     public enum Operation {
         PUBLISH,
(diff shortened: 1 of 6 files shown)


src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -381,6 +376,7 @@ public MqttClient(DeviceConfiguration deviceConfiguration, Spool spool, boolean
 
         this.deviceConfiguration = deviceConfiguration;
         mqttTopics = this.deviceConfiguration.getMQTTNamespace();
+        this.topicValidator = new IotCoreTopicValidator(deviceConfiguration);
```
MikeDombo, line 379: I'd probably simplify and just pass the mqtt version at the lookup time

src/main/java/com/aws/greengrass/mqttclient/IotCoreTopicValidator.java
```diff
@@ -0,0 +1,118 @@
...
+            throw new MqttRequestException(errMsg);
+        }
+    }
+
+    private void validateSharedSubscriptionTopic(String topic) throws MqttRequestException {
+        validateEffectiveTopic(topic.split(SHARED_SUBSCRIPTION_PREFIX, 2)[1]);
+    }
+
+    private void validateDirectIngestTopic(String topic) throws MqttRequestException {
+        validateEffectiveTopic(topic.split(DIRECT_INGEST_PREFIX, 2)[1]);
```
MikeDombo, line 100: validate before indexing, could result in index out of bounds

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
jcosentino11: done

src/main/java/com/aws/greengrass/mqttclient/IotCoreTopicValidator.java
jcosentino11: done

src/test/java/com/aws/greengrass/mqttclient/MqttClientTest.java
```diff
@@ -1010,130 +1011,151 @@ void GIVEN_message_size_exceeds_max_limit_WHEN_publish_THEN_future_complete_exce
...
+                        "mqtt5"),
+                // unreserved topic that's 1 byte greater than 256 bytes
+                Arguments.of("my/example/topic/thats/too/large/00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
+                        "mqtt3"),
+                Arguments.of("my/example/topic/thats/too/large/00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
```
sameerzuberi, line 1118: nit: parameterize the strings of 0s for 256 and 512 bytes and include them using a String.format

jcosentino11: Good point will use padding to clean this up

src/main/java/com/aws/greengrass/mqttclient/IotCoreTopicValidator.java
```diff
@@ -0,0 +1,159 @@
...
+        if (Utils.isEmpty(topic)) {
+            throw new MqttRequestException("Topic must not be empty");
+        }
+
+        if (operation == Operation.PUBLISH && (topic.contains("#") || topic.contains("+"))) {
+            throw new MqttRequestException("Publish topics must not contain wildcard "
+                    + "characters of '#' or '+'");
+        }
+
+        topic = topic.toLowerCase().trim();
```
MikeDombo, line 59: is this necessary? This will be creating a duplicated temporary string. We use `.intern` on MQTT topics to reduce how much memory we need for topics, so this is kind of undoing that.

src/main/java/com/aws/greengrass/mqttclient/IotCoreTopicValidator.java
```diff
@@ -0,0 +1,159 @@
...
+        if (effectiveTopic.chars().filter(num -> num == '/').count() > TOPIC_MAX_NUMBER_OF_FORWARD_SLASHES) {
+            String msg = String.format("The request topic must have no more than %d forward slashes (/)",
+                    TOPIC_MAX_NUMBER_OF_FORWARD_SLASHES);
+            throw new MqttRequestException(msg);
+        }
+        if (effectiveTopic.length() > MAX_LENGTH_OF_TOPIC) {
+            String msg = String.format("The topic size of request must be no "
+                            + "larger than %d bytes of UTF-8 encoded characters. This excludes the first "
+                            + "3 mandatory segments for Basic Ingest topics ($AWS/rules/rule-name/)",
+                    MAX_LENGTH_OF_TOPIC);
```
MikeDombo, line 121: nit: these string errors can be constants. You're always formatting in the same data each time.

jcosentino11: fixed

src/main/java/com/aws/greengrass/mqttclient/IotCoreTopicValidator.java
jcosentino11: ah yeah good point, just addressed

src/main/java/com/aws/greengrass/mqttclient/IotCoreTopicValidator.java
```diff
@@ -148,7 +147,8 @@ private static void validateEffectiveTopic(String effectiveTopic, Operation oper
     private static Optional<String> removePrefix(String topic, String prefixRegex) {
         String[] firstAndRest = topic.split(prefixRegex, 2);
         if (firstAndRest.length == 2) {
-            return Optional.of(firstAndRest[1]);
+            // intern to deduplicate topic in memory
+            return Optional.of(firstAndRest[1].intern());
```
MikeDombo, line 151: this isn't needed. The string is already created, so it is too late. Interning is actually going to spend more time and won't really save memory

src/main/java/com/aws/greengrass/mqttclient/IotCoreTopicValidator.java
```diff
@@ -141,19 +153,12 @@ private static Optional<String> removePrefix(String topic, String prefixRegex) {
...
-     * @param numParts number of topic parts to include in the result
-     * @return topic prefix
-     */
-    private static String prefix(String topic, int numParts) {
-        String[] parts = topic.split("/");
-        return Arrays.stream(parts)
-                .limit(Math.min(parts.length, numParts))
-                .collect(Collectors.joining("/"));
+    private static boolean containsWildcards(String topic) {
+        for (char c : topic.toCharArray()) {
```
MikeDombo, line 157: this creates a copy of the underlying array, seems probably more expensive than just doing a contains check twice

src/main/java/com/aws/greengrass/mqttclient/IotCoreTopicValidator.java
jcosentino11: ah okay, fixed

MikeDombo (approved):

sameerzuberi (approved):

PR #1575, opened 2024-02-07
feat(fss): update component status when recovered from error state

yitingb:
**Issue #, if available:**

**Description of changes:**
We made 2 changes in this pr:
1. We send one more fss message on component status change when the component encountered an error, completes retry and recovered. 
2. We consolidated the trigger type for component errors and broken into a single component status change event.
We enhance customer usability and aim to synchronize the component status across devices and consoles, thereby streamlining operations and providing a seamless experience.

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/status/FleetStatusService.java
+++ b/src/main/java/com/aws/greengrass/status/FleetStatusService.java
@@ -43,6 +43,7 @@ import software.amazon.awssdk.iot.iotjobs.model.JobStatus;
 
 import java.time.Instant;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
@@ -110,12 +111,14 @@ public class FleetStatusService extends GreengrassService {
     private final ConcurrentHashMap<GreengrassService, Instant> serviceFssTracksMap = new ConcurrentHashMap<>();
     private final AtomicBoolean isDeploymentInProgress = new AtomicBoolean(false);
     private final Object periodicUpdateInProgressLock = new Object();
+    private final Set<GreengrassService> erroredComponentsSet = new HashSet<>();
     @Setter // Needed for integration tests.
     @Getter(AccessLevel.PACKAGE) // Needed for unit tests.
     private int periodicPublishIntervalSec;
     private ScheduledFuture<?> periodicUpdateFuture;
     // default to zero so that first Reconnect update would go thru
     private Instant lastReconnectUpdateTime = Instant.EPOCH;
+
 
     @Getter
     public MqttClientConnectionEvents callbacks = new MqttClientConnectionEvents() {
@@ -295,24 +298,41 @@ public class FleetStatusService extends GreengrassService {
                 initialDelay, periodicPublishIntervalSec, TimeUnit.SECONDS);
     }
 
+    private AtomicReference<OverallStatus> getOverallStatus() {
+        Instant now = Instant.now();
+        AtomicReference<OverallStatus> overallStatus = new AtomicReference<>(OverallStatus.HEALTHY);
+        this.kernel.orderedDependencies().forEach(service -> {
+            serviceFssTracksMap.put(service, now);
+            overallStatus.set(getOverallStatusBasedOnServiceState(overallStatus.get(), service));
+        });
+        return overallStatus;
+    }
+
     @SuppressWarnings("PMD.UnusedFormalParameter")
     private void handleServiceStateChange(GreengrassService greengrassService, State oldState,
                                           State newState) {
+
         // Not reporting status of other components in errored component message because a deployment could
         // be in-progress and other component's fleet config arn may not have updated
         if (newState.equals(State.ERRORED)) {
-            Set<GreengrassService> erroredComponentSet = new HashSet<>();
-            erroredComponentSet.add(greengrassService);
+            if (!isDeploymentInProgress.get()) {
+                //Do not update the recovery state later for ongoing deployment
+                erroredComponentsSet.add(greengrassService);
+            }
 
             // Evaluate overall status based on current state of all components
-            Instant now = Instant.now();
-            AtomicReference<OverallStatus> overAllStatus = new AtomicReference<>(OverallStatus.HEALTHY);
-            this.kernel.orderedDependencies().forEach(service -> {
-                serviceFssTracksMap.put(service, now);
-                overAllStatus.set(getOverallStatusBasedOnServiceState(overAllStatus.get(), service));
-            });
+            uploadFleetStatusServiceData(new HashSet<>(Arrays.asList(greengrassService)),
+                    getOverallStatus().get(), null,
+                    Trigger.COMPONENT_STATUS_CHANGE);
+            return;
+        }
 
-            uploadFleetStatusServiceData(erroredComponentSet, overAllStatus.get(), null, Trigger.ERRORED_COMPONENT);
+        // if reported errored component is recovered, remove component from set and update the component status
+        if (!isDeploymentInProgress.get() && greengrassService.reachedDesiredState()
+                && erroredComponentsSet.remove(greengrassService)) {
+            uploadFleetStatusServiceData(new HashSet<>(Arrays.asList(greengrassService)),
+                    getOverallStatus().get(), null,
+                    Trigger.COMPONENT_STATUS_CHANGE);
             return;
         }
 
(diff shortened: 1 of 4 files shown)


src/main/java/com/aws/greengrass/status/FleetStatusService.java
```diff
@@ -295,24 +296,36 @@ public void schedulePeriodicFleetStatusDataUpdate(boolean isDuringConnectionResu
...
+        });
+        return overAllStatus;
+    }
+
     @SuppressWarnings("PMD.UnusedFormalParameter")
     private void handleServiceStateChange(GreengrassService greengrassService, State oldState,
                                           State newState) {
+
+        Set<GreengrassService> erroredComponentSet = new HashSet<>();
+
```
junfuchen99, line 314: what is this set

src/main/java/com/aws/greengrass/status/FleetStatusService.java
```diff
@@ -295,24 +296,36 @@ public void schedulePeriodicFleetStatusDataUpdate(boolean isDuringConnectionResu
...
-            });
+            uploadFleetStatusServiceData(erroredComponentSet, getOverAllStatus().get(), null, Trigger.COMPONENT_STATUS_CHANGE);
+            return;
+        }
 
-            uploadFleetStatusServiceData(erroredComponentSet, overAllStatus.get(), null, Trigger.ERRORED_COMPONENT);
+        // if the reported errored component is covered, remove the component from the set and update the component status
+        if (greengrassService.reachedDesiredState() && erroredComponentsSet.remove(greengrassService) ) {
+            uploadFleetStatusServiceData(erroredComponentSet, getOverAllStatus().get(), null, Trigger.COMPONENT_STATUS_CHANGE);
             return;
```
junfuchen99, line 329: 1. If a deployment is in progress, should this message be sent? What happens when an errored component is being reinstalled?
2. If the component is broken, you will also need to remove it from errored component set.
3. Is there any scenario where you want to send this message but reachedDesiredState is false?

src/main/java/com/aws/greengrass/status/FleetStatusService.java
```diff
@@ -116,6 +116,7 @@ public class FleetStatusService extends GreengrassService {
     private ScheduledFuture<?> periodicUpdateFuture;
     // default to zero so that first Reconnect update would go thru
     private Instant lastReconnectUpdateTime = Instant.EPOCH;
+    private static Set<GreengrassService> erroredComponentsSet = new HashSet<>();
```
junfuchen99, line 119: It doesn't need to be static

src/main/java/com/aws/greengrass/status/FleetStatusService.java
yitingb: 1. Right it shouldn't. When an errored component is being reinstalled, that means either Nucleus restart or another deployment. If Nucleus restarts, the list will be emptied out and start over. If there is another deployment, the component will be removed already since it will either enter broken or desired state from last deployment. Deployment can't be cancelled when component starts install (after config merge)
2. Yes, will update it
3. The only scenario we consider to use this message is when a RUNNING component gets errored and being recovered. And here we define recovery is when the component finish it's lifecycle and reached RUNNING state again. So I think the whole story needs to rely on reachedDesiredState is true

src/main/java/com/aws/greengrass/status/FleetStatusService.java
```diff
@@ -295,24 +296,40 @@ public void schedulePeriodicFleetStatusDataUpdate(boolean isDuringConnectionResu
                 initialDelay, periodicPublishIntervalSec, TimeUnit.SECONDS);
     }
 
+    private AtomicReference<OverallStatus> getOverAllStatus() {
```
alter-mage, line 299: i think `overall` is just one word so your method should be named `getOverallStatus`

src/main/java/com/aws/greengrass/status/FleetStatusService.java
```diff
@@ -295,24 +296,40 @@ public void schedulePeriodicFleetStatusDataUpdate(boolean isDuringConnectionResu
...
+        Instant now = Instant.now();
+        AtomicReference<OverallStatus> overAllStatus = new AtomicReference<>(OverallStatus.HEALTHY);
```
alter-mage, line 301: same as above, `overallStatus`

src/main/java/com/aws/greengrass/status/FleetStatusService.java
```diff
@@ -295,24 +296,40 @@ public void schedulePeriodicFleetStatusDataUpdate(boolean isDuringConnectionResu
...
+            overAllStatus.set(getOverallStatusBasedOnServiceState(overAllStatus.get(), service));
+        });
+        return overAllStatus;
+    }
+
     @SuppressWarnings("PMD.UnusedFormalParameter")
     private void handleServiceStateChange(GreengrassService greengrassService, State oldState,
                                           State newState) {
+
+        Set<GreengrassService> erroredComponentSet = new HashSet<>();
```
alter-mage, line 313: Why do you need a copy of the `erroredComponentsSet`? is it not possible to reuse the existing set?

alter-mage: Also, it looks like you are adding instances of services which recovered in the `erroredComponentSet`. Maybe, rename with a more relevant name? like `recoveredComponentSet`?

src/main/java/com/aws/greengrass/status/FleetStatusService.java
```diff
@@ -295,24 +296,40 @@ public void schedulePeriodicFleetStatusDataUpdate(boolean isDuringConnectionResu
...
-            });
+            uploadFleetStatusServiceData(erroredComponentSet, getOverAllStatus().get(), null, Trigger.COMPONENT_STATUS_CHANGE);
+            return;
+        }
 
-            uploadFleetStatusServiceData(erroredComponentSet, overAllStatus.get(), null, Trigger.ERRORED_COMPONENT);
+        // if the reported errored component is recovered, remove the component from the set and update the component status
+        if (!isDeploymentInProgress.get() && greengrassService.reachedDesiredState() && erroredComponentsSet.remove(greengrassService)) {
+            erroredComponentSet.add(greengrassService);
+            uploadFleetStatusServiceData(erroredComponentSet, getOverAllStatus().get(), null, Trigger.COMPONENT_STATUS_CHANGE);
```
alter-mage, line 332: So, lets say that you have 2 services which recover from an errored state outside a deployment. FSS state change handler will receive 2 updates.
1. Update 1 adds ServiceA to `erroredComponentState` and triggers upload to FSS. ServiceA recovery is published.
2. Update 2 adds ServiceB to `erroredComponentState` and triggers upload to FSS again. This time both ServiceA and ServiceB recovery is published.

Is this the desired design?

yitingb: Yes, we only update partial fss message to that specific component, the design logic is pretty much the same as errored component, we update whichever failed. The only difference is we only update the recovery ones when they happen outside of deployments

src/main/java/com/aws/greengrass/status/FleetStatusService.java
yitingb: Updated it to one liner

src/main/java/com/aws/greengrass/status/FleetStatusService.java
yitingb: Updated

src/main/java/com/aws/greengrass/status/FleetStatusService.java
yitingb: Updated

src/main/java/com/aws/greengrass/status/FleetStatusService.java
```diff
@@ -295,24 +298,41 @@ public void schedulePeriodicFleetStatusDataUpdate(boolean isDuringConnectionResu
...
+    private AtomicReference<OverallStatus> getOverallStatus() {
+        Instant now = Instant.now();
+        AtomicReference<OverallStatus> overallStatus = new AtomicReference<>(OverallStatus.HEALTHY);
+        this.kernel.orderedDependencies().forEach(service -> {
+            serviceFssTracksMap.put(service, now);
+            overallStatus.set(getOverallStatusBasedOnServiceState(overallStatus.get(), service));
+        });
+        return overallStatus;
+    }
+
```
junfuchen99, line 310: There are other places calling `getOverallStatusBasedOnServiceState` and looks like you refactor those as well

src/main/java/com/aws/greengrass/status/FleetStatusService.java
```diff
@@ -322,9 +342,10 @@ private void handleServiceStateChange(GreengrassService greengrassService, State
 
         // if there is no ongoing deployment and we encounter a BROKEN component, update the fleet status as UNHEALTHY.
         if (!isDeploymentInProgress.get() && newState.equals(State.BROKEN)) {
+            erroredComponentsSet.remove(greengrassService);
```
junfuchen99, line 345: Add a comment here explaining why you are removing it

PR #1577, opened 2024-02-09
feat(c2cuc): add basic functionality for updating configuration of other components

aws-kevinrickard:
**Issue #, if available:**
GG-44771
GG-44748

**Description of changes:**
* use snapshot version of aws-iot-device-sdk for new UpdateConfiguration field
* add basic functionality for updating configuration of other components
* add logging
  * of note: I considered whether to put this in the UpdateConfiguration request handler vs in the Authorization handler. I decided on the UpdateConfiguration request handler but am open to changing it- let me know. The point is we want any component-to-component configuration updates to be logged, so that if an unintended configuration update happens it can be traced back to the requesting component.
* The error message produced by GetConfiguration when a requested component can not be found was misleading as it read `Key not found`, so I updated it to also use the error message I introduced for UpdateConfiguration when a component is not found.

As a developer building components to run in Greengrass, I want to update a component A's config from an other component B, so that B can adjust the behavior of A at runtime.

* for now, merging to a feature branch (`c2c-update-configuration`)
* Authorization changes are necessary for this feature and will follow

**Documentation Checklist:**

**Compatibility Checklist:**
- [X] Any modification or deletion of public interfaces does not impact other plugin components.


src/main/java/com/aws/greengrass/builtin/services/configstore/ConfigStoreIPCEventStreamAgent.java
```diff
@@ -270,7 +274,14 @@ public UpdateConfigurationResponse handleRequest(UpdateConfigurationRequest requ
 
                 Object value = request.getValueToMerge();
 
-                Topics serviceTopics = kernel.findServiceTopic(serviceName);
+                logger.atInfo().kv(SERVICE_NAME, serviceName).log(String.format(
```
MikeDombo, line 277: should not info log. This is at best debug, and possibly even trace.

MikeDombo: Do not use String.format, our logger will correctly format everything for you. Use the logger's formatting.

src/main/java/com/aws/greengrass/builtin/services/configstore/ConfigStoreIPCEventStreamAgent.java
```diff
@@ -270,7 +274,14 @@ public UpdateConfigurationResponse handleRequest(UpdateConfigurationRequest requ
...
+                        "Updating configuration for target component %s, key path %s, and value keys {%s}",
+                        targetServiceName, //todo-krickar fix value replacement
+                        Arrays.toString(keyPath),
+                        String.join(", ", request.getValueToMerge().keySet())
+                ));
+
+                Topics serviceTopics = kernel.findServiceTopic(targetServiceName);
```
MikeDombo, line 284: you need to check for authorization before starting to do the lookup and applying changes.

src/test/java/com/aws/greengrass/builtin/services/configstore/ConfigStoreIPCEventStreamAgentTest.java
```diff
@@ -265,6 +266,20 @@ void GIVEN_agent_running_WHEN_update_config_request_for_ACL_THEN_update_fails()
...
+        when(mockAuthenticationData.getIdentityLabel()).thenReturn(TEST_COMPONENT_A);
+        UpdateConfigurationRequest request = new UpdateConfigurationRequest();
+        request.setComponentName(TEST_COMPONENT_B);
+        request.setKeyPath(Collections.EMPTY_LIST);
+        request.setValueToMerge(Collections.singletonMap(ACCESS_CONTROL_NAMESPACE_TOPIC, Collections.singletonMap(
+                "aws.greengrass.ipc.pubsub", "policy")));
+        request.setTimestamp(Instant.now());
+        Exception e = assertThrows(InvalidArgumentsError.class,
+                () -> agent.getUpdateConfigurationHandler(mockContext).handleRequest(request));
+        assertTrue(e.getMessage().contains("Config update is not allowed for following fields"));
```
MikeDombo, line 280: use hamcrest. `assertThat(e.getMessage(), containsString(...)`

src/main/java/com/aws/greengrass/builtin/services/configstore/ConfigStoreIPCEventStreamAgent.java
```diff
@@ -327,14 +338,17 @@ private void validateRequest(UpdateConfigurationRequest request) {
...
 
-            Topics serviceTopics = kernel.findServiceTopic(serviceName);
-            if (serviceTopics == null) {
-                throw new InvalidArgumentsError("Component config not found for component " + serviceName);
+            final String targetServiceName =
+                    request.getComponentName() == null ? serviceName : request.getComponentName();
+
+            Topics targetServiceTopics = kernel.findServiceTopic(targetServiceName);
+            if (targetServiceTopics == null) {
+                throw new InvalidArgumentsError(String.format(COMPONENT_NOT_FOUND_ERROR_FORMAT, targetServiceName));
```
MikeDombo, line 346: switch this to `ResourceNotFound` if the updateConfiguration API is allowed to throw that type.

aws-kevinrickard: It is not- allowed types are:
                 ServiceError, 
                 UnauthorizedError, 
                 ConflictError, 
                 FailedUpdateConditionCheckError, 
                 InvalidArgumentsError

src/main/java/com/aws/greengrass/builtin/services/configstore/ConfigStoreIPCEventStreamAgent.java
aws-kevinrickard: Will do

src/main/java/com/aws/greengrass/builtin/services/configstore/ConfigStoreIPCEventStreamAgent.java
MikeDombo: Thank you for checking

src/test/java/com/aws/greengrass/builtin/services/configstore/ConfigStoreIPCEventStreamAgentTest.java
aws-kevinrickard: Will do

src/main/java/com/aws/greengrass/builtin/services/configstore/ConfigStoreIPCEventStreamAgent.java
aws-kevinrickard: The intent is we want any component-to-component configuration updates to be logged, so that if an unintended configuration update happens it can be traced back to the requesting component. If that is done at debug or trace, those logs would not be saved under normal circumstances.

However, I think there may be a better place for info level logging that meets that need. Junfu mentioned that info level logging may already be present for IPC operations. And I was also considering having the logging happen somewhere in the Authorization flow. I plan on looking more into those, unless you have other thoughts on this

MikeDombo: We should not be logging everything at info. Info is the default level that we have enabled. IPC is used far too often to have this type logged at info level, it must be debug.

aws-kevinrickard: I see; sounds good. I suppose for the use case of tracing unwanted actions, it's reasonable to enable debug level logging first, and then monitor for it.

I'll make this log here debug. And if I find a better place for it when I'll update it at that point.

aws-kevinrickard:
Authorization changes still WIP, setting this PR to draft for that

PR #1578, opened 2024-02-15
feat(fss): update fss trigger type when component status change

yitingb:
**Issue #, if available:**

**Description of changes:**
Combining the fss trigger type `ERRORED_COMPONENT` and `BROKEN_COMPONENT` to one type: `COMPONENT_STATUS_CHANGE` 
We currently only have ERROED and BROKEN trigger types, and it's not enough to cover all component lifecycle. 

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/status/model/Trigger.java
+++ b/src/main/java/com/aws/greengrass/status/model/Trigger.java
@@ -11,8 +11,7 @@ public enum Trigger {
     LOCAL_DEPLOYMENT,
     THING_DEPLOYMENT,
     THING_GROUP_DEPLOYMENT,
-    ERRORED_COMPONENT,
-    BROKEN_COMPONENT,
+    COMPONENT_STATUS_CHANGE,
     // when mqtt connection resumes
     RECONNECT,
     // when nucleus initially connects IoT Core, a complete FSS update is sent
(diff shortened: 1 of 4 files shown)


alter-mage (approved):

tiancishen (approved):

src/main/java/com/aws/greengrass/status/model/Trigger.java
```diff
@@ -11,8 +11,7 @@ public enum Trigger {
     LOCAL_DEPLOYMENT,
     THING_DEPLOYMENT,
     THING_GROUP_DEPLOYMENT,
-    ERRORED_COMPONENT,
-    BROKEN_COMPONENT,
+    COMPONENT_STATUS_CHANGE,
```
pradheCloud, line 14: Add the new field in the middle of the enum will cause issue with older nucleus versions, which are not aware of this field. This new enum item should be added to the end of the list.

Enums use ordering internally to generate `int` as ordinals for comparision. 
When enums are marshalled, the int is used to the digest. Same goes for any comparision. 
In this case; ordinal value of `COMPONENT_STATUS_CHANGE` is 5, but for older nucleus without this trigger type, ordinal 5 is `RECONNECT`.

Example playground code: https://onecompiler.com/java/425mnfkbw

Java spec: https://docs.oracle.com/javase/8/docs/api/java/lang/Enum.html#ordinal--

Particular line of interest: _"The natural order implemented by this method is the order in which the constants are declared."_

MikeDombo: That is only true if you're using the ordinal somwhere. We should always be passing enums as strings

pradheCloud: Ack. 

`.toString` will return the name of the enum in this case. 

Because we use Jackson to marshal the data before sending it out of the process, we should be good.

PR #1581, opened 2024-02-23
fix: timeout http server connections when unable to retrieve credentials

MikeDombo:
**Issue #, if available:**

**Description of changes:**
Timeout credential requests after 30 seconds. Without this, we can end up with tons of threads all waiting for credentials since a new thread is created for each incoming request and those threads are not interrupted even if the client disconnects (due to a client disconnect).

Also avoiding thread creation entirely by using a single thread for all requests, if we do not have credentials to vend out anyway then blocking additional requests should be fine.

**Documentation Checklist:**

**Compatibility Checklist:**
- [x] For external library version updates, I have reviewed its change logs and Nucleus does not consume 

### DIFF
--- a/src/main/java/com/aws/greengrass/tes/CredentialRequestHandler.java
+++ b/src/main/java/com/aws/greengrass/tes/CredentialRequestHandler.java
@@ -40,6 +40,8 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicReference;
 import javax.inject.Inject;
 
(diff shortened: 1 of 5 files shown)


src/main/java/com/aws/greengrass/tes/CredentialRequestHandler.java
```diff
@@ -174,6 +179,44 @@ public void handle(final HttpExchange exchange) throws IOException {
...
+            try {
+                // block along with any other threads so we don't send multiple requests
+                if (timeout == 0) {
+                    future.get();
+                } else {
+                    future.get(timeout, timeUnit);
+                }
+            } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+            } catch (ExecutionException ignore) {
```
MikeDombo, line 208: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

**Problem**: An exception has been ignored. The catch-block hides information about the stack trace. Not logging or forwarding the stack trace of an exception can mask unexpected errors and complicated debugging.

**Fix**: Use a catch block to handle the exception and write a log message declaring that the exception happened. 
[Learn more.](https://www.ibm.com/support/pages/best-practice-catching-and-re-throwing-java-exceptions)

Suggested remediation:
Handle all exceptions by creating a log for each exception using the message thrown.
```
@@ -209,2 +209,3 @@
                 // We never complete the future exceptionally
+                log.error(&#34;ExecutionException is caught.&#34;, ignore);
             }

```

jcosentino11 (approved):

tiancishen (approved):

junfuchen99 (approved):

PR #1584, opened 2024-02-29
feat(fss): update component status for terminal states

yitingb:
**Issue #, if available:**

**Description of changes:**
1. Once Nucleus finished restarting process and all components are up and running, Nucleus sends an additional fss message to console with all components status
2. Once a component being recovered, Nucleus waits for all other components being recovered and send an additional message notifying all components has been recovered.

We would like to improve our customer experience on sending extra messages indicating when all components in Nucleus are up and running

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/status/ComponentStatusChangeFleetStatusServiceTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/status/ComponentStatusChangeFleetStatusServiceTest.java
@@ -1,0 +1,214 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.aws.greengrass.integrationtests.status;
+
+import com.aws.greengrass.dependency.State;
+import com.aws.greengrass.deployment.DeviceConfiguration;
+import com.aws.greengrass.integrationtests.util.ConfigPlatformResolver;
+import com.aws.greengrass.lifecyclemanager.Kernel;
+import com.aws.greengrass.mqttclient.MqttClient;
+import com.aws.greengrass.mqttclient.PublishRequest;
+import com.aws.greengrass.status.FleetStatusService;
+import com.aws.greengrass.status.model.FleetStatusDetails;
+import com.aws.greengrass.status.model.OverallStatus;
+import com.aws.greengrass.status.model.Trigger;
+import com.aws.greengrass.testcommons.testutilities.GGExtension;
+import com.aws.greengrass.testcommons.testutilities.NoOpPathOwnershipHandler;
+import com.fasterxml.jackson.databind.JsonMappingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.DisabledOnOs;
+import org.junit.jupiter.api.condition.OS;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
+
+import static com.aws.greengrass.status.FleetStatusService.FLEET_STATUS_SERVICE_TOPICS;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.when;
+
+@ExtendWith({GGExtension.class, MockitoExtension.class})
+public class ComponentStatusChangeFleetStatusServiceTest {
+
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+    private static DeviceConfiguration deviceConfiguration;
+    private static Kernel kernel;
+    @Mock
+    private MqttClient mqttClient;
+
+    private AtomicReference<List<FleetStatusDetails>> fleetStatusDetailsList;
+
+    private CountDownLatch statusChange;
+
+    private final static String SERVICE_A = "ServiceA";
+    private final static String SERVICE_B = "ServiceB";
+    private final static String SERVICE_C = "ServiceC";
+
+    @BeforeEach
+    void setupKernel() throws Exception {
+        fleetStatusDetailsList = new AtomicReference<>(new ArrayList<>());
+        kernel = new Kernel();
+        NoOpPathOwnershipHandler.register(kernel);
+        kernel.getContext().put(MqttClient.class, mqttClient);
+
+        when(mqttClient.publish(any(PublishRequest.class))).thenAnswer(i -> {
+            Object argument = i.getArgument(0);
+            PublishRequest publishRequest = (PublishRequest) argument;
+            try {
+                FleetStatusDetails fleetStatusDetails = OBJECT_MAPPER.readValue(publishRequest.getPayload(), FleetStatusDetails.class);
+                if (fleetStatusDetails.getTrigger().equals(Trigger.COMPONENT_STATUS_CHANGE)) {
+                    statusChange.countDown();
+                }
+                fleetStatusDetailsList.get().add(fleetStatusDetails);
+            } catch (JsonMappingException ignored) {
+            }
+            return CompletableFuture.completedFuture(0);
+        });
+    }
+    @AfterEach
+    void afterEach() {
+        if (kernel != null) {
+            kernel.shutdown();
+        }
+    }
+    @Test
+    void GIVEN_one_component_errored_and_recovered_THEN_fss_should_send_recovery_message() throws Exception {
+        statusChange = new CountDownLatch(2);
+        deviceConfiguration = new DeviceConfiguration(kernel, "ThingName", "xxxxxx-ats.iot.us-east-1.amazonaws.com",
+                "xxxxxx.credentials.iot.us-east-1.amazonaws.com", "privKeyFilePath", "certFilePath", "caFilePath",
+                "us-east-1", "roleAliasName");
+        kernel.getContext().put(DeviceConfiguration.class, deviceConfiguration);
+        ConfigPlatformResolver.initKernelWithMultiPlatformConfig(kernel,
+                ComponentStatusChangeFleetStatusServiceTest.class.getResource("fss_service_recovery.yaml"));
+        kernel.launch();
+
+        FleetStatusService fss = (FleetStatusService) kernel.locateIgnoreError(FLEET_STATUS_SERVICE_TOPICS);
+        fss.setWaitBetweenPublishDisabled(true);
+
+        assertTrue(statusChange.await(20, TimeUnit.SECONDS));
+
+        // we expect a total of 3 messages; 1 nucleus launch and 2 component status change (error + recovery)
+        assertEquals(3, fleetStatusDetailsList.get().size());
+
+        // the first message should be nucleus launch
+        FleetStatusDetails fleetStatusDetails = fleetStatusDetailsList.get().get(0);
+        assertEquals(OverallStatus.HEALTHY, fleetStatusDetails.getOverallStatus());
+        assertEquals(Trigger.NUCLEUS_LAUNCH, fleetStatusDetails.getTrigger());
+
+        // the second message should be errored component
+        fleetStatusDetails = fleetStatusDetailsList.get().get(1);
+        assertEquals(Trigger.COMPONENT_STATUS_CHANGE, fleetStatusDetails.getTrigger());
+        componentStatusAssertion(fleetStatusDetails, SERVICE_A, State.ERRORED);
+
+        // the third message should be component recovery
+        fleetStatusDetails = fleetStatusDetailsList.get().get(2);
+        assertEquals(Trigger.COMPONENT_STATUS_CHANGE, fleetStatusDetails.getTrigger());
+        componentStatusAssertion(fleetStatusDetails, SERVICE_A, State.RUNNING);
+    }
+
+    @Test
+    void GIVEN_two_components_errored_and_recovered_THEN_fss_should_send_only_one_recovery_message() throws Exception {
+        statusChange  = new CountDownLatch(3);
+        deviceConfiguration = new DeviceConfiguration(kernel, "ThingName", "xxxxxx-ats.iot.us-east-1.amazonaws.com",
+                "xxxxxx.credentials.iot.us-east-1.amazonaws.com", "privKeyFilePath", "certFilePath", "caFilePath",
+                "us-east-1", "roleAliasName");
+        kernel.getContext().put(DeviceConfiguration.class, deviceConfiguration);
+        ConfigPlatformResolver.initKernelWithMultiPlatformConfig(kernel,
+                ComponentStatusChangeFleetStatusServiceTest.class.getResource("fss_service_5s_recovery.yaml"));
+        kernel.launch();
+
+        FleetStatusService fss = (FleetStatusService) kernel.locateIgnoreError(FLEET_STATUS_SERVICE_TOPICS);
+        fss.setWaitBetweenPublishDisabled(true);
+        //Increase this for windows testing
+        assertTrue(statusChange.await(60, TimeUnit.SECONDS));
+        // we expect a total of 4 messages, 1 Nucleus launch, 3 component status change includes:
+        // 1 Errored from A, 1 Errored B, 1 recovery message for both
+        assertEquals(4, fleetStatusDetailsList.get().size());
+
+        // the first message should be nucleus launch
+        FleetStatusDetails fleetStatusDetails = fleetStatusDetailsList.get().get(0);
+        assertEquals(OverallStatus.HEALTHY, fleetStatusDetails.getOverallStatus());
+        assertEquals(Trigger.NUCLEUS_LAUNCH, fleetStatusDetails.getTrigger());
+
+        // the second and third message should be errored component
+        fleetStatusDetails = fleetStatusDetailsList.get().get(1);
+        assertEquals(Trigger.COMPONENT_STATUS_CHANGE, fleetStatusDetails.getTrigger());
+        componentStatusAssertion(fleetStatusDetails, SERVICE_B, State.ERRORED);
+
+        fleetStatusDetails = fleetStatusDetailsList.get().get(2);
+        assertEquals(Trigger.COMPONENT_STATUS_CHANGE, fleetStatusDetails.getTrigger());
+        componentStatusAssertion(fleetStatusDetails, SERVICE_A, State.ERRORED);
+
+        // the fourth message should be component recovery
+        fleetStatusDetails = fleetStatusDetailsList.get().get(3);
+        assertEquals(Trigger.COMPONENT_STATUS_CHANGE, fleetStatusDetails.getTrigger());
+        componentStatusAssertion(fleetStatusDetails, SERVICE_A, State.RUNNING);
+        componentStatusAssertion(fleetStatusDetails, SERVICE_B, State.RUNNING);
+    }
+    @Test
+    @DisabledOnOs(OS.WINDOWS)
+    void GIVEN_three_components_errored_and_recovered_THEN_fss_should_send_only_one_recovery_message_and_one_errored_message_should_contain_recovery_message() throws Exception {
+        statusChange  = new CountDownLatch(4);
+        deviceConfiguration = new DeviceConfiguration(kernel, "ThingName", "xxxxxx-ats.iot.us-east-1.amazonaws.com",
+                "xxxxxx.credentials.iot.us-east-1.amazonaws.com", "privKeyFilePath", "certFilePath", "caFilePath",
+                "us-east-1", "roleAliasName");
+        kernel.getContext().put(DeviceConfiguration.class, deviceConfiguration);
+        ConfigPlatformResolver.initKernelWithMultiPlatformConfig(kernel,
+                ComponentStatusChangeFleetStatusServiceTest.class.getResource("fss_service_10s_recovery.yaml"));
+        kernel.launch();
+
+        FleetStatusService fss = (FleetStatusService) kernel.locateIgnoreError(FLEET_STATUS_SERVICE_TOPICS);
+        fss.setWaitBetweenPublishDisabled(true);
+        //Increase this for windows testing
+        assertTrue(statusChange.await(60, TimeUnit.SECONDS));
+        // we expect a total of 5 messages, 1 Nucleus launch, 4 component status change includes:
+        // 1 Errored from A with B reovered, 1 Errored B, 1 Errored C, 1 recovery message for the rest of non recovery ones
+        assertEquals(5, fleetStatusDetailsList.get().size());
+
+        // the first message should be nucleus launch
+        FleetStatusDetails fleetStatusDetails = fleetStatusDetailsList.get().get(0);
+        assertEquals(OverallStatus.HEALTHY, fleetStatusDetails.getOverallStatus());
+        assertEquals(Trigger.NUCLEUS_LAUNCH, fleetStatusDetails.getTrigger());
+
+        // the second should be errored component C
+        fleetStatusDetails = fleetStatusDetailsList.get().get(1);
+        assertEquals(Trigger.COMPONENT_STATUS_CHANGE, fleetStatusDetails.getTrigger());
+        componentStatusAssertion(fleetStatusDetails, SERVICE_C, State.ERRORED);
+
+        // the 3rd message should be errored component B
+        fleetStatusDetails = fleetStatusDetailsList.get().get(2);
+        assertEquals(Trigger.COMPONENT_STATUS_CHANGE, fleetStatusDetails.getTrigger());
+        componentStatusAssertion(fleetStatusDetails, SERVICE_B, State.ERRORED);
+
+        // the 4th message should be errrored component A with recovery of B
+        fleetStatusDetails = fleetStatusDetailsList.get().get(3);
+        assertEquals(Trigger.COMPONENT_STATUS_CHANGE, fleetStatusDetails.getTrigger());
+        componentStatusAssertion(fleetStatusDetails, SERVICE_A, State.ERRORED);
+        assertTrue(fleetStatusDetails.getComponentDetails().stream()
+                .anyMatch(c -> "ServiceB".equals(c.getComponentName()) && !State.ERRORED.equals(c.getState())));
+
+        // the 5th message should be rest component recovery
+        fleetStatusDetails = fleetStatusDetailsList.get().get(4);
+        assertEquals(Trigger.COMPONENT_STATUS_CHANGE, fleetStatusDetails.getTrigger());
+        componentStatusAssertion(fleetStatusDetails, SERVICE_C, State.RUNNING);
+        componentStatusAssertion(fleetStatusDetails, SERVICE_A, State.RUNNING);
+    }
+    private void componentStatusAssertion(FleetStatusDetails fleetStatusDetails, String componentName, State state) {
+        assertTrue(fleetStatusDetails.getComponentDetails().stream()
+                .anyMatch(c -> componentName.equals(c.getComponentName()) && state.equals(c.getState())));
+    }
+}
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/status/EventFleetStatusServiceTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/status/EventFleetStatusServiceTest.java
@@ -82,6 +82,7 @@ import java.util.function.Consumer;
 
 import static com.aws.greengrass.deployment.IotJobsHelper.UPDATE_DEPLOYMENT_STATUS_ACCEPTED;
 import static com.aws.greengrass.deployment.model.Deployment.DeploymentType;
+import static com.aws.greengrass.status.FleetStatusService.FLEET_STATUS_SERVICE_TOPICS;
 import static com.aws.greengrass.testcommons.testutilities.ExceptionLogProtector.ignoreExceptionOfType;
 import static com.aws.greengrass.testcommons.testutilities.ExceptionLogProtector.ignoreExceptionUltimateCauseOfType;
 import static com.aws.greengrass.testcommons.testutilities.TestUtils.createCloseableLogListener;
@@ -126,9 +127,7 @@ class EventFleetStatusServiceTest extends BaseITCase {
     private ThingGroupHelper thingGroupHelper;
 
     private AtomicReference<List<FleetStatusDetails>> fleetStatusDetailsList;
-
     private final CountDownLatch mainFinished = new CountDownLatch(1);
-
     @Captor
     private ArgumentCaptor<Consumer<UpdateJobExecutionResponse>> jobsAcceptedHandlerCaptor;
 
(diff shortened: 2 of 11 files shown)


src/main/java/com/aws/greengrass/status/FleetStatusService.java
```diff
@@ -298,38 +306,79 @@ public void schedulePeriodicFleetStatusDataUpdate(boolean isDuringConnectionResu
...
                 overAllStatus.set(getOverallStatusBasedOnServiceState(overAllStatus.get(), service));
             });
-
-            uploadFleetStatusServiceData(erroredComponentSet, overAllStatus.get(), null,
+            // Not reporting status of other components in errored component message because a deployment could
+            // be in-progress and other component's fleet config arn may not have updated
+            Set<GreengrassService> erroredComponent = new HashSet<>();
+            erroredComponent.add(greengrassService);
+            if (isDeploymentInProgress.get()) {
+                uploadFleetStatusServiceData(erroredComponent, overAllStatus.get(), null,
```
alter-mage, line 322: since you are not using erroredComponentSet anywhere else in this method, can you create an inline set in the call to `uploadFleetStatusServiceData`?

like 
```
uploadFleetStatusServiceData(Collections.singleton(greengrassService)), ...
```

src/main/java/com/aws/greengrass/status/FleetStatusService.java
```diff
@@ -298,38 +306,79 @@ public void schedulePeriodicFleetStatusDataUpdate(boolean isDuringConnectionResu
...
+            if (State.BROKEN.equals(state)) {
+                continue;
+            }
+            allReachedTerminalState = false;
+        }
+        //2. If all other components are in terminal state, update status for all component
+        // Skip updates when kernel shutdown since service ends with FINISHED for stopping
+        if (allReachedTerminalState && !isDeploymentInProgress.get()
+                && !kernelLifecycle.getIsShutdownInitiated().get()) {
+            updateFleetStatusUpdateForAllComponents(Trigger.COMPONENT_STATUS_CHANGE);
```
alter-mage, line 376: do you need to update fss data for all components here? why not update for per component?

src/main/java/com/aws/greengrass/status/FleetStatusService.java
```diff
@@ -298,38 +306,79 @@ public void schedulePeriodicFleetStatusDataUpdate(boolean isDuringConnectionResu
...
+            Set<GreengrassService> erroredComponent = new HashSet<>();
+            erroredComponent.add(greengrassService);
+            if (isDeploymentInProgress.get()) {
+                uploadFleetStatusServiceData(erroredComponent, overAllStatus.get(), null,
+                        Trigger.COMPONENT_STATUS_CHANGE);
+                return;
+            }
+            // Report status of other components, in case recovery duration takes too long and other components need
+            // to wait until all components to reach terminal state to update status
+            uploadFleetStatusServiceData(updatedGreengrassServiceSet, overAllStatus.get(), null,
```
junfuchen99, line 328: why are you using two different messages for this update? it should be merged within the errored component update

src/main/java/com/aws/greengrass/status/FleetStatusService.java
```diff
@@ -298,38 +306,79 @@ public void schedulePeriodicFleetStatusDataUpdate(boolean isDuringConnectionResu
...
+            // if there is no ongoing deployment and we encounter a BROKEN component,
+            // update the fleet status as UNHEALTHY.
+            if (newState.equals(State.BROKEN)) {
+                synchronized (updatedGreengrassServiceSet) {
+                    uploadFleetStatusServiceData(updatedGreengrassServiceSet, OverallStatus.UNHEALTHY,
+                            null, Trigger.COMPONENT_STATUS_CHANGE);
+                }
+            }
+            // Check and update component status when a component reaches to terminal state
+            if (greengrassService.reachedDesiredState() || newState.equals(State.BROKEN)) {
```
junfuchen99, line 346: why do you need to check newState.equals(State.BROKEN)

src/main/java/com/aws/greengrass/status/FleetStatusService.java
```diff
@@ -298,38 +306,79 @@ public void schedulePeriodicFleetStatusDataUpdate(boolean isDuringConnectionResu
...
+                handleServiceReachedTerminalState();
             }
         }
     }
 
+    private void handleServiceReachedTerminalState() {
+        //1. Check all other component status
```
junfuchen99, line 353: this part of checking all services in terminal state can move to Kernel class

src/main/java/com/aws/greengrass/status/FleetStatusService.java
```diff
@@ -298,38 +306,79 @@ public void schedulePeriodicFleetStatusDataUpdate(boolean isDuringConnectionResu
...
+        List<GreengrassService> servicesToTrack = kernel.orderedDependencies().stream()
+                    .filter(GreengrassService::shouldAutoStart).collect(Collectors.toList());
+        boolean allReachedTerminalState = true;
+        for (GreengrassService service : servicesToTrack) {
+            State state = service.getState();
+            if (!service.reachedDesiredState()) {
+                allReachedTerminalState = false;
+                break;
+            }
+            if (State.RUNNING.equals(state) || State.FINISHED.equals(state) || !service.shouldAutoStart()
```
junfuchen99, line 363: can you explain this if condition?

src/main/java/com/aws/greengrass/status/FleetStatusService.java
junfuchen99, line 376: do we update all components? should this be only the components that have changed state?

src/main/java/com/aws/greengrass/status/FleetStatusService.java
yitingb: I'm trying to cover the scenario where the last errored component eventually becomes BROKEN.

src/main/java/com/aws/greengrass/status/FleetStatusService.java
yitingb: We don't want to change the existing deployment behavior right?

src/main/java/com/aws/greengrass/status/FleetStatusService.java
yitingb: Yes, we only send one fss update for all components reaches to terminal state. We want to avoid the scenario where 100 components recovered within 1s at the same time. And also trying to align with the behavior with deployment: only one message when all components are up and running

src/main/java/com/aws/greengrass/status/FleetStatusService.java
yitingb: Updated

src/main/java/com/aws/greengrass/status/FleetStatusService.java
yitingb: Updated

src/main/java/com/aws/greengrass/status/FleetStatusService.java
yitingb: Updated

src/main/java/com/aws/greengrass/status/FleetStatusService.java
yitingb: Updated

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
```diff
@@ -592,4 +594,31 @@ public void shutdown(int timeoutSeconds) {
...
+        List<GreengrassService> servicesToTrack = kernel.orderedDependencies().stream()
+                .filter(GreengrassService::shouldAutoStart).collect(Collectors.toList());
+        boolean allReachedTerminalState = true;
+        for (GreengrassService service : servicesToTrack) {
+            State state = service.getState();
+            if (!service.reachedDesiredState()) {
+                allReachedTerminalState = false;
+                break;
+            }
+            if (State.RUNNING.equals(state) || State.FINISHED.equals(state) || !service.shouldAutoStart()
```
alter-mage, line 613: `!service.shouldAutoStart()`
this is always false, right?

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
```diff
@@ -592,4 +594,31 @@ public void shutdown(int timeoutSeconds) {
...
+                    && service.reachedDesiredState()) {
```
alter-mage, line 614: `service.reachedDesiredState()`
i think your check on 609 covers this

src/main/java/com/aws/greengrass/status/FleetStatusService.java
```diff
@@ -366,13 +412,10 @@ public void triggerFleetStatusUpdateAtKernelLaunch() {
...
         Set<GreengrassService> greengrassServiceSet = new HashSet<>();
-        AtomicReference<OverallStatus> overAllStatus = new AtomicReference<>(OverallStatus.HEALTHY);
-
-        // Get all running services from the Nucleus to update the fleet status.
-        this.kernel.orderedDependencies().forEach(greengrassService -> {
-            greengrassServiceSet.add(greengrassService);
-            overAllStatus.set(getOverallStatusBasedOnServiceState(overAllStatus.get(), greengrassService));
-        });
+        AtomicReference<OverallStatus> overAllStatus = updateDeviceOverallStatus();
+        for (GreengrassService service : serviceFssTracksMap.keySet()) {
```
alter-mage, line 416: i believe you can replace this entire loop with a bulk set constructor call like
`Set<GreengrassService> greengrassServiceSet = new HashSet<>(serviceFssTracksMap.keySet());`

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
yitingb: Yes, but if we look at the loop from `waitForServicesToStart` it's also checking this condition. I think the meaning for it to be here is in future, if we have some components should not auto start but in the desired state, we also want to consider that service is running

alter-mage: Sure, but you are already filtering out services with should auto start as true at 604. If your filtered list of services will always autostart then `!service.shouldAutoStart()` will always be false, right?

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
yitingb: Updated

src/main/java/com/aws/greengrass/status/FleetStatusService.java
yitingb: Updated

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
yitingb: Right, updated

src/main/java/com/aws/greengrass/status/FleetStatusService.java
yitingb: Removed the should autostart check since already filtered before the loop

src/integrationtests/java/com/aws/greengrass/integrationtests/status/EventFleetStatusServiceTest.java
```diff
@@ -126,9 +126,8 @@ class EventFleetStatusServiceTest extends BaseITCase {
     private ThingGroupHelper thingGroupHelper;
 
     private AtomicReference<List<FleetStatusDetails>> fleetStatusDetailsList;
-
-    private final CountDownLatch mainFinished = new CountDownLatch(1);
-
+    //during kernel launch, main component enters finished once then shutdown and finish again
```
alter-mage, line 129: what is the reason for this?

src/integrationtests/java/com/aws/greengrass/integrationtests/status/EventFleetStatusServiceTest.java
```diff
@@ -214,13 +213,40 @@ void setupKernel(ExtensionContext context) throws Exception {
...
 
     @AfterEach
     void afterEach() {
         kernel.shutdown();
     }
 
+    @Test
+    void GIVEN_Nucleus_launch_THEN_fss_send_launch_message_and_running_message() throws Exception {
+        Thread.sleep(10000); // make sure Nucleus finished launching then verify
+        FleetStatusDetails fleetStatusDetails = fleetStatusDetailsList.get().get(0);
```
alter-mage, line 240: why check the overall status from the zeroth element of the list?
i think you should assert that overall status remained healthy throughout nucleus launch

src/integrationtests/java/com/aws/greengrass/integrationtests/status/EventFleetStatusServiceTest.java
```diff
@@ -214,13 +213,40 @@ void setupKernel(ExtensionContext context) throws Exception {
...
+        // 2 additional messages will be added to fleetStatusDetailsList. After Nucleus launch, components up and running and then main stop and finish
     }
 
     @AfterEach
     void afterEach() {
         kernel.shutdown();
     }
 
+    @Test
+    void GIVEN_Nucleus_launch_THEN_fss_send_launch_message_and_running_message() throws Exception {
```
alter-mage, line 238: overall, i think this test should assert that a device sends a nucleus launch message first and then component change messages

src/main/java/com/aws/greengrass/status/FleetStatusService.java
```diff
@@ -298,36 +304,71 @@ public void schedulePeriodicFleetStatusDataUpdate(boolean isDuringConnectionResu
...
             this.kernel.orderedDependencies().forEach(service -> {
                 serviceFssTracksMap.put(service, now);
                 overAllStatus.set(getOverallStatusBasedOnServiceState(overAllStatus.get(), service));
             });
-
-            uploadFleetStatusServiceData(erroredComponentSet, overAllStatus.get(), null,
-                    Trigger.COMPONENT_STATUS_CHANGE);
+            // Not reporting status of other components in errored component message because a deployment could
+            // be in-progress and other component's fleet config arn may not have updated
+            Set<GreengrassService> erroredcomponet = new HashSet<>();
```
alter-mage, line 320: nit: erroredComponent

src/main/java/com/aws/greengrass/status/FleetStatusService.java
```diff
@@ -298,36 +304,71 @@ public void schedulePeriodicFleetStatusDataUpdate(boolean isDuringConnectionResu
...
-                        null, Trigger.COMPONENT_STATUS_CHANGE);
+    private void handleServiceReachedTerminalState() {
+        //1. check all other components reached terminal state
+        boolean allReachedTerminalState = kernelLifecycle.allServicesInTerminalState();
+        //2. If all other components are in terminal state, update status for all component
+        // Skip updates when kernel shutdown since service ends with FINISHED for stopping
+        synchronized (updatedGreengrassServiceSet) {
+            if (!updatedGreengrassServiceSet.isEmpty() && allReachedTerminalState && !isDeploymentInProgress.get()
+                    && !kernelLifecycle.getIsShutdownInitiated().get()) {
+                Instant now = Instant.now();
```
alter-mage, line 360: nit: seems like this is a duplicate of the code at 312. maybe extract this into a method?

src/main/java/com/aws/greengrass/status/FleetStatusService.java
```diff
@@ -568,8 +612,12 @@ private void uploadFleetStatusServiceData(Set<GreengrassService> greengrassServi
                 .messageType(MessageType.fromTrigger(trigger))
                 .deploymentInformation(deploymentInformation)
                 .build();
-
-        publisher.publish(fleetStatusDetails, components);
+        try {
+            Thread.sleep(10_000); // Prevent messages sent too frequently and cloud received them out of order
```
alter-mage, line 616: how does this prevent frequent message publish?

src/main/java/com/aws/greengrass/status/FleetStatusService.java
```diff
@@ -568,8 +612,12 @@ private void uploadFleetStatusServiceData(Set<GreengrassService> greengrassServi
...
+            publisher.publish(fleetStatusDetails, components);
+        } catch (InterruptedException e) {
+            throw new RuntimeException(e);
```
alter-mage, line 619: why runtime exception?

alter-mage: if you catch an interrupted exception, you should try to interrupt that thread using `Thread.currentThread().interrupt()`

src/main/java/com/aws/greengrass/status/FleetStatusService.java
yitingb: Before sending each message, sleep for 10 seconds to make sure the previous message is sent already

src/integrationtests/java/com/aws/greengrass/integrationtests/status/EventFleetStatusServiceTest.java
yitingb: I want to make sure the main component reaches to FINISHED state twice

src/integrationtests/java/com/aws/greengrass/integrationtests/status/EventFleetStatusServiceTest.java
yitingb: Because after Nucleus launch, there will be 2 messages added to `fleetStatusDetaiilsList`, and I want to verify the first message containing overall status of healthy

src/main/java/com/aws/greengrass/status/FleetStatusService.java
yitingb: Extracting them into a method creates bugs :(

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
```diff
@@ -592,4 +594,31 @@ public void shutdown(int timeoutSeconds) {
...
     }
+
+    /**
+     * Check if all services has reached to terminal state: RUNNING, FINISHED or BROKEN.
+     *
+     */
+    public boolean allServicesInTerminalState() {
+        //Check all other component status, false if any component is not at the terminal state
+        List<GreengrassService> servicesToTrack = kernel.orderedDependencies().stream()
+                .filter(GreengrassService::shouldAutoStart).collect(Collectors.toList());
```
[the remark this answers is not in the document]
yitingb: It's basically all components, it's always returning true and we never set it to false

src/main/java/com/aws/greengrass/status/FleetStatusService.java
```diff
@@ -298,36 +304,71 @@ public void schedulePeriodicFleetStatusDataUpdate(boolean isDuringConnectionResu
...
+                uploadFleetStatusServiceData(erroredcomponet, overAllStatus.get(), null,
+                        Trigger.COMPONENT_STATUS_CHANGE);
+                return;
+            }
+            // Report status of other components, in case recovery duration takes too long and other components need
+            // to wait until all components to reach terminal state to update status
+            synchronized (updatedGreengrassServiceSet) {
+                uploadFleetStatusServiceData(updatedGreengrassServiceSet, overAllStatus.get(), null,
+                        Trigger.COMPONENT_STATUS_CHANGE);
+            }
```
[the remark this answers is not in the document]
yitingb: No, for non-deployment, we want to send terminal state info also in the errored component updates, and then send the remaining when everything recovers. 
For deployment, we only want to send the info for that single component errored, the other terminal states only means the lifecycle is up and running, it doesn't mean some component is recovered

src/main/java/com/aws/greengrass/status/FleetStatusService.java
yitingb: With refactoring, the `updatedGreengrassServiceSet` will not always contain the correct values, I observed that sometimes deployment fss message doesn't contain any component details, which means `updatedGreengrassServiceSet` already got cleared. There will be thread timing issue

src/integrationtests/java/com/aws/greengrass/integrationtests/status/EventFleetStatusServiceTest.java
yitingb: Updated

src/main/java/com/aws/greengrass/status/FleetStatusService.java
yitingb: Updated

src/main/java/com/aws/greengrass/status/FleetStatusService.java
yitingb: Updated

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
```diff
@@ -592,4 +594,31 @@ public void shutdown(int timeoutSeconds) {
...
+            State state = service.getState();
+            //Consider RUNNING, FINISHED and BROKEN as terminal state, as component reached to the end of lifecycle
+            if (State.BROKEN.equals(state)) {
+                continue;
+            }
+            if (!service.reachedDesiredState()) {
+                allReachedTerminalState = false;
+                break;
+            }
+            if (State.RUNNING.equals(state) || State.FINISHED.equals(state)) {
```
alter-mage, line 617: nit: you can combine this with condition on 610

src/main/java/com/aws/greengrass/status/FleetStatusService.java
```diff
@@ -568,10 +617,25 @@ private void uploadFleetStatusServiceData(Set<GreengrassService> greengrassServi
...
+        // make sure messages are not published too frequently to cause message drop from cloud
+        if (lastFSSPublishTime.plusSeconds(FLEET_STATUS_MESSAGE_PUBLISH_MAX_WAIT_TIME_SEC)
+                .isBefore(Instant.now())) {
+            publisher.publish(fleetStatusDetails, components);
+            lastFSSPublishTime = Instant.now();
+            logger.atInfo().event("fss-status-update-published").kv("trigger", trigger)
+                    .log("Status update published to FSS");
+        } else {
+            // Calculate the time that next message needs to wait.
+            Duration duration = Duration.between(Instant.now(), lastFSSPublishTime
```
alter-mage, line 629: I dont understand how this prevents messages from being dropped by FSS. This is simply scheduling multiple messages to be published at the same time, right?

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
yitingb: But broken is not entering desired state, is that more confusing to write it as `if((reachedDesiredstate && RUNNING || FINISHED) || BROKEN)`

alter-mage: I am not suggesting to club all the conditions together, only club conditions 1 and 3 since they are repetitive

alter-mage (dismissed):

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
yitingb: Updated

src/integrationtests/java/com/aws/greengrass/integrationtests/status/EventFleetStatusServiceTest.java
```diff
@@ -214,13 +217,129 @@ void setupKernel(ExtensionContext context) throws Exception {
...
+        kernel.getContext().put(DeviceConfiguration.class, deviceConfiguration);
+        // pre-load contents to package store
+        Path localStoreContentPath =
+                Paths.get(EventFleetStatusServiceTest.class.getResource("local_store_content").toURI());
+        PreloadComponentStoreHelper.preloadRecipesFromTestResourceDir(localStoreContentPath.resolve("recipes"), kernel.getNucleusPaths().recipePath());
+        copyFolderRecursively(localStoreContentPath.resolve("artifacts"), kernel.getNucleusPaths().artifactPath(), REPLACE_EXISTING);
+        kernel.launch();
+
+        Thread.sleep(25000); // make sure Nucleus finished launching then verify\
+        // assertTrue(fssRunning.await(10, TimeUnit.SECONDS));
```
junfuchen99, line 322: clean this up? you can just wait for count down latches to finish instead of sleeping

src/integrationtests/java/com/aws/greengrass/integrationtests/status/EventFleetStatusServiceTest.java
```diff
@@ -236,10 +355,10 @@ void GIVEN_jobs_deployment_WHEN_deployment_finishes_THEN_status_is_uploaded_to_c
             }
         };
         try (AutoCloseable ignoredListener = createCloseableLogListener(logListener)) {
-
+            fleetStatusDetailsList.get().clear(); // Remove the 2 messages introduced with Nucleus launch
             offerSampleIoTJobsDeployment("FleetStatusServiceConfig.json", TEST_JOB_ID_1);
             assertTrue(fssPublishLatch.await(60, TimeUnit.SECONDS));
-
+            Thread.sleep(25000);
```
junfuchen99, line 361: I don't think you need all these sleeps.

src/main/java/com/aws/greengrass/status/FleetStatusService.java
```diff
@@ -239,6 +249,7 @@ private void setUpFSS() throws DeviceConfigurationException {
                     FLEET_STATUS_SERVICE_TOPICS);
             this.deploymentStatusKeeper.registerDeploymentStatusConsumer(SHADOW, deploymentStatusChanged,
                     FLEET_STATUS_SERVICE_TOPICS);
+            ses = getContext().get(ScheduledExecutorService.class);
```
junfuchen99, line 252: nit: move this to constructor

src/main/java/com/aws/greengrass/status/FleetStatusService.java
junfuchen99: I think we need to address this duplicate code. I don't understand why updatedGreengrassServiceSet can cause race condition. You can extract these codes in a way that does not depend on this set

src/main/java/com/aws/greengrass/status/FleetStatusService.java
```diff
@@ -568,10 +617,27 @@ private void uploadFleetStatusServiceData(Set<GreengrassService> greengrassServi
...
+                publisher.publish(fleetStatusDetails, components);
+                lastFSSPublishTime = Instant.now();
+                logger.atInfo().event("fss-status-update-published").kv("trigger", trigger)
+                        .log("Status update published to FSS");
+            } else {
+                // Calculate the time that next message needs to wait.
+                Duration duration = Duration.between(Instant.now(), lastFSSPublishTime
+                        .plusSeconds(FLEET_STATUS_MESSAGE_PUBLISH_MAX_WAIT_TIME_SEC));
+                lastFSSPublishTime = Instant.now().plusSeconds(duration.getSeconds());
+                ses.schedule(() -> {
```
junfuchen99, line 634: Do you have a test scenario that spams FSS messages and make sure this publish happens every x seconds?

src/integrationtests/java/com/aws/greengrass/integrationtests/status/PeriodicFleetStatusServiceTest.java
junfuchen99, line 1: what is this empty change

src/integrationtests/java/com/aws/greengrass/integrationtests/status/ComponentStatusChangeFleetStatusServiceTest.java
```diff
@@ -0,0 +1,202 @@
...
+
+        // the third message should be component recovery
+        fleetStatusDetails = fleetStatusDetailsList.get().get(2);
+        assertEquals(Trigger.COMPONENT_STATUS_CHANGE, fleetStatusDetails.getTrigger());
+        assertTrue(fleetStatusDetails.getComponentDetails().stream()
+                .anyMatch(c -> "ServiceA".equals(c.getComponentName()) && State.RUNNING.equals(c.getState())));
+    }
+
+    @Test
+    void GIVEN_two_components_errored_and_recovered_THEN_fss_should_send_only_one_recovery_message() throws Exception {
```
junfuchen99, line 116: I don't think this scenario is needed?

src/integrationtests/java/com/aws/greengrass/integrationtests/status/ComponentStatusChangeFleetStatusServiceTest.java
```diff
@@ -0,0 +1,202 @@
...
+        // the first message should be nucleus launch
+        FleetStatusDetails fleetStatusDetails = fleetStatusDetailsList.get().get(0);
+        assertEquals(OverallStatus.HEALTHY, fleetStatusDetails.getOverallStatus());
+        assertEquals(Trigger.NUCLEUS_LAUNCH, fleetStatusDetails.getTrigger());
+
+        // the second should be errored component C
+        fleetStatusDetails = fleetStatusDetailsList.get().get(1);
+        assertEquals(Trigger.COMPONENT_STATUS_CHANGE, fleetStatusDetails.getTrigger());
+        assertTrue(fleetStatusDetails.getComponentDetails().stream()
+                .anyMatch(c -> "ServiceC".equals(c.getComponentName()) && State.ERRORED.equals(c.getState())));
```
junfuchen99, line 178: nit: can we refactor this into a function? like `assertComponentStatus`

src/integrationtests/java/com/aws/greengrass/integrationtests/status/EventFleetStatusServiceTest.java
```diff
@@ -163,6 +163,9 @@ void setupKernel(ExtensionContext context) throws Exception {
                 // filter all event-triggered fss messages
                 if (fleetStatusDetails.getMessageType() == MessageType.PARTIAL) {
                     fleetStatusDetailsList.get().add(fleetStatusDetails);
+                    if (fleetStatusDetails.getTrigger() == Trigger.COMPONENT_STATUS_CHANGE) {
```
junfuchen99, line 166: nit: use `.equals`

src/integrationtests/java/com/aws/greengrass/integrationtests/status/EventFleetStatusServiceTest.java
```diff
@@ -212,8 +215,19 @@ void setupKernel(ExtensionContext context) throws Exception {
         PreloadComponentStoreHelper.preloadRecipesFromTestResourceDir(localStoreContentPath.resolve("recipes"), kernel.getNucleusPaths().recipePath());
         copyFolderRecursively(localStoreContentPath.resolve("artifacts"), kernel.getNucleusPaths().artifactPath(), REPLACE_EXISTING);
         kernel.launch();
+        FleetStatusService fss = (FleetStatusService) kernel.locateIgnoreError(FLEET_STATUS_SERVICE_TOPICS);
+        fss.setWaitBetweenPublishDisabled(true);
         assertTrue(fssRunning.await(10, TimeUnit.SECONDS));
         assertTrue(deploymentServiceRunning.await(10, TimeUnit.SECONDS));
+        assertTrue(mainFinished.await(20, TimeUnit.SECONDS));
+        assertTrue(componentStatusChange.await(60, TimeUnit.SECONDS));
```
junfuchen99, line 223: do these really need 80 seconds?

yitingb: Updated

src/integrationtests/java/com/aws/greengrass/integrationtests/status/EventFleetStatusServiceTest.java
yitingb: Updated

src/integrationtests/java/com/aws/greengrass/integrationtests/status/ComponentStatusChangeFleetStatusServiceTest.java
yitingb: Updated

src/integrationtests/java/com/aws/greengrass/integrationtests/status/ComponentStatusChangeFleetStatusServiceTest.java
yitingb: I think we should still keep it since it's the basic requirement

src/integrationtests/java/com/aws/greengrass/integrationtests/status/PeriodicFleetStatusServiceTest.java
yitingb: updated

junfuchen99 (dismissed):

alter-mage (approved):

junfuchen99 (approved):

PR #1592, opened 2024-03-29
fix: avoid publish until fss is set up

alter-mage:
**Issue #, if available:**

**Description of changes:**
Avoid publishing component status change message until FSS is set up and Nucleus launch message is published.

**Why is this change necessary:**
A deadlock situation can arise if FSS is being set up while a component status change wants to be published during Nucleus startup.

**How was this change tested:**
- [ ] Updated or added new unit tests.
- [ ] Updated or added new integration tests.
- [ ] Updated or added new end-to-end tests.
- [ ] If my code makes a remote network call, it was tested with a proxy.

**Any additional information or context required to review the change:**

**Documentation Checklist:**
 - [ ] Updated the README if applicable.

**Compatibility Checklist:**
- [ ] I confirm that the change is backwards compatible.
- [ ] Any modification or deletion of public interfaces does not impact other plugin components.
- [ ] For external library version updates, I have reviewed its change logs and Nucleus does not consume 
  any deprecated method or type.

Refer to [Compatibility Guidelines](/COMPATIBILITY.md) for more information.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1593, opened 2024-04-02
fix: synchronize on outer Context before inner Value

aws-kevinrickard:
**Issue #, if available:**
n/a

**Description of changes:**
A Context.Value should only be locked after acquiring a lock on the outer Context. 

This prevents circular wait deadlocks when multiple threads are using the Context / Context.Values at the same time.

Ran existing unit tests

Related to https://github.com/aws-greengrass/aws-greengrass-nucleus/pull/1592 , but this is a lower-level fix.

**Documentation Checklist:**
 - [ ] Updated the README if applicable. (n/a)

**Compatibility Checklist:**
- [X] Any modification or deletion of public interfaces does not impact other plugin components.

### DIFF
--- a/src/main/java/com/aws/greengrass/dependency/Context.java
+++ b/src/main/java/com/aws/greengrass/dependency/Context.java
@@ -436,7 +436,6 @@ public class Context implements Closeable {
             return constructObjectWithInjection();
         }
 
-
         /**
          * Put a new object instance and inject fields with pre and post actions, if the new object is not equal to
          * current one.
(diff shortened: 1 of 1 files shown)


src/main/java/com/aws/greengrass/dependency/Context.java
```diff
@@ -553,13 +563,17 @@ private Constructor<T> pickConstructor(Class<T> clazz) throws NoSuchMethodExcept
...
+        public final <E extends Exception> T computeObjectIfEmpty(
                 CrashableFunction<Value, T, E> mappingFunction) throws E {
-            if (object != null) {
-                return object;
-            }
+            synchronized (Context.this) {
+                synchronized (this) {
+                    if (object != null) {
+                        return object;
+                }
```
MikeDombo, line 572: you need to keep holding the lock when calling put and inject, or else the semantics of this change do not match the previous implementation

MikeDombo (commented):
The `synchronized` methods in `Context` can mostly be removed; they don't actually need any synchronization because they're already using threadsafe constructs or are called while holding a different global lock (eg. see caller of globalNotifyStateChanged)

src/main/java/com/aws/greengrass/dependency/Context.java
aws-kevinrickard: Oops, IDE had this collapsed so the indent is wrong. Will fix. The putAndInject is included in the lock scope

Issue #1618, opened 2024-05-08
Greengrass installer not working with AWS SSO Profile


Sourish1997:
**Describe the bug**
The Greengrass nucleus installer is unable to load AWS credentials when using AWS SSO profiles.

**To Reproduce**
1. Create a default profile in `~/.aws/config` and obtain temporary credentials using `aws sso login`.
2. Download and run the Greengrass core installer as per the instructions [here](https://docs.aws.amazon.com/greengrass/v2/developerguide/quick-installation.html#run-greengrass-core-v2-installer)

```
curl -s https://d2s8p88vqu9w66.cloudfront.net/releases/greengrass-nucleus-latest.zip \
> greengrass-nucleus-latest.zip && unzip greengrass-nucleus-latest.zip -d GreengrassInstaller

sudo -E java -Droot="/greengrass/v2" -Dlog.store=FILE -jar ./GreengrassInstaller/lib/Greengrass.jar \
    --aws-region us-east-1 \
    --thing-name "$ROBOT_ID" \
    --thing-group-name "$THING_GROUP_NAME" \
    --tes-role-name "$TES_ROLE_NAME" \
    --tes-role-alias-name "$TES_ROLE_ALIAS_NAME" \
    --provision true \
    --setup-system-service true \
```

**Expected behavior**
The Greengrass core software is successfully installed on the device.

**Actual behavior**
The installer is unable to load AWS credentials. I get the following error:

```
Error while trying to setup Greengrass Nucleus
[...]

PR #1619, opened 2024-05-08
feat: support sso for greengrass installation

jcosentino11:
**Issue #, if available:**

Closes https://github.com/aws-greengrass/aws-greengrass-nucleus/issues/1618

**Description of changes:**

Add AWS SDK SSO dependencies so SSO login can be used during Greengrass installer. See above issue for details.

Built locally and verified with dummy sso profiles in ~/.aws/config that aws sso provider was being used

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -198,6 +198,27 @@
                 </exclusion>
             </exclusions>
         </dependency>
+        <!-- To support SSO login for Greengrass Installer -->
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>sso</artifactId>
+            <exclusions>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -198,6 +198,27 @@
                 </exclusion>
             </exclusions>
         </dependency>
+        <!-- To support SSO login for Greengrass Installer -->
```
MikeDombo, line 201: I don't think we really want a new dependency unless there's a strong justification. We have many ways for customers to provision gg and to get credentials one way or another.
Let's try and help the customer with alternatives first.

PR #1631, opened 2024-06-19
chore: general changes for cancel deployment

tiancishen:
**Issue #, if available:**

**Description of changes:**

1. Use executor service to execute update action for deployment in a thread to prevent direct interruption
3. Before taking a config snapshot check if the deployment has been cancelled
4. Life Cycle should allow requests to restart / reinstall components that are closing

General changes needed to improve cancelling deployments

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
+++ b/src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
@@ -40,6 +40,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
@@ -66,6 +67,7 @@ public class DeploymentConfigMerger {
     private Kernel kernel;
     private DeviceConfiguration deviceConfiguration;
     private DynamicComponentConfigurationValidator validator;
+    private ExecutorService executorService;
 
     /**
      * Merge in new configuration values and new services.
@@ -106,7 +108,11 @@ public class DeploymentConfigMerger {
         } else {
             logger.atInfo().log("Deployment is configured to skip update policy check,"
                     + " not waiting for disruptable time to update");
-            updateActionForDeployment(newConfig, deployment, activator, totallyCompleteFuture);
+            // use executor service to execute updateActionForDeployment
+            // prevents default deployment cancellation from directly interrupting
+            executorService.execute(() -> updateActionForDeployment(newConfig, deployment, activator,
+                    totallyCompleteFuture));
+
         }
 
         return totallyCompleteFuture;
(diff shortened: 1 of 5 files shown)


src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
```diff
@@ -106,7 +108,11 @@ public Future<DeploymentResult> mergeInNewConfig(Deployment deployment,
         } else {
             logger.atInfo().log("Deployment is configured to skip update policy check,"
                     + " not waiting for disruptable time to update");
-            updateActionForDeployment(newConfig, deployment, activator, totallyCompleteFuture);
+            // use executor service to execute updateActionForDeployment
+            // prevents default deployment cancellation from directly interrupting
```
jcosentino11, line 112: Just for my understanding, where's the interrupt happening?

tiancishen: mergeInNewConfig is called here:
https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/main/src/main/java/com/aws/greengrass/deployment/DefaultDeploymentTask.java#L154

and the interruption is caught just below that:
https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/main/src/main/java/com/aws/greengrass/deployment/DefaultDeploymentTask.java#L178

If it's not running in a different thread, it will be interrupted when we try to cancel a deployment here:
https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/main/src/main/java/com/aws/greengrass/deployment/DeploymentService.java#L500

jcosentino11: `updateActionForDeployment` doesn't throw interrupted exception though?

also I'm not entirely clear on the use case, what's the behavior that we're trying to prevent? e.g. deployments aren't getting cancelled properly?

tiancishen: updateActionForDeployment doesn't throw interrupted exception no, but currently it's running on the same thread as the Default Deployment task, which means it will be interrupted when we cancel the deployment Future.

For reference `resolveDependencies` and `preparePackages` both execute in their own thread.

- https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/main/src/main/java/com/aws/greengrass/deployment/DefaultDeploymentTask.java#L124
- https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/main/src/main/java/com/aws/greengrass/componentmanager/ComponentManager.java#L350

When we begin the config merge we don't want it to be interrupted, instead we want to cleanly exit as soon as possible by checking if the CompletableFuture is cancelled or not.

jcosentino11 (approved):

yitingb (approved):

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
```diff
@@ -106,7 +108,11 @@ public Future<DeploymentResult> mergeInNewConfig(Deployment deployment,
...
+            executorService.execute(() -> updateActionForDeployment(newConfig, deployment, activator,
```
junfuchen99, line 113: Do we need to wait for this future to complete?

tiancishen: we do not need to wait for this future to complete, instead we wait on `CompletableFuture totallyCompleteFuture` that's being returned

PR #1637, opened 2024-07-10
fix: discard update action on notify only

saranyailla:
**Issue #, if available:**

**Description of changes:**

**Documentation Checklist:**

**Compatibility Checklist:**


src/main/java/com/aws/greengrass/deployment/DeploymentService.java
```diff
@@ -504,10 +505,12 @@ private void cancelCurrentDeployment() {
                     .isCancellable()) {
                 logger.atInfo().log("Deployment already finished processing or cannot be cancelled");
             } else {
-                boolean canCancelDeployment = context.get(UpdateSystemPolicyService.class).discardPendingUpdateAction(
```
saranyailla, line 507: Currently, this returns `true` for the deployments that do not have the update policy set to  `DeploymentComponentUpdatePolicyAction.NOTIFY_COMPONENTS` which is wrong. We should discarding an update only when the policy is set to notify.

src/main/java/com/aws/greengrass/lifecyclemanager/UpdateSystemPolicyService.java
```diff
@@ -96,20 +94,21 @@ public Set<String> getPendingActions() {
...
-            for (Map.Entry<String, UpdateAction> todo : pendingActions.entrySet()) {
-                try {
-                    actionInProgress.set(todo.getKey());
-                    todo.getValue().getAction().run();
-                    logger.atDebug().setEventType("service-update-action").addKeyValue("action", todo.getKey()).log();
-                } catch (Throwable t) {
-                    logger.atError().setEventType("service-update-action-error").addKeyValue("action", todo.getKey())
-                            .setCause(t).log();
-                }
+            final UpdateAction pendingUpdateAction = pendingActions.remove(deploymentId);
```
saranyailla, line 97: - Pending actions that only correspond to the specified deployment id should be performed. 
- Rely on  `pendingActions` map to check if the update action is in progress/completed or never added.

src/main/java/com/aws/greengrass/lifecyclemanager/UpdateSystemPolicyService.java
```diff
@@ -149,39 +145,41 @@ protected void startup() throws InterruptedException {
...
             logger.atDebug().setEventType("service-update-pending").addKeyValue("numOfUpdates", pendingActions.size())
                     .log();
-
-            boolean ggcRestarting = false;
-            for (UpdateAction action : pendingActions.values()) {
-                if (action.isGgcRestart()) {
-                    ggcRestarting = true;
-                    break;
-                }
-            }
```
saranyailla, line 159: Restarting GG feature flag should be set only when for a single deployment ID.

src/main/java/com/aws/greengrass/lifecyclemanager/UpdateSystemPolicyService.java
```diff
@@ -149,39 +145,41 @@ protected void startup() throws InterruptedException {
...
-
-            PreComponentUpdateEvent preComponentUpdateEvent = new PreComponentUpdateEvent();
-            preComponentUpdateEvent.setIsGgcRestarting(ggcRestarting);
-            String deploymentId = pendingActions.values().stream().map(UpdateAction::getDeploymentId).findFirst().get();
```
saranyailla, line 163: Fix: At this point, `pendingActions` can be empty and result in NPE.

src/main/java/com/aws/greengrass/lifecyclemanager/UpdateSystemPolicyService.java
```diff
@@ -121,16 +120,13 @@ protected void runUpdateActions(String deploymentId) {
      *         false if update actions were already in progress
      */
     public boolean discardPendingUpdateAction(String tag) {
-        if (tag.equals(actionInProgress.get())) {
+        final UpdateAction pendingUpdateAction = pendingActions.remove(tag);
+        if (pendingUpdateAction == null) {
+            // Update action is never added or already in progress.
             return false;
         }
-        final UpdateAction pendingUpdateAction = pendingActions.get(tag);
```
saranyailla, line 127: Fix - race: A thread can reach this point before the update action is completed (but in progress) resulting in returning an incorrect value for `discardPendingUpdateAction` and sending two post update events (one via `runUpdateActions` and other via `discardPendingUpdateAction`). This is because 1/ `runUpdateActions`removes the action from the map only after it is completed 2/ No locking used while accessing/modifying the map . 

Use `remove` to ensure that the updated state is reflected across threads.

src/main/java/com/aws/greengrass/lifecyclemanager/UpdateSystemPolicyService.java
alter-mage: Does this mean that a deferred deployment cannot be cancelled? Ideally, if a component is deferring a config merge due a deployment, this deployment should still be cancellable given that defer period is still ongoing.

src/main/java/com/aws/greengrass/lifecyclemanager/UpdateSystemPolicyService.java
```diff
@@ -149,39 +145,41 @@ protected void startup() throws InterruptedException {
...
+                    throw new RuntimeException(e);
                 }
+            });
+        }
+    }
+
+    private void sendComponentUpdateEvents(UpdateAction action) throws InterruptedException {
+        PreComponentUpdateEvent preComponentUpdateEvent = new PreComponentUpdateEvent();
+        boolean ggcRestarting = action.isGgcRestart();
+        preComponentUpdateEvent.setIsGgcRestarting(ggcRestarting);
```
alter-mage, line 162: nit: use `withIsGgRestarting`

src/main/java/com/aws/greengrass/lifecyclemanager/UpdateSystemPolicyService.java
```diff
@@ -149,39 +145,41 @@ protected void startup() throws InterruptedException {
...
+
+        preComponentUpdateEvent.setDeploymentId(action.getDeploymentId());
```
alter-mage, line 164: nit: same as above

src/main/java/com/aws/greengrass/lifecyclemanager/UpdateSystemPolicyService.java
saranyailla: - Deferred deployments are still cancellable. That code path is not modified. 
- `runUpdateActions` is run only after we waited enough for the deferred time.
- While we're waiting for the deferred time, deployment pending actions can be discarded upon request.

PR #1638, opened 2024-07-11
feat(fips): enable fips support deployment flow

yitingb:
**Issue #, if available:**

**Description of changes:**
1. Combined update device configuration based on existing system property
2. When fips mode is true, download root ca 3 to support iot endpoint
3. When fips is enabled, override the `GreengrassV2DataClient` endpoint with `endpointProvider` 
4. Refresh S3 cached client when `fipsMode` change

This is to support new config fipsMode and to ensure all clients in Nucleus supports fips endpoint

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
+++ b/src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
@@ -30,6 +30,7 @@ import com.aws.greengrass.util.Coerce;
 import com.aws.greengrass.util.LockFactory;
 import com.aws.greengrass.util.LockScope;
 import com.aws.greengrass.util.NucleusPaths;
+import com.aws.greengrass.util.RootCAUtils;
 import com.aws.greengrass.util.Utils;
 import com.aws.greengrass.util.exceptions.TLSAuthException;
 import com.aws.greengrass.util.platforms.Platform;
@@ -49,6 +50,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.concurrent.locks.Lock;
 import javax.inject.Inject;
@@ -130,6 +132,7 @@ public class DeviceConfiguration {
 
     private final Validator deTildeValidator;
     private final Validator regionValidator;
+    private final AtomicBoolean rootCA3Downloaded = new AtomicBoolean(false);
     private final AtomicReference<Boolean> deviceConfigValidateCachedResult = new AtomicReference<>();
 
     private Topics loggingTopics;
@@ -156,10 +159,7 @@ public class DeviceConfiguration {
         handleLoggingConfig();
         getComponentStoreMaxSizeBytes().dflt(COMPONENT_STORE_MAX_SIZE_DEFAULT_BYTES);
         getDeploymentPollingFrequencySeconds().dflt(DEPLOYMENT_POLLING_FREQUENCY_DEFAULT_SECONDS);
-        if (System.getProperty(S3_ENDPOINT_PROP_NAME) != null
-                && System.getProperty(S3_ENDPOINT_PROP_NAME).equalsIgnoreCase(S3EndpointType.REGIONAL.name())) {
-            gets3EndpointType().withValue(S3EndpointType.REGIONAL.name());
-        }
+        handleExistingSystemProperty();
         // reset the cache when device configuration changes
         onAnyChange((what, node) -> deviceConfigValidateCachedResult.set(null));
     }
@@ -193,11 +193,6 @@ public class DeviceConfiguration {
         getRootCAFilePath().withValue(rootCaFilePath);
         getAWSRegion().withValue(awsRegion);
         getIotRoleAlias().withValue(tesRoleAliasName);
-
-        if (System.getProperty(S3_ENDPOINT_PROP_NAME) != null
-                && System.getProperty(S3_ENDPOINT_PROP_NAME).equalsIgnoreCase(S3EndpointType.REGIONAL.name())) {
-            gets3EndpointType().withValue(S3EndpointType.REGIONAL.name());
-        }
         validate();
     }
 
@@ -370,6 +365,11 @@ public class DeviceConfiguration {
 
             // Get the current FIPS mode for the AWS SDK. Default will be false (no FIPS).
             String useFipsMode = Boolean.toString(Coerce.toBoolean(getFipsMode()));
+            //Download CA3 to support iotDataEndpoint
+            if (Coerce.toBoolean(getFipsMode()) && !rootCA3Downloaded.get()) {
+                rootCA3Downloaded.set(RootCAUtils.downloadRootCAsWithPath(Coerce.toString(getRootCAFilePath()),
+                        RootCAUtils.AMAZON_ROOT_CA_3_URL));
+            }
             // Set the FIPS property so our SDK clients will use this FIPS mode by default.
             // This won't change any client that exists already.
             System.setProperty(SdkSystemSetting.AWS_USE_FIPS_ENDPOINT.property(), useFipsMode);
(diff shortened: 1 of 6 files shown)


src/main/java/com/aws/greengrass/util/GreengrassServiceClientFactory.java
```diff
@@ -183,19 +188,38 @@ private void configureClient(DeviceConfiguration deviceConfiguration) {
...
+                @Override
+                public CompletableFuture<Endpoint> resolveEndpoint(GreengrassV2DataEndpointParams endpointParams) {
+                    return CompletableFuture.supplyAsync(() -> Endpoint.builder()
+                            .url(URI.create(greengrassServiceEndpoint))
+                            .build());
+                }
+            };
+            clientBuilder = GreengrassV2DataClient.builder()
+                // Use an empty credential provider because our requests don't need SigV4
+                // signing, as they are going through IoT Core instead
```
junfuchen99, line 206: Why does fips differ from regular request?

junfuchen99: where do you set fips mode? is that already implemented in nucleus?

yitingb: Yes, we can only override the ggdataClient endpoint to iotData fips endpoint by applying the `endpointProvider`, the `endpointOverride` function doesn't have a higher priority than the built-in function `useFips` when creating aws sdk client. 

Yes, the fips mode set up is implemented https://github.com/aws-greengrass/aws-greengrass-nucleus/commit/f188e273f27dd78ebbba4b80551a656d863b1f3d

src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
```diff
@@ -791,4 +793,39 @@ public KeyManager[] getDeviceIdentityKeyManagers() throws TLSAuthException {
...
+     */
+    private void handleExistingSystemProperty() {
+        //handle s3 endpoint type
+        if (System.getProperty(S3_ENDPOINT_PROP_NAME) != null
+                && System.getProperty(S3_ENDPOINT_PROP_NAME).equalsIgnoreCase(S3EndpointType.REGIONAL.name())) {
+            gets3EndpointType().withValue(S3EndpointType.REGIONAL.name());
+        }
+        //handle fips mode
+        String useFipsMode = System.getProperty(SdkSystemSetting.AWS_USE_FIPS_ENDPOINT.property());
+        if (useFipsMode != null && Coerce.toBoolean(useFipsMode)) {
```
saranyailla, line 808: Nit: `Coerce.toBoolean` can handle null. 

Shouldn't we set the system properties based on the device config?
In L 376, we're setting the system property for fips with the device config. So, why are we updating the device config for fips based on the system property again? [getRegionValidator() comes before handleExistingSystemProperty()](https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/fa6cedb36481062cf4fd8213bd4fe31e5f1d9e63/src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java#L160-L164)

src/main/java/com/aws/greengrass/util/S3SdkClientFactory.java
```diff
@@ -89,33 +91,54 @@ public S3Client getS3Client() throws DeviceConfigurationException {
...
-     * Set s3 endpoint type.
+     * Update system property and refresh client.
+     */
+    private void handleSystemPropertyUpdate() {
+        // handle s3 endpoint type
+        boolean refreshCache = handleS3EndpointType(Coerce.toString(deviceConfiguration.gets3EndpointType()));
+
+        // handle fips endpoint, check current useFips value is consistent with deviceConfig fipsMode value,
+        // refresh client if they are inconsistent
+        if (useFips != Coerce.toBoolean(deviceConfiguration.getFipsMode())) {
```
saranyailla, line 110: You should be checking fips system property with the mode in device config. 

Nit: If the s3 client is global and L106 is false, we can return as fips mode is not supported by s3 global endpoint. But, I think it's ok to refresh the client like this.

src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
```diff
@@ -791,4 +793,39 @@ public KeyManager[] getDeviceIdentityKeyManagers() throws TLSAuthException {
...
+            getFipsMode().withValue(useFipsMode);
+        }
+    }
+
+    private void downloadRootCA3() {
```
saranyailla, line 813: Nit: move this method out of this class.

src/main/java/com/aws/greengrass/util/S3SdkClientFactory.java
saranyailla: Add lock for useFips.

saranyailla: or use atomic compareAndSet() as the value obtained for `deviceConfiguration.getFipsMode()` in L110 may not be the same as L112.

yitingb: Updated

src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
yitingb: Updated

src/main/java/com/aws/greengrass/util/S3SdkClientFactory.java
```diff
@@ -89,33 +92,53 @@ public S3Client getS3Client() throws DeviceConfigurationException {
...
-     * Set s3 endpoint type.
+     * Update system property and refresh client.
+     */
+    private void handleSystemPropertyUpdate() {
+        // handle s3 endpoint type
+        boolean refreshCache = handleS3EndpointType(Coerce.toString(deviceConfiguration.gets3EndpointType()));
+
+        // handle fips endpoint, check current useFips value is consistent with deviceConfig fipsMode value,
+        // refresh client if they are inconsistent
+        if (useFips.compareAndSet(Coerce.toBoolean(deviceConfiguration.getFipsMode()), useFips.get())) {
```
saranyailla, line 111: Switch the args to `compareAndSet(boolean expect, boolean update)`

saranyailla:
```
refreshCache = useFips.compareAndSet(useFips.get(), Coerce.toBoolean(deviceConfiguration.getFipsMode()))
```

yitingb: If the fips doesn't change but s3 endpoint changed, the `refreshCache` would be false, but we do want to refresh the client in this case

saranyailla: Right, you need to update that accordingly. That was a suggestion to use the compareAndSet value to refresh client. 
```
if(!isS3EndpointTypeModified(Coerce.toString(deviceConfiguration.gets3EndpointType())){
 if (!useFips.compareAndSet(useFips.get(), Coerce.toBoolean(deviceConfiguration.getFipsMode()))){
  return;
 }
}
refreshClientCache();

```

saranyailla: Right, you need to update that accordingly. That was a suggestion to use the compareAndSet value to refresh client. 
```
if(isS3EndpointTypeModified(Coerce.toString(deviceConfiguration.gets3EndpointType()) ||  useFips.compareAndSet(useFips.get(), Coerce.toBoolean(deviceConfiguration.getFipsMode()))){
  refreshClientCache();
}
return;
```

yitingb: Updated

saranyailla (dismissed):

saranyailla (approved):

junfuchen99 (approved):

PR #1639, opened 2024-07-12
feat(fips): accept endpoints from external config

yitingb:
**Issue #, if available:**

**Description of changes:**
1. Get the endpoints from external config file before creating a thing
2. Assign endpoints accordingly

Currently, when customer uses auto provisioning and provide endpoints with init-config, the endpoints will be overwritten when creating thing info. We now check the external config first before creating endpoints for auto provisioning flow

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/easysetup/DeviceProvisioningHelper.java
+++ b/src/main/java/com/aws/greengrass/easysetup/DeviceProvisioningHelper.java
@@ -188,7 +188,7 @@ public class DeviceProvisioningHelper {
      */
     public ThingInfo createThingForE2ETests() {
         return createThing(iotClient, E2E_TESTS_POLICY_NAME_PREFIX,
-                E2E_TESTS_THING_NAME_PREFIX + UUID.randomUUID().toString());
+                E2E_TESTS_THING_NAME_PREFIX + UUID.randomUUID().toString(), "", "");
     }
 
     /**
@@ -197,9 +197,13 @@ public class DeviceProvisioningHelper {
      * @param client     iotClient to use
      * @param policyName policyName
      * @param thingName  thingName
+     * @param iotDataEndpoint  iotDataEndpoint
+     * @param iotCredEndpoint  iotCredEndpoint
      * @return created thing info
      */
-    public ThingInfo createThing(IotClient client, String policyName, String thingName) {
+    @SuppressWarnings("PMD.UseObjectForClearerAPI")
+    public ThingInfo createThing(IotClient client, String policyName, String thingName,
+                                 String iotDataEndpoint, String iotCredEndpoint) {
         // Find or create IoT policy
         try {
             client.getPolicy(GetPolicyRequest.builder().policyName(policyName).build());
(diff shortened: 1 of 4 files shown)


junfuchen99 (approved):

tiancishen (dismissed):

src/main/java/com/aws/greengrass/easysetup/DeviceProvisioningHelper.java
```diff
@@ -216,6 +220,16 @@ public ThingInfo createThing(IotClient client, String policyName, String thingNa
                     .build());
         }
 
+        // handle endpoints
+        if (iotDataEndpoint == null || iotDataEndpoint.isEmpty()) {
```
saranyailla, line 224: Nit: Use `Utils.isEmpty` instead.

saranyailla: Nit: With this change, if customer's set this endpoint to a non-empty invalid value, we won't use and update it to the correct ones as we used to do before.

yitingb: Updated

src/main/java/com/aws/greengrass/easysetup/GreengrassSetup.java
```diff
@@ -510,9 +517,20 @@ private String peekArg() {
...
+        String iotDataEndpoint = "";
+        String iotCredEndpoint = "";
+        if (Coerce.toBoolean(kernel.getConfig().find(SERVICES_NAMESPACE_TOPIC,
+                DEFAULT_NUCLEUS_COMPONENT_NAME, CONFIGURATION_CONFIG_KEY, DEVICE_PARAM_FIPS_MODE))) {
+            iotDataEndpoint  = Coerce.toString(kernel.getConfig().find(SERVICES_NAMESPACE_TOPIC,
+                    DEFAULT_NUCLEUS_COMPONENT_NAME, CONFIGURATION_CONFIG_KEY, DEVICE_PARAM_IOT_DATA_ENDPOINT));
+            iotCredEndpoint = Coerce.toString(kernel.getConfig().find(SERVICES_NAMESPACE_TOPIC,
+                    DEFAULT_NUCLEUS_COMPONENT_NAME, CONFIGURATION_CONFIG_KEY, DEVICE_PARAM_IOT_CRED_ENDPOINT));
+        }
+
```
saranyailla, line 530: I agree with your previous code to update the endpoints only if those are not provided irrespective of the fips flag. 

Here, you can use findOrDefault with default making DescribeEndpointRequest

yitingb: Sure, I prefer to change it back, because all requests made through iot client are handled in `DeviceProvisioningHelper` and `GreengrassSetup` handles configs with kernel. I'm a bit hesitate to inject kernel into `DeviceProvisioningHelper`

yitingb: Updated

saranyailla (approved):

junfuchen99 (approved):

PR #1642, opened 2024-07-15
feat: allow bootstrap deployments to cancel while waiting for service…

tiancishen:
**Issue #, if available:**

**Description of changes:**
1. Removed old waitForServicesToStart function
2. Added default no-op cancel function to DeploymentTask
3. Enabled Kernel Update Deployments to be cancellable
4. Remove boolean cancellable from DeploymentTaskMetadata
5. Kernel Update Deployments overrides DeploymentTask cancel function to cancel the CompletableFuture
6. Refactor Kernel Update Deployments to use new waitForServicesToStart function

Allows deployments with bootstrap to be cancelled

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/deployment/DeploymentService.java
+++ b/src/main/java/com/aws/greengrass/deployment/DeploymentService.java
@@ -228,8 +228,7 @@ public class DeploymentService extends GreengrassService {
                     // Handle IoT Jobs cancellation
 
                     if (currentDeploymentTaskMetadata != null && currentDeploymentTaskMetadata.getDeploymentType()
-                            .equals(nextDeployment.getDeploymentType())
-                            && currentDeploymentTaskMetadata.isCancellable()) {
+                            .equals(nextDeployment.getDeploymentType())) {
                         // Cancel the current deployment if it's an IoT Jobs deployment
                         // that is in progress and still cancellable.
                         logger.atInfo().kv(DEPLOYMENT_ID_LOG_KEY_NAME, currentDeploymentTaskMetadata.getDeploymentId())
@@ -238,8 +237,7 @@ public class DeploymentService extends GreengrassService {
                                 .log("Canceling current deployment");
                         // Send interrupt signal to the deployment task.
                         cancelCurrentDeployment();
-                    } else if (currentDeploymentTaskMetadata != null
-                            && !currentDeploymentTaskMetadata.isCancellable()) {
+                    } else if (currentDeploymentTaskMetadata != null) {
                         // Ignore the cancelling signal if the deployment is NOT cancellable any more.
                         logger.atInfo().kv(DEPLOYMENT_ID_LOG_KEY_NAME, currentDeploymentTaskMetadata.getDeploymentId())
                                 .kv(GG_DEPLOYMENT_ID_LOG_KEY_NAME,
@@ -500,22 +498,23 @@ public class DeploymentService extends GreengrassService {
     private void cancelCurrentDeployment() {
         if (currentDeploymentTaskMetadata.getDeploymentResultFuture() != null && !currentDeploymentTaskMetadata
                 .getDeploymentResultFuture().isCancelled()) {
-            if (currentDeploymentTaskMetadata.getDeploymentResultFuture().isDone() || !currentDeploymentTaskMetadata
-                    .isCancellable()) {
+            if (currentDeploymentTaskMetadata.getDeploymentResultFuture().isDone()) {
                 logger.atInfo().log("Deployment already finished processing or cannot be cancelled");
             } else {
-                boolean canCancelDeploymentUpdateAction = context.get(UpdateSystemPolicyService.class)
-                        .discardPendingUpdateAction(((DefaultDeploymentTask) currentDeploymentTaskMetadata
-                                .getDeploymentTask()).getDeployment().getGreengrassDeploymentId());
-                if (!canCancelDeploymentUpdateAction) {
-                    logger.atWarn().kv(DEPLOYMENT_ID_LOG_KEY_NAME, currentDeploymentTaskMetadata.getDeploymentId())
-                            .kv(GG_DEPLOYMENT_ID_LOG_KEY_NAME,
-                                    currentDeploymentTaskMetadata.getGreengrassDeploymentId())
-                            .log("Cancelling deployment, changes may already have been applied. "
-                                    + "Make a new deployment to revert the update");
+                DeploymentTask deploymentTask = currentDeploymentTaskMetadata.getDeploymentTask();
+                if (deploymentTask instanceof  DefaultDeploymentTask) {
+                    DefaultDeploymentTask defaultDeploymentTask = (DefaultDeploymentTask) deploymentTask;
+                    context.get(UpdateSystemPolicyService.class)
+                            .discardPendingUpdateAction(defaultDeploymentTask.getDeployment()
+                                    .getGreengrassDeploymentId());
                 }
+                logger.atWarn().kv(DEPLOYMENT_ID_LOG_KEY_NAME, currentDeploymentTaskMetadata.getDeploymentId())
+                        .kv(GG_DEPLOYMENT_ID_LOG_KEY_NAME,
+                                currentDeploymentTaskMetadata.getGreengrassDeploymentId())
+                        .log("Cancelling deployment, changes may already have been applied. "
+                                + "Make a new deployment to revert the update");
 
-                currentDeploymentTaskMetadata.getDeploymentResultFuture().cancel(true);
+                currentDeploymentTaskMetadata.cancel(true);
                 DeploymentType deploymentType = currentDeploymentTaskMetadata.getDeploymentType();
                 if (DeploymentType.SHADOW.equals(deploymentType) || DeploymentType.LOCAL.equals(deploymentType)) {
                     deploymentStatusKeeper.persistAndPublishDeploymentStatus(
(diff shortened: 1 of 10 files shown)


src/main/java/com/aws/greengrass/deployment/DeploymentService.java
```diff
@@ -544,7 +546,7 @@ private void createNewDeployment(Deployment deployment) {
             deploymentTask = createDefaultNewDeployment(deployment);
         } else {
             deploymentTask = createKernelUpdateDeployment(deployment);
-            cancellable = false;
+            cancellable = true;
```
junfuchen99, line 549: do we still need this?

src/main/java/com/aws/greengrass/deployment/KernelUpdateDeploymentTask.java
```diff
@@ -38,49 +42,72 @@ public class KernelUpdateDeploymentTask implements DeploymentTask {
...
 
     @SuppressWarnings({"PMD.AvoidDuplicateLiterals"})
     @Override
     public DeploymentResult call() {
+        executorService.execute(this::waitForServicesToStart);
+        DeploymentResult result;
+        try {
+            result = deploymentResultCompletableFuture.get();
+        } catch (InterruptedException | ExecutionException | CancellationException e) {
+            return null;
```
junfuchen99, line 75: why are you returning null here? is this result not being handled anywhere else? should you throw this exception out instead?

junfuchen99 (commented):
Could you also add an integration test?

src/main/java/com/aws/greengrass/deployment/KernelUpdateDeploymentTask.java
tiancishen: If we want to throw the exception then it would only be the interrupted exception.

The other two will occur when a deployment is cancelled, because the exceptions are already handled when waiting for services to start

src/main/java/com/aws/greengrass/deployment/DeploymentService.java
```diff
@@ -504,18 +504,20 @@ private void cancelCurrentDeployment() {
...
-                boolean canCancelDeploymentUpdateAction = context.get(UpdateSystemPolicyService.class)
-                        .discardPendingUpdateAction(((DefaultDeploymentTask) currentDeploymentTaskMetadata
-                                .getDeploymentTask()).getDeployment().getGreengrassDeploymentId());
-                if (!canCancelDeploymentUpdateAction) {
-                    logger.atWarn().kv(DEPLOYMENT_ID_LOG_KEY_NAME, currentDeploymentTaskMetadata.getDeploymentId())
-                            .kv(GG_DEPLOYMENT_ID_LOG_KEY_NAME,
-                                    currentDeploymentTaskMetadata.getGreengrassDeploymentId())
-                            .log("Cancelling deployment, changes may already have been applied. "
-                                    + "Make a new deployment to revert the update");
+                DeploymentTask deploymentTask = currentDeploymentTaskMetadata.getDeploymentTask();
```
saranyailla, line 507: Are we adding this additional check?
For `KernelDeploymentTask` tasks, `isCancellable` is false.

src/main/java/com/aws/greengrass/deployment/DeploymentService.java
```diff
@@ -848,7 +850,8 @@ public static ComponentRecipe copyRecipeFileToComponentStore(ComponentStore comp
     }
 
     private KernelUpdateDeploymentTask createKernelUpdateDeployment(Deployment deployment) {
-        return new KernelUpdateDeploymentTask(kernel, logger.createChild(), deployment, componentManager);
+        return new KernelUpdateDeploymentTask(kernel, logger.createChild(), deployment, componentManager,
+                executorService);
```
saranyailla, line 854: Don't pass executorService.

src/main/java/com/aws/greengrass/deployment/KernelUpdateDeploymentTask.java
```diff
@@ -38,49 +42,72 @@ public class KernelUpdateDeploymentTask implements DeploymentTask {
...
             long mergeTimestamp = kernel.getConfig().lookup("system", "rootpath").getModtime();
-            logger.atDebug().kv("serviceToTrack", servicesToTrack).kv("mergeTime", mergeTimestamp)
+
+            logger.atInfo().kv("serviceToTrack", servicesToTrack).kv("mergeTime", mergeTimestamp)
                     .log("Nucleus update workflow waiting for services to complete update");
-            DeploymentConfigMerger.waitForServicesToStart(servicesToTrack, mergeTimestamp, kernel);
+            DeploymentConfigMerger.waitForServicesToStart(servicesToTrack, mergeTimestamp, kernel,
+                    deploymentResultCompletableFuture);
+            if (deploymentResultCompletableFuture.isCancelled()) {
+                return;
```
saranyailla, line 95: Add a log here that we won't proceed with the deployment as it is cancelled.

src/main/java/com/aws/greengrass/deployment/DeploymentService.java
tiancishen: > Why are we adding this additional check?

We are adding this check because only `DefaultDeploymentTask` allows you to get the Deployment object, `KernelDeploymentTask` does not have a getter for it.

> For KernelDeploymentTask tasks, isn't isCancellable false?

We want to make `KernelDeploymentTask` cancellable as part of this code change

src/main/java/com/aws/greengrass/deployment/KernelUpdateDeploymentTask.java
```diff
@@ -38,49 +42,72 @@ public class KernelUpdateDeploymentTask implements DeploymentTask {
...
+            }
 
-            DeploymentResult result = null;
```
saranyailla, line 71: Nit: You can do this initialization outside of `try block`, update result value like how it was before but now, update that in catch blocks as well. In L-147, complete the future with result.

src/main/java/com/aws/greengrass/deployment/KernelUpdateDeploymentTask.java
tiancishen: We already have enough logging.
DeploymentConfigMerger.waitForServicesToStart already logs that deployment was cancelled while waiting for services to start
DeploymentService already logs that we are trying to cancel the deployment then deployment is cancelled

I can add a debug log here for "kernel udpate deployment is cancelled"

src/main/java/com/aws/greengrass/deployment/DeploymentService.java
tiancishen: removed

src/main/java/com/aws/greengrass/deployment/DeploymentService.java
tiancishen: ack, retrieving from kernel context instead

src/main/java/com/aws/greengrass/deployment/KernelUpdateDeploymentTask.java
tiancishen: updated

src/main/java/com/aws/greengrass/deployment/DeploymentService.java
```diff
@@ -228,8 +228,7 @@ protected void startup() throws InterruptedException {
                     // Handle IoT Jobs cancellation
 
                     if (currentDeploymentTaskMetadata != null && currentDeploymentTaskMetadata.getDeploymentType()
-                            .equals(nextDeployment.getDeploymentType())
-                            && currentDeploymentTaskMetadata.isCancellable()) {
+                            .equals(nextDeployment.getDeploymentType())) {
```
saranyailla, line 231: We can still keep this check.

saranyailla: I see that we're removing `cancellable` field altogether.

src/main/java/com/aws/greengrass/deployment/model/DeploymentTask.java
```diff
@@ -10,4 +10,7 @@
 public interface DeploymentTask extends Callable<DeploymentResult> {
     @Override
     DeploymentResult call() throws InterruptedException;
+
+    default void cancel() {
+    }
```
saranyailla, line 15: Do we need this?

src/main/java/com/aws/greengrass/deployment/KernelUpdateDeploymentTask.java
```diff
@@ -141,4 +161,9 @@ private DeploymentException getDeploymentStatusDetails() {
 
         return new DeploymentException(deployment.getStageDetails(), errorStack, errorTypes);
     }
+
+    @Override
+    public void cancel() {
+        deploymentResultCompletableFuture.cancel(false);
+    }
```
saranyailla, line 168: We're already canceling the `deploymentResultCompletableFuture` in `DeploymentTaskMetada`. No need to have the `cancel` method here to do the same.

src/main/java/com/aws/greengrass/deployment/DeploymentService.java
tiancishen: The boolean field for this was removed because both Default Deployment and Kernel Update Deployment are cancellable now, it will always evaluate to true

src/main/java/com/aws/greengrass/deployment/model/DeploymentTask.java
tiancishen: Default Deployment also inherits from DeploymentTask.
We can either define a no-op default function here, for define it as no-op in Default Deployment

src/main/java/com/aws/greengrass/deployment/KernelUpdateDeploymentTask.java
tiancishen: The cancel function in DeploymentTaskMetadata calls this cancel function first to cancel the CompletableFuture for waitForServicesToStart and the Future that waits for KernelUpdateDeployment to return the deployment results.

It's two different futures, both needs to be cancelled for KernelUpdateDeployment

alter-mage (approved):

saranyailla (approved):

PR #1657, opened 2024-08-28
fix: fail deployment if prepare bootstrap fails

alter-mage:
**Issue #, if available:**

**Description of changes:**
Do not restart Nucleus if kernel update deployment fails while preparing bootstrap. Instead, fail the deployment and start up all the services currently configured.

Something which is a bit unknown is what should the behavior be like if rollback fails.

Nucleus restart is unnecessary if preparing bootstrap fails and adds the possibility of a Nucleus Restart Failure when the Nucleus does eventually start up.

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/deployment/activator/KernelUpdateActivator.java
+++ b/src/main/java/com/aws/greengrass/deployment/activator/KernelUpdateActivator.java
@@ -89,7 +89,13 @@ public class KernelUpdateActivator extends DeploymentActivator {
             kernelAlternatives.prepareBootstrap(deploymentDocument.getDeploymentId());
         } catch (IOException e) {
             // TODO: better handling of error codes for different IO operations
-            rollback(deployment, e);
+            rollback(deployment, e, false);
+            lifecycle.startupAllServices();
+
+            totallyCompleteFuture.complete(
+                    new DeploymentResult(DeploymentResult.DeploymentStatus.FAILED_NO_STATE_CHANGE,
+                            new DeploymentException("Unable to process deployment. Greengrass could not prepare for"
+                                    + " bootstrap.", e)));
             return;
         }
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/aws/greengrass/deployment/activator/KernelUpdateActivator.java
```diff
@@ -89,7 +89,13 @@ public void activate(Map<String, Object> newConfig, Deployment deployment,
             kernelAlternatives.prepareBootstrap(deploymentDocument.getDeploymentId());
         } catch (IOException e) {
             // TODO: better handling of error codes for different IO operations
-            rollback(deployment, e);
+            rollback(deployment, e, false);
```
jcosentino11, line 92: Unit tests?

alter-mage: oops :(

src/main/java/com/aws/greengrass/deployment/activator/KernelUpdateActivator.java
```diff
@@ -120,10 +126,13 @@ void rollback(Deployment deployment, Throwable failureCause) {
         deployment.setErrorTypes(errorReport.getRight());
         deployment.setStageDetails(Utils.generateFailureMessage(failureCause));
 
-        final boolean bootstrapOnRollbackRequired = kernelAlternatives.prepareBootstrapOnRollbackIfNeeded(
-                kernel.getContext(), deploymentDirectoryManager, bootstrapManager);
+        // Persist deployment metadata only if restart is required
+        if (requiresRestart) {
```
yitingb, line 130: QQ: when bootstrap prepared failed, and we don't persist the deployment metadata, if we use systemctl to restart Nucleus, would it still look for deployment metadata when restarting?

alter-mage: We will still persist deployment metadata but only the launch directory of the deployment that failed. After rollback, the launch directory will revert back to previous one (before current deployment was received). So after restart, yes, Nucleus will still look for deployment metadata but its fine if it does not find it and will launch in default mode.

PR #1659, opened 2024-09-06
chore: override state transitions into topic

alter-mage:
**Issue #, if available:**

**Description of changes:**
Override changes to lifecycle state transition into the Nucleus config tree.

Today, we first compare the timestamps of the previously persisted lifecycle state and the incoming lifecycle state. If, due to clock skew, previously persisted state is in the future compared to the incoming state, the incoming state is rejected. This can cause an endless loop of timestamp comparisons between persisted and incoming state, blocking the main thread, bricking the device.

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/config/Topic.java
+++ b/src/main/java/com/aws/greengrass/config/Topic.java
@@ -128,6 +128,10 @@ public class Topic extends Node {
         return withNewerValue(this.modtime, nv);
     }
 
+    private Topic overrideValue(Object nv, long proposedModTime) {
+        return withNewerValue(proposedModTime, nv);
+    }
+
     /**
      * Update the value in place without changing the timestamp.
      * @param nv new value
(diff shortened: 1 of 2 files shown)


src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
```diff
@@ -441,7 +441,7 @@ void setState(State current, StateTransitionEvent stateTransitionEvent) {
         // Sync on State.class to make sure the order of setValue and globalNotifyStateChanged
         // are consistent across different services.
         try (LockScope ls = LockScope.lock(globalLock)) {
-            stateTopic.withValue(newState.ordinal());
+            stateTopic.overrideValue(newState.ordinal());
             statusCodeTopic.withValue(stateTransitionEvent.getStatusCode().name());
             statusReasonTopic.withValue(stateTransitionEvent.getStatusReason());
```
MikeDombo, line 446: the statuses should be overridden too

alter-mage: oh yeah, silly me

yitingb (approved):

junfuchen99 (approved):

src/main/java/com/aws/greengrass/config/Topic.java
```diff
@@ -128,6 +128,10 @@ private Topic overrideValue(Object nv) {
         return withNewerValue(this.modtime, nv);
     }
 
+    private Topic overrideValue(Object nv, long proposedModTime) {
```
MikeDombo, line 131: This doesn't make sense, it won't do anything different from with newer value.
The point of overriding is that the value provided here is going to be set no matter what the timestamp is on the node. This implementation won't do that, this is a straight up alias for with newer value.

MikeDombo (changes requested):

Issue #1670, opened 2024-11-15
(ipc): resources are not completely freed when closing subscription


felixeriksson:
**Describe the bug**
Memory leaks when an IPC subscription is created and closed.

Heap dumps from affected machines show that the memory is hogged by the object `Map<String, SubscriptionTrie<K>> children` in an instance of the class `SubscriptionTrie` (see [source](https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/main/src/main/java/com/aws/greengrass/builtin/services/pubsub/SubscriptionTrie.java)). The `SubscriptionTrie` instance that hogs the memory is called `listeners` which is a member of the class `PubSubIPCEventStreamAgent` ([source](https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/main/src/main/java/com/aws/greengrass/builtin/services/pubsub/PubSubIPCEventStreamAgent.java)).

![386201870-9902e383-248d-42a7-882a-79171019d7fb](https://github.com/user-attachments/assets/25d133a1-946b-49f7-9ad7-1049c8570655)

In `PubSubIPCEventStreamAgent.handleUnsubscribeToTopicRequest()` (which I believe is invoked when calling `close()` on `SubscribeToTopicOperation` [in the Python SDK](https://github.com/aws/aws-iot-device-sdk-python-v2/blob/4338c4caa392cc42a126f7f0e030e3500a54f23a/awsiot/greengrasscoreipc/client.py#L1098), which we're using in our application), a call to `remove()` on `listeners` is made. When inspecting [the source](https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/2554a4202025f0553174a1e8912c42c5b24581fd/src/main/java/com/aws/greengrass/builtin/services/pubsub/SubscriptionTrie.java#L71) for `remove()`, it seems that the member `subscriptionCallbacks` is cleaned up, but not `children`. `subscriptionCallbacks` and `children` are the only non-`static final` members of `SubscriptionTrie`.
**To Reproduce**
Using the Python SDK:
```
ipc_client = GreengrassCoreIPCClientV2()

count = 0
while True:
    request_id = uuid.uuid1()
    response_topic = f"dummy_method-response-{request_id}"
    def response_listener(message: SubscriptionResponseMessage) -> None:
        log.info("Response listener")
    def error_listener(_: Exception) -> Union[None, bool]:
        log.info("Error listener")
        return True
[...]

PR #1671, opened 2024-11-20
fix: prune subscriptions trie when topics are unsubscribed

saranyailla:
**Issue #, if available:**
https://github.com/aws-greengrass/aws-greengrass-nucleus/issues/1670

**Description of changes:**
----
**Background**

`PubSubIPCEventStreamAgent` handles all the IPC local pubsub requests. It implements the handlers for `SubscribeToTopicRequest` and `PublishToTopicRequest` where the topics are stored in a `SubscriptionTrie` trie. 

When a topic is subscribed to, it is added to the `SubscriptionTrie` trie where the topic levels as represented as nodes of the trie. When you subscribe to a topic with a handler, the handler is registered as a listener of that topic where the last topic level node of that topic increments its listeners count by 1. So, for a valid topic in a trie, the number of subscriptions count at the last topic level of that topic is greater than or equal to one.

Eg. When topics like a/b, a/b/c, a/# and d/+ are registered with a callback, the trie representation is as follows where the nodes `b`, `c`, `#` and `+` maintain the number of subscription handlers held by a topic that ends at their topic level. 
```
        root
       /  \
      a    d
    /   \    \
    b(1) #(1) +(1)
  / 
  c(1)
 ```

**Problem**

When an IPC subscription for a topic is closed, it is expected that topic references are removed from the trie if there are no others listeners for the same topic. But, currently, we only remove the registered subscription handler from the list of handlers maintained at that topic level, leaving behind all the nodes as is.  That means, once a topic is subscribed to, it is never removed from the heap memory even when it is closed/unsubscribed to and no longer used. 

----
**Changes**

When a topic is unsubscribed to, we clear up the references from the `SubscriptionTrie`, which manages topics along with their listeners for matching and filtering. 

If a topic node has no children or no callbacks registered, it means that the path to that topic node is not a complete topic or a prefix of other topic(s). So, it can be removed. 

We modify the trie in the following way when a topic is unsubscribed to or a subscription is closed:

- We first identify the leaf node of the unsubscribed/closed topic. We do this by simply going one level after another till we reach the end of the topic. While doing this, we also store the last node along the topic path that has one or more callbacks registered along with the next topic level. In this process, if a node becomes null, that means the topic doesn't exist in the trie. So, we remove it. 
- Once the final node of the topic is reached, we remove the unsubscribed/closed handlers of the topic from the list of handlers maintained at that level. 
- After removing the handlers, if the final topic node cannot be removed, it means that it's still a valid topic or prefix of other topic(s) in the trie. So, do nothing but return true. 
- If the topic node can be removed, then remove all the topic nodes after the last known node with callbacks in the topic path, as they no longer form a valid topic or prefix of other topics.  
- After removing the unneeded topic nodes, if the last known node to keep can also be removed, then prune the trie along the given topic path recursively where the topic nodes are removed in the reverse order. 

https://github.com/aws-greengrass/aws-greengrass-nucleus/issues/1670

The heap memory  keeps growing by retaining the object references even when they're no longer needed. If the heap size continues to grows due to `SubscribeToTopicRequest` object, the IPC event loop thread which handles IPC requests is killed due to out of memory error leaving the JVM in a bad state.

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/builtin/services/pubsub/SubscriptionTrie.java
+++ b/src/main/java/com/aws/greengrass/builtin/services/pubsub/SubscriptionTrie.java
@@ -12,6 +12,7 @@ import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 /**
  * Trie to manage subscriptions.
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/main/java/com/aws/greengrass/builtin/services/pubsub/SubscriptionTrie.java [hunk 1]
```diff
@@ -69,11 +71,61 @@ public boolean remove(String topic, K cb) {
      * @return if changed after removal
      */
     public boolean remove(String topic, Set<K> cbs) {
-        SubscriptionTrie<K> sub = lookup(topic);
-        if (sub == null) {
+        if (Utils.isEmpty(topic) || Utils.isEmpty(cbs)) {
             return false;
         }
-        return sub.subscriptionCallbacks.removeAll(cbs);
+        SubscriptionTrie<K> current = this;
```
alter-mage, line 77: nit: I do not disagree with this approach but I think we can accomplish this with one recursive method only. If I understand this correctly, we are passing over the path generated by the topic to remove twice, once in `remove` and again in `pruneRecursively`.

saranyailla: Removing(L100) is a necessary step. But `pruneRecursively`(L103) is optional and it is needed only if 'removing' leads to a leaf with no children or no subscriptions.

alter-mage (dismissed):

yitingb (dismissed):

src/main/java/com/aws/greengrass/builtin/services/pubsub/SubscriptionTrie.java
```diff
@@ -69,11 +72,43 @@ public boolean remove(String topic, K cb) {
...
         }
-        return sub.subscriptionCallbacks.removeAll(cbs);
+
+        AtomicBoolean isSubscriptionRemoved = new AtomicBoolean(false);
+        removeRecursively(topic.split(TOPIC_LEVEL_SEPARATOR), this, cbs, 0, isSubscriptionRemoved);
+        return isSubscriptionRemoved.get();
+    }
+
+    private boolean canRemove(SubscriptionTrie<K> node) {
+        // returns false if the topic level is not prefix of another topic or has callbacks registered
```
robcmann, line 85: nit: should this be returns _true_ if the topic level is not the prefix of another topic

alter-mage: thats right, that is a typo in the comment. Will update in 2 mins.

alter-mage (approved):

robcmann (approved):

PR #1675, opened 2024-12-03
fix: link launch dir with existing nucleus package if does not exist

alter-mage:
**Issue #, if available:**

**Description of changes:**
Re create (if not present) and re link the current launch directory symbolic link with the Nucleus package which is currently being executed. This will only be executed if the current launch directory symbolic link does not exist.

If the current Nucleus package in execution does not have a loader file in its un-archived path, un-archive the the current Nucleus package artifact to re-populate the loader paths.

This change is necessary to mitigate deployment failures due to Launch Dir Corrupted.

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/componentmanager/ComponentManager.java
+++ b/src/main/java/com/aws/greengrass/componentmanager/ComponentManager.java
@@ -61,6 +61,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ExecutorService;
@@ -71,6 +72,7 @@ import javax.inject.Inject;
 
 import static com.aws.greengrass.componentmanager.KernelConfigResolver.PREV_VERSION_CONFIG_KEY;
 import static com.aws.greengrass.componentmanager.KernelConfigResolver.VERSION_CONFIG_KEY;
+import static com.aws.greengrass.deployment.DeviceConfiguration.DEFAULT_NUCLEUS_COMPONENT_NAME;
 import static com.aws.greengrass.deployment.converter.DeploymentDocumentConverter.ANY_VERSION;
 import static org.apache.commons.io.FileUtils.ONE_MB;
 
@@ -294,6 +296,37 @@ public class ComponentManager implements InjectionActions {
             // This should never happen as SHA-256 is mandatory for every default JVM provider
             throw new HashingAlgorithmUnavailableException("No security provider found for message digest", e);
         }
+    }
+
+    /**
+     * Un-archives the artifacts for the current Nucleus version package.
+     *
+     * @return list of un-archived paths
+     * @throws PackageLoadingException when unable to load current Nucleus
+     */
+    public List<Path> unArchiveCurrentNucleusVersionArtifacts() throws PackageLoadingException {
+        String currentNucleusVersion = deviceConfiguration.getNucleusVersion();
+        ComponentIdentifier nucleusComponentIdentifier =
+                new ComponentIdentifier(DEFAULT_NUCLEUS_COMPONENT_NAME, new Semver(currentNucleusVersion));
+        List<File> nucleusArtifactFileNames =
+                componentStore.getArtifactFiles(nucleusComponentIdentifier, artifactDownloaderFactory);
+        return nucleusArtifactFileNames.stream()
+                .map(file -> {
+                    try {
+                        Path unarchivePath =
+                                nucleusPaths.unarchiveArtifactPath(nucleusComponentIdentifier, getFileName(file));
+                        /*
+                        Using a hard-coded ZIP un-archiver as today this code path is only used to un-archive a Nucleus
+                        .zip artifact.
+                         */
+                        unarchiver.unarchive(Unarchive.ZIP, file, unarchivePath);
+                        return unarchivePath;
+                    } catch (IOException e) {
+                        logger.atDebug().setCause(e).kv("comp-id", nucleusComponentIdentifier)
+                                .log("Could not un-archive Nucleus artifact");
+                        return null;
+                    }
+                }).filter(Objects::nonNull).collect(Collectors.toList());
     }
 
     private Optional<ComponentIdentifier> findBestCandidateLocally(String componentName,
--- a/src/main/java/com/aws/greengrass/lifecyclemanager/KernelAlternatives.java
+++ b/src/main/java/com/aws/greengrass/lifecyclemanager/KernelAlternatives.java
@@ -5,6 +5,8 @@
 
 package com.aws.greengrass.lifecyclemanager;
 
+import com.aws.greengrass.componentmanager.ComponentManager;
+import com.aws.greengrass.componentmanager.exceptions.PackageLoadingException;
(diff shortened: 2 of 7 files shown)


src/main/java/com/aws/greengrass/lifecyclemanager/KernelAlternatives.java
```diff
@@ -169,10 +169,22 @@ public boolean isLaunchDirSetup() {
      * @throws DeploymentException when user is not allowed to change file permission
      */
     public void validateLaunchDirSetupVerbose() throws DirectoryValidationException, DeploymentException {
-        Path currentDir = getCurrentDir();
-        if (!Files.isSymbolicLink(currentDir)) {
-            throw new DirectoryValidationException("Missing symlink to current nucleus launch directory");
+        if (!Files.isSymbolicLink(getCurrentDir())) {
```
junfuchen99, line 172: what if current exists but loader path does not exist?

alter-mage: I am finding it hard to imagine a scenario where current points to a valid Nucleus package but loader does not exist in that package

src/main/java/com/aws/greengrass/lifecyclemanager/KernelAlternatives.java
```diff
@@ -162,31 +169,69 @@ public boolean isLaunchDirSetup() {
...
+        if (Files.exists(currentNucleusExecutablePath.resolve(KERNEL_BIN_DIR)
+                .resolve(Platform.getInstance().loaderFilename()))) {
+            logger.atDebug().kv("path", currentNucleusExecutablePath)
+                    .log("Current Nucleus executable is valid, setting up launch dir");
+            relinkInitLaunchDir(currentNucleusExecutablePath, true);
+            return true;
+        }
+
+        List<Path> localNucleusExecutablePaths = componentManager.findValidLocalNucleusPackage();
+        if (!localNucleusExecutablePaths.isEmpty()) {
```
junfuchen99, line 190: aren't you calling isEmpty on null object?

alter-mage: Updated. This will still be an empty check in case artifacts dir is empty (initial installation) but this code path should only hit if a customer purposefully deletes the loader script from the unarchived dir. Imo, this is low risk.

src/main/java/com/aws/greengrass/lifecyclemanager/KernelAlternatives.java
```diff
@@ -162,31 +169,69 @@ public boolean isLaunchDirSetup() {
...
+            Optional<Path> validNucleusExecutablePath = localNucleusExecutablePaths.stream()
```
saranyailla, line 191: How many nucleus artifacts exist at any given point including the current one that's looking up these paths? 
Should we be choosing the latest one? 
What if the one chosen bricks the device?

alter-mage: There can be only 2 nucleus versions at any point. 
It is not likely that the locally present nucleus versions can brick a device because these versions were running correctly on the device at some point and they were updated due to an OTA update.

src/main/java/com/aws/greengrass/componentmanager/ComponentManager.java
```diff
@@ -296,6 +298,33 @@ private void storeRecipeDigestInConfigStoreForPlugin(
...
+        ComponentIdentifier nucleusComponentIdentifier =
+                new ComponentIdentifier(DEFAULT_NUCLEUS_COMPONENT_NAME, new Semver(currentNucleusVersion));
+        List<File> nucleusArtifactFileNames =
+                componentStore.getArtifactFiles(nucleusComponentIdentifier, artifactDownloaderFactory);
+        return nucleusArtifactFileNames.stream()
+                .map(file -> {
+                    try {
+                        Path unarchivePath =
+                                nucleusPaths.unarchiveArtifactPath(nucleusComponentIdentifier, getFileName(file));
+                        unarchiver.unarchive(Unarchive.ZIP, file, unarchivePath);
```
saranyailla, line 318: Should we check if the archive file exists on the disk before unarchiving?

src/main/java/com/aws/greengrass/componentmanager/ComponentManager.java
```diff
@@ -296,6 +298,33 @@ private void storeRecipeDigestInConfigStoreForPlugin(
...
+        String currentNucleusVersion = deviceConfiguration.getNucleusVersion();
+        ComponentIdentifier nucleusComponentIdentifier =
+                new ComponentIdentifier(DEFAULT_NUCLEUS_COMPONENT_NAME, new Semver(currentNucleusVersion));
+        List<File> nucleusArtifactFileNames =
+                componentStore.getArtifactFiles(nucleusComponentIdentifier, artifactDownloaderFactory);
+        return nucleusArtifactFileNames.stream()
+                .map(file -> {
+                    try {
+                        Path unarchivePath =
+                                nucleusPaths.unarchiveArtifactPath(nucleusComponentIdentifier, getFileName(file));
```
saranyailla, line 317: I understand that we have zip artifacts only in the nucleus component versions, but this line would always copy over the artifacts to unarchived-artifacts folder irrespective of the Unarchive type of the artifact in the recipe.

src/main/java/com/aws/greengrass/componentmanager/ComponentManager.java
alter-mage: That is not necessary, if the file does not exist, unarchiver will throw an IOException which ComponentManager will catch and just return null. The stream will filter out nonNull objects later on. If no artifact file existed then the list returned by ComponentManager will be empty and KernelAlternatives will handle that appropriately.

src/main/java/com/aws/greengrass/componentmanager/ComponentManager.java
alter-mage: Yep, since this code path is used only for Nucleus artifacts in the near future, I do not see a harm in hard coding the unarchive as ZIP. But I can modify to fetch unarchive from the recipe if you have a strong opinion on it.

src/main/java/com/aws/greengrass/lifecyclemanager/KernelAlternatives.java
```diff
@@ -162,31 +169,69 @@ public boolean isLaunchDirSetup() {
...
+        Path currentNucleusExecutablePath = locateCurrentKernelUnpackDir();
+        if (Files.exists(currentNucleusExecutablePath.resolve(KERNEL_BIN_DIR)
+                .resolve(Platform.getInstance().loaderFilename()))) {
+            logger.atDebug().kv("path", currentNucleusExecutablePath)
+                    .log("Current Nucleus executable is valid, setting up launch dir");
+            relinkInitLaunchDir(currentNucleusExecutablePath, true);
+            return true;
+        }
+
+        List<Path> localNucleusExecutablePaths = componentManager.unArchiveCurrentNucleusVersionArtifacts();
```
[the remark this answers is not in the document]
alter-mage: Yep, that's right

saranyailla (approved):

yitingb (approved):

PR #1677, opened 2024-12-06
feat: greengrass.service.template: add After=systemd-time-wait-sync.service

thomas-roos:
This will allow if systemd-time-wait-sync.service is enabled to wait for a synced clock. If this is not enabled it will not wait. Especially in systems without a RTC (like RPi) it will take some time after boot till NTP will sync clock and if Greengrass is already started this time jump will cause issues.

**Issue #, if available:**

**Description of changes:**

**Documentation Checklist:**

**Compatibility Checklist:**


aws-kevinrickard (approved):

scripts/greengrass.service.template
```diff
@@ -1,6 +1,7 @@
 [Unit]
 Description=Greengrass Core
 After=network.target
+After=systemd-time-wait-sync.service
```
kstribrnAmzn, line 4: Should this be a weaker dependency type such as `Wants` for example? [Doc link](https://www.freedesktop.org/software/systemd/man/latest/systemd.unit.html#Wants=)

thomas-roos: Right, this is NOT adding a dependency. Just a order if available - see here:  https://serverfault.com/questions/812584/in-systemd-whats-the-difference-between-after-and-requires and as I understand the shared doc section you shared.

thomas-roos: To use different words - this will cause GG to start after the clock is synced if such a service is present, if not it don't care. If sync fails it will delayed max 90s and then start (hopefully still successfully without a accurate clock)

kstribrnAmzn (approved):

PR #1701, opened 2025-03-28
fix: prevent double injection of service class

saranyailla:
**Issue #, if available:**

**Description of changes:**
When Nucleus is started, all the plugin/builtin Greengrass services are injected twice. These changes help 

**Current flow:**

When Greengrass is restarted, it starts all the auto-startable services in order after locating them from the context. When a service doesn't already exist in the Context, it is created and injected in to the context in the following way:

At a high-level, a new instance of the Service is created as the following:
  1. If it exists in the config file under service topic dependencies, it is created after its dependencies are created. 
  2. For an external plugin, it is loaded from the plugins path.
  3. For a builtin/plugin service, if a constructor with `Topics` exists, an instance of it is created from the constructor(Not injected) and returned. Otherwise, an instance of the service is created from `@Inject` constructor if it exists (or a default constructor otherwise) and injection hooks (preInject, Inject, PostInject) are executed where the object is injected into the context along with all the injectable args of the class. 
  4. If it is not a plugin or a builtin service, a new instance of `GenericExternalService` is created. 

After the new instance of a service is created, all context injection hooks  (preInject, Inject and postInject) are executed where the object is injected into the context along with all the injectable args of the class. 

Thus the builtin/plugin service injection hooks are executed twice resulting in other side effects such as 
- When a service subscribes to config changes in post inject, two listeners are created and fired for the same events. 
- Any config writes in those hooks is written twice to the config.tlog file. This is applied to logging as well. 

**Fixed flow**
When a new instance of the builtin/plugin service is requested, we only return the constructed instance without injecting itself or its dependencies/args as they will be injected later.  

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/dependency/Context.java
+++ b/src/main/java/com/aws/greengrass/dependency/Context.java
@@ -580,8 +580,6 @@ public class Context implements Closeable {
                 if (object != null) {
                     return object;
                 }
-                // Mapping function should make sure that the object is not injected already. Otherwise, it results in
-                // double injection of the same object.
                 return putAndInjectFields(mappingFunction.apply(this));
             }
         }
--- a/src/main/java/com/aws/greengrass/dependency/Context.java
+++ b/src/main/java/com/aws/greengrass/dependency/Context.java
@@ -121,7 +121,7 @@ public class Context implements Closeable {
     }
 
     public <T> T newInstance(Class<T> cl) {
-        return new Value<>(cl, null).get();
+        return new Value<>(cl, null).getObjectWithoutInjection();
     }
 
     /**
@@ -434,6 +434,17 @@ public class Context implements Closeable {
             return constructObjectWithInjection();
         }
 
+        /**
+         * Constructs an object without injection.
+         * @return object
+         */
+        public final T getObjectWithoutInjection() {
+            if (object != null && injectionCompleted) {
+                return object;
+            }
+            return constructObject();
+        }
+
 
         /**
          * Put a new object instance and inject fields with pre and post actions, if the new object is not equal to
@@ -462,7 +473,7 @@ public class Context implements Closeable {
         }
 
         @SuppressWarnings("PMD.AvoidCatchingThrowable")
-        private T constructObjectWithInjection() {
+        private T constructObject() {
             try (LockScope ls = LockScope.lock(lock)) {
                 if (object != null) {
                     return object;
@@ -483,14 +494,20 @@ public class Context implements Closeable {
                     int paramCount = pickedConstructor.getParameterCount();
                     if (paramCount == 0) {
                         // no arg constructor
-                        return putAndInjectFields(pickedConstructor.newInstance());
+                        return pickedConstructor.newInstance();
                     }
 
                     Object[] args = getOrCreateArgInstances(clazz, pickedConstructor, paramCount);
-                    return putAndInjectFields(pickedConstructor.newInstance(args));
+                    return pickedConstructor.newInstance(args);
                 } catch (Throwable ex) {
                     throw new IllegalArgumentException("Can't create instance of " + targetClass.getName(), ex);
                 }
+            }
+        }
+
+        private T constructObjectWithInjection() {
+            try (LockScope ls = LockScope.lock(lock)) {
+                return putAndInjectFields(constructObject());
             }
         }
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/aws/greengrass/dependency/Context.java
```diff
@@ -483,17 +494,21 @@ private T constructObjectWithInjection() {
...
-                    return putAndInjectFields(pickedConstructor.newInstance(args));
+                    return pickedConstructor.newInstance(args);
                 } catch (Throwable ex) {
                     throw new IllegalArgumentException("Can't create instance of " + targetClass.getName(), ex);
                 }
             }
         }
 
+        private T constructObjectWithInjection() {
+            return putAndInjectFields(constructObject());
```
aws-kevinrickard, line 509: This method should be guarded by the lock as well so that two threads calling `Value::get` can't both construct their own objects (one after the other) and then inject them both (one after the other).

Locking would work here because constructObject has the check and quick return on line 479, as well as putAndInjectFields on line 459

src/main/java/com/aws/greengrass/dependency/Context.java
```diff
@@ -434,6 +434,17 @@ public final T get() {
             return constructObjectWithInjection();
         }
 
+        /**
+         * Constructs an object without injection.
+         * @return object
+         */
+        public final T getObjectWithoutInjection() {
+            if (object != null && injectionCompleted) {
+                return object;
```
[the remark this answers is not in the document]
saranyailla: For a new instance, object will always be null, so, this `if` condition evaluates to false for that. But for other usages of this method, it makes sense to return an existing object.

src/main/java/com/aws/greengrass/dependency/Context.java
saranyailla: Updated

src/main/java/com/aws/greengrass/dependency/Context.java
```diff
@@ -563,7 +580,8 @@ public final <E extends Exception> T computeObjectIfEmpty(
                 if (object != null) {
                     return object;
                 }
-
+                // Mapping function should make sure that the object is not injected already. Otherwise, it results in
```
aws-kevinrickard, line 583: If object is null then that means it is not injected already right?

saranyailla: Yes. This comment is for the mapping function in line 585 as we're injecting the object in the same line.

aws-kevinrickard: We're holding the lock so the object will not change, and we know the object is null at this point (and thus not injected already), so the mappingFunction can safely assume the object is not injected already. This comment says the opposite

saranyailla: The `object` here is not same the as the object created and injected already by the mapping function. Mapping function creates the object and injects it. Now we made changes to not.

alter-mage (dismissed):

src/main/java/com/aws/greengrass/dependency/Context.java
aws-kevinrickard: I see that now. Sorry for the back and forth on a comment

Not a big deal, approving now; but for the javadoc change discussed in standup, I meant it should be noted the javadoc spec of the mappingFunction provided to computeObjectIfEmpty, as that is part of the API that was being misused (when it was being passed a function that called newInstance that did also inject)

aws-kevinrickard (approved):

src/main/java/com/aws/greengrass/lifecyclemanager/Kernel.java
```diff
@@ -481,7 +481,7 @@ private void executeBootstrapTasksAndShutdown(BootstrapManager bootstrapManager,
 
     @SuppressWarnings(
             {"UseSpecificCatch", "PMD.AvoidCatchingThrowable", "PMD.AvoidDeeplyNestedIfStmts", "PMD.ConfusingTernary"})
-    private GreengrassService locateGreengrassService(Context.Value v, String name, CrashableFunction<String,
+    private GreengrassService createGreengrassServiceInstance(Context.Value v, String name, CrashableFunction<String,
```
MikeDombo, line 484: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

The cyclomatic complexity of this method is 24. By comparison, 99% of the methods in the CodeGuru reference dataset have a lower cyclomatic complexity. This indicates the method has a high number of decisions and it can make the logic difficult to understand and test.

We recommend that you simplify this method or break it into multiple methods. For example, consider extracting the code block on lines 537-592 into a separate method.

alter-mage (approved):

aws-kevinrickard (approved):

PR #1706, opened 2025-04-08
feat: add deploymentConfigurationTimeSource configuration

aws-kevinrickard:
**Issue #, if available:**
P200769160

**Description of changes:**

Introduce a new configuration for the nucleus component: deploymentConfigurationTimeSource with two possible values: 

* deploymentCreationTime - (Default, current behavior) - When a deployment is processed on a device, the configuration update uses the deployment creation timestamp to resolve conflicts on config keys existing on the device. This ensures a consistent config state when multiple deployments are applied to the same device at the same time. A config value from a deployment will not be overwritten by other deployments with older creation times while this behavior is selected, even if the deployment is processed on the device at a later (more recent) time.
* deploymentProcessingTime - When a deployment is processed on a device, the configuration update uses the current device local timestamp to resolve conflicts on config keys existing on the device. In effect, config will always be updated and overwrite during any conflicts while this behavior is selected because the deployment processing timestamp will be greater than any deployment creation timestamp, assuming the device clock is calibrated. This means that the config state can be non-deterministic when multiple deployments that set the same config key to different values are active and queued to be deployed to the device at the same time, because the deployments could happen in any order- these situations should be avoided (and we strongly recommend avoiding them) when using this option.

Have an option for more-recently-processed deployments to always update config (regardless of configs set by previously processed deployments which may have had priority by deployment creation timestamp).

  - Attempted but aborted as the existing unit tests for these modules don't actually check these code paths
  - Attempted but aborted in favor of end-to-end tests
- [X] Updated or added new end-to-end tests.
  - See CR-189181118

**Documentation Checklist:**
   - n/a

**Compatibility Checklist:**
  - n/a
  - n/a

### DIFF
--- a/src/main/java/com/aws/greengrass/componentmanager/KernelConfigResolver.java
+++ b/src/main/java/com/aws/greengrass/componentmanager/KernelConfigResolver.java
@@ -143,16 +143,18 @@ public class KernelConfigResolver {
      * @param componentsToDeploy package identifiers for resolved packages of complete dependency graph across groups
      * @param document           deployment document
      * @param rootPackages       root level packages
+     * @param configMergeTimestamp timestamp to use for configuration merge
      * @return a kernel config map
      * @throws PackageLoadingException if any service package was unable to be loaded
      * @throws IOException             for directory issues
      */
     public Map<String, Object> resolve(List<ComponentIdentifier> componentsToDeploy, DeploymentDocument document,
-            List<String> rootPackages) throws PackageLoadingException, IOException {
+            List<String> rootPackages, long configMergeTimestamp) throws PackageLoadingException, IOException {
         Map<String, Object> servicesConfig = new HashMap<>();
         // resolve configuration
         for (ComponentIdentifier componentToDeploy : componentsToDeploy) {
-            servicesConfig.put(componentToDeploy.getName(), getServiceConfig(componentToDeploy, document));
+            servicesConfig.put(componentToDeploy.getName(), getServiceConfig(componentToDeploy, document,
+                    configMergeTimestamp));
         }
 
         // Interpolate configurations
@@ -217,10 +219,12 @@ public class KernelConfigResolver {
      *
      * @param componentIdentifier         target component id
      * @param document                    deployment doc for the current deployment
+     * @param configMergeTimestamp        timestamp to use during configuration merge
      * @return a built map representing the kernel config under "services" key for a particular component
      * @throws PackageLoadingException if any service package was unable to be loaded
      */
-    private Map<String, Object> getServiceConfig(ComponentIdentifier componentIdentifier, DeploymentDocument document)
+    private Map<String, Object> getServiceConfig(ComponentIdentifier componentIdentifier, DeploymentDocument document,
+                                                 long configMergeTimestamp)
             throws PackageLoadingException {
 
         ComponentRecipe componentRecipe = componentStore.getPackageRecipe(componentIdentifier);
@@ -262,7 +266,7 @@ public class KernelConfigResolver {
         }
 
         Map<String, Object> resolvedConfiguration = resolveConfigurationToApply(optionalConfigUpdate.orElse(null),
-                componentRecipe, document);
+                componentRecipe, configMergeTimestamp);
 
         // merge resolved param and resolved configuration for backward compatibility
         resolvedServiceConfig
@@ -333,13 +337,13 @@ public class KernelConfigResolver {
      *
      * @param configurationUpdateOperation nullable component configuration update operation.
      * @param componentRecipe              component recipe containing default configuration.
-     * @param document                     deployment document
+     * @param configMergeTimestamp         timestamp to use during configuration merge
      * @return resolved configuration for this component. non null.
      */
     @SuppressWarnings("PMD.ConfusingTernary")
     private Map<String, Object> resolveConfigurationToApply(
             @Nullable ConfigurationUpdateOperation configurationUpdateOperation, ComponentRecipe componentRecipe,
-            DeploymentDocument document) {
+            long configMergeTimestamp) {
 
         // try read the running service config
         try (Context context = new Context()) {
@@ -379,7 +383,7 @@ public class KernelConfigResolver {
 
             // Merge in the requested config updates
             if (configurationUpdateOperation != null && configurationUpdateOperation.getValueToMerge() != null) {
-                currentRunningConfig.mergeMap(document.getTimestamp(), configurationUpdateOperation.getValueToMerge());
+                currentRunningConfig.mergeMap(configMergeTimestamp, configurationUpdateOperation.getValueToMerge());
             }
 
             return currentRunningConfig.toPOJO();
--- a/src/main/java/com/aws/greengrass/componentmanager/KernelConfigResolver.java
+++ b/src/main/java/com/aws/greengrass/componentmanager/KernelConfigResolver.java
@@ -151,6 +151,8 @@ public class KernelConfigResolver {
     public Map<String, Object> resolve(List<ComponentIdentifier> componentsToDeploy, DeploymentDocument document,
             List<String> rootPackages, long configMergeTimestamp) throws PackageLoadingException, IOException {
         Map<String, Object> servicesConfig = new HashMap<>();
+        LOGGER.atDebug().kv("Components to deploy", componentsToDeploy).kv("Root packages", rootPackages)
+                .log("Resolving services configuration");
         // resolve configuration
         for (ComponentIdentifier componentToDeploy : componentsToDeploy) {
             servicesConfig.put(componentToDeploy.getName(), getServiceConfig(componentToDeploy, document,
@@ -187,6 +189,7 @@ public class KernelConfigResolver {
         servicesConfig.putIfAbsent(nucleusComponentName, getNucleusComponentConfig(nucleusComponentName));
         servicesConfig.put(kernel.getMain().getName(), getMainConfig(rootPackages, nucleusComponentName));
 
+        LOGGER.atDebug().kv("Services Configuration", servicesConfig).log("Resolved services configuration.");
         // Services need to be under the services namespace in kernel config
         return Collections.singletonMap(SERVICES_NAMESPACE_TOPIC, servicesConfig);
     }
--- a/src/main/java/com/aws/greengrass/deployment/DefaultDeploymentTask.java
+++ b/src/main/java/com/aws/greengrass/deployment/DefaultDeploymentTask.java
@@ -17,6 +17,7 @@ import com.aws.greengrass.config.Topics;
 import com.aws.greengrass.deployment.exceptions.DeploymentTaskFailureException;
 import com.aws.greengrass.deployment.model.Deployment;
 import com.aws.greengrass.deployment.model.DeploymentDocument;
+import com.aws.greengrass.deployment.model.DeploymentPackageConfiguration;
 import com.aws.greengrass.deployment.model.DeploymentResult;
 import com.aws.greengrass.deployment.model.DeploymentTask;
 import com.aws.greengrass.logging.api.Logger;
@@ -42,6 +43,8 @@ import java.util.stream.Collectors;
 
 import static com.aws.greengrass.deployment.DeploymentConfigMerger.DEPLOYMENT_ID_LOG_KEY;
 import static com.aws.greengrass.deployment.DeploymentService.GROUP_TO_ROOT_COMPONENTS_VERSION_KEY;
+import static com.aws.greengrass.deployment.DeviceConfiguration.DEPLOYMENT_CONFIGURATION_TIME_SOURCE_DEPLOYMENT_PROCESSING_TIME;
+import static com.aws.greengrass.deployment.DeviceConfiguration.DEVICE_PARAM_DEPLOYMENT_CONFIGURATION_TIME_SOURCE;
 import static com.aws.greengrass.deployment.converter.DeploymentDocumentConverter.LOCAL_DEPLOYMENT_GROUP_NAME;
 
 /**
@@ -59,6 +62,7 @@ public class DefaultDeploymentTask implements DeploymentTask {
     private final DeploymentConfigMerger deploymentConfigMerger;
     private final ExecutorService executorService;
     private final Logger logger;
+    private final DeviceConfiguration deviceConfiguration;
     @Getter
     private final Deployment deployment;
     private final Topics deploymentServiceConfig;
@@ -78,7 +82,8 @@ public class DefaultDeploymentTask implements DeploymentTask {
      * @param deploymentServiceConfig      Deployment service configuration Topics
      * @param executorService              Executor service
      * @param deploymentDocumentDownloader download large deployment document.
-     * @param thingGroupHelper             Executor service
+     * @param thingGroupHelper             Thing Group Helper / Retriever
+     * @param deviceConfiguration          Device Configuration Information
      */
     @SuppressWarnings("PMD.ExcessiveParameterList")
     public DefaultDeploymentTask(DependencyResolver dependencyResolver, ComponentManager componentManager,
@@ -86,7 +91,8 @@ public class DefaultDeploymentTask implements DeploymentTask {
                                  DeploymentConfigMerger deploymentConfigMerger, Logger logger, Deployment deployment,
                                  Topics deploymentServiceConfig, ExecutorService executorService,
                                  DeploymentDocumentDownloader deploymentDocumentDownloader,
-                                 ThingGroupHelper thingGroupHelper) {
+                                 ThingGroupHelper thingGroupHelper,
+                                 DeviceConfiguration deviceConfiguration) {
         this.dependencyResolver = dependencyResolver;
         this.componentManager = componentManager;
         this.kernelConfigResolver = kernelConfigResolver;
@@ -97,6 +103,7 @@ public class DefaultDeploymentTask implements DeploymentTask {
         this.executorService = executorService;
         this.deploymentDocumentDownloader = deploymentDocumentDownloader;
         this.thingGroupHelper = thingGroupHelper;
+        this.deviceConfiguration = deviceConfiguration;
     }
 
     @Override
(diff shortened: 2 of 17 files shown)


src/main/java/com/aws/greengrass/componentmanager/KernelConfigResolver.java
```diff
@@ -379,7 +380,17 @@ private Map<String, Object> resolveConfigurationToApply(
 
             // Merge in the requested config updates
             if (configurationUpdateOperation != null && configurationUpdateOperation.getValueToMerge() != null) {
-                currentRunningConfig.mergeMap(document.getTimestamp(), configurationUpdateOperation.getValueToMerge());
+                // default to using the deployment creation time as the timestamp
+                long timestamp = document.getTimestamp();
+                String deploymentConfigurationTimeSource =
+                        Coerce.toString(deviceConfiguration.getDeploymentConfigurationTimeSource());
```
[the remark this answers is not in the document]
aws-kevinrickard: From prior deployment

src/main/java/com/aws/greengrass/deployment/activator/KernelUpdateActivator.java
```diff
@@ -101,7 +106,20 @@ public void activate(Map<String, Object> newConfig, Deployment deployment,
...
         lifecycle.softShutdown(30);
 
-        updateConfiguration(deploymentDocument.getTimestamp(), newConfig);
+        // default to using the deployment creation time as the timestamp
+        long timestamp = deploymentDocument.getTimestamp();
+        Object deploymentConfigurationTimeSourceObject = kernel.getConfig().lookup(
+                SERVICES_NAMESPACE_TOPIC,
+                DEFAULT_NUCLEUS_COMPONENT_NAME,
+                CONFIGURATION_CONFIG_KEY,
+                DEVICE_PARAM_DEPLOYMENT_CONFIGURATION_TIME_SOURCE).getOnce();
```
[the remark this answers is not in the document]
aws-kevinrickard: Thanks, will update

src/main/java/com/aws/greengrass/deployment/activator/DefaultActivator.java
```diff
@@ -59,7 +64,20 @@ public void activate(Map<String, Object> newConfig, Deployment deployment,
...
+                DeviceConfiguration.DEFAULT_NUCLEUS_COMPONENT_NAME,
+                CONFIGURATION_CONFIG_KEY,
+                DEVICE_PARAM_DEPLOYMENT_CONFIGURATION_TIME_SOURCE).getOnce();
+        if (deploymentConfigurationTimeSourceObject != null
+                && Coerce.toString(deploymentConfigurationTimeSourceObject)
+                .equals(DEPLOYMENT_CONFIGURATION_TIME_SOURCE_DEPLOYMENT_PROCESSING_TIME)) {
+            // Optional override to use the deployment processing time as the timestamp
+            timestamp = System.currentTimeMillis();
+        }
+        updateConfiguration(timestamp, newConfig);
```
saranyailla, line 80: Can we deduplicate this timestamp resolution?

src/main/java/com/aws/greengrass/deployment/activator/DefaultActivator.java
```diff
@@ -59,7 +64,20 @@ public void activate(Map<String, Object> newConfig, Deployment deployment,
...
+        // default to using the deployment creation time as the timestamp
+        long timestamp = deploymentDocument.getTimestamp();
+        Object deploymentConfigurationTimeSourceObject = kernel.getConfig().lookup(
+                SERVICES_NAMESPACE_TOPIC,
+                DeviceConfiguration.DEFAULT_NUCLEUS_COMPONENT_NAME,
+                CONFIGURATION_CONFIG_KEY,
+                DEVICE_PARAM_DEPLOYMENT_CONFIGURATION_TIME_SOURCE).getOnce();
+        if (deploymentConfigurationTimeSourceObject != null
+                && Coerce.toString(deploymentConfigurationTimeSourceObject)
+                .equals(DEPLOYMENT_CONFIGURATION_TIME_SOURCE_DEPLOYMENT_PROCESSING_TIME)) {
```
saranyailla, line 76: Nit: Flip this comparison.

src/main/java/com/aws/greengrass/deployment/activator/KernelUpdateActivator.java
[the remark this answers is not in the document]
aws-kevinrickard: Updated

src/main/java/com/aws/greengrass/deployment/activator/DefaultActivator.java
aws-kevinrickard: Addressed via https://github.com/aws-greengrass/aws-greengrass-nucleus/pull/1706#discussion_r2052990079

src/main/java/com/aws/greengrass/deployment/activator/DefaultActivator.java
aws-kevinrickard: We can, but it's only a couple lines in one other place- in this case I don't think it's worth the abstraction. Do you think it should be abstracted?

saranyailla: If we ever make changes to the timestamp resolution in one of the activators, how can we ensure that it is updated or at least verified in the other place as well?  Does it make sense to add `getConfigResolutionTimestamp` method to `DeploymentActivator` and keep it like this
```
updateConfiguration(getConfigResolutionTimestamp(), newConfig);
```

aws-kevinrickard: Yes, de-duplication helps ensure that they are updated in sync- but adding an abstraction in the form of a helper function comes will the usual downsides as well (Makes reading, following, understanding, debugging, refactoring the code more difficult). 

Ideally none of these modules should be computing which timestamp should be used in the first place. The timestamp should be computed in one place in the deployment processing flow and passed into each of these places that need it. That deduplicates it without adding in an abstraction. But that is a more involved refactor and I didn't want to take that on or increase the scope of this change yet- I'm still new to the deployment processing parts of the code. Do you think we should do that now? Maybe it will be easier than I expected and resolve this concern in the best way.

Or do you still think we should in a helper function? It is not a big deal and I can agree to do it. It's just not what I would choose in this case.

saranyailla (dismissed):

src/main/java/com/aws/greengrass/componentmanager/KernelConfigResolver.java
```diff
@@ -379,7 +380,14 @@ private Map<String, Object> resolveConfigurationToApply(
 
             // Merge in the requested config updates
             if (configurationUpdateOperation != null && configurationUpdateOperation.getValueToMerge() != null) {
-                currentRunningConfig.mergeMap(document.getTimestamp(), configurationUpdateOperation.getValueToMerge());
+                // default to using the deployment creation time as the timestamp
+                long timestamp = document.getTimestamp();
+                if (DEPLOYMENT_CONFIGURATION_TIME_SOURCE_DEPLOYMENT_PROCESSING_TIME.equals(
```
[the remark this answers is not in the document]
aws-kevinrickard: Will it work depends on the expected behavior, so I came up with the following examples:

### An example of a customer switching to using deploymentProcessingTime:

A user is considering changing the behavior of their fleet to deploymentProcessingTime. They create a deployment targeting a subset of their fleet that includes the behavior change and some arbitrary other config at timestamp 1:

```
deploymentConfigurationUpdateTimeSource = deploymentProcessingTime
myconfig = B
```

Then they continue to roll out changes to the rest of their fleet over time, and make another change to that arbitrary config, issuing this deployment at timestamp 2:
```
myconfig = A
```
Consider a device that was not part of that test fleet using deploymentProcessingTime, which has received the deployment from timestamp 2; that device’s configuration will look like:
```
myconfig = A (TS 2)
```
Then, the user wants to change the behavior of their whole fleet to use the new behavior of deploymentProcessingTime and applies that deployment from timestamp 1 to their whole fleet. The device receives the deployment from timestamp 1 at timestamp 3. Under the expected new behavior of deploymentProcessingTime , they want myconfig = B to be set effectively regardless of timestamps. So they want the resulting device configuration to be:
```
deploymentConfigurationUpdateTimeSource = deploymentProcessingTime (TS 3)
myconfig = B (TS 3)
```
In order to achieve this, the device needs to be considering and using the changed deploymentConfigurationUpdateTimeSource behavior during the during the deployment which sets it.

### An example of a customer switching to using deploymentCreationTime:

Say the user from the previous example has started using deploymentProcessingTime behavior for a bit, but has now decided to switch back to the old/default deploymentCreationTime behavior in their fleet. Their devices are still all running with this config:
```
deploymentConfigurationUpdateTimeSource = deploymentProcessingTime (TS 3)
myconfig = B (TS 3)
```
And they create a new deployment to test a subset of their fleet with the old behavior at timestamp 4, which also happens to change some arbitrary config:
```
deploymentConfigurationUpdateTimeSource = deploymentCreationTime
myconfig = C
```
But let's say the device we’ve been following is not part of this test fleet, and does not receive that deployment yet.

Then the main part of their fleet has another arbitrary deployment created at timestamp 5:
```
myconfig = D
```
And at timestamp 6 the device processes it and would have:
```
deploymentConfigurationUpdateTimeSource = deploymentProcessingTime (TS 3)
myconfig = D (TS 6)
```
Then, the user changes the behavior of their whole fleet by applying the deployment from their test fleet created at timestamp 4 to all the devices in the fleet. At timestamp 7 the device receives it, and under the expected behavior of deploymentCreationTime existing configs with newer timestamps should be preserved, so they would want the resulting config to be:
```
deploymentConfigurationUpdateTimeSource = deploymentCreationTime (TS 4)
myconfig = D (TS 6)
```
Again, in order to achieve this, the device needs to be considering and using the changed deploymentConfigurationUpdateTimeSource behavior during the during the deployment which sets it.

----

However there are more edge cases at play. Take a more complicated example:

A device has the following config (where TS = timestamp):
```
deploymentConfigurationUpdateTimeSource = deploymentProcessingTime (TS 2)
myconfig = B (TS 2)
```
Then the device receives a deployment created at timestamp 1, and processes it at timestamp 3: merge:

```
deploymentConfigurationUpdateTimeSource = deploymentCreationTime
myconfig = A
```

Here we can say that the behavior is undefined. I feel confident in saying that if you’re doing something like this, there is no reasonable answer (unlike the previous scenarios). It will be implemented however is most convenient. If there exists an argument for what the resulting config should be in this situation, or any other situation not covered by these examples, please let me know.

----

The following implementation should satisfy all of the above requirements:

When a deployment is processed, temporarily use the deploymentConfigurationUpdateTimeSource behavior specified in that deployment to apply that deployment (including updates to deploymentConfigurationUpdateTimeSource itself).

So I'll put this change in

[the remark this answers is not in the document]
aws-kevinrickard: During testing of the above use case, we uncovered an unexpected nucleus behavior. That is, during deployment processing, for all components which are determined to be part of the services list and thus have services configuration and the nucleus component itself, all configurations for these components will have their timestamps set to the deployment processing time. This means that the use case we came up with is not possible because previously set (and older) configs will have increased timestamp on the device and thus priority is not consistent. Config merge conflict resolution is undefined when multiple deployments specify overlapping configurations and do not specify all configurations in all deployments.

So the above use case is not valid, and we are back to not having a reason for the behavior to be one way vs the other.

The implementation of this feature will still be updated to have a behavior to use the deploymentConfigurationUpdateTimeSource behavior specified in that deployment to apply that deployment (including updates to deploymentConfigurationUpdateTimeSource itself). As was to satisfy the use case we came up with. This behavior is not required anymore, but will be kept for convenience of implementation, as some of the code has been refactored for the better.

src/main/java/com/aws/greengrass/deployment/activator/KernelUpdateActivator.java
```diff
@@ -66,7 +68,7 @@ public KernelUpdateActivator(Kernel kernel, BootstrapManager bootstrapManager) {
     }
 
     @Override
-    public void activate(Map<String, Object> newConfig, Deployment deployment,
+    public void activate(Map<String, Object> newConfig, Deployment deployment, long configMergeTimestamp,
```
MikeDombo, line 69: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

The cyclomatic complexity of this method is 12. By comparison, 98% of the methods in the CodeGuru reference dataset have a lower cyclomatic complexity. This indicates the method has a high number of decisions and it can make the logic difficult to understand and test.

We recommend that you simplify this method or break it into multiple methods. For example, consider extracting the code block on lines 72-98 into a separate method.

src/main/java/com/aws/greengrass/deployment/activator/DefaultActivator.java
aws-kevinrickard: I put in the ideal refactor described above, now the DefaultDeploymentTask determines the timestamp to be used and passes it through to the downstream processing which need to know the timestamp.

src/main/java/com/aws/greengrass/deployment/activator/KernelUpdateActivator.java
```diff
@@ -66,7 +66,7 @@ public KernelUpdateActivator(Kernel kernel, BootstrapManager bootstrapManager) {
...
```
MikeDombo, line 69: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

The cyclomatic complexity of this method is 12. By comparison, 98% of the methods in the CodeGuru reference dataset have a lower cyclomatic complexity. This indicates the method has a high number of decisions and it can make the logic difficult to understand and test.

We recommend that you simplify this method or break it into multiple methods. For example, consider extracting the code block on lines 70-96 into a separate method.

src/main/java/com/aws/greengrass/componentmanager/KernelConfigResolver.java
```diff
@@ -143,16 +143,20 @@ public KernelConfigResolver(ComponentStore componentStore, Kernel kernel, Nucleu
...
+     * @param configMergeTimestamp timestamp to use for configuration merge
      * @return a kernel config map
      * @throws PackageLoadingException if any service package was unable to be loaded
      * @throws IOException             for directory issues
      */
     public Map<String, Object> resolve(List<ComponentIdentifier> componentsToDeploy, DeploymentDocument document,
-            List<String> rootPackages) throws PackageLoadingException, IOException {
+            List<String> rootPackages, long configMergeTimestamp) throws PackageLoadingException, IOException {
         Map<String, Object> servicesConfig = new HashMap<>();
+        LOGGER.atDebug().kv("Components to deploy", componentsToDeploy).kv("Root packages", rootPackages)
```
alter-mage, line 154: nit: Do we need this debug log?

src/main/java/com/aws/greengrass/componentmanager/KernelConfigResolver.java
```diff
@@ -185,6 +189,7 @@ public Map<String, Object> resolve(List<ComponentIdentifier> componentsToDeploy,
         servicesConfig.putIfAbsent(nucleusComponentName, getNucleusComponentConfig(nucleusComponentName));
         servicesConfig.put(kernel.getMain().getName(), getMainConfig(rootPackages, nucleusComponentName));
 
+        LOGGER.atDebug().kv("Services Configuration", servicesConfig).log("Resolved services configuration.");
```
alter-mage, line 192: nit: same

src/main/java/com/aws/greengrass/componentmanager/KernelConfigResolver.java
```diff
@@ -334,12 +341,13 @@ private void updateRunWith(RunWith runWith, Map<String, Object> resolvedServiceC
      * @param configurationUpdateOperation nullable component configuration update operation.
      * @param componentRecipe              component recipe containing default configuration.
      * @param document                     deployment document
+     * @param configMergeTimestamp         timestamp to use during configuration merge
```
alter-mage, line 343: nit: since you removed the document var here, remove it from the method doc as well

src/main/java/com/aws/greengrass/deployment/DefaultDeploymentTask.java
```diff
@@ -146,12 +153,72 @@ public DeploymentResult call() throws InterruptedException {
...
+                    && incomingNucleusComponentConfiguration
+                        .get()
+                        .getConfigurationUpdateOperation()
+                        .getValueToMerge() != null
+                    && incomingNucleusComponentConfiguration
+                        .get()
+                        .getConfigurationUpdateOperation()
+                        .getValueToMerge()
+                        .containsKey(DEVICE_PARAM_DEPLOYMENT_CONFIGURATION_TIME_SOURCE)) {
+                logger.atDebug(DEPLOYMENT_TASK_EVENT_TYPE).log(
```
alter-mage, line 183: nit: same

src/main/java/com/aws/greengrass/deployment/activator/DeploymentActivator.java
```diff
@@ -145,6 +145,8 @@ protected UpdateBehaviorTree createDeploymentMergeBehavior(long deploymentTimest
         insideServiceMergeBehavior.getChildOverride().put(
                 CONFIGURATION_CONFIG_KEY, serviceConfigurationMergeBehavior);
 
+        logger.atDebug().kv("Root merge behavior", rootMergeBehavior)
```
alter-mage, line 148: nit: same

alter-mage (dismissed):
LGTM 
Are you planning on adding new unit tests?

src/main/java/com/aws/greengrass/componentmanager/KernelConfigResolver.java
aws-kevinrickard: This was very helpful for understanding and debugging the deployment process, I think it adds significant value. Is there any downside/concern such as log spam at the debug level? I would think not

aws-kevinrickard:
Re: Unit tests

No plans for unit tests since we have UATs

src/main/java/com/aws/greengrass/componentmanager/KernelConfigResolver.java
aws-kevinrickard: Will update

PR #1707, opened 2025-04-08
chore: update ubuntu version

yitingb:
chore: update ubuntu version

**Issue #, if available:**

**Description of changes:**
Update ubuntu version to latest due to ubuntu 20 version closing down soon https://github.blog/changelog/2025-01-15-github-actions-ubuntu-20-runner-image-brownout-dates-and-other-breaking-changes/#ubuntu-20-image-is-closing-down

Deleted flaky unit test step where testing when shutting down Nucleus, soft dependency does not wait for customer app to close. The unit test itself is testing the shutdown order, which is not useful to this scenario. 

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/deployment/DeploymentTaskIntegrationTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/deployment/DeploymentTaskIntegrationTest.java
@@ -52,6 +52,7 @@ import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.MethodOrderer;
 import org.junit.jupiter.api.Order;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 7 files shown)


junfuchen99 (approved):

src/integrationtests/java/com/aws/greengrass/integrationtests/deployment/DeploymentTaskIntegrationTest.java
```diff
@@ -867,6 +868,7 @@ void GIVEN_services_running_WHEN_service_added_and_deleted_THEN_add_remove_servi
      * Start a service running with a user, then deploy an update to change the user and ensure the correct user stops
      * the process and starts the new one.
      */
+    @Disabled
```
alter-mage, line 871: nit: This is not present in the PR description, but why are we disabling these tests and do we plan on re-enabling them in the future?

yitingb: Yes, after the cgroupv2 is supported, we'll re-enable them. We also test these features in UAT, so disabling them here should not be a problem

alter-mage: What do you mean by `cgroupv2 is supported`? Is that something not supported on ubuntu-latest?

junfuchen99: cgroupv2 is enabled on ubuntu latest by default. Greengrass does not support cgroup v2 and requires v1 for setting system resource limit. It is not easy to reconfigure ubuntu latest to use cgroup v1 because it would require a device reboot, so we propose to disable system resource limit tests with ubuntu latest until we support cgroup v2.

aws-kevinrickard (approved):

PR #1717, opened 2025-06-16
fix: we skip download if the same version of nucleus is being deployed

thomas-roos:
**Issue #, if available:**

**Description of changes:**

During installation the gg nucleus zip is not copied into the artifacts dir, as the installer is contained in the zip file. Even if the zip file is copied, it does not match the expected checksum, as there is some code signature added in the zip file provided by the zip file coming from cloudfront.

This is now using the version that is provided by the localCandidate.getVersion() function to determine if a download is necessary.
So the nucleus component is treated special.

As not necessary downloads should be avoided.

**Documentation Checklist:**

**Compatibility Checklist:**


alter-mage (dismissed):
The PR looks good to me. I made some logging improvements but that's just nit.

alter-mage (approved):

PR #1738, opened 2025-10-15
fix: account for recheck value to be null

muboheng:
**Issue #, if available:**

**Description of changes:**
the sdk allows you to not pass a value for the recheck time. If this happens the nucleus is stuck in a broken state. Adding a null check and a default value to prevent this.
```
java.lang.NullPointerException
        at com.aws.greengrass.lifecyclemanager.UpdateSystemPolicyService.getTimeToReCheck(UpdateSystemPolicyService.java:212)
        at com.aws.greengrass.lifecyclemanager.UpdateSystemPolicyService.startup(UpdateSystemPolicyService.java:168)
        at com.aws.greengrass.lifecyclemanager.Lifecycle.lambda$handleStateTransitionStartingToRunningAsync$9(Lifecycle.java:601)
        at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
        at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
        at java.base/java.lang.Thread.run(Thread.java:829)
```

This issue is hard to require from and requires access to the device. 

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>com.aws.greengrass</groupId>
     <artifactId>nucleus</artifactId>
-    <version>2.14.0-SNAPSHOT</version>
+    <version>2.15.0-SNAPSHOT</version>
     <packaging>jar</packaging>
 
     <licenses>
@@ -858,7 +858,7 @@
         <excludedGroups>E2E,E2E-INTRUSIVE</excludedGroups>
         <groups></groups>
         <greengrassjar.name>Greengrass</greengrassjar.name>
-        <lastVersion>2.13.0-SNAPSHOT</lastVersion>
+        <lastVersion>2.14.0-SNAPSHOT</lastVersion>
     </properties>
     <distributionManagement>
         <snapshotRepository>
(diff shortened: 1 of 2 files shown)


src/main/java/com/aws/greengrass/lifecyclemanager/UpdateSystemPolicyService.java
```diff
@@ -209,12 +211,14 @@ private long getTimeToReCheck(long timeout, String deploymentId,
                     try {
                         DeferComponentUpdateRequest deferRequest = fut.get();
                         if (deploymentId.equals(deferRequest.getDeploymentId())) {
-                            long timeToRecheck = currentTimeMillis + deferRequest.getRecheckAfterMs();
+                            Long recheckAfterMs = deferRequest.getRecheckAfterMs();
+                            long recheckValue = recheckAfterMs == null ? DEFAULT_RECHECK_VALUE : recheckAfterMs;
```
alter-mage: Why do we need a default value? I think we should just reject if the defer request presents a null recheck timeout.

muboheng: why not have a default value? there is still a request to defer the deployment, although in hindsight 0 might be a better value.

PR #1740, opened 2025-10-17
fix: remove warning when limit is set to 0

limingyaoo:
**Issue #, if available:**
The default input value for cpu limit and memory limit is 0, so there shouldn't be warning regarding limit set to 0.

**Description of changes:**
Updated resource limit validation

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/util/platforms/unix/linux/LinuxSystemResourceController.java
+++ b/src/main/java/com/aws/greengrass/util/platforms/unix/linux/LinuxSystemResourceController.java
@@ -111,12 +111,15 @@ public class LinuxSystemResourceController implements SystemResourceController {
                 initializeCgroup(component, memoryManager);
             }
             if (resourceLimit.containsKey(MEMORY_KEY)) {
-                long memoryLimitInKB = Coerce.toLong(resourceLimit.get(MEMORY_KEY));
-                if (memoryLimitInKB > 0) {
-                    memoryManager.setMemoryLimit(component.getServiceName(), memoryLimitInKB);
-                } else {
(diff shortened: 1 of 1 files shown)


muboheng (approved):

src/main/java/com/aws/greengrass/util/platforms/unix/linux/LinuxSystemResourceController.java
```diff
@@ -114,7 +114,7 @@ public void updateResourceLimits(GreengrassService component, Map<String, Object
                 long memoryLimitInKB = Coerce.toLong(resourceLimit.get(MEMORY_KEY));
                 if (memoryLimitInKB > 0) {
                     memoryManager.setMemoryLimit(component.getServiceName(), memoryLimitInKB);
-                } else {
+                } else if (memoryLimitInKB < 0) {
```
junfuchen99: what if the cx provided value is actually 0? Should you try to do null check before calling Coerce.toLong/toDouble?

limingyaoo: Great point. I've tested the null check approach and found that when we configure the deployment without setting (for example) memory limit, the resourceLimit map still has value = 0, (`DEBUG: resourceLimit map. {resourceLimit={cpus=0.5, memory=0}}`). Which means we can't distinguish between customer-provided 0 vs default 0. 
Given that, I think the current behavior (warning on all invalid values including 0) is correct and acceptable. I'll close this pr and keep the existing warning behavior.

PR #1743, opened 2025-10-31
fix: force shutdown unclosed services

limingyaoo:
**Issue #, if available:**

**Description of changes:**
Force cleanup of component processes when kernel shutdown times out.

When stopAllServices() exceeds the 30-second timeout during nucleus bootstrap, components with hard dependencies may remain blocked in waitForDependersToExit(), preventing them from shutdown. This changes adds process-level cleanup to ensure all component processes are terminated before executor service shutdown.

It could prevent resource leaks that cause restart failures.

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
+++ b/src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
@@ -504,7 +504,23 @@ public class KernelLifecycle {
             List<String> unclosedServices =
                     IntStream.range(0, arr.length).filter((i) -> !arr[i].isDone() || arr[i].isCompletedExceptionally())
                             .mapToObj((i) -> d[i].getName()).collect(Collectors.toList());
-            logger.atError("services-shutdown-errored", e).kv("unclosedServices", unclosedServices).log();
+
+            logger.atError("services-shutdown-errored", e).kv("unclosedServices", unclosedServices)
+                    .log("Force closing services");
+
+            for (String unclosedService : unclosedServices) {
+                try {
+                    GreengrassService service = kernel.locate(unclosedService);
+                    // for generic external service, kill processes and clean up
+                    if (service instanceof GenericExternalService) {
+                        GenericExternalService genericService = (GenericExternalService) service;
+                        genericService.forceShutdown();
+                    }
+                } catch (ServiceLoadException ex) {
+                    logger.atWarn().log("Could not locate service: {} for forced process cleanup. "
+                            + "Skipping cleanup for this service.", unclosedService, ex);
+                }
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
```diff
@@ -504,7 +504,22 @@ public void stopAllServices(int timeoutSeconds) {
...
+            for (String unclosedService : unclosedServices) {
+                try {
+                    GreengrassService service = kernel.locate(unclosedService);
+                    // for generic external service, kill processes and clean up
+                    if (service instanceof GenericExternalService) {
+                        GenericExternalService genericService = (GenericExternalService) service;
+                        genericService.forceShutdown();
+                    }
+                } catch (ServiceLoadException ex) {
+                    logger.atWarn().log("Could not locate service: {}", unclosedService, ex);
```
junfuchen99: nit: logging could be more specific?

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
```diff
@@ -504,7 +504,22 @@ public void stopAllServices(int timeoutSeconds) {
             List<String> unclosedServices =
                     IntStream.range(0, arr.length).filter((i) -> !arr[i].isDone() || arr[i].isCompletedExceptionally())
                             .mapToObj((i) -> d[i].getName()).collect(Collectors.toList());
-            logger.atError("services-shutdown-errored", e).kv("unclosedServices", unclosedServices).log();
+
+            logger.atError("services-shutdown-errored", e).kv("unclosedServices", unclosedServices)
+                    .log("Force closing services");
+
+            for (String unclosedService : unclosedServices) {
```
saranyailla, line 511: Is there an ordering needed here?

junfuchen99 (dismissed):

aws-kevinrickard (dismissed):

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
junfuchen99: since we are killing everything without calling shutdown lifecycle, the order should not matter

saranyailla: Looking more closely at the code, found that the services shutdown completion is not ordered even when the shutdown is triggered in order(for the ones that have dependency relationship). So, this order doesn't matter either.

saranyailla (dismissed):

junfuchen99 (approved):

PR #1751, opened 2025-11-20
feat: add ability to configure tes error cache expire time

rauthsuj:
**Issue #, if available:**

**Description of changes:**
Add ability for users to configure TES error cache expire time

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/.github/workflows/externalPR.yml
+++ b/.github/workflows/externalPR.yml
@@ -6,6 +6,9 @@ on:
     types:
       - completed
 
+permissions:
+  pull-requests: write
+
 jobs:
   comment:
     runs-on: ubuntu-latest
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -93,8 +93,8 @@ jobs:
           mkdir -p ./pr/japicmp
           mkdir -p target/japicmp
           mkdir -p target/jacoco-report
-          echo "$PR_NUMBER" > ./pr/NR
-          echo "$PR_SHA" > ./pr/SHA
+          echo "$PR_NUMBER" | tr -cd '0-9' > ./pr/NR
+          echo "$PR_SHA" | tr -cd 'a-fA-F0-9' > ./pr/SHA
           cp -R target/japicmp/default-cli.xml ./pr/japicmp/default-cli.xml
           cp target/jacoco-report/cobertura.xml ./pr/jacoco-report/cobertura.xml
           cp target/jacoco-report/cobertura-it.xml ./pr/jacoco-report/cobertura-it.xml
--- a/pom.xml
+++ b/pom.xml
@@ -84,12 +84,12 @@
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
-            <version>2.15.2</version>
+            <version>2.16.0</version>
         </dependency>
         <dependency>
             <groupId>com.fasterxml.jackson.dataformat</groupId>
             <artifactId>jackson-dataformat-yaml</artifactId>
-            <version>2.15.2</version>
+            <version>2.16.0</version>
         </dependency>
         <dependency>
             <groupId>javax.inject</groupId>


src/main/java/com/aws/greengrass/tes/CredentialRequestHandler.java
```diff
@@ -68,10 +68,18 @@ public class CredentialRequestHandler implements HttpHandler {
...
-    public static final int CLOUD_5XX_ERROR_CACHE_IN_MIN = 1;
-    public static final int UNKNOWN_ERROR_CACHE_IN_MIN = 5;
+    public static final int TIME_BEFORE_CACHE_EXPIRE_IN_SEC = 300;
+    public static final int CLOUD_4XX_ERROR_CACHE_IN_SEC = 120;
+    public static final int CLOUD_5XX_ERROR_CACHE_IN_SEC = 60;
+    public static final int UNKNOWN_ERROR_CACHE_IN_SEC = 300;
+
+    public static final String CLOUD_4XX_ERROR_CACHE_TOPIC = "cloud4xxErrorCacheInSec";
+    public static final String CLOUD_5XX_ERROR_CACHE_TOPIC = "cloud5xxErrorCacheInSec";
+    public static final String UNKNOWN_ERROR_CACHE_TOPIC = "unknownErrorCacheInSec";
```
saranyailla: Nit: Move these to TES class file and rename them to reflect the configuration topics.
The configuration topic names are too implementation specific. These expiration times are when TES will retry fetching credentials again, right? Can we instead call them as `error4xxCredentialRetryInSec`?

rauthsuj: I moved them to TES class and renamed them

src/main/java/com/aws/greengrass/tes/TokenExchangeService.java
```diff
@@ -75,6 +83,56 @@ public TokenExchangeService(Topics topics,
 
         this.authZHandler = authZHandler;
         this.credentialRequestHandler = credentialRequestHandler;
+
+        cloud4xxErrorCache = validateCacheConfig(Coerce.toInt(config.lookup(
+                        CONFIGURATION_CONFIG_KEY, CLOUD_4XX_ERROR_CACHE_TOPIC).dflt(
```
saranyailla: Let's replace this with `findOrDefault`? 

`lookup` and `dflt` create configuration nodes (topics). We don't always need to write the default values to config file for optimization reasons (such as having less no of writes and callback events, and config file remains smaller without losing information).

src/main/java/com/aws/greengrass/tes/TokenExchangeService.java
```diff
@@ -75,6 +83,56 @@ public TokenExchangeService(Topics topics,
...
+
+        // Subscribe to cache configuration changes
+        config.subscribe((why, node) -> {
+            if (node != null && (node.childOf(CLOUD_4XX_ERROR_CACHE_TOPIC)
+                    || node.childOf(CLOUD_5XX_ERROR_CACHE_TOPIC)
+                    || node.childOf(UNKNOWN_ERROR_CACHE_TOPIC))) {
+                logger.atDebug("tes-cache-config-change").kv("node", node).kv("why", why).log();
+
+                int newCloud4xxErrorCache = validateCacheConfig(Coerce.toInt(config.lookup(
+                        CONFIGURATION_CONFIG_KEY, CLOUD_4XX_ERROR_CACHE_TOPIC).dflt(
```
saranyailla: Same comment as above to not use dflt as this will trigger the subscribe callback again with the updated value when the node value is null.

src/main/java/com/aws/greengrass/tes/TokenExchangeService.java
```diff
@@ -130,6 +188,14 @@ private void validateConfig() {
         }
     }
 
+    private int validateCacheConfig(int newCacheValue, int oldCacheValue) {
+        if (newCacheValue < MINIMUM_ERROR_CACHE_IN_SEC) {
+            logger.atError().log("Error cache value must be at least {}", MINIMUM_ERROR_CACHE_IN_SEC);
+            return oldCacheValue;
```
saranyailla: Instead of the old value, we'd want to use default value for each of these or the minimum error cache value. We don't know if the old value itself is valid or not.

src/main/java/com/aws/greengrass/tes/TokenExchangeService.java
```diff
@@ -75,6 +83,56 @@ public TokenExchangeService(Topics topics,
...
+                CredentialRequestHandler.CLOUD_5XX_ERROR_CACHE_IN_SEC);
+        unknownErrorCache = validateCacheConfig(Coerce.toInt(config.lookup(
+                        CONFIGURATION_CONFIG_KEY, UNKNOWN_ERROR_CACHE_TOPIC).dflt(
+                        CredentialRequestHandler.UNKNOWN_ERROR_CACHE_IN_SEC)),
+                CredentialRequestHandler.UNKNOWN_ERROR_CACHE_IN_SEC);
+
+        credentialRequestHandler.configureCacheSettings(cloud4xxErrorCache, cloud5xxErrorCache, unknownErrorCache);
+
+        // Subscribe to cache configuration changes
+        config.subscribe((why, node) -> {
```
saranyailla, line 92: This should be combined with the existing subscribe callback for port changes as it will otherwise create two callback executions when port and error cache config are updated at the same time . 

If both port and error cache config are updated at the same time, it is enough to restart TES once.

src/main/java/com/aws/greengrass/tes/TokenExchangeService.java
rauthsuj: replaced all the new instances of`lookup` and `dflt` with `findOrDefault`

src/main/java/com/aws/greengrass/tes/TokenExchangeService.java
rauthsuj: Done

src/main/java/com/aws/greengrass/tes/TokenExchangeService.java
rauthsuj: Changed the logic, if a value less than minimum is given, only minimum value is set along with an error message stating the allowed minimum value.

src/main/java/com/aws/greengrass/tes/TokenExchangeService.java
rauthsuj: Combined both error cache and port config into one subscribe call back. Checked the scenario using a local nucleus build a UAT test. Now it should only restart once even if both the values are changed at once

src/main/java/com/aws/greengrass/tes/TokenExchangeService.java
```diff
@@ -57,24 +65,66 @@ public TokenExchangeService(Topics topics,
...
+
+        credentialRequestHandler.configureCacheSettings(cloud4xxErrorCache, cloud5xxErrorCache, unknownErrorCache);
+
+        // Subscribe to cache configuration changes
         config.subscribe((why, node) -> {
-            if (node != null && node.childOf(PORT_TOPIC)) {
+            if (node != null && (node.childOf(PORT_TOPIC)
+                    || node.childOf(CLOUD_4XX_ERROR_CACHE_TOPIC)
+                    || node.childOf(CLOUD_5XX_ERROR_CACHE_TOPIC)
+                    || node.childOf(UNKNOWN_ERROR_CACHE_TOPIC))) {
```
saranyailla: Nit: We don't need to execute this callback if the what happened events are irrelevant to this. For eg, see how we don't take an action if the what happened event doesn't show a change in value.

https://github.com/aws-greengrass/aws-greengrass-shadow-manager/blob/132bf7438a2dedbbf71c7787ad633a35db7fbd39/src/main/java/com/aws/greengrass/shadowmanager/ShadowManager.java#L247-L249

saranyailla (approved):

aws-kevinrickard:
We had an offline discussion about if the component should enter errored state when configuration is not valid, instead of printing an error but proceeding to succeed operation by using the minimum value rather than the configured value.

We agreed it is preferable to enter errored state. So that the component provides feedback to greengrass about the configuration not being valid, let greengrass decide how to proceed, and provide immediate feedback to the customer about the invalid configuration.

Currently there is only one way for TES config to be updated, that is through deployments. In this case greengrass can fail the deployment and roll back to a previously working state if configured to do so.

In the future if we introduce the ability for TES config to be updated outside of a deployment (i.e. during runtime) (e.g. by calling UpdateConfiguration for TES from some custom component), then it seems like entering errored state would cause the device to become unhealthy and not auto-recover at that time. This problem is left to be addressed in the future.

src/main/java/com/aws/greengrass/tes/CredentialRequestHandler.java
```diff
@@ -421,16 +431,16 @@ private String parseExpiryFromResponse(final String credentials) throws AWSIotEx
     }
 
     private Instant getExpiryPolicyForErr(int statusCode) {
-        int expiryTime = UNKNOWN_ERROR_CACHE_IN_MIN; // In case of unrecognized cloud errors, back off
+        int expiryTime = unknownErrorCacheInSec; // In case of unrecognized cloud errors, back off
```
aws-kevinrickard, line 442: nit: it was already like this, but expiryDuration is more appropriate than expiryTime

src/main/java/com/aws/greengrass/tes/TokenExchangeService.java
```diff
@@ -57,24 +66,69 @@ public TokenExchangeService(Topics topics,
...
+                    cloud5xxErrorCache = newCloud5xxErrorCache;
+                    unknownErrorCache = newUnknownErrorCache;
+
+                    credentialRequestHandler.configureCacheSettings(
+                            newCloud4xxErrorCache, newCloud5xxErrorCache, newUnknownErrorCache);
+
+                    logger.atInfo("tes-config-change")
+                            .kv("node", node).kv("why", why)
+                            .log("Restarting TES server due to config change");
                     requestRestart();
```
aws-kevinrickard, line 137: I believe you should be able to change the error cache time during runtime and without restart. I don't see why not. This would change the subscription handler logic to only request restart if the port changes. This would be useful because TES restarts cause a lot of dependent component restarts, which we want to avoid if possible.

src/main/java/com/aws/greengrass/tes/CredentialRequestHandler.java
```diff
@@ -68,10 +68,14 @@ public class CredentialRequestHandler implements HttpHandler {
     public static final String AUTH_HEADER = "Authorization";
     public static final String IOT_CREDENTIALS_HTTP_VERB = "GET";
     public static final String SUPPORTED_REQUEST_VERB = "GET";
-    public static final int TIME_BEFORE_CACHE_EXPIRE_IN_MIN = 5;
-    public static final int CLOUD_4XX_ERROR_CACHE_IN_MIN = 2;
-    public static final int CLOUD_5XX_ERROR_CACHE_IN_MIN = 1;
-    public static final int UNKNOWN_ERROR_CACHE_IN_MIN = 5;
+    public static final int TIME_BEFORE_CACHE_EXPIRE_IN_SEC = 300;
+    public static final int CLOUD_4XX_ERROR_CACHE_IN_SEC = 120;
```
aws-kevinrickard: nit: DEFAULT_CLOUD_4XX_ERROR_CACHE_IN_SEC , etc.

src/main/java/com/aws/greengrass/tes/TokenExchangeService.java
```diff
@@ -57,24 +66,61 @@ public TokenExchangeService(Topics topics,
...
         port = Coerce.toInt(config.lookup(CONFIGURATION_CONFIG_KEY, PORT_TOPIC).dflt(DEFAULT_PORT));
+
+        deviceConfiguration.getIotRoleAlias().subscribe((why, newv) -> {
+            iotRoleAlias = Coerce.toString(newv);
+        });
+
+        this.authZHandler = authZHandler;
+        this.credentialRequestHandler = credentialRequestHandler;
+
+        int[] errorCacheConfig = getErrorCacheConfigValues();
```
aws-kevinrickard: I don't think we want to throw any errors for invalid config during the constructor? I'm familiar with what happens to the service lifecycle that case. Please let me know if that was intentional / how that works if so.

We don't have to do it the following way if there is a better way, I don't love what I am about to propose. But I was expecting it this way, based on how the existing code validates the port, and think this is a reasonable way to do it:

1. In constructor, fetch the config if it exists or use default if not exists. Set/store the fetched value in CRH. No validation yet.
2. In startup script, fetch the values from the CRH state and validate them. If invalid, call serviceErrored; if valid proceed running.
3. In config update callback, fetch the new config if exists or use default if not exists, and validate immediately. If invalid, call serviceErrored; if valid update the values stored in CRH and keep running.

You have most of this already implemented, but the differences are:
1. (This line) In the constructor you are also implicitly doing validation and not handling the possible IllegalArgumentException. 
2. In the startup script, instead of validating the values currently set in the CRH, we are fetching the value from config again, and then validating that. I can't think of a scenario where these would be different so not a big deal but I think it would be better to validate what we've already fetched in the constructor.
3. In the config update callback, we should handle the invalid case (IllegalArgumentException) and call serviceErrored in that case.

So in order to do this, I think you'll need to separate your function which gets the values from the config, from the function which validates the values.

src/main/java/com/aws/greengrass/tes/TokenExchangeService.java
```diff
@@ -123,13 +121,90 @@ private void setEnvVariablesForDependencies(final int serverPort) {
...
+    }
+
     private void validateConfig() {
         // Validate roleAlias
         if (Utils.isEmpty(iotRoleAlias)) {
             throw new IllegalArgumentException(String.format(TES_CONFIG_ERROR_STR, IOT_ROLE_ALIAS_TOPIC));
         }
     }
 
+    private void onConfigurationChange() {
```
aws-kevinrickard: This function name is confusing, cause it does not fire on configuration changes- it only fires once on startup
Suggest rename to something like `loadConfig`, and call it from startup directly so it isn't called inside of subscribeToConfigChanges

aws-kevinrickard (dismissed):

src/main/java/com/aws/greengrass/tes/TokenExchangeService.java
rauthsuj: Removed it and moved back subscribe call to constructor like how it was before.

aws-kevinrickard:
as a follow up to https://github.com/aws-greengrass/aws-greengrass-nucleus/pull/1751#issuecomment-3578024725 , 

It was further discussed and decided that we want to make the TES component less likely to break because of its importance to Greengrass and deployments (currently, a broken TES component prevents future deployments from working). So we prefer to log the error message and succeed in running with a valid value of our choosing. Part of the justification for why this is okay is that these cache configs are within our service domain and thus customers are unlikely to have a problem if we adjust the value for them to make it valid. There is still a communication/feedback problem in that a customer will not know that this is happening unless they happen to check or detect this in their logs, and that could be improved in the future by introducing some kind of communication channel or mechanism for non-blocking errors like this.

src/main/java/com/aws/greengrass/tes/CredentialRequestHandler.java
```diff
@@ -68,10 +68,14 @@ public class CredentialRequestHandler implements HttpHandler {
...
-    public static final int TIME_BEFORE_CACHE_EXPIRE_IN_MIN = 5;
-    public static final int CLOUD_4XX_ERROR_CACHE_IN_MIN = 2;
-    public static final int CLOUD_5XX_ERROR_CACHE_IN_MIN = 1;
-    public static final int UNKNOWN_ERROR_CACHE_IN_MIN = 5;
+    public static final int DEFAULT_TIME_BEFORE_CACHE_EXPIRE_IN_SEC = 300;
+    public static final int DEFAULT_CLOUD_4XX_ERROR_CACHE_IN_SEC = 120;
+    public static final int DEFAULT_CLOUD_5XX_ERROR_CACHE_IN_SEC = 60;
+    public static final int DEFAULT_UNKNOWN_ERROR_CACHE_IN_SEC = 300;
+
+    private int cloud4xxErrorCacheInSec = DEFAULT_CLOUD_4XX_ERROR_CACHE_IN_SEC;
```
aws-kevinrickard: we should set these as `volatile` to guarantee an updated value is used after update

PR #1756, opened 2025-12-11
chore: upgrade jackson-databind version to 2.16.0

yitingb:
**Issue #, if available:**

**Description of changes:**
Update jackson-databind version to 2.16.0 which is the current latest
There is a CVE opened against our current jackson-databind version 2.15.2

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/.github/workflows/externalPR.yml
+++ b/.github/workflows/externalPR.yml
@@ -6,6 +6,9 @@ on:
     types:
       - completed
 
+permissions:
+  pull-requests: write
+
 jobs:
   comment:
     runs-on: ubuntu-latest


pom.xml
```diff
@@ -84,12 +84,12 @@
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
-            <version>2.15.2</version>
+            <version>2.16.0</version>
```
alter-mage, line 87: just out of curiosity, why 2.16.0?

yitingb: Upgrade to 2.16.0 but not the latest because we're using [UpperCamelCaseStrategy](https://www.javadoc.io/static/com.fasterxml.jackson.core/jackson-databind/2.17.0/com/fasterxml/jackson/databind/PropertyNamingStrategy.UpperCamelCaseStrategy.html) in component-common snapshot. And this method got deprecated with version 2.17.0 and above. Upgrading to latest doesn't really change the way we use the library in our snapshots, so I just chose 2.16.0 for compatibility reason

alter-mage (approved):

PR #1761, opened 2026-01-20
feat: add uninstall lifecycle

muboheng:
**Issue #, if available:**

**Description of changes:**

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/lifecyclemanager/GenericExternalServiceIntegTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/lifecyclemanager/GenericExternalServiceIntegTest.java
@@ -884,6 +884,97 @@ class GenericExternalServiceIntegTest extends BaseITCase {
         assertFalse(freezerManager.isComponentFrozen(component.getServiceName()));
     }
 
+    // AC1 & AC3: End-to-end uninstall with environment variables
+    @Test
+    void GIVEN_component_with_uninstall_script_WHEN_uninstall_THEN_script_executes_successfully() throws Exception {
+        ConfigPlatformResolver.initKernelWithMultiPlatformConfig(kernel, 
+            getClass().getResource("uninstall_success_config.yaml"));
+
+        CountDownLatch finishedLatch = new CountDownLatch(1);
+        CountDownLatch uninstalledLatch = new CountDownLatch(1);
+        
+        kernel.getContext().addGlobalStateChangeListener((service, oldState, newState) -> {
+            if (service.getName().equals("UninstallTestComponent")) {
+                if (newState.equals(State.FINISHED)) {
+                    finishedLatch.countDown();
+                } else if (newState.equals(State.UNINSTALLED)) {
+                    uninstalledLatch.countDown();
+                }
+            }
+        });
+
+        kernel.launch();
+        assertTrue(finishedLatch.await(30, TimeUnit.SECONDS), "component should reach FINISHED");
+
+        GenericExternalService component = (GenericExternalService) kernel.locate("UninstallTestComponent");
+        component.requestUninstall();
+        
+        assertTrue(uninstalledLatch.await(30, TimeUnit.SECONDS), "component should reach UNINSTALLED");
+        assertThat(component.getState(), is(State.UNINSTALLED));
+    }
+
+    // AC2: Uninstall script failure doesn't fail deployment
+    @Test
+    void GIVEN_component_with_failing_uninstall_script_WHEN_uninstall_THEN_component_still_uninstalled() throws Exception {
+        ConfigPlatformResolver.initKernelWithMultiPlatformConfig(kernel, 
+            getClass().getResource("uninstall_failure_config.yaml"));
+
+        CountDownLatch finishedLatch = new CountDownLatch(1);
+        CountDownLatch uninstalledLatch = new CountDownLatch(1);
+        
+        kernel.getContext().addGlobalStateChangeListener((service, oldState, newState) -> {
+            if (service.getName().equals("FailingUninstallComponent")) {
+                if (newState.equals(State.FINISHED)) {
+                    finishedLatch.countDown();
+                } else if (newState.equals(State.UNINSTALLED)) {
+                    uninstalledLatch.countDown();
+                }
+            }
+        });
+
+        kernel.launch();
+        assertTrue(finishedLatch.await(30, TimeUnit.SECONDS), "component should reach FINISHED");
+
+        GenericExternalService component = (GenericExternalService) kernel.locate("FailingUninstallComponent");
+        component.requestUninstall();
+        
+        assertTrue(uninstalledLatch.await(30, TimeUnit.SECONDS), "component should reach UNINSTALLED despite script failure");
+        assertThat(component.getState(), is(State.UNINSTALLED));
+    }
+
+    // AC7: State transitions
+    @Test
+    void GIVEN_component_WHEN_uninstall_requested_THEN_goes_through_UNINSTALLING_to_UNINSTALLED_states() throws Exception {
+        ConfigPlatformResolver.initKernelWithMultiPlatformConfig(kernel, 
+            getClass().getResource("uninstall_state_transition_config.yaml"));
+
+        List<State> stateTransitions = new CopyOnWriteArrayList<>();
+        CountDownLatch finishedLatch = new CountDownLatch(1);
+        CountDownLatch uninstalledLatch = new CountDownLatch(1);
+        
+        kernel.getContext().addGlobalStateChangeListener((service, oldState, newState) -> {
+            if (service.getName().equals("StateTransitionComponent")) {
+                stateTransitions.add(newState);
+                if (newState.equals(State.FINISHED)) {
+                    finishedLatch.countDown();
+                } else if (newState.equals(State.UNINSTALLED)) {
+                    uninstalledLatch.countDown();
+                }
+            }
+        });
+
+        kernel.launch();
+        assertTrue(finishedLatch.await(30, TimeUnit.SECONDS), "component should reach FINISHED");
+
+        GenericExternalService component = (GenericExternalService) kernel.locate("StateTransitionComponent");
+        component.requestUninstall();
+        
+        assertTrue(uninstalledLatch.await(30, TimeUnit.SECONDS), "component should reach UNINSTALLED");
+        
+        assertTrue(stateTransitions.contains(State.UNINSTALLING), "Should transition through UNINSTALLING");
+        assertTrue(stateTransitions.contains(State.UNINSTALLED), "Should reach UNINSTALLED");
+    }
+
     private boolean isCgroupV2Supported() {
         return Files.exists(Paths.get("/sys/fs/cgroup/cgroup.controllers"));
     }
--- a/src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
+++ b/src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
@@ -392,6 +392,11 @@ public class DeploymentConfigMerger {
                 return true;
             }).collect(Collectors.toSet());
             logger.atInfo(MERGE_CONFIG_EVENT_KEY).kv("service-to-remove", servicesToRemove).log("Removing services");
+
+            // Request uninstall for each service before closing
+            for (GreengrassService service : ggServicesToRemove) {
+                service.requestUninstall();
+            }
             // waiting for removed service to close before removing reference and config entry
             for (GreengrassService service : ggServicesToRemove) {
                 try {
--- a/src/main/java/com/aws/greengrass/lifecyclemanager/GenericExternalService.java
+++ b/src/main/java/com/aws/greengrass/lifecyclemanager/GenericExternalService.java
@@ -63,6 +63,7 @@ public class GenericExternalService extends GreengrassService {
     private static final String SKIP_COMMAND_REGEX = "(exists|onpath) +(.+)";
     private static final Pattern SKIPCMD = Pattern.compile(SKIP_COMMAND_REGEX);
     private static final String CONFIG_NODE = "configNode";
+    public static final String COMPONENT_VERSION_ENV_NAME = "GREENGRASS_COMPONENT_VERSION";
     // Logger which write to a file for just this service
     protected final Logger separateLogger;
     protected final Platform platform;
@@ -597,6 +598,34 @@ public class GenericExternalService extends GreengrassService {
                 logger.atInfo().setEventType("generic-service-shutdown").log();
             }
             resetRunWith(); // reset runWith - a deployment can change user info
+        }
+    }
+
+    /**
+     * Execute the uninstall lifecycle script for permanent component removal.
+     */
+    @Override
+    protected void uninstall() {
+        try (LockScope ls = LockScope.lock(lock)) {
+            logger.atInfo().log("Shutdown initiated");
+
+            try {
+                RunResult result = run(Lifecycle.LIFECYCLE_UNINSTALL_NAMESPACE_TOPIC, null, lifecycleProcesses, false);
+                if (result.getExec() != null) {
+                    Topic versionTopic = getConfig().find(VERSION_CONFIG_KEY);
+                    if (versionTopic != null) {
+                        result.getExec().setenv(COMPONENT_VERSION_ENV_NAME, Coerce.toString(versionTopic));
+                    }
+                    if (result.getDoExec() != null) {
+                        result.getDoExec().apply();
+                    }
+                }
+            } catch (InterruptedException ex) {
+                logger.atWarn("generic-service-uninstall-interrupted")
+                        .kv("componentName:", getServiceName())
+                        .log("Thread interrupted while uninstalling service");
+                Thread.currentThread().interrupt();
+            }
         }
     }
 
--- a/src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
+++ b/src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
@@ -59,6 +59,7 @@ public class Lifecycle {
     public static final String LIFECYCLE_INSTALL_NAMESPACE_TOPIC = "install";
     public static final String LIFECYCLE_STARTUP_NAMESPACE_TOPIC = "startup";
     public static final String LIFECYCLE_SHUTDOWN_NAMESPACE_TOPIC = "shutdown";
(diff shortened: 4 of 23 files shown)


src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
```diff
@@ -675,7 +689,50 @@ private void handleCurrentStateFinished(Optional<State> desiredState) {
...
+        }
+
+    }
+
+    /**
+     * Handle UNINSTALLING state - execute uninstall script and exit lifecycle thread.
+     */
+    private void handleCurrentStateUninstalled(Optional<State> desiredState) {
+        //do nothing, unless its reinstall
+        if (State.NEW.equals(desiredState.get())) {
```
MikeDombo: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

It appears that you are using `Optional` value which can hold either a value or not. The value held in the `Optional` can be accessed using the `get()` method, but it will throw a `NoSuchElementException` if there is no value present. To avoid the exception, calling the `isPresent()` or `ifPresent()` or `!isEmpty()` method should always be done before any call to `get()`. Alternatively, other methods such as `orElse(...)`, `orElseGet(...)` or `orElseThrow(...)` can be used to specify what to do with an empty `Optional`.

[CWE-476](https://cwe.mitre.org/data/definitions/476)

src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
```diff
@@ -675,7 +690,48 @@ private void handleCurrentStateFinished(Optional<State> desiredState) {
...
     }
 
+    /**
+     * Handle UNINSTALLING state - execute uninstall script and exit lifecycle thread.
+     */
+    private void handleCurrentStateUninstalling() throws InterruptedException {
+        Future<?> uninstallFuture = greengrassService.getContext().get(ExecutorService.class).submit(() -> {
+            try {
+                greengrassService.uninstall();
+            } catch (InterruptedException i) {
```
MikeDombo, line 700: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

Learn more about [interrupts](https://docs.oracle.com/javase/tutorial/essential/concurrency/interrupt.html)

src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
```diff
@@ -823,6 +890,10 @@ void setClosed(boolean b) {
         isClosed.set(b);
     }
 
+    void setUninstall(boolean b) {
+        requestedUninstall.set(b);
```
alter-mage: nit: b is always true, right?

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
```diff
@@ -392,6 +392,11 @@ public void removeObsoleteServices() throws InterruptedException, ServiceUpdateE
                 return true;
             }).collect(Collectors.toSet());
             logger.atInfo(MERGE_CONFIG_EVENT_KEY).kv("service-to-remove", servicesToRemove).log("Removing services");
+
+            // Request uninstall for each service before closing
+            for (GreengrassService service : ggServicesToRemove) {
```
alter-mage, line 397: to me it makes more sense to stop a service before uninstalling it. Imo, this uninstall loop should come after the next loop.

This will also simplify your requestUninstall method in Lifecycle

src/integrationtests/java/com/aws/greengrass/integrationtests/lifecyclemanager/GenericExternalServiceIntegTest.java
```diff
@@ -884,6 +884,97 @@ void GIVEN_running_service_WHEN_pause_resume_requested_THEN_pause_resume_Service
         assertFalse(freezerManager.isComponentFrozen(component.getServiceName()));
     }
 
+    // AC1 & AC3: End-to-end uninstall with environment variables
+    @Test
+    void GIVEN_component_with_uninstall_script_WHEN_uninstall_THEN_script_executes_successfully() throws Exception {
```
alter-mage, line 889: I was expecting an integ test where you deploy a component with uninstall and then deploy another deployment which removes it.

You can take inspiration from these [tests](https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/86a3562ee24908cbee8a61753768c768b1c6dcf9/src/integrationtests/java/com/aws/greengrass/integrationtests/deployment/DeploymentServiceIntegrationTest.java#L149)

src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
```diff
@@ -346,11 +349,13 @@ private void enqueueStateEvent(StateEvent event) {
     private void startStateTransition() throws InterruptedException {
         AtomicReference<Predicate<Object>> asyncFinishAction = new AtomicReference<>((stateEvent) -> true);
         State prevState = getState();
-        while (!(isClosed.get() && getState().isClosable())) {
+        // if uninstall is requested, the wait for uninstalled state, else see if is closable
+        while (requestedUninstall.get()
```
alter-mage, line 353: why do you need the additional check?

src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
```diff
@@ -479,6 +487,10 @@ private void handleCurrentStateBroken(Optional<State> desiredState, State previo
                     stopBackingTask();
                 }
                 break;
+            case UNINSTALLING:
```
alter-mage, line 490: why would a service go from uninstalling / uninstalled to broken?

src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
```diff
@@ -490,6 +502,9 @@ private void handleCurrentStateBroken(Optional<State> desiredState, State previo
         if (State.NEW.equals(desiredState.get())) {
             internalReportState(State.NEW);
             stateToErroredCount.clear();
+        } else if (State.UNINSTALLED.equals(desiredState.get()) && getState() != State.UNINSTALLED) {
```
alter-mage, line 505: I dont think this code path will ever get hit. Reason being, uninstall is only triggered when a deployment removes a component from a device. The method `handleCurrentStateBroken` is invoked whenever a component transitions to BROKEN. Unless, a component is broken while trying to stop, sure.

But I have left a comment somewhere here where I want Nucleus to stop a component before uninstalling it so this edge case should also not exist.

src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
```diff
@@ -740,6 +803,10 @@ private void serviceTerminatedMoveToDesiredState(@Nonnull State desiredState) {
             case FINISHED:
                 internalReportState(State.FINISHED);
                 break;
+            case UNINSTALLED:
+            case UNINSTALLING:
```
saranyailla: switch the order of these two?

src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
saranyailla, line 353: What's the behavior here when an uninstalled service is rolled back from a failed deployment?

src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
```diff
@@ -675,6 +690,45 @@ private void handleCurrentStateFinished(Optional<State> desiredState) {
...
+        } finally {
+            internalReportState(State.UNINSTALLED);
+        }
+
+    }
+
+    /**
+     * Handle UNINSTALLING state - execute uninstall script and exit lifecycle thread.
+     */
+    private void handleCurrentStateUninstalled(Optional<State> desiredState) {
```
saranyailla, line 727: The UNINSTALLED service instance must be removed from GG context in this case. This is handled during the deployment execution itself. So, we don't need to handle the NEW state of the service instance here. This is similar to handling FINISHED state.

src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
```diff
@@ -725,6 +783,11 @@ private void handleCurrentStateErrored(Optional<State> desiredState, State prevS
      */
     @SuppressWarnings("PMD.MissingBreakInSwitch")
     private void serviceTerminatedMoveToDesiredState(@Nonnull State desiredState) {
+        if (requestedUninstall.get()) {
```
saranyailla: `requestedUninstall.get()` returns `True` for a service that is already in `UNINSTALLED` state. So, reporting the state as UNINSTALLING is not correct.

src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
```diff
@@ -857,6 +928,10 @@ final void requestReinstall() {
             logger.atWarn("service-shutdown-in-progress")
                     .log("Requesting service to reinstall while it is closing");
         }
+        if (requestedUninstall.compareAndSet(true, false)) {
```
saranyailla, line 934: If we handle Uninstall only via deployments, is this something that we still run into?

src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
muboheng: this comes from the logic that we can have both closing components and uninstalling components. closing component that is trying to uninstall should not exit out of this loop, instead should run until the state is UNINSTALLED. Closing component can just exit when ever it reaches on of the satisfied states.

src/integrationtests/java/com/aws/greengrass/integrationtests/lifecyclemanager/GenericExternalServiceIntegTest.java
muboheng: ack

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
muboheng: This is what I thought initially too, but there is a situation where a component can close without trying to uninstall. specifically in version upgrades. So we need a way to differentiate before the life cycle finishes. simply calling close before uninstall would cause the component to exit out of the lifecycle loop and not attempt the uninstall step.

src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
muboheng: a reinstall/install will be requested which will reset this boolean. then the install script will run again.

src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
muboheng: It shouldn't happen, but we will try to report uninstalled anyways.

src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
muboheng: This is also valid for when the component has already reached the Broken state then we try to uninstall. In this current design we are stopping the component before uninstalling it. It does so by the lifecycle stages. where we ensure that we're either in finished or broken state before transitioning to uninstalling

src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
muboheng: b is not always true. our usage right now is to only call true, yes.

src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
muboheng: There is a case where the uninstall takes a long time to finish, and a new deployment comes in with a new version of the removed component

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
alter-mage: I thought we were not doing uninstall during a component version update in this iteration. Is that not true?

src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
alter-mage: Right, does not make sense to pass the variable `b` if the `setUninstall` method always sets the `requestedUninstall` to true.

src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
alter-mage: Is there a race condition here where a component is marked as closed but not yet marked for uninstall?

src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
alter-mage: Why?

src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
alter-mage: I see, that makes sense

src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
muboheng: if it marked for closed not for uninstall then we will simply actually close it when it reaches finish.

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
muboheng: yes, so this is specifically designed to guard against that. 
there are also other situations where a component can close, without uninstalling.

src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
muboheng: uninstalling is a subset of closing. so if its only closing, we will exit when it reaches finished/broken, but if its closing + uninstalling, we will only exit at uninstalled.

src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
muboheng: as a fail safe. so it does not get stuck in this broken state when theres a deployment that comes in to remove the component. By design this should not be possible, but just in case.

src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
muboheng: sure I can flatten this. I was aligning this with how isClosed is implemented, but it seems like isClosed is also access outside of this class.

src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
muboheng: it returns true for when a component is attempting to uninstall, this is true for how isClosed is interpreted too. so its valid here to report uninstalling. What I should do is to ensure that the state is in FINISHED before doing so.

src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
muboheng: This is to account for the edge case where a reinstall is requested when a new deployment (after the one to remove the component) comes in with a new version of the component. so we do not stay in this limbo uninstalled state. you can see something similar in [here](https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/main/src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java#L734)

src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
```diff
@@ -675,6 +690,45 @@ private void handleCurrentStateFinished(Optional<State> desiredState) {
...
 
+    /**
+     * Handle UNINSTALLING state - execute uninstall script and exit lifecycle thread.
+     */
+    private void handleCurrentStateUninstalling() throws InterruptedException {
+        Future<?> uninstallFuture = greengrassService.getContext().get(ExecutorService.class).submit(() -> {
+            try {
+                greengrassService.uninstall();
+            } catch (InterruptedException i) {
+                logger.atWarn("service-shutdown-interrupted").log("Service interrupted while running shutdown");
```
saranyailla: Fix the state in log.

src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
```diff
@@ -675,6 +690,45 @@ private void handleCurrentStateFinished(Optional<State> desiredState) {
...
+            uninstallFuture.cancel(true);
+            greengrassService.serviceErrored(ComponentStatusCode.UNINSTALL_TIMEOUT, "Timeout in uninstall");
+        } finally {
+            internalReportState(State.UNINSTALLED);
+        }
+
+    }
+
+    /**
+     * Handle UNINSTALLING state - execute uninstall script and exit lifecycle thread.
```
saranyailla: Nit: UNINSTALLED state.

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
alter-mage: What other situations?

src/main/java/com/aws/greengrass/lifecyclemanager/Lifecycle.java
saranyailla: Can we reset the `requestedUninstall` flag here?

src/main/java/com/aws/greengrass/lifecyclemanager/GenericExternalService.java
```diff
@@ -600,6 +601,34 @@ protected void shutdown() {
         }
     }
 
+    /**
+     * Execute the uninstall lifecycle script for permanent component removal.
+     */
+    @Override
+    protected void uninstall() {
+        try (LockScope ls = LockScope.lock(lock)) {
+            logger.atInfo().log("Shutdown initiated");
```
saranyailla, line 610: nit: uninstall

PR #1776, opened 2026-04-01
feat: detect endpoint-switch deployments and persist rollback metadata (Step 1)

Vipul-5517:
### Summary

Adds the foundation for IoT endpoint switch deployments in Greengrass Nucleus Classic. When a deployment changes `iotDataEndpoint`, the deployment service detects it, persists the source endpoint for later status reporting and rollback, and forces config snapshot creation regardless of the deployment's `FailureHandlingPolicy`.

This is Step 1 of the IoT Endpoint Switch design.

### Changes

- **`DeploymentConfigMerger`** — Add `isEndpointSwitchDeployment()` to detect when incoming config changes `iotDataEndpoint`. Persist source endpoint metadata before activation.
- **`DeploymentDirectoryManager`** — Add `persistSourceEndpoints()`, `readSourceEndpoints()`, `hasEndpointSwitchMetadata()` for endpoint switch metadata lifecycle.
- **`DefaultActivator`** — Force config snapshot and rollback support for endpoint-switch deployments even when `FailureHandlingPolicy` is `DO_NOTHING`.
- **`EndpointSwitchMetadata`** — Model class for persisted source endpoint data.

### E2E Test Results (14/14 pass)

Tested with locally built Nucleus 2.16.0-SNAPSHOT in Podman containers, cross-region endpoint switch (us-east-1 → us-west-2).

| # | Scenario | Result | Status | Config Changed |
|---|----------|--------|--------|----------------|
| 1 | Happy path endpoint switch | ✅ | SUCCEEDED | Yes → us-west-2 |
| 2 | No IoT resources in dest | ✅ | FAILED_NO_STATE_CHANGE | No |
| 3 | No IoT policy in dest | ✅ | FAILED_NO_STATE_CHANGE | No |
| 4 | Non-endpoint deployment | ✅ | SUCCEEDED | No |
| 5 | Same endpoint value | ✅ | SUCCEEDED | No |
| 6 | Only data endpoint (no cred) | ✅ | FAILED_NO_STATE_CHANGE | No |
| 7 | DO_NOTHING forces snapshot | ✅ | SUCCEEDED | Yes → us-west-2 |
| 8 | Timeout (non-routable IP) | ✅ | FAILED_NO_STATE_CHANGE | No |
| 9 | Invalid endpoint format | ✅ | FAILED_NO_STATE_CHANGE | No |
| 10 | Region mismatch | ✅ | FAILED_NO_STATE_CHANGE | No |
| 11 | Metadata survives restart | ✅ | SUCCEEDED | Yes → us-west-2 |
| 12 | Back-to-back switches | ✅ | SUCCEEDED ×2 | Yes (both ways) |
| 13 | Endpoint + logging config | ✅ | SUCCEEDED | Yes + DEBUG |
| 14 | Retry behavior verification | ✅ | FAILED_NO_STATE_CHANGE | No |

### Testing

- Unit tests added for `isEndpointSwitchDeployment`, metadata persistence, and forced rollback.
- Existing tests updated for new `DeploymentDirectoryManager` constructor parameter.

### Related

- Step 2 (pre-flight MQTT check) will follow in a separate PR.

### DIFF
--- a/src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
+++ b/src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
@@ -37,6 +37,7 @@ import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
@@ -60,6 +61,7 @@ public class DeploymentConfigMerger {
     public static final String MERGE_ERROR_LOG_EVENT_KEY = "config-update-error";
     public static final String DEPLOYMENT_ID_LOG_KEY = "deploymentId";
     public static final String SERVICE_NAME_LOG_KEY = "serviceName";
+    public static final String SOURCE_IOT_DATA_ENDPOINT_KEY = "sourceIotDataEndpoint";
     protected static final int WAIT_SVC_START_POLL_INTERVAL_MILLISEC = 1000;
 
     private static final Logger logger = LogManager.getLogger(DeploymentConfigMerger.class);
@@ -152,6 +154,22 @@ public class DeploymentConfigMerger {
         // Validate the AWS region, IoT credentials endpoint as well as the IoT data endpoint.
         if (!validateNucleusConfig(totallyCompleteFuture, nucleusConfig)) {
             return;
+        }
+
+        // Persist source endpoint before activation for endpoint-switch deployments.
+        // Stored under services.DeploymentService.runtime (internal deployment state, not customer config).
+        // Persisted to config.tlog so it survives device crashes mid-endpoint-switch.
+        // Step 5 uses this value to report deployment status back to the source account.
+        // Cleared after status reporting; stale keys are harmless and overwritten by the next switch.
+        if (isEndpointSwitchDeployment(nucleusConfig, deviceConfiguration)) {
+            String currentDataEndpoint = Coerce.toString(deviceConfiguration.getIotDataEndpoint());
+            String newDataEndpoint = Coerce.toString(nucleusConfig.get(DEVICE_PARAM_IOT_DATA_ENDPOINT));
+            logger.atInfo().setEventType(MERGE_CONFIG_EVENT_KEY)
+                    .kv("currentIotDataEndpoint", currentDataEndpoint)
+                    .kv("newIotDataEndpoint", newDataEndpoint)
+                    .log("Endpoint switch deployment detected, persisting source endpoint for rollback");
+            kernel.getContext().get(DeploymentService.class).getRuntimeConfig()
+                    .lookup(SOURCE_IOT_DATA_ENDPOINT_KEY).withValue(currentDataEndpoint);
         }
 
         logger.atInfo(MERGE_CONFIG_EVENT_KEY).kv("deployment", deploymentId)
@@ -253,6 +271,22 @@ public class DeploymentConfigMerger {
         return iotDataEndpoint;
     }
 
+    /**
+     * Detect whether a deployment changes the IoT data endpoint.
+     *
+     * @param nucleusConfig       the incoming nucleus configuration map, may be null
+     * @param deviceConfiguration the current device configuration
+     * @return true if the deployment changes iotDataEndpoint
+     */
+    static boolean isEndpointSwitchDeployment(Map<String, Object> nucleusConfig,
+                                              DeviceConfiguration deviceConfiguration) {
+        if (nucleusConfig == null || !nucleusConfig.containsKey(DEVICE_PARAM_IOT_DATA_ENDPOINT)) {
+            return false;
+        }
+        String current = Coerce.toString(deviceConfiguration.getIotDataEndpoint());
+        String incoming = Coerce.toString(nucleusConfig.get(DEVICE_PARAM_IOT_DATA_ENDPOINT));
+        return !Objects.equals(current, incoming);
+    }
 
     @Getter
     @AllArgsConstructor(access = AccessLevel.PRIVATE)
--- a/src/main/java/com/aws/greengrass/deployment/activator/DefaultActivator.java
+++ b/src/main/java/com/aws/greengrass/deployment/activator/DefaultActivator.java
@@ -5,7 +5,9 @@
 
 package com.aws.greengrass.deployment.activator;
 
+import com.aws.greengrass.config.Topic;
 import com.aws.greengrass.deployment.DeploymentConfigMerger;
+import com.aws.greengrass.deployment.DeploymentService;
 import com.aws.greengrass.deployment.exceptions.ServiceUpdateException;
 import com.aws.greengrass.deployment.model.Deployment;
 import com.aws.greengrass.deployment.model.DeploymentDocument;
@@ -13,6 +15,7 @@ import com.aws.greengrass.deployment.model.DeploymentResult;
 import com.aws.greengrass.lifecyclemanager.GreengrassService;
 import com.aws.greengrass.lifecyclemanager.Kernel;
 import com.aws.greengrass.lifecyclemanager.exceptions.ServiceLoadException;
+import com.aws.greengrass.util.Coerce;
 
 import java.util.HashMap;
 import java.util.Map;
@@ -49,7 +52,18 @@ public class DefaultActivator extends DeploymentActivator {
         }
 
         DeploymentDocument deploymentDocument = deployment.getDeploymentDocumentObj();
-        if (isAutoRollbackRequested(deploymentDocument) && !takeConfigSnapshot(totallyCompleteFuture)) {
+        boolean autoRollback = isAutoRollbackRequested(deploymentDocument);
+
+        // Endpoint switch deployments force config snapshot and rollback regardless of FailureHandlingPolicy,
+        // so that we can restore the original endpoint if the new one is unreachable after applying the change.
+        // Only check when auto-rollback is not already requested to avoid redundant config store reads.
+        boolean forceSnapshotForEndpointSwitch = !autoRollback && isEndpointSwitch();
+        if (forceSnapshotForEndpointSwitch) {
+            logger.atInfo(MERGE_CONFIG_EVENT_KEY)
+                    .log("Endpoint switch deployment: forcing config snapshot and rollback support");
+        }
+        if ((autoRollback || forceSnapshotForEndpointSwitch)
+                && !takeConfigSnapshot(totallyCompleteFuture)) {
             return;
         }
 
@@ -114,7 +128,9 @@ public class DefaultActivator extends DeploymentActivator {
                                Throwable failureCause) {
         logger.atError(MERGE_CONFIG_EVENT_KEY).kv(DEPLOYMENT_ID_LOG_KEY, deploymentDocument.getDeploymentId())
                 .setCause(failureCause).log("Deployment failed");
-        if (isAutoRollbackRequested(deploymentDocument)) {
+        if (isAutoRollbackRequested(deploymentDocument) || isEndpointSwitch()) {
+            logger.atInfo(MERGE_CONFIG_EVENT_KEY).kv(DEPLOYMENT_ID_LOG_KEY, deploymentDocument.getDeploymentId())
+                    .log("Initiating rollback for failed deployment");
             rollback(deploymentDocument, totallyCompleteFuture, failureCause,
                     servicesChangeManager.createRollbackManager());
         } else {
@@ -178,6 +194,17 @@ public class DefaultActivator extends DeploymentActivator {
         } catch (InterruptedException | ServiceUpdateException | ServiceLoadException e) {
             handleFailureRollback(totallyCompleteFuture, failureCause, e);
         }
+    }
+
+    /**
+     * Check if the current deployment is an endpoint switch by looking for persisted source endpoint
+     * in DeploymentService runtime config.
+     */
+    private boolean isEndpointSwitch() {
+        Topic t = kernel.getContext().get(DeploymentService.class).getRuntimeConfig()
+                .find(DeploymentConfigMerger.SOURCE_IOT_DATA_ENDPOINT_KEY);
+        String source = t == null ? null : Coerce.toString(t);
+        return source != null && !source.isEmpty();
     }
 
     private void handleFailureRollback(CompletableFuture totallyCompleteFuture, Throwable deploymentFailureCause,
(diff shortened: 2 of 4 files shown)


src/main/java/com/aws/greengrass/deployment/DeploymentDirectoryManager.java
```diff
@@ -175,6 +177,48 @@ public Deployment readDeploymentMetadata() throws IOException {
...
+     * Read persisted source IoT endpoint values.
+     *
+     * @return the endpoint switch metadata
+     * @throws IOException on I/O error
+     */
+    public EndpointSwitchMetadata readSourceEndpoints() throws IOException {
+        Path filePath = getDeploymentDirectoryPath().resolve(ENDPOINT_SWITCH_METADATA_FILE);
+        AtomicReference<EndpointSwitchMetadata> ref = new AtomicReference<>();
+        CommitableReader.of(filePath).read(in -> {
+            ref.set(SerializerFactory.getFailSafeJsonObjectMapper().readValue(in, EndpointSwitchMetadata.class));
```
MikeDombo: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

sahith:
The addition of `deploymentDirectoryManager` to `DeploymentConfigMerger` breaks the integration tests.

src/main/java/com/aws/greengrass/deployment/activator/DefaultActivator.java
```diff
@@ -180,6 +182,15 @@ void rollback(DeploymentDocument deploymentDocument, CompletableFuture<Deploymen
         }
     }
 
+    private boolean isEndpointSwitchDeployment() {
+        try {
+            return deploymentDirectoryManager.hasEndpointSwitchMetadata();
+        } catch (IOException e) {
+            logger.atWarn().setCause(e).log("Failed to check endpoint switch metadata");
+            return false;
```
sahith: IOException is being swallowed here. If the metadata file was persisted but can't be read (broken symlink, disk error), the deployment silently loses rollback protection — no snapshot, no rollback on failure. The device could get stuck with an unreachable endpoint and no way to recover. Consider letting the IOException propagate and failing the deployment with `FAILED_NO_STATE_CHANGE` instead, since no config has been applied yet.

sahith (dismissed):

src/main/java/com/aws/greengrass/deployment/DeploymentDirectoryManager.java
Vipul-5517: Updated

src/main/java/com/aws/greengrass/deployment/activator/DefaultActivator.java
Vipul-5517: Make sense. Updated

src/main/java/com/aws/greengrass/deployment/activator/DefaultActivator.java
```diff
@@ -49,7 +50,22 @@ public void activate(Map<String, Object> newConfig, Deployment deployment, long
...
+            logger.atError(MERGE_CONFIG_EVENT_KEY).setCause(e)
+                    .log("Failed to check endpoint switch metadata");
+            totallyCompleteFuture.complete(
+                    new DeploymentResult(DeploymentResult.DeploymentStatus.FAILED_NO_STATE_CHANGE, e));
+            return;
+        }
+        if (endpointSwitch) {
+            logger.atInfo(MERGE_CONFIG_EVENT_KEY)
+                    .log("Endpoint switch deployment: forcing config snapshot and rollback support");
+        }
```
sahith: This logic is essentially to trigger a rollback on endpoint switch deployment irrespective of auto rollback failure handling policy configured (`ROLLBACK` or `DO_NOTHING`). We can add a comment clarifying it. Also, how about executing this logic only if `isAutoRollbackRequested(deploymentDocument)` is `false` so that it doesn't get triggered on every endpoint switch deployment

src/main/java/com/aws/greengrass/deployment/activator/DefaultActivator.java
```diff
@@ -180,6 +198,22 @@ void rollback(DeploymentDocument deploymentDocument, CompletableFuture<Deploymen
...
     }
 
+    private boolean isEndpointSwitchDeployment() throws IOException {
+        return deploymentDirectoryManager.hasEndpointSwitchMetadata();
+    }
+
+    /**
+     * Safe variant for use after config is applied — defaults to true (attempt rollback) on IO error.
+     */
+    private boolean isEndpointSwitchSafe() {
```
sahith: Why have a separate method `isEndpointSwitchSafe` instead of re-using `isEndpointSwitchDeployment`?

src/main/java/com/aws/greengrass/deployment/DeploymentDirectoryManager.java
```diff
@@ -35,6 +36,7 @@ public class DeploymentDirectoryManager {
     static final String BOOTSTRAP_TASK_FILE = "bootstrap_task.json";
     static final String ROLLBACK_BOOTSTRAP_TASK_FILE = "rollback_bootstrap_task.json";
     static final String DEPLOYMENT_METADATA_FILE = "deployment_metadata.json";
+    static final String ENDPOINT_SWITCH_METADATA_FILE = "endpoint_switch_metadata.json";
```
saranyailla: Have we considered writing source endpoint info to config.tlog?

alter-mage:
I am concerned if we are introducing race conditions through some edge cases. I am looking at scenarios where
- a deployment gets cancelled
- Nucleus dies due to a power loss while processing a deployment

Let's talk about situations where a bad endpoint change gets deployed. Specifically, Nucleus loses all connectivity if it is configured to use this endpoint. That is, its bricked.

For the cancellation scenario, have we considered implementing a forced rollback if Nucleus cancels a deployment between config update and waiting for services to start?
Reason being, Nucleus could cancel a deployment as soon as it receives the cancellation notice. It does not block on service lifecycle threads from ongoing deployment. If a bad endpoint change gets committed to the Nucleus config and deployment is cancelled before this commit can be verified, we are at risk of bricking the device.

For the power loss scenario, have we considered what happens if rollback snapshot is stale?
Reason being, a customer's device could suffer a power loss at any point, including a situation where a bad endpoint change is committed to the Nucleus config but not verified. Nucleus boots up with this bad change and loses all connectivity, getting bricked. I believe a similar situation can occur during rollback as well.

src/main/java/com/aws/greengrass/deployment/DeploymentDirectoryManager.java
sahith: Discussed it offline. There is a room for error if another file is being introduced to be managed by Greengrass Nucleus. Instead, we can have the source IoT data endpoint written to runtime configuration within `config.tlog` file

sahith:
> I am concerned if we are introducing race conditions through some edge cases. I am looking at scenarios where
> 
> * a deployment gets cancelled
> * Nucleus dies due to a power loss while processing a deployment
> 
> Let's talk about situations where a bad endpoint change gets deployed. Specifically, Nucleus loses all connectivity if it is configured to use this endpoint. That is, its bricked.
> 
> For the cancellation scenario, have we considered implementing a forced rollback if Nucleus cancels a deployment between config update and waiting for services to start? Reason being, Nucleus could cancel a deployment as soon as it receives the cancellation notice. It does not block on service lifecycle threads from ongoing deployment. If a bad endpoint change gets committed to the Nucleus config and deployment is cancelled before this commit can be verified, we are at risk of bricking the device.
> 
> For the power loss scenario, have we considered what happens if rollback snapshot is stale? Reason being, a customer's device could suffer a power loss at any point, including a situation where a bad endpoint change is committed to the Nucleus config but not verified. Nucleus boots up with this bad change and loses all connectivity, getting bricked. I believe a similar situation can occur during rollback as well.

Discussed this offline. The risk of bad IoT data endpoint being deployed is offset by having the pre-flight MQTT connectivity check to verify if the device is successfully able to establish the connectivity.

src/main/java/com/aws/greengrass/deployment/DeploymentDirectoryManager.java
Vipul-5517: Addressed in the new revision

src/main/java/com/aws/greengrass/deployment/activator/DefaultActivator.java
Vipul-5517: Addressed in the new revision

src/main/java/com/aws/greengrass/deployment/activator/DefaultActivator.java
Vipul-5517: Addressed in the new revision

src/main/java/com/aws/greengrass/deployment/activator/DefaultActivator.java
```diff
@@ -49,7 +50,18 @@ public void activate(Map<String, Object> newConfig, Deployment deployment, long
         }
 
         DeploymentDocument deploymentDocument = deployment.getDeploymentDocumentObj();
-        if (isAutoRollbackRequested(deploymentDocument) && !takeConfigSnapshot(totallyCompleteFuture)) {
+        boolean autoRollback = isAutoRollbackRequested(deploymentDocument);
+
+        // Endpoint switch deployments force config snapshot and rollback regardless of FailureHandlingPolicy,
+        // so that we can restore the original endpoint if the new one is unreachable after applying the change.
+        // Only check when auto-rollback is not already requested to avoid redundant metadata reads.
+        boolean endpointSwitch = !autoRollback && isEndpointSwitch();
```
sahith: If autoRollback is set to `true` and isEndpointSwitch() is false, then essentially endpointSwitch variable would be set to `false`. This would be confusing. 

The main purpose of defining this variable is to know whether config snapshot should be taken or not. We can name it accordingly instead of `endpointSwitch`

Vipul-5517: updated

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
```diff
@@ -154,6 +155,22 @@ private void updateActionForDeployment(Map<String, Object> newConfig, Deployment
             return;
         }
 
+        // Persist source endpoint before activation for endpoint-switch deployments.
+        // The source endpoint is stored in config.tlog under a distinct key (sourceIotDataEndpoint)
+        // so the MQTT BatchedSubscriber does not trigger reconnection. Step 5 uses this value to
+        // report deployment status back to the source account. clearSourceIotDataEndpoint() is
+        // called after status reporting; if cleanup is skipped (crash, Step 5 not yet implemented),
```
sahith: This is talking about the future implementation steps. We can update the comment to not have any step details

Vipul-5517: this is transient state so should be fine to keep it here as before final merge this will be removed

src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java
```diff
@@ -476,6 +477,34 @@ public Topic getIotCredentialEndpoint() {
...
+
+    /**
+     * Remove the persisted source IoT data endpoint after deployment completes.
+     */
+    public void clearSourceIotDataEndpoint() {
+        Topic t = config.find(DEVICE_PARAM_SOURCE_IOT_DATA_ENDPOINT);
+        if (t != null) {
+            t.remove();
+        }
+    }
```
sahith: `sourceIotDataEndpoint` is internal deployment state, not a customer-facing Nucleus config parameter. It should live under `services.DeploymentService.runtime` — persisted to config.tlog so it survives device crashes mid-endpoint-switch, and not customer-configurable.

src/main/java/com/aws/greengrass/deployment/activator/DefaultActivator.java
```diff
@@ -5,14 +5,17 @@
 
 package com.aws.greengrass.deployment.activator;
 
+import com.aws.greengrass.config.Topic;
 import com.aws.greengrass.deployment.DeploymentConfigMerger;
+import com.aws.greengrass.deployment.DeploymentService;
```
alter-mage, line 10: Please do not import the service class, this may cause unnecessary object injections.

Please use `kernel.getContext().get(DeploymentService.class)` to load the object in-place.

Vipul-5517: kernel.getContext().get(DeploymentService.class) still requires import com.aws.greengrass.deployment.DeploymentService, or else I need to do kernel.findServiceTopic("DeploymentService") which will introduce a magic string which is more risky at compile time

alter-mage: Oh right, that makes sense. I hallucinated there for a bit.

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
```diff
@@ -60,6 +61,7 @@ public class DeploymentConfigMerger {
     public static final String MERGE_ERROR_LOG_EVENT_KEY = "config-update-error";
     public static final String DEPLOYMENT_ID_LOG_KEY = "deploymentId";
     public static final String SERVICE_NAME_LOG_KEY = "serviceName";
+    public static final String SOURCE_IOT_DATA_ENDPOINT_KEY = "sourceIotDataEndpoint";
```
alter-mage, line 64: nit: the data endpoint is a device configuration and not related to the deployment configuration merger. Please move this to the [device configuration](https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/main/src/main/java/com/aws/greengrass/deployment/DeviceConfiguration.java) class.

alter-mage (approved):
Left a nitpicking comment but the overall PR looks good to me

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
Medivh0416, line 64: nit: do we need sourceIotCredEndpoint?

Medivh0416 (approved):

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
Vipul-5517: We only persist the data endpoint — the credential endpoint isn't needed for rollback or status reporting.

Vipul-5517: For Step 1, DeploymentConfigMerger is where the constant is used (it writes the value), so having it there isn't wrong — it's just not the ideal long-term home. In Step 2, when we introduce the pre-flight
validation class, the constant naturally moves there since that class will both write and read it.

PR #1781, opened 2026-04-14
feat: standalone MQTT connection and pre-flight connectivity check

Vipul-5517:
### Summary

Adds pre-flight MQTT connectivity verification for endpoint-switch deployments. Before applying a new IoT endpoint configuration, the deployment service creates a temporary MQTT connection to the target endpoint to verify reachability. If pre-flight fails, the deployment reports `FAILED_NO_STATE_CHANGE` — no device state is modified.

Also addresses Step 1 PR #1776 review comments from Siddhant and Mitchell.

This is Step 2 of the IoT Endpoint Switch design.

### Changes

- **StandaloneMqttConnection** — New reusable standalone MQTT connection class in `mqttclient` package. Supports connect with retries/exponential backoff, publish, and close. Maps CRT exceptions to deployment error codes (`MQTT_CONNECTION_FAILED`, `TLS_HANDSHAKE_FAILURE`, `MISSING_MQTT_CONNECT_POLICY`). Not thread-safe — designed for single-use one-shot operations.
- **DeploymentConfigMerger** — Add `verifyMqttConnectivity()` pre-flight check using device certs. Respects proxy config (same pattern as MqttClient with noProxy filtering). Configurable timeout via `mqtt.standaloneMqttTimeoutMs` (default 60s). Pre-flight runs BEFORE persisting source endpoint so failure leaves zero state change.
- **DeploymentService** — `SOURCE_IOT_DATA_ENDPOINT_KEY` constant moved here from `DeploymentConfigMerger` (Siddhant review: constant belongs with the service that owns the runtime config).
- **DefaultActivator** — `isEndpointSwitch()` updated to reference `DeploymentService.SOURCE_IOT_DATA_ENDPOINT_KEY`.
- **DeploymentErrorCode** — 3 new error codes: `MQTT_CONNECTION_FAILED` (NETWORK_ERROR), `TLS_HANDSHAKE_FAILURE` (NETWORK_ERROR), `MISSING_MQTT_CONNECT_POLICY` (PERMISSION_ERROR).
- **README_CONFIG_SCHEMA.md** — Document `mqtt.standaloneMqttTimeoutMs` config key.

### Step 1 PR Review Comments Addressed

- Siddhant: Moved `SOURCE_IOT_DATA_ENDPOINT_KEY` from `DeploymentConfigMerger` to `DeploymentService`
- Mitchell: Confirmed we only persist `sourceIotDataEndpoint`, not `sourceIotCredEndpoint`

### E2E Test Results (22/22 pass)

Tested with locally built Nucleus 2.16.0-SNAPSHOT in Podman containers, fresh device per test.

| # | Scenario | Result | Status | Config Changed |
|---|---|---|---|---|
| 1 | Happy path us-east-1 → us-west-2 | ✅ | SUCCEEDED | Yes → us-west-2 |
| 2 | Reverse us-west-2 → us-east-1 | ✅ | SUCCEEDED | Yes → us-east-1 |
| 3 | Non-ATS endpoint fails at pre-flight DNS | ✅ | FAILED_NO_STATE_CHANGE | No |
| 4 | Unreachable fake endpoint | ✅ | FAILED_NO_STATE_CHANGE | No |
| 5 | DO_NOTHING policy forces snapshot | ✅ | SUCCEEDED | Yes → us-west-2 |
| 6 | Back-to-back switch (W2→E1→W2) | ✅ | SUCCEEDED ×2 | Yes (both ways) |
| 7 | Non-endpoint deployment unaffected | ✅ | SUCCEEDED | No |
| 8 | Endpoint + component config change | ✅ | SUCCEEDED | Yes → us-west-2 |
| 9 | Cred-only change not detected as switch | ✅ | SUCCEEDED | No |
| 10 | Same endpoint redeployed not detected | ✅ | SUCCEEDED | No |
| 11 | sourceIotDataEndpoint persisted in config.tlog | ✅ | SUCCEEDED | Yes → us-west-2 |
| 12 | Empty iotDataEndpoint is no-op | ✅ | SUCCEEDED | No |
| 13 | Malformed URL device unchanged | ✅ | SUCCEEDED | No |
| 14 | Cross-region cert in both regions | ✅ | SUCCEEDED | Yes → us-west-2 |
| 15 | Pre-flight: unreachable endpoint | ✅ | FAILED_NO_STATE_CHANGE | No |
| 16 | Pre-flight: cert not registered (validation catches) | ✅ | FAILED_NO_STATE_CHANGE | No |
| 17 | Custom standaloneMqttTimeoutMs=5s | ✅ | FAILED_NO_STATE_CHANGE | No |
| 18 | Cloud-side FAILED status visible | ✅ | FAILED | No |
| 19 | Custom MQTT port 443 | ✅ | SUCCEEDED | Yes → us-west-2 |
| 20 | Bad component fails at resolution before pre-flight | ✅ | FAILED_NO_STATE_CHANGE | No |
| 21 | Concurrent deployment queued correctly | ✅ | SUCCEEDED | No |
| 22 | Network disconnect mid-pre-flight | ✅ | FAILED_NO_STATE_CHANGE | No |

### Testing

- Unit tests: 38/38 pass (StandaloneMqttConnectionTest 13, DeploymentConfigMergerTest 19, DefaultActivatorTest 3, DeploymentErrorCodeTest 3)
- Checkstyle clean, japicmp clean (no backward compatibility breaks)
- E2E: 22/22 pass (Podman, fresh device per test, account 986036454283)

### Related

- Step 1: PR #1776 (merged)
- Step 3 (credential endpoint validation) will follow in a separate PR
- Design: [Detailed Design — Section F.1 Step 4](https://quip-amazon.com/B49kAM8Mb4xI)

### DIFF
--- a/README_CONFIG_SCHEMA.md
+++ b/README_CONFIG_SCHEMA.md
@@ -177,6 +177,7 @@ services:
         pingTimeoutMs: 30000
         operationTimeoutMs: 30000
         maxInFlightPublishes: 5
+        standaloneMqttTimeoutMs: 60000
         spooler:
           keepQos0WhenOffline: false
           maxSizeInBytes: 2621440
@@ -195,6 +196,22 @@ services:
         periodicAggregateMetricsIntervalSeconds: 3600
         periodicPublishMetricsIntervalSeconds: 86400
 ```
+
+### Endpoint Switch Configuration
+
+`mqtt.standaloneMqttTimeoutMs` (default: `60000`) — Timeout budget in milliseconds for the pre-flight
+MQTT connectivity check during endpoint-switch deployments. Controls the number of connection attempts
+(timeoutMs / 10s per attempt) with exponential backoff between retries. Actual elapsed time may exceed
+this value due to backoff sleep between attempts. Values ≤ 0 are ignored and the default is used.
+
+**IoT policy requirement:** The pre-flight check connects to the target endpoint using the MQTT client ID
+`<thingName>#endpoint-switch`. If your IoT policy restricts `iot:Connect` to an exact thing name match
+(e.g., `client/${iot:Connection.Thing.ThingName}`), you must update it to allow the suffix pattern:
+```
+"Resource": "arn:aws:iot:*:*:client/${iot:Connection.Thing.ThingName}*"
+```
+This is the same wildcard pattern required by IoT Device Shadow and IoT Jobs clients. Without this,
+endpoint-switch deployments will fail with `MISSING_MQTT_CONNECT_POLICY`.
 
 Setting a custom path to relocate the $GG_ROOT/ipc.socket to another location
  on the filesystem, and it doesn't apply to Windows.
--- a/src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
+++ b/src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
@@ -15,6 +15,7 @@ import com.aws.greengrass.deployment.activator.KernelUpdateActivator;
 import com.aws.greengrass.deployment.errorcode.DeploymentErrorCode;
 import com.aws.greengrass.deployment.errorcode.DeploymentErrorCodeUtils;
 import com.aws.greengrass.deployment.exceptions.ComponentConfigurationValidationException;
+import com.aws.greengrass.deployment.exceptions.DeploymentException;
 import com.aws.greengrass.deployment.exceptions.ServiceUpdateException;
 import com.aws.greengrass.deployment.model.Deployment;
 import com.aws.greengrass.deployment.model.DeploymentDocument;
@@ -27,6 +28,9 @@ import com.aws.greengrass.lifecyclemanager.UpdateSystemPolicyService;
 import com.aws.greengrass.lifecyclemanager.exceptions.ServiceLoadException;
 import com.aws.greengrass.logging.api.Logger;
 import com.aws.greengrass.logging.impl.LogManager;
+import com.aws.greengrass.mqttclient.StandaloneMqttConnector;
+import com.aws.greengrass.security.SecurityService;
+import com.aws.greengrass.security.exceptions.MqttConnectionProviderException;
 import com.aws.greengrass.util.Coerce;
 import lombok.AccessLevel;
 import lombok.AllArgsConstructor;
@@ -61,8 +65,11 @@ public class DeploymentConfigMerger {
     public static final String MERGE_ERROR_LOG_EVENT_KEY = "config-update-error";
     public static final String DEPLOYMENT_ID_LOG_KEY = "deploymentId";
     public static final String SERVICE_NAME_LOG_KEY = "serviceName";
-    public static final String SOURCE_IOT_DATA_ENDPOINT_KEY = "sourceIotDataEndpoint";
     protected static final int WAIT_SVC_START_POLL_INTERVAL_MILLISEC = 1000;
+    static final long DEFAULT_PREFLIGHT_MQTT_TIMEOUT_MS = 60_000;
+    static final String STANDALONE_MQTT_TIMEOUT_KEY = "standaloneMqttTimeoutMs";
+    private static final String ENDPOINT_LOG_KEY = "endpoint";
+    private static final String ENDPOINT_SWITCH_CLIENT_SUFFIX = "#endpoint-switch";
 
     private static final Logger logger = LogManager.getLogger(DeploymentConfigMerger.class);
 
@@ -156,9 +163,10 @@ public class DeploymentConfigMerger {
             return;
         }
 
-        // Persist source endpoint before activation for endpoint-switch deployments.
-        // Stored under services.DeploymentService.runtime (internal deployment state, not customer config).
-        // Persisted to config.tlog so it survives device crashes mid-endpoint-switch.
+        // Endpoint-switch deployments: pre-flight connectivity check, then persist source endpoint.
+        // Pre-flight runs BEFORE persisting so that on failure, no device state has changed.
+        // Source endpoint is stored under services.DeploymentService.runtime (internal deployment state,
+        // not customer config). Persisted to config.tlog so it survives device crashes mid-endpoint-switch.
         // Step 5 uses this value to report deployment status back to the source account.
         // Cleared after status reporting; stale keys are harmless and overwritten by the next switch.
         if (isEndpointSwitchDeployment(nucleusConfig, deviceConfiguration)) {
@@ -167,9 +175,31 @@ public class DeploymentConfigMerger {
             logger.atInfo().setEventType(MERGE_CONFIG_EVENT_KEY)
                     .kv("currentIotDataEndpoint", currentDataEndpoint)
                     .kv("newIotDataEndpoint", newDataEndpoint)
-                    .log("Endpoint switch deployment detected, persisting source endpoint for rollback");
+                    .log("Endpoint switch deployment detected");
+
+            // Pre-flight MQTT connectivity check to new endpoint
+            logger.atInfo().setEventType(MERGE_CONFIG_EVENT_KEY)
+                    .kv(ENDPOINT_LOG_KEY, newDataEndpoint)
+                    .log("Starting pre-flight MQTT connectivity check");
+            long preflightTimeout = Coerce.toLong(deviceConfiguration.getMQTTNamespace()
+                    .findOrDefault(DEFAULT_PREFLIGHT_MQTT_TIMEOUT_MS, STANDALONE_MQTT_TIMEOUT_KEY));
+            if (preflightTimeout <= 0) {
+                logger.atWarn().kv("configured", preflightTimeout)
+                        .kv("using", DEFAULT_PREFLIGHT_MQTT_TIMEOUT_MS)
+                        .log("Invalid standaloneMqttTimeoutMs, using default");
+                preflightTimeout = DEFAULT_PREFLIGHT_MQTT_TIMEOUT_MS;
+            }
+            if (!verifyMqttConnectivity(totallyCompleteFuture, newDataEndpoint, ENDPOINT_SWITCH_CLIENT_SUFFIX,
+                    preflightTimeout)) {
+                return;
+            }
+            logger.atInfo().setEventType(MERGE_CONFIG_EVENT_KEY)
+                    .kv(ENDPOINT_LOG_KEY, newDataEndpoint)
+                    .log("Pre-flight MQTT connectivity check passed");
+
+            // Persist source endpoint only after pre-flight succeeds — no state change on failure
             kernel.getContext().get(DeploymentService.class).getRuntimeConfig()
-                    .lookup(SOURCE_IOT_DATA_ENDPOINT_KEY).withValue(currentDataEndpoint);
+                    .lookup(DeploymentService.SOURCE_IOT_DATA_ENDPOINT_KEY).withValue(currentDataEndpoint);
         }
 
         logger.atInfo(MERGE_CONFIG_EVENT_KEY).kv("deployment", deploymentId)
@@ -193,6 +223,55 @@ public class DeploymentConfigMerger {
             }
         }
         return true;
+    }
+
+    @SuppressWarnings({"PMD.AvoidCatchingGenericException"})
+    @edu.umd.cs.findbugs.annotations.SuppressFBWarnings(value = "REC_CATCH_EXCEPTION",
+            justification = "Defensive catch for unexpected errors from CRT builder and proxy config")
+    boolean verifyMqttConnectivity(CompletableFuture<DeploymentResult> totallyCompleteFuture,
+                                   String endpoint, String clientIdSuffix, long timeoutMs) {
+        try (StandaloneMqttConnector conn = StandaloneMqttConnector.of(
+                kernel.getContext().get(SecurityService.class), deviceConfiguration,
+                endpoint, clientIdSuffix)) {
+            conn.connect(timeoutMs);
+            return true;
+        } catch (MqttConnectionProviderException e) {
+            logger.atError().setEventType(MERGE_ERROR_LOG_EVENT_KEY).setCause(e)
+                    .kv(ENDPOINT_LOG_KEY, endpoint)
+                    .log("Pre-flight MQTT connectivity check failed: device identity configuration error");
+            totallyCompleteFuture.complete(
+                    new DeploymentResult(DeploymentResult.DeploymentStatus.FAILED_NO_STATE_CHANGE,
+                            new DeploymentException("Device identity configuration error", e,
+                                    DeploymentErrorCode.TLS_HANDSHAKE_FAILURE)));
+            return false;
+        } catch (DeploymentException e) {
+            if (e.getErrorCodes().contains(DeploymentErrorCode.MISSING_MQTT_CONNECT_POLICY)) {
+                String thingName = Coerce.toString(deviceConfiguration.getThingName());
+                logger.atError().setEventType(MERGE_ERROR_LOG_EVENT_KEY).setCause(e)
+                        .kv(ENDPOINT_LOG_KEY, endpoint)
+                        .log("Pre-flight MQTT connectivity check failed: MQTT CONNECT rejected. "
+                                + "Possible causes: (1) IoT policy does not allow client ID \""
+                                + thingName + clientIdSuffix + "\" — update iot:Connect resource to "
+                                + "\"arn:aws:iot:*:*:client/" + thingName + "*\", or "
+                                + "(2) device certificate is not authorized in the target account");
+            } else {
+                logger.atError().setEventType(MERGE_ERROR_LOG_EVENT_KEY).setCause(e)
+                        .kv(ENDPOINT_LOG_KEY, endpoint)
+                        .log("Pre-flight MQTT connectivity check failed");
+            }
+            totallyCompleteFuture.complete(
+                    new DeploymentResult(DeploymentResult.DeploymentStatus.FAILED_NO_STATE_CHANGE, e));
+            return false;
+        } catch (Exception e) {
+            logger.atError().setEventType(MERGE_ERROR_LOG_EVENT_KEY).setCause(e)
+                    .kv(ENDPOINT_LOG_KEY, endpoint)
+                    .log("Pre-flight MQTT connectivity check failed");
+            totallyCompleteFuture.complete(
+                    new DeploymentResult(DeploymentResult.DeploymentStatus.FAILED_NO_STATE_CHANGE,
+                            new DeploymentException("MQTT pre-flight connection to " + endpoint + " failed",
+                                    e, DeploymentErrorCode.MQTT_CONNECTION_FAILED)));
+            return false;
+        }
     }
 
     /**
--- a/src/main/java/com/aws/greengrass/deployment/errorcode/DeploymentErrorCode.java
+++ b/src/main/java/com/aws/greengrass/deployment/errorcode/DeploymentErrorCode.java
@@ -41,6 +41,9 @@ public enum DeploymentErrorCode {
 
     /* Network / http */
     NETWORK_ERROR(DeploymentErrorType.NETWORK_ERROR),
+    MQTT_CONNECTION_FAILED(DeploymentErrorType.NETWORK_ERROR),
+    TLS_HANDSHAKE_FAILURE(DeploymentErrorType.NETWORK_ERROR),
+    MISSING_MQTT_CONNECT_POLICY(DeploymentErrorType.PERMISSION_ERROR),
     HTTP_REQUEST_ERROR(DeploymentErrorType.HTTP_ERROR),
     DOWNLOAD_DEPLOYMENT_DOCUMENT_ERROR(DeploymentErrorType.HTTP_ERROR),
     GET_GREENGRASS_ARTIFACT_SIZE_ERROR(DeploymentErrorType.HTTP_ERROR),
--- a/src/main/java/com/aws/greengrass/mqttclient/StandaloneMqttConnector.java
+++ b/src/main/java/com/aws/greengrass/mqttclient/StandaloneMqttConnector.java
@@ -1,0 +1,206 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.aws.greengrass.mqttclient;
+
+import com.aws.greengrass.deployment.DeviceConfiguration;
+import com.aws.greengrass.deployment.errorcode.DeploymentErrorCode;
+import com.aws.greengrass.deployment.exceptions.DeploymentException;
+import com.aws.greengrass.logging.api.Logger;
+import com.aws.greengrass.logging.impl.LogManager;
+import com.aws.greengrass.security.SecurityService;
+import com.aws.greengrass.security.exceptions.MqttConnectionProviderException;
+import com.aws.greengrass.util.Coerce;
+import com.aws.greengrass.util.ProxyUtils;
+import com.aws.greengrass.util.RetryUtils;
+import com.aws.greengrass.util.Utils;
+import software.amazon.awssdk.crt.http.HttpProxyOptions;
+import software.amazon.awssdk.crt.mqtt.MqttClientConnection;
+import software.amazon.awssdk.iot.AwsIotMqttConnectionBuilder;
+
+import java.io.Closeable;
+import java.time.Duration;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Locale;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * A standalone MQTT connection for one-shot operations (pre-flight checks, status reporting)
+ * independent of the main MqttClient. Not thread-safe — callers must not share instances across threads.
+ */
+@SuppressWarnings({"PMD.NullAssignment"})
+public class StandaloneMqttConnector implements Closeable {
+    private static final Logger logger = LogManager.getLogger(StandaloneMqttConnector.class);
+    private static final long DISCONNECT_TIMEOUT_MS = 5000;
+    private static final long PER_ATTEMPT_TIMEOUT_MS = 10_000;
+
+    private final AwsIotMqttConnectionBuilder builder;
+    private final String clientId;
+    private MqttClientConnection connection;
+
+    /**
+     * Create a standalone MQTT connection.
+     *
+     * @param builder  configured MQTT connection builder (with certs, CA, endpoint)
+     * @param clientId MQTT client ID
+     */
+    public StandaloneMqttConnector(AwsIotMqttConnectionBuilder builder, String clientId) {
+        this.builder = builder;
+        this.clientId = clientId;
+    }
+
+    /**
+     * Build a StandaloneMqttConnector configured with device identity, proxy, and the given endpoint.
+     * Extracts builder setup from the deployment layer so callers don't need to know about proxy/TLS details.
+     *
+     * <p>Proxy options are built directly rather than via {@code ProxyUtils.getHttpProxyOptions()} because
+     * that method requires a {@code @NonNull ClientTlsContext} for HTTPS proxy tunneling. Standalone
+     * connections use HTTP CONNECT tunneling (the common case for MQTT) which works without a TLS context.</p>
+     *
+     * @param securityService     provides the MQTT connection builder with device certs
+     * @param deviceConfiguration provides endpoint, proxy, port, and CA configuration
+     * @param endpoint            the IoT data endpoint to connect to
+     * @param clientIdSuffix      suffix appended to thing name for the MQTT client ID
+     * @return a configured StandaloneMqttConnector ready to call {@link #connect(long)}
+     * @throws MqttConnectionProviderException if the device identity builder cannot be created
+     */
+    @SuppressWarnings("PMD.AvoidCatchingGenericException")
+    public static StandaloneMqttConnector of(SecurityService securityService,
+                                              DeviceConfiguration deviceConfiguration,
+                                              String endpoint, String clientIdSuffix)
+            throws MqttConnectionProviderException {
+        String thingName = Coerce.toString(deviceConfiguration.getThingName());
+        String clientId = thingName + clientIdSuffix;
+        AwsIotMqttConnectionBuilder mqttBuilder = securityService.getDeviceIdentityMqttConnectionBuilder();
+        int mqttPort = Coerce.toInt(deviceConfiguration.getMQTTNamespace()
+                .findOrDefault(MqttClient.DEFAULT_MQTT_PORT, MqttClient.MQTT_PORT_KEY));
+        mqttBuilder.withEndpoint(endpoint)
+                .withCertificateAuthorityFromPath(null,
+                        Coerce.toString(deviceConfiguration.getRootCAFilePath()))
+                .withPort((short) mqttPort)
+                .withCleanSession(true);
+        configureProxy(mqttBuilder, deviceConfiguration, endpoint);
+        return new StandaloneMqttConnector(mqttBuilder, clientId);
+    }
+
+    private static void configureProxy(AwsIotMqttConnectionBuilder mqttBuilder,
+                                       DeviceConfiguration deviceConfiguration, String endpoint) {
+        String proxyUrl = deviceConfiguration.getProxyUrl();
+        if (Utils.isEmpty(proxyUrl)) {
+            return;
+        }
+        HttpProxyOptions httpProxyOptions = new HttpProxyOptions();
+        httpProxyOptions.setHost(ProxyUtils.getHostFromProxyUrl(proxyUrl));
+        httpProxyOptions.setPort(ProxyUtils.getPortFromProxyUrl(proxyUrl));
+        httpProxyOptions.setConnectionType(HttpProxyOptions.HttpProxyConnectionType.Tunneling);
+        String proxyUsername = deviceConfiguration.getProxyUsername();
+        if (Utils.isNotEmpty(proxyUsername)) {
+            httpProxyOptions.setAuthorizationType(HttpProxyOptions.HttpProxyAuthorizationType.Basic);
+            httpProxyOptions.setAuthorizationUsername(proxyUsername);
+            httpProxyOptions.setAuthorizationPassword(deviceConfiguration.getProxyPassword());
+        }
+        String noProxy = Coerce.toString(deviceConfiguration.getNoProxyAddresses());
+        if (Utils.isNotEmpty(noProxy) && Utils.isNotEmpty(endpoint)
+                && Arrays.stream(noProxy.split(",")).anyMatch(endpoint::matches)) {
+            return;
+        }
+        mqttBuilder.withHttpProxyOptions(httpProxyOptions);
+    }
+
+    /**
+     * Connect to the MQTT endpoint with retries using exponential backoff.
+     * The timeout controls the number of attempts (timeoutMs / perAttemptTimeout), not a hard wall-clock
+     * deadline — actual elapsed time may exceed timeoutMs due to backoff sleep between attempts.
+     *
+     * <p>All exceptions are retryable to support JITR/JITP provisioning flows where the first connection
+     * attempt fails and triggers provisioning, then subsequent retries succeed.</p>
+     *
+     * @param timeoutMs timeout budget in milliseconds used to derive the number of connection attempts
+     * @throws DeploymentException with appropriate error code if all attempts fail
+     */
+    @SuppressWarnings("PMD.AvoidCatchingGenericException")
+    public void connect(long timeoutMs) throws DeploymentException {
+        if (timeoutMs <= 0) {
+            throw new DeploymentException("MQTT connection timeout must be positive, got " + timeoutMs,
+                    DeploymentErrorCode.MQTT_CONNECTION_FAILED);
+        }
+        long perAttemptTimeout = Math.min(timeoutMs, PER_ATTEMPT_TIMEOUT_MS);
+        int maxAttempts = Math.max(1, (int) (timeoutMs / perAttemptTimeout));
+        RetryUtils.RetryConfig retryConfig = RetryUtils.RetryConfig.builder()
+                .initialRetryInterval(Duration.ofSeconds(1))
+                .maxRetryInterval(Duration.ofSeconds(10))
+                .maxAttempt(maxAttempts)
+                // All exceptions are retryable — supports JITR/JITP where first connect triggers provisioning
+                .retryableExceptions(Collections.singletonList(Exception.class))
+                .build();
+        try {
+            RetryUtils.runWithRetry(retryConfig, () -> {
+                connectionCleanup();
+                builder.withClientId(clientId);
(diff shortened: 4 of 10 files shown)


src/main/java/com/aws/greengrass/mqttclient/StandaloneMqttConnection.java
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    static DeploymentErrorCode mapExceptionToErrorCode(Throwable cause) {
+        if (cause == null) {
+            return DeploymentErrorCode.MQTT_CONNECTION_FAILED;
+        }
+        // Unwrap ExecutionException from CompletableFuture.get()
+        Throwable actual = cause instanceof java.util.concurrent.ExecutionException && cause.getCause() != null
+                ? cause.getCause() : cause;
+        String message = actual.getMessage() == null ? "" : actual.getMessage().toLowerCase();
```
MikeDombo, line 146: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

#### toLowerCase [](|2045d6a2-68fa-4cde-87a6-566ffddb2b76|)

The parameterless versions of `String.toLowerCase()` and `String.toUpperCase()` use the default locale of the JVM when transforming strings. This can have unintended results, including difficult-to-debug transient failures because case mapping differs based on locale. 

**Even if you aren’t writing code for a user experience that’s displayed to a customer**, system internal code often uses case folding to normalize strings for comparison.
Not doing so can missing locale lead to errors when running on systems that are configured to use a different system locale. This can include developer desktops or build servers (not just your fleet hosts): many hours have been spent debugging errors that only occur on a specific machine due to its locale configuration.

#### Recommended solutions:

Always pass in a `Locale`.

**Language or locale-dependent processing:**

Get the locale from the platform, use the language of the content, or configure it in your code and pass that to the method.
```
String firstName = &#34;Ichabod&#34;; 
Locale locale = // Where you obtain the locale from is platform dependent
String lowerCaseFirstName = firstName.toLowerCase(locale);
``` 

**Internal processing that isn’t locale-dependent:**

*Example: S3 bucket name*

Always specify the language/country-neutral locale `Locale.ROOT`. 

```
String imgTag = &#34;IMG&#34;; 
if (&#34;img&#34;.equals(imgTag.toLowerCase(Locale.ROOT)) {    
// do something 
}
```

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
```diff
@@ -195,6 +222,57 @@ private boolean validateNucleusConfig(CompletableFuture<DeploymentResult> totall
...
+                        mqttBuilder.withHttpProxyOptions(httpProxyOptions);
+                    }
+                }
+            }
+            // Builder is created fresh above — safe to mutate withClientId/withEndpoint per retry.
+            try (StandaloneMqttConnection conn = new StandaloneMqttConnection(mqttBuilder, clientId)) {
+                conn.connect(timeoutMs);
+            }
+            return true;
+        } catch (Exception e) {
```
MikeDombo, line 279: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

src/main/java/com/aws/greengrass/mqttclient/StandaloneMqttConnection.java
```diff
@@ -0,0 +1,157 @@
...
+                connection.connect().get(perAttemptTimeout, TimeUnit.MILLISECONDS);
+                logger.atInfo().kv("clientId", clientId).log("Standalone MQTT connection established");
+                return null;
+            }, "standalone-mqtt-connect", logger);
+        } catch (InterruptedException e) {
+            connectionCleanup();
+            Thread.currentThread().interrupt();
+            throw new DeploymentException("MQTT connection interrupted", e,
+                    DeploymentErrorCode.MQTT_CONNECTION_FAILED);
+        } catch (Exception e) {
```
MikeDombo, line 84: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

src/main/java/com/aws/greengrass/mqttclient/StandaloneMqttConnection.java
```diff
@@ -0,0 +1,157 @@
...
+        }
+    }
+
+    @Override
+    @SuppressWarnings("PMD.AvoidCatchingGenericException")
+    public void close() {
+        if (connection != null) {
+            try {
+                connection.disconnect().get(DISCONNECT_TIMEOUT_MS, TimeUnit.MILLISECONDS);
+            } catch (Exception e) {
```
MikeDombo, line 124: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
```diff
@@ -167,9 +178,25 @@ private void updateActionForDeployment(Map<String, Object> newConfig, Deployment
...
+            if (!verifyMqttConnectivity(totallyCompleteFuture, newDataEndpoint, "#endpoint-switch",
+                    preflightTimeout)) {
+                return;
+            }
+            logger.atInfo().setEventType(MERGE_CONFIG_EVENT_KEY)
+                    .kv("endpoint", newDataEndpoint)
+                    .log("Pre-flight MQTT connectivity check passed");
+
+            // Persist source endpoint only after pre-flight succeeds — no state change on failure
             kernel.getContext().get(DeploymentService.class).getRuntimeConfig()
```
MikeDombo, line 199: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

Untrusted input is passed to a method that executes arbitrary code, creating a code injection vulnerability. This allows attackers to execute malicious code within the application, potentially leading to data theft or system compromise. Validate input using allowlists with `Arrays.asList().contains()`, use regex matching with `String.matches()`, or parse input safely with `Integer.parseInt()` or `Boolean.parseBoolean()`. [Learn more](https://owasp.org/www-community/attacks/Code_Injection)

src/main/java/com/aws/greengrass/deployment/errorcode/DeploymentErrorCode.java
```diff
@@ -41,6 +41,9 @@ public enum DeploymentErrorCode {
 
     /* Network / http */
     NETWORK_ERROR(DeploymentErrorType.NETWORK_ERROR),
+    MQTT_CONNECTION_FAILED(DeploymentErrorType.NETWORK_ERROR),
+    TLS_HANDSHAKE_FAILURE(DeploymentErrorType.NETWORK_ERROR),
+    MISSING_MQTT_CONNECT_POLICY(DeploymentErrorType.PERMISSION_ERROR),
```
Medivh0416, line 46: Are we able to get detail error or just get a access denied?

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
```diff
@@ -167,9 +179,25 @@ private void updateActionForDeployment(Map<String, Object> newConfig, Deployment
...
-                    .log("Endpoint switch deployment detected, persisting source endpoint for rollback");
+                    .log("Endpoint switch deployment detected");
+
+            // Pre-flight MQTT connectivity check to new endpoint
+            logger.atInfo().setEventType(MERGE_CONFIG_EVENT_KEY)
+                    .kv(ENDPOINT_LOG_KEY, newDataEndpoint)
+                    .log("Starting pre-flight MQTT connectivity check");
+            long preflightTimeout = Coerce.toLong(deviceConfiguration.getMQTTNamespace()
+                    .findOrDefault(DEFAULT_PREFLIGHT_MQTT_TIMEOUT_MS, STANDALONE_MQTT_TIMEOUT_KEY));
+            if (!verifyMqttConnectivity(totallyCompleteFuture, newDataEndpoint, "#endpoint-switch",
```
Medivh0416, line 190: nit: #endpoint-switch could be a const

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
```diff
@@ -195,6 +223,71 @@ private boolean validateNucleusConfig(CompletableFuture<DeploymentResult> totall
         return true;
     }
 
+    @SuppressWarnings({"PMD.AvoidCatchingGenericException"})
+    @edu.umd.cs.findbugs.annotations.SuppressFBWarnings(value = "REC_CATCH_EXCEPTION",
+            justification = "Defensive catch for unexpected errors from CRT builder and proxy config")
+    boolean verifyMqttConnectivity(CompletableFuture<DeploymentResult> totallyCompleteFuture,
```
Medivh0416, line 229: not blocking: the logic here for setting up the mqtt connection could be extra to a method and reuseable

Medivh0416 (approved):
few minor comment, can be address later

src/main/java/com/aws/greengrass/deployment/errorcode/DeploymentErrorCode.java
sahith: Yes, detailed error is published through deployment failure cause https://docs.aws.amazon.com/greengrass/v2/developerguide/check-deployment-status.html

src/main/java/com/aws/greengrass/mqttclient/StandaloneMqttConnection.java
```diff
@@ -0,0 +1,157 @@
...
+     * Publish a message. Used by Step 5 (status reporting to source account) — remove this comment when used.
+     *
+     * @param topic     MQTT topic
+     * @param payload   message payload
+     * @param qos       quality of service
+     * @param timeoutMs publish timeout in milliseconds
+     * @throws DeploymentException on failure
+     */
+    @SuppressWarnings("PMD.AvoidInstanceofChecksInCatchClause")
+    public void publish(String topic, byte[] payload, QualityOfService qos, long timeoutMs)
```
sahith, line 102: We should have retries configured here to publish status to source account.

src/main/java/com/aws/greengrass/mqttclient/StandaloneMqttConnection.java
```diff
@@ -0,0 +1,157 @@
...
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+/**
+ * A standalone MQTT connection for one-shot operations (pre-flight checks, status reporting)
+ * independent of the main MqttClient. Not thread-safe — callers must not share instances across threads.
+ */
+@SuppressWarnings({"PMD.NullAssignment"})
+public class StandaloneMqttConnection implements Closeable {
```
sahith, line 30: > A standalone MQTT connection for one-shot operations (pre-flight checks, status reporting) independent of the main MqttClient

Can we integrate this logic into main MqttClient? If not, why?

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
```diff
@@ -167,9 +179,25 @@ private void updateActionForDeployment(Map<String, Object> newConfig, Deployment
             logger.atInfo().setEventType(MERGE_CONFIG_EVENT_KEY)
                     .kv("currentIotDataEndpoint", currentDataEndpoint)
                     .kv("newIotDataEndpoint", newDataEndpoint)
-                    .log("Endpoint switch deployment detected, persisting source endpoint for rollback");
+                    .log("Endpoint switch deployment detected");
+
+            // Pre-flight MQTT connectivity check to new endpoint
+            logger.atInfo().setEventType(MERGE_CONFIG_EVENT_KEY)
+                    .kv(ENDPOINT_LOG_KEY, newDataEndpoint)
+                    .log("Starting pre-flight MQTT connectivity check");
```
sahith, line 187: For my reference, does this log which endpoint it is establishing pre-flight connectivity with?

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
sahith: Why can't this logic live in `StandaloneMqttConnection` class?

sahith:
Did we verify if the MQTT connectivity check retries are happening for a default of 60 seconds?

sahith (dismissed):

Vipul-5517:
> Did we verify if the MQTT connectivity check retries are happening for a default of 60 seconds?

Yes — verified in E2E testing. Test 17 specifically validated custom timeout (5s) vs default (60s). With default 60s: maxAttempts = Math.max(1, (int)(60000 / 10000)) = 6 attempts × 10s per-attempt timeout + exponential backoff between retries. The CRT DNS resolution adds overhead, so actual wall-clock time is slightly above the configured timeout (observed ~20s for 5s config, ~65s for 60s config in E2E). All 22 E2E tests passed including timeout scenarios.

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
Vipul-5517: Yes — the .kv(ENDPOINT_LOG_KEY, newDataEndpoint) logs the target endpoint. The log line will read something like: Starting pre-flight MQTT connectivity check {iotDataEndpoint=abc123-ats.iot.us-west-2.amazonaws.com}

src/main/java/com/aws/greengrass/mqttclient/StandaloneMqttConnection.java
Vipul-5517: No — the main MqttClient manages long-lived connections with spooling, BatchedSubscriber, and reconnection logic tied to DeviceConfiguration subscriptions. It's designed for the device's primary cloud connection. StandaloneMqttConnection serves a fundamentally different purpose: short-lived, one-shot connections to arbitrary endpoints (new endpoint for pre-flight, source endpoint for status reporting after switch). Mixing these into MqttClient would add endpoint-switching complexity to a critical path and risk breaking the existing reconnection behavior. The design doc (Section G.4) explicitly calls for a separate temporary connection for this reason.

src/main/java/com/aws/greengrass/mqttclient/StandaloneMqttConnection.java
Vipul-5517: Agreed — retries for publish will be added in Step 5 (status reporting to source account) when this method is actually called. This PR scaffolds the method signature; the retry logic with exponential backoff within a 60s window will come with the Step 5 implementation per the design

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
Vipul-5517: The builder setup requires DeviceConfiguration (for proxy settings) and SecurityService (for the MQTT connection builder with device certs). StandaloneMqttConnection is intentionally decoupled from these deployment-layer dependencies — it accepts a pre-built MqttClientConnection so it can be reused in Step 5 (status reporting) where the builder config differs (source endpoint vs new endpoint). That said, we could extract the builder setup into a factory method on StandaloneMqttConnection that takes DeviceConfiguration + SecurityService as params. Will consider for a follow-up.

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
Vipul-5517: Agreed, will address in a follow-up. The string is used once and serves as a client ID suffix to avoid collisions with the main MQTT client — making it a constant is cleaner.

src/main/java/com/aws/greengrass/mqttclient/StandaloneMqttConnection.java
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    static DeploymentErrorCode mapExceptionToErrorCode(Throwable cause) {
+        if (cause == null) {
+            return DeploymentErrorCode.MQTT_CONNECTION_FAILED;
+        }
+        // Unwrap ExecutionException from CompletableFuture.get()
+        Throwable actual = cause instanceof ExecutionException && cause.getCause() != null
+                ? cause.getCause() : cause;
+        String message = actual.getMessage() == null ? "" : actual.getMessage().toLowerCase();
```
alter-mage: `toLowerCase()` without a locale is unsafe on non-English JVM locales. On `tr_TR`, `"Invalid certificate".toLowerCase()` produces `"ınvalid certıfıcate"` (dotless i), so `contains("certificate")` returns false and the error is misclassified as `MQTT_CONNECTION_FAILED` instead of `TLS_HANDSHAKE_FAILURE`.

Fix: `toLowerCase(Locale.ROOT)` — one-line change.

Separately, consider a follow-up to replace substring matching with type-based classification via `MqttException.getErrorCode()` / `ConnAckPacket` reason codes (pattern already used in `AwsIotMqtt5Client`) — more important once `publish()` is actually called in Step 5.

src/main/java/com/aws/greengrass/mqttclient/StandaloneMqttConnector.java
```diff
@@ -0,0 +1,157 @@
...
+     * @throws DeploymentException with appropriate error code if all attempts fail
+     */
+    @SuppressWarnings("PMD.AvoidCatchingGenericException")
+    public void connect(long timeoutMs) throws DeploymentException {
+        if (timeoutMs <= 0) {
+            throw new DeploymentException("MQTT connection timeout must be positive, got " + timeoutMs,
+                    DeploymentErrorCode.MQTT_CONNECTION_FAILED);
+        }
+        long perAttemptTimeout = Math.min(timeoutMs, PER_ATTEMPT_TIMEOUT_MS);
+        int maxAttempts = Math.max(1, (int) (timeoutMs / perAttemptTimeout));
```
alter-mage, line 132: `standaloneMqttTimeoutMs: 60000` reads as a wall-clock bound, but the actual worst-case is ~85s: 6 attempts × 10s = 60s, plus jittered backoff (1→2→4→8→10s) between attempts = up to ~25s more.

Also, integer division quantizes: `timeoutMs=15000` → `maxAttempts = 15000/10000 = 1` → 1 attempt of 10s, silently discarding 5s of budget.

Suggest tracking `deadline = now + timeoutMs`, checking it before each attempt, and capping per-attempt timeout at `deadline - now`. If too invasive for this PR, at minimum update the README to document that the value is approximate and can be exceeded by up to the max backoff interval.

Side note: `GIVEN_persistent_failure_WHEN_connect_THEN_retries_until_timeout` runs ~37s of real time — CI flake risk; consider a controlled clock.

src/main/java/com/aws/greengrass/mqttclient/StandaloneMqttConnection.java
alter-mage: Suggest removing `publish()` from this PR — it has no callers and commits to the deprecated MQTT v3 `MqttClientConnection.publish` API (flagged by CodeQL). The codebase is migrating to v5 (`com.aws.greengrass.mqttclient.v5`). It also creates inconsistency with `connect()`, which is retry-wrapped, while `publish()` is not — per sahith's comment, retries are needed before any real use. Cleaner to add it in the Step 5 PR built on v5 with retries from the start, rather than land dead code that freezes the wrong shape.

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
```diff
@@ -195,6 +223,71 @@ private boolean validateNucleusConfig(CompletableFuture<DeploymentResult> totall
...
+                            Coerce.toString(deviceConfiguration.getRootCAFilePath()))
+                    .withPort((short) mqttPort)
+                    .withCleanSession(true);
+            // Configure proxy if set — same pattern as MqttClient.
+            // Build proxy options directly to avoid passing null to @NonNull tlsContext param.
+            // HTTPS proxies require a TLS context which we don't create for pre-flight;
+            // HTTP CONNECT tunneling (the common case for MQTT) works without it.
+            String proxyUrl = deviceConfiguration.getProxyUrl();
+            if (Utils.isNotEmpty(proxyUrl)) {
+                HttpProxyOptions httpProxyOptions = new HttpProxyOptions();
```
alter-mage: This proxy block (~22 lines) reimplements what `ProxyUtils.getHttpProxyOptions` already does, but with two behavior divergences:

1. Skips `ProxyUtils.getProxyUsername`/`getProxyPassword` helpers — URL-embedded credentials (`http://user:pass@host`) work for the main `MqttClient` but silently break for pre-flight.
2. Drops the HTTPS-proxy `setTlsContext` branch — HTTPS-proxy customers: main client works, pre-flight fails.

Suggest relaxing `@NonNull` on the `tlsContext` param in `ProxyUtils.getHttpProxyOptions` and guarding `setTlsContext` with a null check. This block then collapses to ~6 lines and stays consistent with `MqttClient`. This was raised by @sahith in review and deferred — given the two divergences, worth fixing in this PR.

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
```diff
@@ -195,6 +223,71 @@ private boolean validateNucleusConfig(CompletableFuture<DeploymentResult> totall
...
+            }
+            return true;
+        } catch (DeploymentException e) {
+            logger.atError().setEventType(MERGE_ERROR_LOG_EVENT_KEY).setCause(e)
+                    .kv(ENDPOINT_LOG_KEY, endpoint)
+                    .log("Pre-flight MQTT connectivity check failed");
+            totallyCompleteFuture.complete(
+                    new DeploymentResult(DeploymentResult.DeploymentStatus.FAILED_NO_STATE_CHANGE, e));
+            return false;
+        } catch (Exception e) {
```
alter-mage, line 265: This generic catch fires for `MqttConnectionProviderException` from `SecurityService.getDeviceIdentityMqttConnectionBuilder()` — a cert/key configuration issue — but wraps it as `MQTT_CONNECTION_FAILED` (NETWORK_ERROR type). Customer sees "network error" when the real problem is a misconfigured cert.

Suggest adding `catch (MqttConnectionProviderException e)` before this catch and mapping it to a cert-appropriate error code, so the granular error codes added in this PR are not undone for this path.

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
```diff
@@ -167,9 +179,25 @@ private void updateActionForDeployment(Map<String, Object> newConfig, Deployment
...
                     .kv("newIotDataEndpoint", newDataEndpoint)
-                    .log("Endpoint switch deployment detected, persisting source endpoint for rollback");
+                    .log("Endpoint switch deployment detected");
+
+            // Pre-flight MQTT connectivity check to new endpoint
+            logger.atInfo().setEventType(MERGE_CONFIG_EVENT_KEY)
+                    .kv(ENDPOINT_LOG_KEY, newDataEndpoint)
+                    .log("Starting pre-flight MQTT connectivity check");
+            long preflightTimeout = Coerce.toLong(deviceConfiguration.getMQTTNamespace()
+                    .findOrDefault(DEFAULT_PREFLIGHT_MQTT_TIMEOUT_MS, STANDALONE_MQTT_TIMEOUT_KEY));
```
alter-mage, line 185: If a customer sets `standaloneMqttTimeoutMs` to 0, negative, or a non-numeric value (coerces to 0), this flows into `StandaloneMqttConnection.connect` which throws, gets caught by the generic `catch (Exception)` in `verifyMqttConnectivity`, and surfaces as `MQTT_CONNECTION_FAILED` — customer sees a network error when the real problem is their config.

Suggest validating here and falling back to the default:
```java
if (preflightTimeout <= 0) {
    logger.atWarn().kv("configured", preflightTimeout)
        .kv("using", DEFAULT_PREFLIGHT_MQTT_TIMEOUT_MS)
        .log("Invalid standaloneMqttTimeoutMs, using default");
    preflightTimeout = DEFAULT_PREFLIGHT_MQTT_TIMEOUT_MS;
}
```

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
alter-mage: The pre-flight uses `thingName + "#endpoint-switch"` as the MQTT client ID. A very common (and AWS-recommended) IoT policy pattern restricts `iot:Connect` to only the thing name (see [Connect and publish policy examples](https://docs.aws.amazon.com/iot/latest/developerguide/connect-and-pub.html)):

```json
{
  "Effect": "Allow",
  "Action": "iot:Connect",
  "Resource": "arn:aws:iot:region:account:client/${iot:Connection.Thing.ThingName}"
}
```

With this policy, `myThing#endpoint-switch` does not match `client/myThing` — the CONNECT is denied, pre-flight fails with `MISSING_MQTT_CONNECT_POLICY`, and every endpoint-switch deployment reports `FAILED_NO_STATE_CHANGE`. The feature is silently broken for any customer following the recommended security posture.

Options to consider:
1. Use the thing name as the client ID — but IoT Core will disconnect the existing main `MqttClient` connection (same client ID = existing connection kicked)
2. Require customers to update their policy to allow `client/${iot:Connection.Thing.ThingName}*` — needs prominent documentation and ideally pre-deployment validation
3. Use a different verification mechanism that does not require a new MQTT CONNECT (e.g., TLS handshake only, or HTTP to the data endpoint)

This needs a decision before merge — correctness issue for a significant portion of customers.

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
alter-mage: Valid finding. The generic `catch (Exception e)` here fires for `MqttConnectionProviderException` from `SecurityService.getDeviceIdentityMqttConnectionBuilder()`, which is a cert/key configuration issue — but it gets wrapped as `MQTT_CONNECTION_FAILED` (NETWORK_ERROR type). Customer sees "network error" when the real problem is a misconfigured cert. Will add a `catch (MqttConnectionProviderException e)` branch before this catch to map it to an appropriate error code.

src/main/java/com/aws/greengrass/mqttclient/StandaloneMqttConnection.java
alter-mage: nit: please rename the class to something like `MqttConnectionValidator`

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
Vipul-5517: Done — extracted to ENDPOINT_SWITCH_CLIENT_SUFFIX constant in DeploymentConfigMerger.

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
Vipul-5517: Moved the builder setup + proxy configuration into StandaloneMqttConnection.of() static factory method. Takes SecurityService + DeviceConfiguration as params so it's reusable for Step 5 (status reporting to
source endpoint with different builder config). verifyMqttConnectivity is now ~15 lines — just calls the factory and connect().

Proxy options are built directly rather than via ProxyUtils.getHttpProxyOptions() because that method requires @NonNull ClientTlsContext for HTTPS proxy tunneling. Standalone connections use HTTP CONNECT
tunneling (the common case for MQTT) which works without a TLS context. Added Javadoc explaining this.

src/main/java/com/aws/greengrass/mqttclient/StandaloneMqttConnection.java
Vipul-5517: Removed publish() from this PR entirely — it was dead code (nothing calls it). Will add it back in Step 5 with proper retry logic when status reporting is implemented.

src/main/java/com/aws/greengrass/mqttclient/StandaloneMqttConnection.java
Vipul-5517: Fixed — using Locale.ROOT throughout mapExceptionToErrorCode().

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
Vipul-5517: Added catch (MqttConnectionProviderException e) before the generic catch. Cert/key configuration errors now map to TLS_HANDSHAKE_FAILURE instead of being misclassified as MQTT_CONNECTION_FAILED. The remaining generic catch is intentional — it's the classification entry point for CRT exceptions via mapExceptionToErrorCode().

src/main/java/com/aws/greengrass/mqttclient/StandaloneMqttConnector.java
```diff
@@ -0,0 +1,157 @@
...
+                connection.connect().get(perAttemptTimeout, TimeUnit.MILLISECONDS);
+                logger.atInfo().kv("clientId", clientId).log("Standalone MQTT connection established");
+                return null;
+            }, "standalone-mqtt-connect", logger);
+        } catch (InterruptedException e) {
+            connectionCleanup();
+            Thread.currentThread().interrupt();
+            throw new DeploymentException("MQTT connection interrupted", e,
+                    DeploymentErrorCode.MQTT_CONNECTION_FAILED);
+        } catch (Exception e) {
```
Vipul-5517: Intentional — the generic catch is the classification entry point. mapExceptionToErrorCode() unwraps ExecutionException and classifies all CRT exceptions into specific DeploymentErrorCode values. All exceptions
are retryable to support JITR/JITP provisioning flows.

src/main/java/com/aws/greengrass/mqttclient/StandaloneMqttConnector.java
```diff
@@ -0,0 +1,157 @@
...
+        }
+    }
+
+    @Override
+    @SuppressWarnings("PMD.AvoidCatchingGenericException")
+    public void close() {
+        if (connection != null) {
+            try {
+                connection.disconnect().get(DISCONNECT_TIMEOUT_MS, TimeUnit.MILLISECONDS);
+            } catch (Exception e) {
```
Vipul-5517: Intentional — close() is best-effort cleanup. Swallowing all exceptions during disconnect is standard practice (same pattern as MqttClient.close()).

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
Vipul-5517: False positive. The value is an IoT endpoint string written to the Greengrass config store (a key-value store), not executed as code. Endpoint format is validated by deviceConfiguration.validateEndpoints() before this point.

src/main/java/com/aws/greengrass/mqttclient/StandaloneMqttConnector.java
Vipul-5517: Good catch. Updated Javadoc to clarify that timeoutMs is an attempt budget, not a wall-clock deadline — actual elapsed time may exceed it due to backoff between attempts. The integer division quantization is acceptable for this use case (pre-flight is a best-effort check, not a precision timer). A deadline-based approach would be cleaner but more invasive — will consider for Step 5 if needed.

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
Vipul-5517: Moved proxy setup into StandaloneMqttConnection.configureProxy(). The two divergences you identified (URL-embedded credentials, HTTPS-proxy TLS context) are valid. However, relaxing @NonNull on ProxyUtils.getHttpProxyOptions() changes a public API used by the main MqttClient — that's a broader change than this PR should make. Filed as a follow-up. The current implementation handles the HTTP CONNECT tunneling case (which is what MQTT-over-proxy uses) correctly

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
Vipul-5517: Done — added catch (MqttConnectionProviderException e) before the generic catch. Maps to TLS_HANDSHAKE_FAILURE with a clear "Device identity configuration error" message.

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
Vipul-5517: Good catch. StandaloneMqttConnection.connect() already validates timeoutMs <= 0 and throws with a clear message. However, the exception gets caught by the generic catch and mapped to MQTT_CONNECTION_FAILED. Will add the validation + fallback to default in DeploymentConfigMerger as you suggested — this is a one-line fix. Done

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
Vipul-5517: This is a valid concern. Customers with ${iot:Connection.Thing.ThingName} exact-match policies will have pre-flight rejected. However:
1. The design explicitly chose MQTT CONNECT as the pre-flight mechanism because a connect failure is catastrophic (device loses cloud connectivity), while a policy issue is recoverable (customer
updates policy, takes effect immediately).
2. Using the bare thing name would disconnect the main MqttClient (IoT Core kicks the existing connection when a new connection uses the same client ID) — this is worse than a policy rejection.
3. The # suffix is the same pattern used by IoT Device Shadow and IoT Jobs clients (thingName#shadow, thingName#jobs). Customers with restrictive policies already need wildcard patterns for these.
4. We document the required policy in README_CONFIG_SCHEMA.md and the deployment failure message includes MISSING_MQTT_CONNECT_POLICY which tells the customer exactly what to fix.

src/main/java/com/aws/greengrass/mqttclient/StandaloneMqttConnection.java
Vipul-5517: The class is designed for reuse beyond validation — Step 5 uses it for publishing deployment status to the source account, and Step 8 for post-apply connectivity verification. StandaloneMqttConnection accurately describes what it is (a standalone MQTT connection independent of the main client). Renaming to MqttConnectionValidator would be misleading when it's used for publishing in Step 5.

src/main/java/com/aws/greengrass/mqttclient/StandaloneMqttConnection.java
Vipul-5517: Fixed — using Locale.ROOT in mapExceptionToErrorCode().

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
Vipul-5517: Added documentation in README_CONFIG_SCHEMA.md with the required policy pattern.

sahith: IoT has a [quota](https://docs.aws.amazon.com/general/latest/gr/iot-core.html#limits_iot) of 50 max subscriptions per connection. Hence, classic uses clientId suffixed with `#N` for additional subscriptions. The [documentation](https://docs.aws.amazon.com/greengrass/v2/developerguide/device-auth.html#greengrass-core-minimal-iot-policy) also recommends defining a connect policy with `core-device-thing-name*`. 

We can have a pre-requisite of defining connect policy with clientId `core-device-thing-name*` for customers to leverage endpoint switch functionality.

README_CONFIG_SCHEMA.md
```diff
@@ -177,6 +177,7 @@ services:
         pingTimeoutMs: 30000
         operationTimeoutMs: 30000
         maxInFlightPublishes: 5
+        standaloneMqttTimeoutMs: 60000
```
sahith, line 180: Why do we need to allow this to be configurable instead of 60s connectivity check for every endpoint switch deployment? Allowing this timeout value to be more would be stalling the deployment. If there is a network issue persists, then a retry of endpoint switch deployment would be done.

sahith: If we allow this to be configurable, there should be guardrails around min and max allowed timeout value.

src/main/java/com/aws/greengrass/mqttclient/StandaloneMqttConnection.java
alter-mage: Sure, the essence of my comment is that the class name should not be a concept which is the case here (`..Connection`). If you notice, all functional class names represent an action (`..Activator`, `..Manager`, `..Merger`, etc.).

Vipul-5517: got it. make sense. I miss-understood that

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
alter-mage: What if a customer's policy is intentionally restrictive? This is a breaking change. This means all endpoint switch deployments will fail for devices running Nucleus version which supports connectivity check with intentionally restrictive policy.

src/main/java/com/aws/greengrass/deployment/activator/DeploymentActivator.java
```diff
@@ -43,6 +43,21 @@ protected DeploymentActivator(Kernel kernel) {
...
 
+    /**
+     * Activate with endpoint switch flag. Default delegates to the 4-arg version.
+     * Overridden by DefaultActivator to use the flag for forced rollback.
+     *
+     * @param newConfig             new configuration to apply
+     * @param deployment            deployment being processed
+     * @param configMergeTimestamp  timestamp for config merge
+     * @param totallyCompleteFuture future to complete with deployment result
+     * @param isEndpointSwitch      true if this deployment changes the IoT data endpoint
```
sahith: Why add `isEndpointSwitch` param to activate method instead of defining `isEndpointSwitch()` private method to know if it is an endpoint switch deployment where required?

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
alter-mage: Agreed that `#endpoint-switch` suffix + documentation is the right approach given the Shadow/Jobs precedent. The remaining gap I'd like addressed before merge:

The current `MISSING_MQTT_CONNECT_POLICY` error code and log message are too generic for this case. When a customer upgrades Nucleus and their endpoint-switch deployment starts failing, they need to be able to self-diagnose. Suggest two changes:

1. A dedicated error code — e.g. `ENDPOINT_SWITCH_CONNECT_POLICY_REQUIRED` — distinct from the generic `MISSING_MQTT_CONNECT_POLICY`, so it's unambiguous in deployment status and metrics
2. A log message that tells the customer exactly what to fix:

```
Pre-flight MQTT connectivity check failed: IoT policy does not permit client ID
"<thingName>#endpoint-switch". To enable endpoint-switch deployments, update your
iot:Connect policy resource to allow "arn:aws:iot:*:*:client/<thingName>*".
```

alter-mage (dismissed):

alter-mage (approved):

sahith (approved):

PR #1788, opened 2026-05-04
test: integ test for KERNEL_ACTIVATION shadow re-enqueue race

alter-mage:
**Issue #, if available:** P422923074

**Description of changes:**
Adds an integration test that replicates the spurious deployment cancellation race after a KERNEL_ACTIVATION resume. Also adds a `triggerShadowUpdatedForTest` accessor to `ShadowDeploymentListener` to allow the test to inject a shadow update from outside the package.

The race was reported by a customer on v2.16.1. The fix landed in #1782 but had no integration test covering the exact scenario. This test confirms the bug is present on v2.16.1 and absent on main.

The test injects a duplicate shadow update into the race window (after `Started deployment execution` but before `currentDeploymentTaskMetadata` is assigned) and asserts the deployment is not spuriously cancelled. Verified it fails on v2.16.1 and passes on main.

Companion to #1782. The fix is already merged; this PR adds the regression test.

**Documentation Checklist:**

**Compatibility Checklist:**
- [ ] For external library version updates, I have reviewed its change logs and Nucleus does not consume any deprecated method or type.


src/integrationtests/java/com/aws/greengrass/integrationtests/deployment/DeploymentServiceIntegrationTest.java
```diff
@@ -698,6 +701,64 @@ void GIVEN_deployment_with_broken_component_WHEN_deployment_is_cancelled_during_
...
+        };
+        try (AutoCloseable l = TestUtils.createCloseableLogListener(startListener)) {
+            assertTrue(startedLatch.await(10, TimeUnit.SECONDS), "Deployment must start");
+        }
+
+        // Inject duplicate shadow update into the race window
+        Map<String, Object> desired = new HashMap<>();
+        desired.put(ShadowDeploymentListener.FLEET_CONFIG_KEY, fleetConfigStr);
+        desired.put(ShadowDeploymentListener.DESIRED_STATUS_KEY, "SUCCEEDED");
+        shadowListener.triggerShadowUpdatedForTest(desired, null, 1);
```
aws-kevinrickard: The cancel log listener is registered *after* `triggerShadowUpdatedForTest` returns:

```java
shadowListener.triggerShadowUpdatedForTest(desired, null, 1);

// ... allocate latch, create consumer ...
try (AutoCloseable l = TestUtils.createCloseableLogListener(cancelListener)) {
    assertFalse(cancelledLatch.await(3, TimeUnit.SECONDS), ...);
}
```

`shadowUpdated` is synchronous — on broken code, the duplicate is already in the `DeploymentQueue` by the time this call returns. With polling frequency 0, the deployment loop can pull it and log `"Canceling current device deployment"` before the listener is attached. The `await` then returns `false`, `assertFalse(false)` passes, and the test gives a false negative on buggy code.

Every other test in this file attaches the listener first, then does the work. Suggest wrapping the whole sequence in a single listener scope:

```java
CountDownLatch cancelledLatch = new CountDownLatch(1);
Consumer<GreengrassLogMessage> cancelListener = m -> {
    if ("Canceling current device deployment".equals(m.getMessage())) {
        cancelledLatch.countDown();
    }
};
try (AutoCloseable l = TestUtils.createCloseableLogListener(cancelListener)) {
    // ... wait for "Started deployment execution" ...
    // ... inject shadow update ...
    assertFalse(cancelledLatch.await(3, TimeUnit.SECONDS),
            "Deployment should NOT be spuriously cancelled after duplicate shadow update");
}
```

This would make the test more likely to give a true positive signal on regression and reduce the chances of false negative signal

src/integrationtests/java/com/aws/greengrass/integrationtests/deployment/DeploymentServiceIntegrationTest.java
```diff
@@ -698,6 +701,64 @@ void GIVEN_deployment_with_broken_component_WHEN_deployment_is_cancelled_during_
...
+                Configuration.class);
+        deploymentConfiguration.setCreationTimestamp(System.currentTimeMillis());
+        deploymentConfiguration.setConfigurationArn(configArn);
+        String fleetConfigStr = OBJECT_MAPPER.writeValueAsString(deploymentConfiguration);
+
+        // Simulate KERNEL_ACTIVATION resume: offer deployment with KERNEL_ACTIVATION stage
+        DeploymentDocument deploymentDocument = convertFromDeploymentConfiguration(deploymentConfiguration);
+        Deployment resumedDeployment = new Deployment(deploymentDocument, DeploymentType.SHADOW, configArn,
+                Deployment.DeploymentStage.KERNEL_ACTIVATION);
+        resumedDeployment.setDeploymentStage(Deployment.DeploymentStage.KERNEL_ACTIVATION);
```
aws-kevinrickard: Nit: This call is redundant when the constructor used in the line above sets this

src/integrationtests/java/com/aws/greengrass/integrationtests/deployment/DeploymentServiceIntegrationTest.java
```diff
@@ -698,6 +701,64 @@ void GIVEN_deployment_with_broken_component_WHEN_deployment_is_cancelled_during_
         }
     }
 
+    @Test
+    void GIVEN_kernel_activation_resume_WHEN_shadow_update_arrives_before_metadata_set_THEN_deployment_not_cancelled(
+            ExtensionContext context) throws Exception {
+        ignoreExceptionOfType(context, NullPointerException.class);
```
aws-kevinrickard: Nit: A narrower ignore using 
ignoreExceptionWithStackTraceContaining(context, NullPointerException.class, "KernelUpdateDeploymentTask") or "waitForServicesToStart" would be 
safer — it would still suppress the expected noise while catching unexpected NPEs in the code path under test

aws-kevinrickard (approved):

yitingb (approved):

PR #1789, opened 2026-05-07
feat(deployment): add pre-flight IoT credential endpoint validation

sahith:
## Summary

Adds pre-flight IoT credential endpoint validation for endpoint-switch deployments (Step 3 of the IoT endpoint switch implementation plan).

When a deployment changes the IoT credential endpoint, the deployment service now validates connectivity by fetching temporary credentials from the new credential endpoint before applying configuration changes. This prevents bricking devices with an unreachable or misconfigured credential endpoint.

### Changes
- Added `validateCredentialEndpoint()` in `DeploymentConfigMerger` that performs an HTTPS request to the new IoT credential endpoint using the device certificate for mTLS
- Added new error codes: `CREDENTIAL_ENDPOINT_AUTH_FAILURE`, `ROLE_ALIAS_NOT_FOUND`, `CREDENTIAL_ENDPOINT_SERVER_ERROR`
- Validation runs after MQTT pre-flight check (Step 2) and only for endpoint-switch deployments
- On failure, reports `FAILED_NO_STATE_CHANGE` — no device state is modified

### Testing Done

**Unit tests:**
- Successful credential fetch (HTTP 200) → deployment proceeds
- HTTP 403 → `CREDENTIAL_ENDPOINT_AUTH_FAILURE` error code
- HTTP 404 → `ROLE_ALIAS_NOT_FOUND` error code
- HTTP 5xx → `CREDENTIAL_ENDPOINT_SERVER_ERROR` error code
- Network timeout → failure reported within configured timeout
- Validation skipped for non-endpoint-switch deployments
- Validation skipped when credential endpoint is unchanged

**Manual validation on GG Classic device:**
- Spun up a Greengrass Classic core device and deployed an endpoint switch with valid IoT credential endpoint → credential validation passed, deployment proceeded (happy path)
- Deployed with a non-existent role alias → deployment failed with `ROLE_ALIAS_NOT_FOUND`, device state unchanged
- Deployed with a credential endpoint where the device certificate policy does not grant credential access → deployment failed with `CREDENTIAL_ENDPOINT_AUTH_FAILURE`, device state unchanged

All existing unit tests pass (`mvn test`). Build passes (`mvn compile`).


src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
```diff
@@ -274,6 +300,82 @@ boolean verifyMqttConnectivity(CompletableFuture<DeploymentResult> totallyComple
...
+            return false;
+        } catch (DeploymentException e) {
+            logger.atError().setEventType(MERGE_ERROR_LOG_EVENT_KEY).setCause(e)
+                    .kv(ENDPOINT_LOG_KEY, credEndpoint)
+                    .kv("reason", e.getMessage())
+                    .log("Pre-flight credential endpoint check failed");
+            totallyCompleteFuture.complete(new DeploymentResult(
+                    DeploymentResult.DeploymentStatus.FAILED_NO_STATE_CHANGE, e));
+            return false;
+        } catch (Exception e) {
```
MikeDombo: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

xuchuwan (dismissed):

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
```diff
@@ -274,6 +300,82 @@ boolean verifyMqttConnectivity(CompletableFuture<DeploymentResult> totallyComple
...
+        String thingName = Coerce.toString(deviceConfiguration.getThingName());
+        String credUrl = String.format(CREDENTIAL_URL_FORMAT, credEndpoint, roleAlias);
+
+        logger.atInfo().setEventType(MERGE_CONFIG_EVENT_KEY)
+                .kv(ENDPOINT_LOG_KEY, credEndpoint)
+                .kv("roleAlias", roleAlias)
+                .log("Starting pre-flight credential endpoint check");
+
+        // Calculate max attempts from timeout with exponential backoff (1s initial, 10s max)
+        int maxAttempts = Math.max(1, (int) (timeoutMs / 10_000));
```
Vipul-5517: nit - Should we have these variables at the top level (1s and 10s). We are using it below as well

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
```diff
@@ -274,6 +300,82 @@ boolean verifyMqttConnectivity(CompletableFuture<DeploymentResult> totallyComple
...
+                        iotConnectionManager, thingName, credUrl, "GET", null);
+                int statusCode = response.getStatusCode();
+                if (statusCode == 200) {
+                    return null;
+                }
+                if (statusCode == 401 || statusCode == 403 || statusCode == 404) {
+                    throw new DeploymentException(
+                            "Credential endpoint returned " + statusCode + " for role alias '"
+                                    + roleAlias + "': " + response
+                                    + ". Possible causes: (1) role alias does not exist in target region,"
```
Vipul-5517: Are these linked to individual error codes ? if yes why not be explicit what is wrong ?

Vipul-5517: The design specifies a separate ROLE_ALIAS_NOT_FOUND error code for HTTP 404, here we are mapping 404 to CREDENTIAL_ENDPOINT_AUTH_FAILURE

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
```diff
@@ -274,6 +300,82 @@ boolean verifyMqttConnectivity(CompletableFuture<DeploymentResult> totallyComple
...
+            totallyCompleteFuture.complete(new DeploymentResult(
+                    DeploymentResult.DeploymentStatus.FAILED_NO_STATE_CHANGE,
+                    new DeploymentException("Credential endpoint check interrupted", e,
+                            DeploymentErrorCode.CREDENTIAL_ENDPOINT_SERVER_ERROR)));
+            return false;
+        } catch (DeploymentException e) {
+            logger.atError().setEventType(MERGE_ERROR_LOG_EVENT_KEY).setCause(e)
+                    .kv(ENDPOINT_LOG_KEY, credEndpoint)
+                    .kv("reason", e.getMessage())
+                    .log("Pre-flight credential endpoint check failed");
```
Vipul-5517: nit - Should we log the same message as below (for consistency ?)

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
```diff
@@ -202,6 +211,23 @@ private void updateActionForDeployment(Map<String, Object> newConfig, Deployment
                     .lookup(DeploymentService.SOURCE_IOT_DATA_ENDPOINT_KEY).withValue(currentDataEndpoint);
         }
 
+        // Pre-flight credential endpoint check — runs whenever iotCredEndpoint or iotRoleAlias changes
```
Vipul-5517: Based of Kiro CLI comment - "request a comment explaining the mTLS
client uses the current device cert, which must be registered in both source and destination accounts"

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
```diff
@@ -202,6 +211,23 @@ private void updateActionForDeployment(Map<String, Object> newConfig, Deployment
...
+        if (nucleusConfig != null) {
+            String currentCredEndpoint = Coerce.toString(deviceConfiguration.getIotCredentialEndpoint());
+            String newCredEndpoint = nucleusConfig.containsKey(DEVICE_PARAM_IOT_CRED_ENDPOINT)
+                    ? Coerce.toString(nucleusConfig.get(DEVICE_PARAM_IOT_CRED_ENDPOINT)) : currentCredEndpoint;
+            String currentRoleAlias = Coerce.toString(deviceConfiguration.getIotRoleAlias());
+            String newRoleAlias = nucleusConfig.containsKey(DeviceConfiguration.IOT_ROLE_ALIAS_TOPIC)
```
Vipul-5517, line 211: suggested by Kiro-cli - "On the block that resolves newRoleAlias — suggest adding a guard: if
roleAlias is empty/null, fail with a clear error before making the HTTP call"

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
```diff
@@ -274,6 +300,82 @@ boolean verifyMqttConnectivity(CompletableFuture<DeploymentResult> totallyComple
...
+
+            RetryUtils.runWithRetry(retryConfig, () -> {
+                IotCloudResponse response = iotCloudHelper.sendHttpRequest(
+                        iotConnectionManager, thingName, credUrl, "GET", null);
+                int statusCode = response.getStatusCode();
+                if (statusCode == 200) {
+                    return null;
+                }
+                if (statusCode == 401 || statusCode == 403 || statusCode == 404) {
+                    throw new DeploymentException(
```
Vipul-5517: from Kiro-cli - "On the line that includes + response in the exception message — suggest truncating to first 200 chars to avoid log pollution from HTML error pages"

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
```diff
@@ -274,6 +300,82 @@ boolean verifyMqttConnectivity(CompletableFuture<DeploymentResult> totallyComple
...
+            logger.atInfo().setEventType(MERGE_CONFIG_EVENT_KEY)
+                    .kv(ENDPOINT_LOG_KEY, credEndpoint)
+                    .log("Pre-flight credential endpoint check passed");
+            return true;
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            totallyCompleteFuture.complete(new DeploymentResult(
+                    DeploymentResult.DeploymentStatus.FAILED_NO_STATE_CHANGE,
+                    new DeploymentException("Credential endpoint check interrupted", e,
+                            DeploymentErrorCode.CREDENTIAL_ENDPOINT_SERVER_ERROR)));
```
Vipul-5517: from kiro-cli - "On the catch block — suggest adding a separate catch (TLSAuthException e) before the generic catch, mapping it to a DEVICE_ERROR type rather than CREDENTIAL_ENDPOINT_SERVER_ERROR"

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
```diff
@@ -274,6 +300,82 @@ boolean verifyMqttConnectivity(CompletableFuture<DeploymentResult> totallyComple
...
+        String credUrl = String.format(CREDENTIAL_URL_FORMAT, credEndpoint, roleAlias);
+
+        logger.atInfo().setEventType(MERGE_CONFIG_EVENT_KEY)
+                .kv(ENDPOINT_LOG_KEY, credEndpoint)
+                .kv("roleAlias", roleAlias)
+                .log("Starting pre-flight credential endpoint check");
+
+        // Calculate max attempts from timeout with exponential backoff (1s initial, 10s max)
+        int maxAttempts = Math.max(1, (int) (timeoutMs / 10_000));
+        RetryUtils.RetryConfig retryConfig = RetryUtils.RetryConfig.builder()
```
Vipul-5517: from kiro-cli - "On the retry config setup — add a comment noting that IotCloudHelper.sendHttpRequest() has its own internal retry (3×200ms), so the effective behavior is nested retries"

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
sahith: Explicit details is dependent on IoT credential endpoint service sending us the exception details. Hence, wrapped it under `CREDENTIAL_ENDPOINT_AUTH_FAILURE` with the plausible reasons for this failure.

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
sahith: The message below is for explicit mention of retries being exhausted which is not the case here

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
sahith: Extracted `CREDENTIAL_CHECK_INITIAL_RETRY_INTERVAL` and `CREDENTIAL_CHECK_MAX_RETRY_INTERVAL` as class-level constants. The maxAttempts calculation now uses CREDENTIAL_CHECK_MAX_RETRY_INTERVAL.toMillis() instead of the magic number.

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
sahith: Added comment explaining the device certificate is used for mTLS authentication and must be registered in the destination account/region.

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
sahith: Added a guard at the top of verifyCredentialEndpoint that fails immediately with a clear error if roleAlias is empty or null. Added unit tests for both cases.

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
sahith: IoT Credential Provider returns short JSON error bodies (e.g., {"message": "..."}), not HTML pages. Truncating would obscure useful diagnostic info. Keeping as-is.

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
sahith: Added comment noting `IotCloudHelper.sendHttpRequest()` has its own internal retry (3×200ms), so the effective behavior is nested retries for transient network errors.

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
sahith: TLS errors at this layer are wrapped by IotCloudHelper/IotConnectionManager into AWSIotException — a raw SSLException won't surface here. The generic catch already handles this correctly.

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
Vipul-5517: IoT credential endpoint service sending us the exception details - Do we know what details are sent or can we get that information to segregate as much as we can ?

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
Vipul-5517: got it, my miss

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
```diff
@@ -68,8 +76,12 @@ public class DeploymentConfigMerger {
     protected static final int WAIT_SVC_START_POLL_INTERVAL_MILLISEC = 1000;
     static final long DEFAULT_PREFLIGHT_MQTT_TIMEOUT_MS = 60_000;
     static final String STANDALONE_MQTT_TIMEOUT_KEY = "standaloneMqttTimeoutMs";
+    static final long DEFAULT_CREDENTIAL_ENDPOINT_VALIDATION_TIMEOUT_MS = 60_000;
```
alter-mage: Should this timeout be customer-configurable, similar to how the MQTT pre-flight uses `standaloneMqttTimeoutMs` from device config? The method signature already accepts `timeoutMs` as a parameter, but the caller hardcodes 60s. For devices on slow or unreliable networks, customers might need to tune this.

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
```diff
@@ -202,6 +214,25 @@ private void updateActionForDeployment(Map<String, Object> newConfig, Deployment
                     .lookup(DeploymentService.SOURCE_IOT_DATA_ENDPOINT_KEY).withValue(currentDataEndpoint);
         }
 
+        // Pre-flight credential endpoint check — uses the device certificate for mTLS authentication,
+        // which must be registered in the destination account/region for this check to succeed.
+        // Runs whenever iotCredEndpoint or iotRoleAlias changes.
+        if (nucleusConfig != null) {
```
alter-mage, line 206: `DeploymentConfigMerger` is accumulating responsibilities beyond config merging — format validation, MQTT connectivity checks, and now credential endpoint probing. The pre-flight checks should be extracted into a dedicated class. The merger's `updateActionForDeployment` is becoming hard to reason about with all the early-return paths.

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
```diff
@@ -274,6 +305,93 @@ boolean verifyMqttConnectivity(CompletableFuture<DeploymentResult> totallyComple
         }
     }
 
+    @SuppressWarnings({"PMD.AvoidCatchingGenericException", "PMD.CloseResource"})
```
alter-mage: Is the `PMD.CloseResource` suppression needed here? I don't see any closeable resources being opened in this method — `IotCloudHelper` and `IotConnectionManager` are singletons from context. Was this carried over from `verifyMqttConnectivity` by accident?

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
```diff
@@ -274,6 +305,93 @@ boolean verifyMqttConnectivity(CompletableFuture<DeploymentResult> totallyComple
...
+            // Note: IotCloudHelper.sendHttpRequest() has its own internal retry (3x200ms),
+            // so the effective behavior is nested retries for transient network errors.
+            RetryUtils.runWithRetry(retryConfig, () -> {
+                IotCloudResponse response = iotCloudHelper.sendHttpRequest(
+                        iotConnectionManager, thingName, credUrl, "GET", null);
+                int statusCode = response.getStatusCode();
+                if (statusCode == 200) {
+                    return null;
+                }
+                if (statusCode == 401 || statusCode == 403 || statusCode == 404) {
```
alter-mage: Nit: Consider using `HttpURLConnection.HTTP_UNAUTHORIZED`, `HTTP_FORBIDDEN`, `HTTP_NOT_FOUND` instead of raw `401`, `403`, `404`.

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
```diff
@@ -274,6 +305,93 @@ boolean verifyMqttConnectivity(CompletableFuture<DeploymentResult> totallyComple
...
+                }
+                // All other non-200 responses (429, 5xx, etc.) are retryable
+                throw new AWSIotException("Credential endpoint returned " + statusCode);
+            }, "verify-credential-endpoint", logger);
+
+            logger.atInfo().setEventType(MERGE_CONFIG_EVENT_KEY)
+                    .kv(ENDPOINT_LOG_KEY, credEndpoint)
+                    .log("Pre-flight credential endpoint check passed");
+            return true;
+        } catch (InterruptedException e) {
```
alter-mage: The `InterruptedException` catch completes the future with `CREDENTIAL_ENDPOINT_SERVER_ERROR`, but an interruption here means the deployment was cancelled (new deployment arrived, IoT Jobs cancel, or device shutdown) — not a server error. The MQTT pre-flight doesn't catch `InterruptedException` separately and lets it propagate to the deployment task runner which handles cancellation at a higher level. Should this do the same?

Also, this catch doesn't log anything before completing the future. The other two catch blocks log at ERROR level. Add a log line here for debuggability — otherwise an interrupted credential check leaves no trace in the logs.

src/test/java/com/aws/greengrass/deployment/DeploymentConfigMergerTest.java
```diff
@@ -822,4 +839,310 @@ void GIVEN_endpoint_switch_WHEN_preflight_fails_THEN_deployment_fails_no_state_c
                 future.get().getDeploymentStatus());
         verify(deploymentActivator, never()).activate(any(), any(), any(Long.class), any());
     }
+
+    @Test
+    void GIVEN_credential_endpoint_returns_200_WHEN_verifyCredentialEndpoint_THEN_returns_true() throws Exception {
+        IotCloudHelper iotCloudHelper = mock(IotCloudHelper.class);
```
alter-mage: The mock setup for `IotCloudHelper`, `IotConnectionManager`, and `thingName` is duplicated across all 8 tests. Extract it into a helper method to reduce noise and make maintenance easier.

src/test/java/com/aws/greengrass/deployment/DeploymentConfigMergerTest.java
```diff
@@ -822,4 +839,310 @@ void GIVEN_endpoint_switch_WHEN_preflight_fails_THEN_deployment_fails_no_state_c
...
+        IotConnectionManager iotConnectionManager = mock(IotConnectionManager.class);
```
alter-mage: Nit: `iotConnectionManager` is never referenced after being registered in context. Consider inlining the mock: `when(context.get(IotConnectionManager.class)).thenReturn(mock(IotConnectionManager.class))`.

src/test/java/com/aws/greengrass/deployment/DeploymentConfigMergerTest.java
```diff
@@ -822,4 +839,310 @@ void GIVEN_endpoint_switch_WHEN_preflight_fails_THEN_deployment_fails_no_state_c
...
+        when(deviceConfiguration.getThingName()).thenReturn(thingNameTopic);
+
+        IotCloudResponse response = new IotCloudResponse("{\"credentials\":{}}".getBytes(), 200);
+        when(iotCloudHelper.sendHttpRequest(any(), any(), any(), any(), any())).thenReturn(response);
+
+        DeploymentConfigMerger merger = new DeploymentConfigMerger(kernel, deviceConfiguration, validator,
+                executorService);
+        CompletableFuture<DeploymentResult> future = new CompletableFuture<>();
+
+        assertTrue(merger.verifyCredentialEndpoint(future, "new.credentials.iot.us-west-2.amazonaws.com", "GreengrassCoreTokenExchangeRoleAlias", 60000));
```
alter-mage: The unit tests call `verifyCredentialEndpoint` directly via package-private access, coupling them to the internal method signature. Consider testing through `mergeInNewConfig` instead (like the two integration tests at the bottom do) so the tests verify behavior rather than implementation.

src/test/java/com/aws/greengrass/deployment/DeploymentConfigMergerTest.java
```diff
@@ -822,4 +839,310 @@ void GIVEN_endpoint_switch_WHEN_preflight_fails_THEN_deployment_fails_no_state_c
...
+        CompletableFuture<DeploymentResult> future = new CompletableFuture<>();
+
+        // Use short timeout to avoid long test
+        assertFalse(merger.verifyCredentialEndpoint(future, "new.credentials.iot.us-west-2.amazonaws.com", "GreengrassCoreTokenExchangeRoleAlias", 20000));
+        assertTrue(future.isDone());
+        assertEquals(DeploymentResult.DeploymentStatus.FAILED_NO_STATE_CHANGE, future.get().getDeploymentStatus());
+        assertTrue(((DeploymentException) future.get().getFailureCause()).getErrorCodes()
+                .contains(DeploymentErrorCode.CREDENTIAL_ENDPOINT_SERVER_ERROR));
+        // Verify retries happened (more than 1 call)
+        verify(iotCloudHelper, times(2)).sendHttpRequest(any(), any(), any(), any(), any());
```
alter-mage, line 972: Nit: The `times(2)` assertion is implicitly derived from `20000 / CREDENTIAL_CHECK_MAX_RETRY_INTERVAL` — could you add a comment explaining that, or derive it from the constant so it stays correct if the interval changes?

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
sahith: Added credentialEndpointTimeoutMs as a configurable key in DeviceConfiguration with `DEFAULT_CREDENTIAL_ENDPOINT_TIMEOUT_MS = 60_000L`. The caller reads it from device config and falls back to the default if invalid.

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
sahith: Created EndpointSwitchPreflightValidator with `verifyMqttConnectivity()` and `verifyCredentialEndpoint()`. DeploymentConfigMerger creates a single instance and delegates both checks to it.

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
sahith: PMD incorrectly flags `IotConnectionManager` (a singleton resolved from DI context) as a closeable resource that needs closing. We don't own its lifecycle. The suppression is justified.

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
sahith: Replaced raw ints with `HTTP_OK`, `HTTP_UNAUTHORIZED`, `HTTP_FORBIDDEN`, `HTTP_NOT_FOUND`.

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
sahith: InterruptedException is now handled within `verifyCredentialEndpoint` itself: restores the interrupt flag, logs at WARN ("Pre-flight credential endpoint check interrupted"), and completes the future with `FAILED_NO_STATE_CHANGE` and null cause — no misleading `CREDENTIAL_ENDPOINT_SERVER_ERROR`. This matches the pattern used by `verifyMqttConnectivity`.

src/test/java/com/aws/greengrass/deployment/DeploymentConfigMergerTest.java
sahith: Created `setupCredentialEndpointMocks()` helper that returns the mocked IotCloudHelper

src/test/java/com/aws/greengrass/deployment/DeploymentConfigMergerTest.java
sahith: `IotConnectionManager` mock is inlined

src/test/java/com/aws/greengrass/deployment/DeploymentConfigMergerTest.java
sahith: Kept direct unit tests for isolated error code mapping (they verify each HTTP status → error code without full deployment setup overhead). Added a comment explaining the rationale. The two integration tests at the bottom already verify end-to-end through `mergeInNewConfig`.

src/test/java/com/aws/greengrass/deployment/DeploymentConfigMergerTest.java
sahith: Added 20000ms timeout / 10000ms max retry interval = 2 attempts above the verify call.

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
```diff
@@ -169,6 +165,8 @@ private void updateActionForDeployment(Map<String, Object> newConfig, Deployment
         // not customer config). Persisted to config.tlog so it survives device crashes mid-endpoint-switch.
         // Step 5 uses this value to report deployment status back to the source account.
         // Cleared after status reporting; stale keys are harmless and overwritten by the next switch.
```
alter-mage, line 167: `EndpointSwitchPreflightValidator` shouldn't be created with `new` on every deployment invocation. Two options:
1. Inject it via the DI context (like other services) so it's a singleton
2. Make the methods static helpers that take `kernel` and `deviceConfiguration` as parameters

As-is, a new object is allocated on every deployment even when no endpoint switch is happening.

src/test/java/com/aws/greengrass/deployment/DeploymentConfigMergerTest.java
```diff
@@ -822,4 +850,268 @@ void GIVEN_endpoint_switch_WHEN_preflight_fails_THEN_deployment_fails_no_state_c
...
+        runnableCaptor.getValue().run();
+
+        // IotCloudHelper SHOULD be called because role alias changed
+        verify(context).get(IotCloudHelper.class);
+        // Verify the URL includes the new role alias
+        verify(iotCloudHelper).sendHttpRequest(any(), any(),
+                contains("NewRoleAlias"), any(), any());
+        verify(deploymentActivator).activate(any(), any(), any(Long.class), any());
+    }
 }
```
alter-mage, line 1203: Missing test coverage: when **both** `iotCredEndpoint` and `iotRoleAlias` change simultaneously. The existing tests only cover one changing at a time. Add a case that verifies the credential URL uses both new values together.

src/main/java/com/aws/greengrass/deployment/EndpointSwitchPreflightValidator.java
```diff
@@ -0,0 +1,193 @@
...
+                                + thingName + clientIdSuffix + "\" — update iot:Connect resource to "
+                                + "\"arn:aws:iot:*:*:client/" + thingName + "*\", or "
+                                + "(2) device certificate is not authorized in the target account");
+            } else {
+                logger.atError().setEventType(MERGE_ERROR_LOG_EVENT_KEY).setCause(e)
+                        .kv(ENDPOINT_LOG_KEY, endpoint)
+                        .log("Pre-flight MQTT connectivity check failed");
+            }
+            totallyCompleteFuture.complete(
+                    new DeploymentResult(DeploymentResult.DeploymentStatus.FAILED_NO_STATE_CHANGE, e));
```
alter-mage, line 96: Nit: `PMD.CloseResource` suppression isn't needed here — this method doesn't open any closeable resource. `IotCloudHelper` and `IotConnectionManager` are singletons from context. The suppression makes sense on `verifyMqttConnectivity` (which opens `StandaloneMqttConnector`) but not here.

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
sahith: Updated `EndpointSwitchPreflightValidator` to a DI singleton resolved via `kernel.getContext().get(EndpointSwitchPreflightValidator.class)`. Constructor is @Inject-annotated with `IotCloudHelper` and `IotConnectionManager` injected as fields.

src/test/java/com/aws/greengrass/deployment/DeploymentConfigMergerTest.java
sahith: Added `GIVEN_both_cred_endpoint_and_role_alias_changed_WHEN_endpoint_switch_THEN_credential_check_uses_both_new_values` — uses ArgumentCaptor to verify the credential URL contains both new values.

src/main/java/com/aws/greengrass/deployment/EndpointSwitchPreflightValidator.java
sahith: With `IotConnectionManager` now stored as a field (injected via constructor), PMD no longer flags it. Removed the suppression — only `PMD.AvoidCatchingGenericException` remains. PMD passes clean.

alter-mage (approved):

xuchuwan (approved):

PR #1794, opened 2026-05-18
feat(deployment): report terminal deployment status to source account after endpoint switch

sahith:
## Summary

After a successful IoT endpoint switch, the main MQTT client connects to the destination endpoint. The source endpoint's IoT Job or deployment shadow never receives the terminal deployment status (SUCCEEDED/FAILED) because the device is no longer connected there. This PR adds source endpoint status reporting via a standalone MQTT connection.

- **IoT Jobs path**: `IotJobsHelper` consumer detects endpoint-switch deployment, publishes SUCCEEDED to source IoT Job via standalone MQTT
- **Shadow path**: `ShadowDeploymentListener` consumer publishes unconditional shadow update (no version field) to source endpoint
- **EndpointSwitchState**: New class managing persisted state (source endpoint, deployment ID) for in-flight endpoint-switch deployments
- **StandaloneMqttConnector.publish()**: Reusable publish with exponential backoff and retry
- **IN_PROGRESS guard**: Dropped after nucleus restart to avoid stale publishes to the destination endpoint

## Design decisions

- Consumer-level routing within existing `DeploymentStatusKeeper` flow (inherits persistence + crash recovery)
- `GsonBuilder().disableHtmlEscaping()` matches `IotJobsClientWrapper` wire format
- Synchronous reporting (deployment is already terminal)
- Best-effort: failure logged but non-fatal (source job times out)
- Shadow update omits `version` field (unconditional update per AWS docs)

## E2E test results

Tested on containerized Greengrass Nucleus Classic devices with cross-region endpoint switch (us-east-1 to us-west-2):

| Test Case | Path | Result |
|-----------|------|--------|
| SUCCEEDED endpoint switch | IoT Jobs (thing group) | Source IoT Job shows SUCCEEDED |
| SUCCEEDED endpoint switch | Shadow (core device) | Source shadow reported.status = SUCCEEDED |
| Pre-flight failure (unreachable endpoint) | IoT Jobs | FAILED via normal path, no state change |
| Normal deployment (no endpoint change) | Shadow | No source routing, existing flow unmodified |
| Round-trip switch (east to west to east) | IoT Jobs | Both directions report correctly |

## Test plan

- [x] `EndpointSwitchStateTest` - 9 tests covering persist, detect, clear, edge cases
- [x] `IotJobsHelperTest` - source routing, ID mismatch, rollback detection, IN_PROGRESS guard, standalone failure, payload shape verification
- [x] `ShadowDeploymentListenerTest` - same patterns for shadow path, payload has no version field
- [x] `DefaultActivatorTest` - endpoint switch forces snapshot/rollback
- [x] `StandaloneMqttConnectorTest` - publish success, retry, failure after exhaustion
- [x] `DeploymentConfigMergerTest` - source keys persisted after pre-flight
- [x] Full test suite: 1174 tests pass, 0 failures

### DIFF
--- a/src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
+++ b/src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
@@ -62,10 +62,7 @@ public class DeploymentConfigMerger {
     public static final String DEPLOYMENT_ID_LOG_KEY = "deploymentId";
     public static final String SERVICE_NAME_LOG_KEY = "serviceName";
     protected static final int WAIT_SVC_START_POLL_INTERVAL_MILLISEC = 1000;
-    static final long DEFAULT_PREFLIGHT_MQTT_TIMEOUT_MS = 60_000;
-    static final String STANDALONE_MQTT_TIMEOUT_KEY = "standaloneMqttTimeoutMs";
     private static final String ENDPOINT_LOG_KEY = "endpoint";
-    private static final String ENDPOINT_SWITCH_CLIENT_SUFFIX = "#endpoint-switch";
 
     private static final Logger logger = LogManager.getLogger(DeploymentConfigMerger.class);
 
@@ -161,10 +158,20 @@ public class DeploymentConfigMerger {
 
         // Endpoint-switch deployments: pre-flight connectivity check, then persist source endpoint.
         // Pre-flight runs BEFORE persisting so that on failure, no device state has changed.
-        // Source endpoint is stored under services.DeploymentService.runtime (internal deployment state,
-        // not customer config). Persisted to config.tlog so it survives device crashes mid-endpoint-switch.
-        // Step 5 uses this value to report deployment status back to the source account.
-        // Cleared after status reporting; stale keys are harmless and overwritten by the next switch.
+        //
+        // Before switching the IoT endpoint, save two values to runtime config:
+        //
+        // 1. sourceIotDataEndpoint — the current IoT data endpoint (where the device is connected
+        //    now). After the switch, the status consumer uses this to publish the deployment result
+        //    back to the originating account via a standalone MQTT connection.
+        //
+        // 2. sourceDeploymentId — this deployment's ID (job ID, config ARN, or UUID depending on
+        //    type). The status consumer matches this against the persisted status entry to identify
+        //    which entry belongs to the endpoint-switch deployment.
+        //
+        // Both are stored under services.DeploymentService.runtime (internal deployment state, not
+        // customer config) and persisted to config.tlog (survives crashes). Removed by the status
+        // consumer after it successfully publishes the deployment result to the source endpoint.
         EndpointSwitchPreflightValidator preflightValidator =
                 kernel.getContext().get(EndpointSwitchPreflightValidator.class);
         if (isEndpointSwitchDeployment(nucleusConfig, deviceConfiguration)) {
@@ -179,16 +186,9 @@ public class DeploymentConfigMerger {
             logger.atInfo().setEventType(MERGE_CONFIG_EVENT_KEY)
                     .kv(ENDPOINT_LOG_KEY, newDataEndpoint)
                     .log("Starting pre-flight MQTT connectivity check");
-            long preflightTimeout = Coerce.toLong(deviceConfiguration.getMQTTNamespace()
-                    .findOrDefault(DEFAULT_PREFLIGHT_MQTT_TIMEOUT_MS, STANDALONE_MQTT_TIMEOUT_KEY));
-            if (preflightTimeout <= 0) {
-                logger.atWarn().kv("configured", preflightTimeout)
-                        .kv("using", DEFAULT_PREFLIGHT_MQTT_TIMEOUT_MS)
-                        .log("Invalid standaloneMqttTimeoutMs, using default");
-                preflightTimeout = DEFAULT_PREFLIGHT_MQTT_TIMEOUT_MS;
-            }
+            long preflightTimeout = deviceConfiguration.getStandaloneMqttTimeout();
             if (!preflightValidator.verifyMqttConnectivity(totallyCompleteFuture, newDataEndpoint,
-                    ENDPOINT_SWITCH_CLIENT_SUFFIX, preflightTimeout)) {
+                    preflightTimeout)) {
                 return;
             }
             logger.atInfo().setEventType(MERGE_CONFIG_EVENT_KEY)
@@ -196,8 +196,7 @@ public class DeploymentConfigMerger {
                     .log("Pre-flight MQTT connectivity check passed");
 
             // Persist source endpoint only after pre-flight succeeds — no state change on failure
-            kernel.getContext().get(DeploymentService.class).getRuntimeConfig()
-                    .lookup(DeploymentService.SOURCE_IOT_DATA_ENDPOINT_KEY).withValue(currentDataEndpoint);
+            kernel.getContext().get(EndpointSwitchState.class).persist(currentDataEndpoint, deployment.getId());
         }
 
         // Pre-flight credential endpoint check — uses the device certificate for mTLS authentication,
--- a/src/main/java/com/aws/greengrass/deployment/DeploymentService.java
+++ b/src/main/java/com/aws/greengrass/deployment/DeploymentService.java
@@ -98,7 +98,6 @@ public class DeploymentService extends GreengrassService {
     public static final String GROUP_TO_LAST_DEPLOYMENT_CONFIG_ARN_KEY = "configArn";
     public static final String GROUP_TO_ROOT_COMPONENTS_VERSION_KEY = "version";
     public static final String GROUP_TO_ROOT_COMPONENTS_GROUP_CONFIG_ARN = "groupConfigArn";
-    public static final String SOURCE_IOT_DATA_ENDPOINT_KEY = "sourceIotDataEndpoint";
     public static final String GROUP_TO_ROOT_COMPONENTS_GROUP_NAME = "groupConfigName";
     public static final String DEPLOYMENT_DETAILED_STATUS_KEY = "detailed-deployment-status";
     public static final String DEPLOYMENT_FAILURE_CAUSE_KEY = "deployment-failure-cause";
--- a/src/main/java/com/aws/greengrass/deployment/EndpointSwitchState.java
+++ b/src/main/java/com/aws/greengrass/deployment/EndpointSwitchState.java
@@ -1,0 +1,107 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.aws.greengrass.deployment;
+
+import com.aws.greengrass.config.Topic;
+import com.aws.greengrass.config.Topics;
+import com.aws.greengrass.util.Coerce;
+import com.aws.greengrass.util.Utils;
+
+import javax.inject.Inject;
+
+/**
+ * Manages persisted state for in-flight endpoint-switch deployments.
+ *
+ * <p>Before an endpoint switch, {@link DeploymentConfigMerger} persists the source IoT data endpoint and
+ * deployment ID into DeploymentService's runtime config. After the switch completes, the status consumers
+ * ({@link IotJobsHelper}, {@link ShadowDeploymentListener}) use this class to identify the endpoint-switch
+ * deployment and report terminal status back to the source endpoint via a standalone MQTT connection.</p>
+ *
+ * <p>Keys are stored under {@code services.DeploymentService.runtime} and persisted to config.tlog
+ * (survives crashes). Cleared after terminal status is reported.</p>
+ */
+public class EndpointSwitchState {
+
+    static final String SOURCE_IOT_DATA_ENDPOINT_KEY = "sourceIotDataEndpoint";
+    static final String SOURCE_DEPLOYMENT_ID_KEY = "sourceDeploymentId";
+    static final String ENDPOINT_SWITCH_CLIENT_SUFFIX = "#endpoint-switch";
+    static final long DEFAULT_STANDALONE_MQTT_TIMEOUT_MS = 60_000;
+    static final String STANDALONE_MQTT_TIMEOUT_KEY = "standaloneMqttTimeoutMs";
+
+    private final DeploymentService deploymentService;
+
+    @Inject
+    EndpointSwitchState(DeploymentService deploymentService) {
+        this.deploymentService = deploymentService;
+    }
+
+    /**
+     * Check if the given deployment is the in-flight endpoint-switch deployment by matching
+     * the deployment ID against the persisted source deployment ID.
+     *
+     * @param deploymentId the deployment ID to check (job ID for IoT Jobs, config ARN for Shadow)
+     * @return true if this is the endpoint-switch deployment
+     */
+    public boolean isEndpointSwitchDeployment(String deploymentId) {
+        String sourceEndpoint = getSourceIotDataEndpoint();
+        if (sourceEndpoint == null) {
+            return false;
+        }
+        String sourceDepId = getSourceDeploymentId();
+        return deploymentId != null && deploymentId.equals(sourceDepId);
+    }
+
+    /**
+     * Read the IoT data endpoint where the endpoint-switch deployment originated.
+     * This is the endpoint the device was connected to before the switch, used to report terminal
+     * deployment status back to the source endpoint via a standalone MQTT connection.
+     *
+     * @return the source IoT data endpoint, or null if no endpoint switch is in progress
+     */
+    public String getSourceIotDataEndpoint() {
+        String sourceEndpoint = Coerce.toString(getRuntimeConfig().find(SOURCE_IOT_DATA_ENDPOINT_KEY));
+        if (Utils.isEmpty(sourceEndpoint)) {
+            return null;
+        }
+        return sourceEndpoint;
+    }
+
+    /**
+     * Persist the source endpoint and deployment ID before applying the endpoint switch.
+     * Called by {@link DeploymentConfigMerger} after pre-flight passes.
+     *
+     * @param sourceEndpoint the current IoT data endpoint (before switch)
+     * @param deploymentId   the deployment ID initiating the switch
+     */
+    public void persist(String sourceEndpoint, String deploymentId) {
+        getRuntimeConfig().lookup(SOURCE_IOT_DATA_ENDPOINT_KEY).withValue(sourceEndpoint);
+        getRuntimeConfig().lookup(SOURCE_DEPLOYMENT_ID_KEY).withValue(deploymentId);
+    }
+
+    /**
+     * Remove the persisted source IoT data endpoint and deployment ID from runtime config.
+     * Called after the endpoint-switch deployment's terminal status has been published to the
+     * source endpoint (or publish failed — keys are cleared regardless to avoid stale state).
+     */
+    public void clear() {
+        Topic sourceEndpointTopic = getRuntimeConfig().find(SOURCE_IOT_DATA_ENDPOINT_KEY);
+        if (sourceEndpointTopic != null) {
+            sourceEndpointTopic.remove();
+        }
+        Topic sourceDeploymentIdTopic = getRuntimeConfig().find(SOURCE_DEPLOYMENT_ID_KEY);
+        if (sourceDeploymentIdTopic != null) {
+            sourceDeploymentIdTopic.remove();
+        }
+    }
+
+    private String getSourceDeploymentId() {
+        return Coerce.toString(getRuntimeConfig().find(SOURCE_DEPLOYMENT_ID_KEY));
+    }
+
+    private Topics getRuntimeConfig() {
+        return deploymentService.getRuntimeConfig();
+    }
+}
--- a/src/main/java/com/aws/greengrass/deployment/IotJobsHelper.java
+++ b/src/main/java/com/aws/greengrass/deployment/IotJobsHelper.java
@@ -17,7 +17,9 @@ import com.aws.greengrass.lifecyclemanager.exceptions.ServiceLoadException;
 import com.aws.greengrass.logging.api.Logger;
 import com.aws.greengrass.logging.impl.LogManager;
 import com.aws.greengrass.mqttclient.MqttClient;
+import com.aws.greengrass.mqttclient.StandaloneMqttConnector;
 import com.aws.greengrass.mqttclient.WrapperMqttClientConnection;
+import com.aws.greengrass.security.SecurityService;
 import com.aws.greengrass.status.FleetStatusService;
 import com.aws.greengrass.status.model.Trigger;
 import com.aws.greengrass.util.Coerce;
@@ -26,6 +28,9 @@ import com.aws.greengrass.util.LockScope;
 import com.aws.greengrass.util.SerializerFactory;
 import com.aws.greengrass.util.Utils;
 import com.fasterxml.jackson.core.JsonProcessingException;
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import lombok.AccessLevel;
 import lombok.Getter;
 import lombok.NoArgsConstructor;
@@ -46,6 +51,7 @@ import software.amazon.awssdk.iot.iotjobs.model.RejectedError;
 import software.amazon.awssdk.iot.iotjobs.model.UpdateJobExecutionRequest;
 import software.amazon.awssdk.iot.iotjobs.model.UpdateJobExecutionSubscriptionRequest;
 
+import java.nio.charset.StandardCharsets;
 import java.time.Duration;
 import java.time.Instant;
 import java.util.HashMap;
@@ -97,6 +103,10 @@ public class IotJobsHelper implements InjectionActions {
             "Interrupted while subscribing to Iot jobs event notifications topic";
     private static final String JOB_ID_LOG_KEY_NAME = "JobId";
     private static final String NEXT_JOB_LITERAL = "$next";
+    /**
+     * IoT Jobs status detail values are limited to 1024 characters by the service.
+     */
+    private static final int JOB_STATUS_DETAIL_MAX_LENGTH = 1024;
     // Sometimes when we are notified that a new job is queued and request the next pending job document immediately,
     // we get an empty response. This unprocessedJobs is to track the number of new queued jobs that we are notified
     // with, and keep retrying the request until we get a non-empty response.
@@ -107,6 +117,7 @@ public class IotJobsHelper implements InjectionActions {
     private static final Logger logger = LogManager.getLogger(IotJobsHelper.class);
     private static final long WAIT_TIME_MS_TO_SUBSCRIBE_AGAIN = Duration.ofMinutes(2).toMillis();
     private static final Random RANDOM = new Random();
+    private static final Gson GSON = new GsonBuilder().disableHtmlEscaping().create();
 
     // setter is only used for testing
     @Setter
@@ -361,20 +372,28 @@ public class IotJobsHelper implements InjectionActions {
     private Boolean deploymentStatusChanged(Map<String, Object> deploymentDetails) {
         String jobId = (String) deploymentDetails.get(DEPLOYMENT_ID_KEY_NAME);
         String status = (String) deploymentDetails.get(DEPLOYMENT_STATUS_KEY_NAME);
+
+        // Endpoint-switch deployment: report SUCCEEDED to the source endpoint via standalone MQTT.
+        //
+        // After a successful switch the main MQTT client is on the destination endpoint, so the
+        // IoT Job at the source endpoint can only be updated via a standalone connection.
+        //
+        // FAILED (rollback): the device is still on the source endpoint (sourceEndpoint ==
+        // currentEndpoint), so the block clears keys and falls through — the existing
+        // updateJobStatus() below publishes FAILED via the main MQTT client which is still
+        // connected to the source endpoint.
+        //
+        // IN_PROGRESS: only reaches this block after a nucleus restart (the source endpoint
+        // already received IN_PROGRESS before the restart). Dropped to avoid publishing to the
+        // destination endpoint where the IoT Job doesn't exist.
+        if (handleEndpointSwitchStatus(jobId, status, deploymentDetails)) {
+            return true;
+        }
+
+        @SuppressWarnings("unchecked")
         Map<String, Object> statusDetails =
                 (Map<String, Object>) deploymentDetails.get(DEPLOYMENT_STATUS_DETAILS_KEY_NAME);
-        HashMap<String, String> jobStatusDetails = new HashMap<>();
-        statusDetails.forEach((k, v) -> {
-            if (v instanceof String) {
-                jobStatusDetails.put(k, (String) v);
-            } else {
-                try {
-                    jobStatusDetails.put(k, SerializerFactory.getFailSafeJsonObjectMapper().writeValueAsString(v));
-                } catch (JsonProcessingException e) {
-                    logger.atWarn().kv("status-detail-value", v).setCause(e).log("Failed to serialize status detail");
-                }
-            }
-        });
+        HashMap<String, String> jobStatusDetails = buildJobStatusDetails(statusDetails);
         logger.atInfo().kv(JOB_ID_LOG_KEY_NAME, jobId).kv(STATUS_LOG_KEY_NAME, status)
                 .kv("StatusDetails", statusDetails).log("Updating status of persisted deployment");
         try {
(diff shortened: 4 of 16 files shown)


src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
```diff
@@ -497,6 +531,47 @@ private MqttClientConnection getMqttClientConnection() {
...
+                    sourceEndpoint, EndpointSwitchState.ENDPOINT_SWITCH_CLIENT_SUFFIX)) {
+                connector.connect(connectTimeout);
+                connector.publish(topic, payloadBytes, QualityOfService.AT_LEAST_ONCE,
+                        EndpointSwitchState.DEFAULT_STANDALONE_MQTT_TIMEOUT_MS);
+            }
+            logger.atInfo().kv("configArn", deploymentDetails.get(CONFIGURATION_ARN_KEY_NAME))
+                    .kv("status", deploymentDetails.get(DEPLOYMENT_STATUS_KEY_NAME))
+                    .kv("sourceEndpoint", sourceEndpoint)
+                    .log("Reported terminal shadow status to source endpoint");
+        } catch (Exception e) {
```
MikeDombo, line 578: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

src/main/java/com/aws/greengrass/deployment/IotJobsHelper.java
```diff
@@ -664,6 +702,61 @@ private void evaluateCancellationAndCancelDeploymentIfNeeded() {
...
+            // Use Gson with HTML escaping disabled to match the IoT Jobs SDK's wire format
+            // (IotJobsClientWrapper uses the same GsonBuilder configuration)
+            byte[] payload = new GsonBuilder().disableHtmlEscaping().create()
+                    .toJson(request).getBytes(StandardCharsets.UTF_8);
+            connector.publish(topic, payload, QualityOfService.AT_LEAST_ONCE,
+                    EndpointSwitchState.DEFAULT_STANDALONE_MQTT_TIMEOUT_MS);
+            logger.atInfo().kv("jobId", jobId).kv("status", status)
+                    .kv("sourceEndpoint", sourceEndpoint)
+                    .log("Reported terminal job status to source endpoint");
+        } catch (Exception e) {
```
MikeDombo, line 744: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

src/main/java/com/aws/greengrass/mqttclient/StandaloneMqttConnector.java
```diff
@@ -164,6 +166,55 @@ public void connect(long timeoutMs) throws DeploymentException {
...
+                connection.publish(message, qos, false).get(perAttemptTimeout, TimeUnit.MILLISECONDS);
+                logger.atDebug().kv("topic", topic).kv("clientId", clientId)
+                        .log("Standalone MQTT publish succeeded");
+                return null;
+            }, "standalone-mqtt-publish", logger);
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throw new DeploymentException("MQTT publish interrupted", e,
+                    DeploymentErrorCode.MQTT_CONNECTION_FAILED);
+        } catch (Exception e) {
```
MikeDombo, line 234: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

src/main/java/com/aws/greengrass/deployment/IotJobsHelper.java
```diff
@@ -664,6 +702,61 @@ private void evaluateCancellationAndCancelDeploymentIfNeeded() {
         }
     }
 
+    @SuppressWarnings("PMD.AvoidCatchingGenericException")
+    @SuppressFBWarnings("REC_CATCH_EXCEPTION")
+    private void publishStatusToSourceEndpoint(String sourceEndpoint, String jobId,
+                                                  Map<String, Object> deploymentDetails) {
```
xuchuwan, line 720: So Map<String, Object> deploymentDetails can be either a string or a JSON object?

src/main/java/com/aws/greengrass/deployment/IotJobsHelper.java
```diff
@@ -664,6 +702,61 @@ private void evaluateCancellationAndCancelDeploymentIfNeeded() {
...
+            if (v instanceof String) {
+                jobStatusDetails.put(k, (String) v);
+            } else {
+                try {
+                    jobStatusDetails.put(k, SerializerFactory.getFailSafeJsonObjectMapper().writeValueAsString(v));
+                } catch (JsonProcessingException e) {
+                    logger.atWarn().setCause(e).log("Failed to serialize status detail");
+                }
+            }
+        });
```
xuchuwan: This logic could be extract to a function

src/main/java/com/aws/greengrass/deployment/IotJobsHelper.java
```diff
@@ -664,6 +702,61 @@ private void evaluateCancellationAndCancelDeploymentIfNeeded() {
...
+                sourceEndpoint, EndpointSwitchState.ENDPOINT_SWITCH_CLIENT_SUFFIX)) {
+            connector.connect(connectTimeout);
+            UpdateJobExecutionRequest request = new UpdateJobExecutionRequest();
+            request.status = JobStatus.valueOf(status);
+            request.statusDetails = jobStatusDetails;
+            request.thingName = thing;
+            request.jobId = jobId;
+            // Use Gson with HTML escaping disabled to match the IoT Jobs SDK's wire format
+            // (IotJobsClientWrapper uses the same GsonBuilder configuration)
+            byte[] payload = new GsonBuilder().disableHtmlEscaping().create()
```
xuchuwan: nit: GsonBuilder can be static

src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
```diff
@@ -497,6 +531,47 @@ private MqttClientConnection getMqttClientConnection() {
...
+
+            HashMap<String, Object> state = new HashMap<>();
+            state.put("reported", reported);
+
+            HashMap<String, Object> payload = new HashMap<>();
+            payload.put("state", state);
+
+            byte[] payloadBytes = SerializerFactory.getFailSafeJsonObjectMapper().writeValueAsBytes(payload);
+
+            try (StandaloneMqttConnector connector = StandaloneMqttConnector.of(
```
xuchuwan: could we move StandaloneMqttConnector connector resource init into try on line 545 so we don;t need another try catch in 556?

src/main/java/com/aws/greengrass/deployment/activator/DefaultActivator.java
```diff
@@ -57,7 +55,7 @@ public void activate(Map<String, Object> newConfig, Deployment deployment, long
         // Endpoint switch deployments force config snapshot and rollback regardless of FailureHandlingPolicy,
         // so that we can restore the original endpoint if the new one is unreachable after applying the change.
         // Only check when auto-rollback is not already requested to avoid redundant config store reads.
-        boolean forceSnapshotForEndpointSwitch = !autoRollback && isEndpointSwitch();
+        boolean forceSnapshotForEndpointSwitch = !autoRollback && isEndpointSwitch(deployment.getId());
```
Vipul-5517: nit - Why are we doing deployment.getId(). We should add deploymentId variable and use it across

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
```diff
@@ -180,24 +187,24 @@ private void updateActionForDeployment(Map<String, Object> newConfig, Deployment
...
-                    .findOrDefault(DEFAULT_PREFLIGHT_MQTT_TIMEOUT_MS, STANDALONE_MQTT_TIMEOUT_KEY));
+                    .findOrDefault(EndpointSwitchState.DEFAULT_STANDALONE_MQTT_TIMEOUT_MS,
+                            EndpointSwitchState.STANDALONE_MQTT_TIMEOUT_KEY));
             if (preflightTimeout <= 0) {
                 logger.atWarn().kv("configured", preflightTimeout)
-                        .kv("using", DEFAULT_PREFLIGHT_MQTT_TIMEOUT_MS)
+                        .kv("using", EndpointSwitchState.DEFAULT_STANDALONE_MQTT_TIMEOUT_MS)
                         .log("Invalid standaloneMqttTimeoutMs, using default");
-                preflightTimeout = DEFAULT_PREFLIGHT_MQTT_TIMEOUT_MS;
+                preflightTimeout = EndpointSwitchState.DEFAULT_STANDALONE_MQTT_TIMEOUT_MS;
```
Vipul-5517: nit - Lets define it before line 194 and use the variable

src/main/java/com/aws/greengrass/deployment/IotJobsHelper.java
```diff
@@ -664,6 +702,61 @@ private void evaluateCancellationAndCancelDeploymentIfNeeded() {
         }
     }
 
+    @SuppressWarnings("PMD.AvoidCatchingGenericException")
+    @SuppressFBWarnings("REC_CATCH_EXCEPTION")
+    private void publishStatusToSourceEndpoint(String sourceEndpoint, String jobId,
+                                                  Map<String, Object> deploymentDetails) {
+        String status = (String) deploymentDetails.get(DEPLOYMENT_STATUS_KEY_NAME);
+        Map<String, Object> statusDetails =
```
Vipul-5517, line 723: Do we need a null point check here ?

src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
Vipul-5517: +1

src/main/java/com/aws/greengrass/mqttclient/StandaloneMqttConnector.java
```diff
@@ -164,6 +166,55 @@ public void connect(long timeoutMs) throws DeploymentException {
         }
     }
 
+    /**
+     * Publish a message on the already-connected standalone connection with retries using exponential backoff.
```
Vipul-5517, line 189: (Caught using AI reviewer) - The standalone connection uses the device's current cert to authenticate to the source endpoint. After the switch, does the source account still recognize this cert? If the switch involves cert rotation or Thing deletion on the source side, this publish will always fail. Document the assumption that the source Thing + cert remain valid post-switch.

src/main/java/com/aws/greengrass/mqttclient/StandaloneMqttConnector.java
```diff
@@ -164,6 +166,55 @@
...
+        int maxAttempts = Math.max(1, (int) (timeoutMs / perAttemptTimeout));
+        RetryUtils.RetryConfig retryConfig = RetryUtils.RetryConfig.builder()
+                .initialRetryInterval(Duration.ofSeconds(1))
+                .maxRetryInterval(Duration.ofSeconds(10))
+                .maxAttempt(maxAttempts)
+                .retryableExceptions(Collections.singletonList(Exception.class))
+                .build();
+        try {
+            RetryUtils.runWithRetry(retryConfig, () -> {
+                MqttMessage message = new MqttMessage(topic, payload);
```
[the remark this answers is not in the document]
sahith: Replaced deprecated `MqttMessage(String, byte[])` constructor + `MqttClientConnection.publish(MqttMessage, QualityOfService, boolean)` with the non-deprecated `MqttMessage(String, byte[], QualityOfService, boolean)` constructor + `MqttClientConnection.publish(MqttMessage)`. The QoS and retain flag are now embedded in the `MqttMessage` object. Updated tests accordingly.

src/main/java/com/aws/greengrass/mqttclient/StandaloneMqttConnector.java
```diff
@@ -164,6 +166,55 @@
...
+                connection.publish(message, qos, false).get(perAttemptTimeout, TimeUnit.MILLISECONDS);
```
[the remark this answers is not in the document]
sahith: Replaced deprecated `MqttMessage(String, byte[])` constructor + `MqttClientConnection.publish(MqttMessage, QualityOfService, boolean)` with the non-deprecated `MqttMessage(String, byte[], QualityOfService, boolean)` constructor + `MqttClientConnection.publish(MqttMessage)`. The QoS and retain flag are now embedded in the `MqttMessage` object. Updated tests accordingly.

src/main/java/com/aws/greengrass/deployment/IotJobsHelper.java
sahith: The `deploymentDetails` map is always `Map<String, Object>` at this point. The value at `DEPLOYMENT_STATUS_DETAILS_KEY_NAME` is always a `Map<String, Object>` — the String vs non-String distinction happens inside the values of that inner map (some values are Strings like "SUCCESSFUL", others are serialized objects like error stacks).

src/main/java/com/aws/greengrass/deployment/IotJobsHelper.java
sahith: Extracted to `buildJobStatusDetails(Map<String, Object>)` — handles serialization, truncation, and null safety in one method.

src/main/java/com/aws/greengrass/deployment/IotJobsHelper.java
sahith: Added `private static final Gson GSON = new GsonBuilder().disableHtmlEscaping().create()` — Gson instances are thread-safe and immutable after creation.

src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
sahith: Moved `StandaloneMqttConnector.of(...)` into the outer `try(...)` declaration, eliminating the nested try-catch.

src/main/java/com/aws/greengrass/deployment/activator/DefaultActivator.java
sahith: Added `String deploymentId = deployment.getId()` and use it across all 3 call sites.

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
sahith: Extracted `long defaultTimeout = EndpointSwitchState.DEFAULT_STANDALONE_MQTT_TIMEOUT_MS` and reuse it in the 3 places it was repeated.

src/main/java/com/aws/greengrass/deployment/IotJobsHelper.java
sahith: The extracted `buildJobStatusDetails()` method returns an empty map if statusDetails is null

src/main/java/com/aws/greengrass/mqttclient/StandaloneMqttConnector.java
sahith: Added Javadoc on `publish()` documenting the assumption: source Thing + cert must remain valid post-switch for the publish to succeed. If they're deregistered, the publish fails non-fatally and the source IoT Job eventually times out.

src/main/java/com/aws/greengrass/mqttclient/StandaloneMqttConnector.java
sahith: These are best-effort status reporting paths — if publish to the source endpoint fails for any reason, we log and move on. The deployment is already complete on the device. We have
`@SuppressWarnings("PMD.AvoidCatchingGenericException")` and `@SuppressFBWarnings("REC_CATCH_EXCEPTION")` annotations to indicate this is deliberate.

src/main/java/com/aws/greengrass/deployment/ShadowDeploymentListener.java
sahith: These are best-effort status reporting paths — if publish to the source endpoint fails for any reason, we log and move on. The deployment is already complete on the device. We have `@SuppressWarnings("PMD.AvoidCatchingGenericException")` and `@SuppressFBWarnings("REC_CATCH_EXCEPTION")` annotations to indicate this is deliberate.

Vipul-5517 (dismissed):

xuchuwan (dismissed):

alter-mage (commented):
Overall this looks good — clean separation with `EndpointSwitchState`, proper resource cleanup via try-with-resources, and solid test coverage. A few suggestions below, mostly around reducing duplication and defensive coding.

src/main/java/com/aws/greengrass/deployment/IotJobsHelper.java
```diff
@@ -361,6 +372,34 @@ private void setupCommWithIotJobs(Boolean isConfigurationUpdate) {
...
+                publishStatusToSourceEndpoint(sourceEndpoint, jobId, deploymentDetails);
+                endpointSwitchState.clear();
+                return true;
+            }
+            endpointSwitchState.clear();
+        }
+
         Map<String, Object> statusDetails =
                 (Map<String, Object>) deploymentDetails.get(DEPLOYMENT_STATUS_DETAILS_KEY_NAME);
         HashMap<String, String> jobStatusDetails = new HashMap<>();
```
alter-mage: Nit: The inline statusDetails conversion logic here (lines 403-416) is now duplicated by `buildJobStatusDetails()` below. Consider refactoring this block to call `buildJobStatusDetails(statusDetails)` as well — it would also give you consistent truncation in both paths.

src/main/java/com/aws/greengrass/deployment/DeploymentService.java
```diff
@@ -422,6 +421,7 @@ private void finishCurrentDeployment() throws InterruptedException {
             logger.atInfo().kv(DEPLOYMENT_ID_LOG_KEY_NAME, currentDeploymentTaskMetadata.getDeploymentId())
                     .kv(GG_DEPLOYMENT_ID_LOG_KEY_NAME, ggDeploymentId).log("Deployment task is cancelled");
         }
+
```
alter-mage: Nit: Stray blank line here (and another at line 1062).

src/test/java/com/aws/greengrass/deployment/ShadowDeploymentListenerTest.java
```diff
@@ -129,4 +144,176 @@ public void testCommunicationWithIotCore_unsuccessful_THEN_retry_on_update() thr
...
+        doReturn(CompletableFuture.completedFuture(null)).when(mockIotShadowClient)
+                .SubscribeToGetNamedShadowAccepted(any(), any(), any(), any());
+
+        when(mockEndpointSwitchState.isEndpointSwitchDeployment("arn:aws:greengrass:us-east-1:123:config")).thenReturn(true);
+        when(mockEndpointSwitchState.getSourceIotDataEndpoint()).thenReturn("source.iot.us-east-1.amazonaws.com");
+
+        Topics mqttNamespace = mock(Topics.class);
+        when(mockDeviceConfiguration.getMQTTNamespace()).thenReturn(mqttNamespace);
+        when(mqttNamespace.findOrDefault(eq(60_000L), eq("standaloneMqttTimeoutMs"))).thenReturn(60_000L);
+
```
alter-mage, line 167: Nit: A few lines in this file exceed 120 chars. Consider extracting the ARN string to a constant like `TEST_CONFIG_ARN`.

src/main/java/com/aws/greengrass/deployment/IotJobsHelper.java
```diff
@@ -361,6 +372,34 @@ private void setupCommWithIotJobs(Boolean isConfigurationUpdate) {
...
+        // FAILED (rollback): the device is still on the source endpoint (sourceEndpoint ==
+        // currentEndpoint), so the block clears keys and falls through — the existing
+        // updateJobStatus() below publishes FAILED via the main MQTT client which is still
+        // connected to the source endpoint.
+        //
+        // IN_PROGRESS: only reaches this block after a nucleus restart (the source endpoint
+        // already received IN_PROGRESS before the restart). Dropped to avoid publishing to the
+        // destination endpoint where the IoT Job doesn't exist.
+        EndpointSwitchState endpointSwitchState = kernel.getContext().get(EndpointSwitchState.class);
+        if (endpointSwitchState.isEndpointSwitchDeployment(jobId)) {
```
alter-mage: Would it be worth extracting the endpoint-switch handling (lines 389-401) into its own method like `handleEndpointSwitchStatus()`? It would keep `deploymentStatusChanged()` focused on the normal path and make the early-return semantics more explicit.

```java
/**
 * @return true if this status update was handled (caller should return), false to continue normal path
 */
private boolean handleEndpointSwitchStatus(String jobId, String status,
                                           Map<String, Object> deploymentDetails) {
    EndpointSwitchState endpointSwitchState = kernel.getContext().get(EndpointSwitchState.class);
    if (!endpointSwitchState.isEndpointSwitchDeployment(jobId)) {
        return false;
    }
    if (JobStatus.IN_PROGRESS.toString().equals(status)) {
        return true;
    }
    String sourceEndpoint = endpointSwitchState.getSourceIotDataEndpoint();
    if (sourceEndpoint != null
            && !sourceEndpoint.equals(Coerce.toString(deviceConfiguration.getIotDataEndpoint()))) {
        publishStatusToSourceEndpoint(sourceEndpoint, jobId, deploymentDetails);
    }
    endpointSwitchState.clear();
    return !Objects.equals(sourceEndpoint, Coerce.toString(deviceConfiguration.getIotDataEndpoint()));
}
```

src/main/java/com/aws/greengrass/deployment/IotJobsHelper.java
```diff
@@ -664,6 +704,66 @@ private void evaluateCancellationAndCancelDeploymentIfNeeded() {
...
+        String status = (String) deploymentDetails.get(DEPLOYMENT_STATUS_KEY_NAME);
+        @SuppressWarnings("unchecked")
+        Map<String, Object> statusDetails =
+                (Map<String, Object>) deploymentDetails.get(DEPLOYMENT_STATUS_DETAILS_KEY_NAME);
+        HashMap<String, String> jobStatusDetails = buildJobStatusDetails(statusDetails);
+
+        String thing = Coerce.toString(deviceConfiguration.getThingName());
+        String topic = String.format(IotJobsClientWrapper.UPDATE_JOB_TOPIC, thing, jobId);
+        long connectTimeout = Coerce.toLong(deviceConfiguration.getMQTTNamespace()
+                .findOrDefault(EndpointSwitchState.DEFAULT_STANDALONE_MQTT_TIMEOUT_MS,
```
alter-mage: The timeout-read-and-validate pattern (read from config, fallback if ≤ 0) is repeated in `DeploymentConfigMerger`, here, and `ShadowDeploymentListener`. Would a helper like `EndpointSwitchState.getStandaloneMqttTimeout(DeviceConfiguration)` make sense to consolidate this?

```java
// In EndpointSwitchState.java:
public long getStandaloneMqttTimeout(DeviceConfiguration deviceConfiguration) {
    long timeout = Coerce.toLong(deviceConfiguration.getMQTTNamespace()
            .findOrDefault(DEFAULT_STANDALONE_MQTT_TIMEOUT_MS, STANDALONE_MQTT_TIMEOUT_KEY));
    return timeout > 0 ? timeout : DEFAULT_STANDALONE_MQTT_TIMEOUT_MS;
}
```

Then callers become:
```java
long connectTimeout = endpointSwitchState.getStandaloneMqttTimeout(deviceConfiguration);
```

src/main/java/com/aws/greengrass/deployment/EndpointSwitchState.java
```diff
@@ -0,0 +1,107 @@
...
+ * deployment and report terminal status back to the source endpoint via a standalone MQTT connection.</p>
+ *
+ * <p>Keys are stored under {@code services.DeploymentService.runtime} and persisted to config.tlog
+ * (survives crashes). Cleared after terminal status is reported.</p>
+ */
+public class EndpointSwitchState {
+
+    static final String SOURCE_IOT_DATA_ENDPOINT_KEY = "sourceIotDataEndpoint";
+    static final String SOURCE_DEPLOYMENT_ID_KEY = "sourceDeploymentId";
+    static final String ENDPOINT_SWITCH_CLIENT_SUFFIX = "#endpoint-switch";
```
alter-mage, line 30: These constants (`DEFAULT_STANDALONE_MQTT_TIMEOUT_MS`, `STANDALONE_MQTT_TIMEOUT_KEY`, `ENDPOINT_SWITCH_CLIENT_SUFFIX`) are accessed from 4 different classes. Should they be `public`, or would encapsulating them behind methods (e.g., the timeout helper suggested on `IotJobsHelper`) be a better approach?

If you go with the helper method for timeout, the only remaining external constant access is `ENDPOINT_SWITCH_CLIENT_SUFFIX`. That could be hidden too:

```java
public StandaloneMqttConnector createSourceConnector(SecurityService securityService,
                                                     DeviceConfiguration deviceConfiguration,
                                                     String sourceEndpoint) {
    return StandaloneMqttConnector.of(securityService, deviceConfiguration,
            sourceEndpoint, ENDPOINT_SWITCH_CLIENT_SUFFIX);
}
```

src/main/java/com/aws/greengrass/deployment/IotJobsHelper.java
```diff
@@ -361,6 +372,34 @@ private void setupCommWithIotJobs(Boolean isConfigurationUpdate) {
...
+        //
+        // IN_PROGRESS: only reaches this block after a nucleus restart (the source endpoint
+        // already received IN_PROGRESS before the restart). Dropped to avoid publishing to the
+        // destination endpoint where the IoT Job doesn't exist.
+        EndpointSwitchState endpointSwitchState = kernel.getContext().get(EndpointSwitchState.class);
+        if (endpointSwitchState.isEndpointSwitchDeployment(jobId)) {
+            if (JobStatus.IN_PROGRESS.toString().equals(status)) {
+                return true;
+            }
+            String sourceEndpoint = endpointSwitchState.getSourceIotDataEndpoint();
```
alter-mage: `getSourceIotDataEndpoint()` is called here after `isEndpointSwitchDeployment()` already confirmed it's non-null internally — but that's two separate reads from config. Is there a concern about a null return on this second read, or is it safe to assume the value can't disappear between the two calls?

One option — add a null guard:
```java
String sourceEndpoint = endpointSwitchState.getSourceIotDataEndpoint();
if (sourceEndpoint == null) {
    endpointSwitchState.clear();
    return false; // state was cleared concurrently, fall through to normal path
}
if (!sourceEndpoint.equals(Coerce.toString(deviceConfiguration.getIotDataEndpoint()))) {
```

Alternatively, `isEndpointSwitchDeployment()` could return the source endpoint directly (null = not an endpoint switch):
```java
String sourceEndpoint = endpointSwitchState.getEndpointSwitchSource(jobId);
if (sourceEndpoint != null) {
    // handle endpoint switch...
}
```

src/main/java/com/aws/greengrass/deployment/IotJobsHelper.java
sahith: Replaced the inline statusDetails conversion with `buildJobStatusDetails(statusDetails)`. Both the normal path and the endpoint-switch path now share the same method.

src/main/java/com/aws/greengrass/deployment/DeploymentService.java
sahith: Removed it

src/test/java/com/aws/greengrass/deployment/ShadowDeploymentListenerTest.java
sahith: Extracted `TEST_CONFIG_ARN` constant and replaced all 8 occurrences

src/main/java/com/aws/greengrass/deployment/IotJobsHelper.java
sahith: Added `EndpointSwitchState.getStandaloneMqttTimeout(DeviceConfiguration)` which encapsulates the read-validate-fallback pattern. All three callers now use it.

src/main/java/com/aws/greengrass/deployment/EndpointSwitchState.java
sahith: Added `EndpointSwitchState.createSourceConnector(...)` factory that hides ENDPOINT_SWITCH_CLIENT_SUFFIX. External callers no longer access the constants directly.

src/main/java/com/aws/greengrass/deployment/IotJobsHelper.java
sahith: Added a null guard in both `handleEndpointSwitchStatus()` implementations. If the value is null (concurrent clear), we clear any remaining state and fall through to the normal path.

src/main/java/com/aws/greengrass/deployment/EndpointSwitchState.java
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    /**
+     * Get the standalone MQTT timeout from device configuration, falling back to the default if the
+     * configured value is invalid (≤ 0).
+     *
+     * @param deviceConfiguration device configuration to read timeout from
+     * @return timeout in milliseconds (always positive)
+     */
+    public static long getStandaloneMqttTimeout(DeviceConfiguration deviceConfiguration) {
```
alter-mage: `EndpointSwitchState` started as a clean state holder (persist/read/clear), but these new `static` methods don't touch any instance state — they're utility/factory methods that live here only because the constants are here. `getStandaloneMqttTimeout` is really about `DeviceConfiguration` (it reads from it, it's about it), and `createSourceConnector` is a factory for `StandaloneMqttConnector`.

Not a blocker, but worth considering whether these belong on their respective classes instead (`DeviceConfiguration.getStandaloneMqttTimeout()` and `StandaloneMqttConnector.forEndpointSwitch(...)`) to keep this class focused on state management.

src/main/java/com/aws/greengrass/deployment/IotJobsHelper.java
```diff
@@ -664,6 +684,90 @@ private void evaluateCancellationAndCancelDeploymentIfNeeded() {
         }
     }
 
+    /**
+     * Handle endpoint-switch status reporting. If this deployment is an endpoint switch,
+     * report terminal status to the source endpoint via standalone MQTT and clear persisted state.
+     *
+     * @return true if this status update was handled (caller should return), false to continue normal path
+     */
+    private boolean handleEndpointSwitchStatus(String jobId, String status,
```
alter-mage, line 693: The `handleEndpointSwitchStatus` implementations in `IotJobsHelper` and `ShadowDeploymentListener` are nearly identical — same structure, same null check, same clear logic. The only difference is the publish call signature. Not a problem at two call sites, but if a third consumer ever appears, consider extracting the shared control flow (e.g., a shared method that takes a publish callback).

alter-mage (commented):
Follow-up review on the refactored changes — looks much cleaner. Two design observations below.

src/main/java/com/aws/greengrass/deployment/EndpointSwitchState.java
```diff
@@ -0,0 +1,140 @@
...
+        long timeout = Coerce.toLong(deviceConfiguration.getMQTTNamespace()
+                .findOrDefault(DEFAULT_STANDALONE_MQTT_TIMEOUT_MS, STANDALONE_MQTT_TIMEOUT_KEY));
+        return timeout > 0 ? timeout : DEFAULT_STANDALONE_MQTT_TIMEOUT_MS;
```
alter-mage: Nit: The original code in `DeploymentConfigMerger` logged a warning when the configured timeout was ≤ 0. This silent fallback loses that observability — a misconfigured `standaloneMqttTimeoutMs` will now be ignored without any log. Consider adding a `logger.atWarn()` here when `timeout <= 0`.

alter-mage (commented):
Follow-up review on the refactored changes — looks much cleaner. A few observations below.

src/main/java/com/aws/greengrass/deployment/IotJobsHelper.java
sahith: The two implementations are intentionally parallel (~10 lines each) with only the publish call differing. Adding a shared method with a callback would introduce a functional interface + generics for minimal benefit at two call sites. If a third consumer appears, we'll extract. Leaving as-is for now.

src/main/java/com/aws/greengrass/deployment/EndpointSwitchState.java
sahith: Moved `getStandaloneMqttTimeout()` to `DeviceConfiguration` (it reads from MQTT namespace, so it naturally belongs there). Moved the connector factory to `StandaloneMqttConnector.forEndpointSwitch(...)` (factory on the class it creates). EndpointSwitchState is now purely a state holder — *persist/read/clear*.

src/main/java/com/aws/greengrass/deployment/EndpointSwitchState.java
sahith: `DeviceConfiguration.getStandaloneMqttTimeout()` now logs a warning when the configured value is ≤ 0, matching the original behavior in `DeploymentConfigMerger`.

alter-mage (approved):

Vipul-5517 (approved):

PR #1797, opened 2026-05-20
fix: remove sudo from reboot command in loader script

### LINKED ISSUE #1795: Loader script: remove `sudo` from reboot command to enable PATH-based override
## Summary

Change line 88 of the Nucleus loader script (`scripts/loader`) from `sudo reboot` to just `reboot`.

## Motivation

Customers would like the ability to override the reboot command used by Greengrass when the Nucleus exits with code 101 (reboot requested). Use cases include custom shutdown sequencing, graceful application teardown, or using alternative reboot mechanisms on embedded/industrial devices.

The loader script cannot be persistently modified because Nucleus updates replace it (the loader re-execs itself from the new distribution on line 83). This means any local patch is lost on upgrade.

If the loader called `reboot` without `sudo`, customers could override the reboot behavior durably via a systemd drop-in that prepends a custom directory to `PATH`:

```ini
# /etc/systemd/system/greengrass.service.d/override.conf
[Service]
Environment="PATH=/opt/greengrass-overrides:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
```

With `sudo` present, this approach is impossible because `sudo` resolves commands via its own `secure_path` (from `/etc/sudoers`), ignoring the caller's `$PATH`.

## Why removing `sudo` is safe

- **Running as root (default installation):** `sudo` is a no-op — root already has permission to reboot.
- **Running as non-root:** The Greengrass installation docs already require configuring the service user with appropriate permissions. The user should have direct reboot capability (via polkit, sudoers NOPASSWD for reboot, or CAP_SYS_BOOT) rather than relying on a hardcoded `sudo` in the loader.

## Proposed change

```diff
-        sudo reboot
+        reboot
```

In `scripts/loader`, line 88.

aws-kevinrickard:
## Summary

Remove the `sudo` prefix from the `reboot` call in `scripts/loader` (line 88).

## Motivation

Customers need the ability to override the reboot command used when the Nucleus exits with code 101. Since the loader script is replaced on every Nucleus update, local patches are not durable.

Without `sudo`, customers can override reboot behavior via a systemd drop-in that prepends a custom directory to `PATH`:

```ini
# /etc/systemd/system/greengrass.service.d/override.conf
[Service]
Environment="PATH=/opt/greengrass-overrides:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
```

With `sudo` present, this is impossible because `sudo` resolves commands via its own `secure_path`, ignoring the caller's `$PATH`.

## Why this is safe

- **Running as root (default):** `sudo` is a no-op.
- **Running as non-root:** The service user should already have reboot permissions configured (polkit, sudoers, or CAP_SYS_BOOT).

## Testing

This is a one-line shell script change with no behavioral difference for the default (root) installation. Non-root installations already require explicit permission configuration.

Fixes #1795

### DIFF
--- a/scripts/loader
+++ b/scripts/loader
@@ -85,7 +85,7 @@ while [ $j -le 3 ] && [ $sigterm_received -eq 0 ];  do
     ;;
   101)
     echo "Rebooting host"
-    sudo reboot
+    reboot
     exit 0
     ;;
   *)


junfuchen99 (approved):

scripts/loader
```diff
@@ -85,7 +85,7 @@ while [ $j -le 3 ] && [ $sigterm_received -eq 0 ];  do
     ;;
   101)
     echo "Rebooting host"
-    sudo reboot
+    reboot
```
muboheng, line 88: will this affect customer that is running as non-root, has sudo permission, but no native reboot permission

aws-kevinrickard: It would but that would mean that the user running greengrass was not set up with the right permission (permission to run reboot should be required).

This is a good call-out though as our documentation does not mention this as a requirement: https://docs.aws.amazon.com/greengrass/v2/developerguide/setup-greengrass-non-root.html ; CAP_SYS_BOOT controls this and is not mentioned. I believe this is a gap that should be fixed by adding it to the required capabilities list. I'll update that as a follow up.

muboheng (approved):

PR #1798, opened 2026-05-20
fix: cancel in-flight subscribes on IPC stream close

archigup:
When a component's IPC stream closes while a SubscribeToIoTCore operation is still pending on the MQTT connection, the pending subscribe future was never cancelled. On reconnect, all accumulated pending subscribes would fire to the broker, leaking subscription count entries and eventually exceeding the 50-subscription-per-client limit.

Track a per-subscribe cancellation flag in MqttClient. On unsubscribe, set the flag so the pending connect().thenCompose chain skips the CRT subscribe call entirely. If the broker subscribe already completed in a race, issue a compensating unsubscribe.

### DIFF
--- a/src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
+++ b/src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
@@ -138,6 +138,7 @@ public class MqttClient implements Closeable {
     private final List<IndividualMqttClient> connections = new CopyOnWriteArrayList<>();
     private final Map<Subscribe, IndividualMqttClient> subscriptions = new ConcurrentHashMap<>();
     private final Map<MqttTopic, IndividualMqttClient> subscriptionTopics = new ConcurrentHashMap<>();
+    private final Map<Subscribe, AtomicBoolean> subscribeCancellations = new ConcurrentHashMap<>();
     private final Set<Integer> activeClientIds = new HashSet<>();
     private final AtomicInteger connectionRoundRobin = new AtomicInteger(0);
     @Getter
@@ -437,7 +438,7 @@ public class MqttClient implements Closeable {
      * @param request subscribe request
      * @throws MqttRequestException if the request is invalid for any reason
      */
-    @SuppressWarnings("PMD.CloseResource")
+    @SuppressWarnings({"PMD.CloseResource", "PMD.AvoidDeeplyNestedIfStmts"})
     public CompletableFuture<SubscribeResponse> subscribe(Subscribe request)
             throws MqttRequestException {
         try (LockScope ls = LockScope.lock(lock)) {
@@ -472,8 +473,28 @@ public class MqttClient implements Closeable {
             // Connection isn't null, so we should subscribe to the topic
             if (connection != null) {
                 IndividualMqttClient finalConnection = connection;
-                return connection.subscribe(request).whenComplete((i, t) -> {
+                AtomicBoolean cancelled = new AtomicBoolean(false);
+                subscribeCancellations.put(request, cancelled);
+                return connection.connect().thenCompose(ignored -> {
+                    if (cancelled.get()) {
+                        return CompletableFuture.completedFuture(null);
+                    }
+                    return finalConnection.subscribe(request);
+                }).whenComplete((i, t) -> {
                     try (LockScope scope = LockScope.lock(connectionLock.readLock())) {
+                        subscribeCancellations.remove(request);
+                        if (cancelled.get()) {
+                            // Cancelled while in-flight; undo if broker subscribe already completed
+                            if (t == null && i != null && i.isSuccessful()) {
+                                finalConnection.unsubscribe(request.getTopic()).whenComplete((r, err) -> {
+                                    if (err != null) {
+                                        logger.atWarn().kv(TOPIC_KEY, request.getTopic())
+                                                .log("Failed to unsubscribe cancelled subscription", err);
+                                    }
+                                });
+                            }
+                            return;
+                        }
                         if (t == null && (i == null || i.isSuccessful())) {
                             subscriptionTopics.put(new MqttTopic(request.getTopic()), finalConnection);
                         } else {
@@ -571,6 +592,10 @@ public class MqttClient implements Closeable {
                 for (Map.Entry<Subscribe, IndividualMqttClient> sub : subscriptions.entrySet()) {
                     if (sub.getKey().getCallback() == request.getSubscriptionCallback() && sub.getKey().getTopic()
                             .equals(request.getTopic())) {
+                        AtomicBoolean cancelled = subscribeCancellations.remove(sub.getKey());
+                        if (cancelled != null) {
+                            cancelled.set(true);
+                        }
                         subscriptions.remove(sub.getKey());
                     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -472,8 +473,28 @@ public CompletableFuture<SubscribeResponse> subscribe(Subscribe request)
...
+                        return CompletableFuture.completedFuture(null);
+                    }
+                    return finalConnection.subscribe(request);
+                }).whenComplete((i, t) -> {
                     try (LockScope scope = LockScope.lock(connectionLock.readLock())) {
+                        subscribeCancellations.remove(request);
+                        if (cancelled.get()) {
+                            // Cancelled while in-flight; undo if broker subscribe already completed
+                            if (t == null && i != null && i.isSuccessful()) {
+                                finalConnection.unsubscribe(request.getTopic()).whenComplete((r, err) -> {
```
MikeDombo, line 489: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

This code is using non-async method `whenComplete` on `CompletableFuture` (or `CompletionStage`). Please be aware that task subscribed to `CompletableFuture`(or `CompletionStage`) through non-async method may be executed by the thread that completes the current `CompletableFuture` (or `CompletionStage`), or the thread that calls this non-async method to add subscription. In other words, the thread that executes the subscription is non-deterministic. If you prefer having fully control over threads, you may consider using the async variants instead. In addition, please make sure tasks subscribed to `CompletableFuture` (or `CompletionStage`) are short lived and non-blocking to avoid deadlock. Learn more: [https://docs.oracle.com/en/java/javase/18/docs/api/java.base/java/util/concurrent/CompletableFuture.html](https://docs.oracle.com/en/java/javase/18/docs/api/java.base/java/util/concurrent/CompletableFuture.html)

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -472,8 +473,28 @@ public CompletableFuture<SubscribeResponse> subscribe(Subscribe request)
             // Connection isn't null, so we should subscribe to the topic
             if (connection != null) {
                 IndividualMqttClient finalConnection = connection;
-                return connection.subscribe(request).whenComplete((i, t) -> {
+                AtomicBoolean cancelled = new AtomicBoolean(false);
+                subscribeCancellations.put(request, cancelled);
+                return connection.connect().thenCompose(ignored -> {
```
MikeDombo, line 478: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

This code is using non-async method `whenComplete` on `CompletableFuture` (or `CompletionStage`). Please be aware that task subscribed to `CompletableFuture`(or `CompletionStage`) through non-async method may be executed by the thread that completes the current `CompletableFuture` (or `CompletionStage`), or the thread that calls this non-async method to add subscription. In other words, the thread that executes the subscription is non-deterministic. If you prefer having fully control over threads, you may consider using the async variants instead. In addition, please make sure tasks subscribed to `CompletableFuture` (or `CompletionStage`) are short lived and non-blocking to avoid deadlock. Learn more: [https://docs.oracle.com/en/java/javase/18/docs/api/java.base/java/util/concurrent/CompletableFuture.html](https://docs.oracle.com/en/java/javase/18/docs/api/java.base/java/util/concurrent/CompletableFuture.html)

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -472,8 +473,28 @@ public CompletableFuture<SubscribeResponse> subscribe(Subscribe request)
             // Connection isn't null, so we should subscribe to the topic
             if (connection != null) {
                 IndividualMqttClient finalConnection = connection;
-                return connection.subscribe(request).whenComplete((i, t) -> {
+                AtomicBoolean cancelled = new AtomicBoolean(false);
```
alter-mage, line 476: Could you add a test that exercises the cancellation path? For example: start a subscribe with a delayed/mocked `connect()` future, call `unsubscribe()` before it completes, then verify that `subscriptionTopics` has no orphaned entry and the cleanup unsubscribe fires if the broker subscribe already succeeded.

alter-mage (commented):
Logic looks correct — the two-phase cancellation check (before subscribe + in whenComplete) covers the race. One request below.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
MikeDombo, line 478: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

This code is using non-async method `thenCompose` on `CompletableFuture` (or `CompletionStage`). Please be aware that task subscribed to `CompletableFuture`(or `CompletionStage`) through non-async method may be executed by the thread that completes the current `CompletableFuture` (or `CompletionStage`), or the thread that calls this non-async method to add subscription. In other words, the thread that executes the subscription is non-deterministic. If you prefer having fully control over threads, you may consider using the async variants instead. In addition, please make sure tasks subscribed to `CompletableFuture` (or `CompletionStage`) are short lived and non-blocking to avoid deadlock. Learn more: [https://docs.oracle.com/en/java/javase/18/docs/api/java.base/java/util/concurrent/CompletableFuture.html](https://docs.oracle.com/en/java/javase/18/docs/api/java.base/java/util/concurrent/CompletableFuture.html)

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -437,7 +438,7 @@ private void validateAndSetMqttPublishConfiguration() {
      * @param request subscribe request
      * @throws MqttRequestException if the request is invalid for any reason
      */
-    @SuppressWarnings("PMD.CloseResource")
+    @SuppressWarnings({"PMD.CloseResource", "PMD.AvoidDeeplyNestedIfStmts"})
     public CompletableFuture<SubscribeResponse> subscribe(Subscribe request)
```
aws-kevinrickard, line 442: nit: The future can complete with a `null` SubscribeResponse — worth documenting that in the Javadoc:

```java
@return future that completes with the broker's response, or {@code null} if the
       subscribe was not sent to the broker
```

This is an existing thing of course, but I think it's worth adding now that we're adding more null return paths.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
aws-kevinrickard, line 131: nit (pre-existing): `connectionLock` is a raw `ReentrantReadWriteLock` that bypasses `LockFactory`'s `CycleDetectingLockFactory`. The other locks in this class (`lock`, `httpProxyLock`, `spoolingFutureLock`) all go through the factory, so ordering violations involving them get caught in tests — but any cycle involving `connectionLock` won't.

Guava's `CycleDetectingLockFactory` supports `newReentrantReadWriteLock(name)`. Could add a wrapper to `LockFactory` and use it here so the `lock` → `connectionLock` ordering invariant is enforced at test time.

Not blocking since it's pre-existing and the ordering in this PR is correct.

aws-kevinrickard (approved):

alter-mage (approved):

muboheng (approved):

PR #1807, opened 2026-06-22
fix: prevent preflight endpoint check from failing on unprovisioned devices

rauthsuj:
**Issue #, if available:**

**Description of changes:**
Normalize null and empty-string to equivalent values when comparing IoT credential endpoint and role alias during deployment pre-flight checks. Added effectivelyChanged() helper that treats null and "" as the same "not configured" state, and applied it to both the credential endpoint comparison and isEndpointSwitchDeployment.

On unprovisioned devices, DeviceConfiguration getters use Topic.dflt("") which lazily initializes null Topic values to "". However, the resolved deployment config map (built via toPOJO()) reads the raw Topic value before dflt() has been triggered. This causes a false-positive change detection ("" != null) that fires the EndpointSwitchPreflightValidator credential endpoint check, which immediately fails with "Role alias must not be empty" and blocking all local deployments on unprovisioned devices.

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
+++ b/src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
@@ -28,6 +28,7 @@ import com.aws.greengrass.lifecyclemanager.exceptions.ServiceLoadException;
 import com.aws.greengrass.logging.api.Logger;
 import com.aws.greengrass.logging.impl.LogManager;
 import com.aws.greengrass.util.Coerce;
+import com.aws.greengrass.util.Utils;
 import lombok.AccessLevel;
 import lombok.AllArgsConstructor;
 import lombok.Getter;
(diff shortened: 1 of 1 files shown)


muboheng (approved):

src/main/java/com/aws/greengrass/deployment/DeploymentConfigMerger.java
```diff
@@ -322,6 +323,12 @@ private String tryGetIoTDataEndpointFromNewConfig(Map<String, Object> kernelConf
         return iotDataEndpoint;
     }
 
+    private static boolean effectivelyChanged(String current, String incoming) {
+        String normalizedCurrent = Utils.isEmpty(current) ? "" : current;
+        String normalizedIncoming = Utils.isEmpty(incoming) ? "" : incoming;
+        return !Objects.equals(normalizedCurrent, normalizedIncoming);
```
aws-kevinrickard, line 329: nit: the normalized parameters can no longer be null, so `!normalizedCurrent.equals(normalizedIncoming)` should be sufficient

aws-kevinrickard (approved):

PR #1813, opened 2026-07-01
feat(ipc): add SubscribeToIoTCoreConnectionStatus operation support

rawalexe:
**Issue #, if available:**

**Description of changes:**
Add support for the SubscribeToIoTCoreConnectionStatus IPC operation introduced in aws-iot-device-sdk 1.33.0. Components can subscribe to receive the IoT Core MQTT connection status (`CONNECTED`/`DISCONNECTED`) as a stream of `IoTCoreConnectionStatusEvent` messages.

Bump dependencies:
* AWS IoT Device SDK v2: 1.31.0 -> 1.33.0

aws-crt is intentionally kept at 0.45.0 (not bumped to the 0.47.0 pinned by device SDK 1.33.0).

Vendor the generated server-side stubs for the new operation:
* GreengrassCoreIPCService: add operation constant, register it in the service operation set, and add setSubscribeToIoTCoreConnectionStatusHandler
* GeneratedAbstractSubscribeToIoTCoreConnectionStatusOperationHandler (new)

Implement the operation in the MQTT proxy (`MqttProxyIPCAgent` / `MqttProxyIPCService`):
* Register shared connection callbacks with the MQTT client once, on first subscription, via `CallbackEventManager` (which deduplicates events across underlying connections)
* Send the current connection status as the initial stream event so subscribers learn the present state without waiting for the next transition
* Guarantee the non-streaming subscribe response precedes any stream event, and that a status transition racing subscription setup is stashed and delivered as the initial event rather than dropped — both paths hold a single lock, so nothing is lost and wire order matches status order (same approach as client-device-auth's `SubscribeToCertificateUpdatesOperationHandler`)
* Remove subscribers on stream close and self-heal on stream-send failure to prevent listener leaks

Design decision — no authorization required: this is a local informational operation that exposes no MQTT topic data, following the same pattern as other ungated informational IPC operations (e.g. SubscribeToConfigurationUpdate). It is intentionally not registered with the AuthorizationHandler, and components do not need an `accessControl` policy to use it.

Components (e.g. stream processors, offline-first apps) need a supported way to react to the core device's IoT Core connectivity state without probing the network themselves. The device SDKs (Java v2 1.33.0+) already ship the client-side model for this operation; this PR adds the nucleus server-side implementation.

Unit tests (`MqttProxyIPCAgentTest`, 12 tests) cover the full event sequencing contract: no events forwarded before the subscribe response is sent, initial status delivered after the response, DISCONNECTED forwarded on connection interruption, CONNECTED forwarded on resume/initial connect, and no events after stream close. A dedicated test covers the race where a transition arrives during subscription setup, asserting it is delivered as the initial event exactly once instead of being dropped.

End-to-end verification on a real core device (aarch64 container, nucleus built from this branch, provisioned against IoT Core): a Java SDK v2 test component subscribed via IPC and observed the complete lifecycle —
```
[INITIAL] Connection status: CONNECTED      (initial event on subscribe)
[CHANGED] Connection status: DISCONNECTED   (network severed; detected via MQTT keepalive)
[CHANGED] Connection status: CONNECTED      (network restored)
```
Exactly one event per transition, in order, with no duplicates.

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/builtin/services/mqttproxy/MqttProxyIPCAgent.java
+++ b/src/main/java/com/aws/greengrass/builtin/services/mqttproxy/MqttProxyIPCAgent.java
@@ -10,6 +10,7 @@ import com.aws.greengrass.authorization.Permission;
 import com.aws.greengrass.authorization.exceptions.AuthorizationException;
 import com.aws.greengrass.logging.api.Logger;
 import com.aws.greengrass.logging.impl.LogManager;
+import com.aws.greengrass.mqttclient.CallbackEventManager;
 import com.aws.greengrass.mqttclient.MqttClient;
 import com.aws.greengrass.mqttclient.MqttRequestException;
 import com.aws.greengrass.mqttclient.spool.SpoolerStoreException;
@@ -22,8 +23,12 @@ import com.aws.greengrass.util.Utils;
 import lombok.AccessLevel;
 import lombok.Setter;
 import software.amazon.awssdk.aws.greengrass.GeneratedAbstractPublishToIoTCoreOperationHandler;
+import software.amazon.awssdk.aws.greengrass.GeneratedAbstractSubscribeToIoTCoreConnectionStatusOperationHandler;
 import software.amazon.awssdk.aws.greengrass.GeneratedAbstractSubscribeToIoTCoreOperationHandler;
+import software.amazon.awssdk.aws.greengrass.model.ConnectionStatus;
+import software.amazon.awssdk.aws.greengrass.model.ConnectionStatusEvent;
 import software.amazon.awssdk.aws.greengrass.model.InvalidArgumentsError;
+import software.amazon.awssdk.aws.greengrass.model.IoTCoreConnectionStatusEvent;
 import software.amazon.awssdk.aws.greengrass.model.IoTCoreMessage;
 import software.amazon.awssdk.aws.greengrass.model.MQTTMessage;
 import software.amazon.awssdk.aws.greengrass.model.PayloadFormat;
@@ -31,14 +36,19 @@ import software.amazon.awssdk.aws.greengrass.model.PublishToIoTCoreRequest;
 import software.amazon.awssdk.aws.greengrass.model.PublishToIoTCoreResponse;
 import software.amazon.awssdk.aws.greengrass.model.QOS;
 import software.amazon.awssdk.aws.greengrass.model.ServiceError;
+import software.amazon.awssdk.aws.greengrass.model.SubscribeToIoTCoreConnectionStatusRequest;
+import software.amazon.awssdk.aws.greengrass.model.SubscribeToIoTCoreConnectionStatusResponse;
 import software.amazon.awssdk.aws.greengrass.model.SubscribeToIoTCoreRequest;
 import software.amazon.awssdk.aws.greengrass.model.SubscribeToIoTCoreResponse;
 import software.amazon.awssdk.aws.greengrass.model.UnauthorizedError;
+import software.amazon.awssdk.crt.mqtt.MqttClientConnectionEvents;
 import software.amazon.awssdk.crt.mqtt5.packets.SubAckPacket;
 import software.amazon.awssdk.eventstreamrpc.OperationContinuationHandlerContext;
 import software.amazon.awssdk.eventstreamrpc.model.EventStreamJsonMessage;
 
+import java.util.Set;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CopyOnWriteArraySet;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
@@ -67,6 +77,29 @@ public class MqttProxyIPCAgent {
     @Setter(AccessLevel.PACKAGE)
     private AuthorizationHandler authorizationHandler;
 
+    private final Set<SubscribeToIoTCoreConnectionStatusOperationHandler> connectionStatusListeners =
+            new CopyOnWriteArraySet<>();
+
+    private final AtomicBoolean connectionStatusCallbacksRegistered = new AtomicBoolean(false);
+
+    // Broadcasts IoT Core connection status changes to all IPC subscribers. CallbackEventManager
+    // deduplicates events across the underlying connections, so these fire once per online/offline
+    // transition of the overall MQTT client.
+    private final MqttClientConnectionEvents connectionStatusCallbacks = new MqttClientConnectionEvents() {
+        @Override
+        public void onConnectionInterrupted(int errorCode) {
+            forwardConnectionStatusToListeners(ConnectionStatus.DISCONNECTED);
+        }
+
+        @Override
+        public void onConnectionResumed(boolean sessionPresent) {
+            forwardConnectionStatusToListeners(ConnectionStatus.CONNECTED);
+        }
+    };
+
+    private final CallbackEventManager.OnConnectCallback onInitialConnect =
+            connectionStatusCallbacks::onConnectionResumed;
+
     public PublishToIoTCoreOperationHandler getPublishToIoTCoreOperationHandler(
             OperationContinuationHandlerContext context) {
         return new PublishToIoTCoreOperationHandler(context);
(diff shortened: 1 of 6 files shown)


src/main/java/com/aws/greengrass/builtin/services/mqttproxy/MqttProxyIPCAgent.java
```diff
@@ -275,6 +328,98 @@ private void forwardToSubscriber(Publish m) {
...
+                return;
+            }
+
+            sendStatusEvent(status);
+        }
+
+        private void sendStatusEvent(ConnectionStatus status) {
+            IoTCoreConnectionStatusEvent event = new IoTCoreConnectionStatusEvent()
+                    .withConnectionStatusEvent(new ConnectionStatusEvent().withStatus(status));
+            this.sendStreamEvent(event).exceptionally((t) -> {
```
MikeDombo, line 410: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

This code is using non-async method `exceptionally` on `CompletableFuture` (or `CompletionStage`). Please be aware that task subscribed to `CompletableFuture`(or `CompletionStage`) through non-async method may be executed by the thread that completes the current `CompletableFuture` (or `CompletionStage`), or the thread that calls this non-async method to add subscription. In other words, the thread that executes the subscription is non-deterministic. If you prefer having fully control over threads, you may consider using the async variants instead. In addition, please make sure tasks subscribed to `CompletableFuture` (or `CompletionStage`) are short lived and non-blocking to avoid deadlock. Learn more: [https://docs.oracle.com/en/java/javase/18/docs/api/java.base/java/util/concurrent/CompletableFuture.html](https://docs.oracle.com/en/java/javase/18/docs/api/java.base/java/util/concurrent/CompletableFuture.html)

rawalexe: Keeping the non-async `exceptionally` here, for three reasons:

1. The callback is short and non-blocking — a log statement plus a `ConcurrentLinkedQueue.remove`, safe to run on whichever thread completes the future.
2. It matches the established pattern for these same eventstream futures: the embedded `software.amazon.awssdk.eventstreamrpc` layer in this repo chains non-async `whenComplete`/`exceptionally` on `sendStreamEvent`/`sendMessage` futures in six places (e.g. `OperationContinuationHandler` lines 207/248/277/336/349).
3. `exceptionallyAsync` is Java 12+; this project targets Java 8 (`maven.compiler.target` 1.8), so the suggested variant isn't available without hand-rolling `handle(...).thenComposeAsync` — extra complexity with no benefit given (1).

The Coral note doesn't apply — this is a CRT eventstream continuation future, not a CoralClient one.

src/main/java/com/aws/greengrass/builtin/services/mqttproxy/MqttProxyIPCAgent.java
```diff
@@ -275,6 +328,92 @@ private void forwardToSubscriber(Publish m) {
...
+        public void afterHandleRequest() {
+            state.set(STATE_SENDING_INITIAL);
+            // Send the current connection status as the initial event so that subscribers
+            // learn the present state without waiting for the next transition.
+            sendStatusEvent(
+                    mqttClient.getMqttOnline().get() ? ConnectionStatus.CONNECTED
+                            : ConnectionStatus.DISCONNECTED);
+            // Note: a status change arriving during SENDING_INITIAL is dropped; the subscriber
+            // will catch up on the next transition.
+            state.set(STATE_ACTIVE);
```
saranyailla:
```
synchronized (firstStreamingEvent) {
      if (firstSent.compareAndSet(false, true)) {
          ConnectionStatus status = firstStreamingEvent.get();
          if (status == null) {
              status = mqttClient.getMqttOnline().get() ? ConnectionStatus.CONNECTED
                      : ConnectionStatus.DISCONNECTED;
          }
          sendStatusEvent(status);
          firstStreamingEvent.set(null);
      }
}
```

See this for reference https://github.com/aws-greengrass/aws-greengrass-client-device-auth/blob/main/src/main/java/com/aws/greengrass/ipc/SubscribeToCertificateUpdatesOperationHandler.java#L186

src/main/java/com/aws/greengrass/builtin/services/mqttproxy/MqttProxyIPCAgent.java
```diff
@@ -275,6 +328,92 @@ private void forwardToSubscriber(Publish m) {
...
+        void forwardConnectionStatus(ConnectionStatus status) {
+            // Only allow forwarding events if our initial response has been sent already.
+            // If we don't do this, the callback may be invoked and send the streaming response
+            // before the non-streaming SubscribeToIoTCoreConnectionStatusResponse which will
+            // cause a client error.
+            if (state.get() != STATE_ACTIVE) {
+                LOGGER.atWarn().kv(COMPONENT_NAME, serviceName).kv("status", status)
+                        .log("Not forwarding connection status because subscription response "
+                                + "not yet sent or initial status in flight");
+                return;
```
saranyailla: The subscription response has two types of event responses where the first response sent must be the non-streaming `SubscribeToIoTCoreConnectionStatusResponse`, and only after that can streaming `ConnectionStatusEvents` go out. If this fires before `afterHandleRequest` has sent the initial event, stash the status instead of sending it directly; `afterHandleRequest` will send it as the initial event.
```
  synchronized (firstStreamingEvent) {
     if (firstResponseSent.get()) {
          sendStatusEvent(status);
      } else {
          firstStreamingEvent.set(status);
      }
  }
```

src/main/java/com/aws/greengrass/builtin/services/mqttproxy/MqttProxyIPCAgent.java
saranyailla: The 3-state machine has a gap where a connection status transition that arrives while the initial status is being sent gets silently dropped. By using this pattern, we can preserve ordering guarantee without these states.

rawalexe: Adopted — thanks for the CDA pointer, that pattern is strictly better and the state machine is gone.

`afterHandleRequest` now mirrors `SubscribeToCertificateUpdatesOperationHandler`, with the one adaptation this operation needs: unlike a certificate update, there may be no stashed event at all (no transition raced setup), so it falls back to polling `mqttClient.getMqttOnline()` for the current status:

```java
synchronized (firstStreamingEvent) {
    if (subscriptionResponseSent.compareAndSet(false, true)) {
        ConnectionStatus status = firstStreamingEvent.get();
        if (status == null) {
            status = mqttClient.getMqttOnline().get() ? ConnectionStatus.CONNECTED
                    : ConnectionStatus.DISCONNECTED;
        }
        sendStatusEvent(status);
        firstStreamingEvent.set(null);
    }
}
```

You were right about the gap, and my own comment admitted it (`a status change arriving during SENDING_INITIAL is dropped`). Worth spelling out why the new shape closes it rather than just narrowing the window: both paths hold the same monitor, and `afterHandleRequest` holds it across the `sendStatusEvent` call, so a callback arriving mid-send blocks rather than racing, then observes `subscriptionResponseSent == true` and sends normally. Nothing is dropped, and wire order still matches status order.

Pushed as `df9641d4`.

src/main/java/com/aws/greengrass/builtin/services/mqttproxy/MqttProxyIPCAgent.java
rawalexe: Done, exactly as written:

```java
synchronized (firstStreamingEvent) {
    if (subscriptionResponseSent.get()) {
        sendStatusEvent(status);
    } else {
        firstStreamingEvent.set(status);
    }
}
```

The stash branch also logs at debug so the deferral is visible in a support log rather than silent. (It replaced a `WARN` that fired on every subscribe with a racing transition — now it is genuinely benign, since the status is delivered rather than discarded.)

One behavioral consequence worth flagging for review: a transition that arrives during setup now *becomes* the initial event, replacing the polled value. That is the correct outcome — the stashed status is the newer one — and it changed an existing unit test assertion, so I updated that test and added `GIVEN_status_changes_during_subscription_setup_WHEN_response_sent_THEN_it_becomes_initial_event` to pin the behavior down: it asserts the initial event is `DISCONNECTED` (the stashed value) and that exactly one event is sent. `getMqttOnline()` is deliberately left unstubbed there, so a regression back to polling on this path fails the test.

`MqttProxyIPCAgentTest`: 12/12 passing.

src/main/java/com/aws/greengrass/builtin/services/mqttproxy/MqttProxyIPCAgent.java
rawalexe: Follow-up on the exact shape, since the pushed code differs from your snippet in one detail: locking on the `AtomicReference` fails this repo's build. SpotBugs `JLM_JSR166_UTILCONCURRENT_MONITORENTER` rejects it —

```
[ERROR] Synchronization performed on java.util.concurrent.atomic.AtomicReference in
  ...SubscribeToIoTCoreConnectionStatusOperationHandler.afterHandleRequest()
```

client-device-auth does not run this SpotBugs check, so the pattern is fine there but not portable here as-is. Rather than suppress the rule, I kept the semantics and moved the monitor to a dedicated lock object, matching the existing idiom in this repo (`GreengrassService.dependencyReadyLock`):

```java
private final Object initialEventLock = new Object();
private boolean subscriptionResponseSent;
private ConnectionStatus deferredStatus;
```

Both fields are only ever touched while holding `initialEventLock`, so the atomics became redundant once the lock was explicit — `compareAndSet` collapses to `if (!subscriptionResponseSent)`. Ordering and the no-drop guarantee are unchanged; only the monitor identity moved.

Head is now `c5f7548d`. SpotBugs and `MqttProxyIPCAgentTest` (12/12) verified locally before pushing.

saranyailla (approved):

alter-mage (approved):

philbegg:
Which published Nucleus release will include SubscribeToIoTCoreConnectionStatus?
  
We tested against Nucleus 2.18.3, using the JavaScript SDK built from commit 4f055b36867eb6341dddfa60d074d155110f2a97.
 
 IPC authentication succeeds, but activating the subscription returns:
  ```
  {
  "error": "Unsupported Operation",
  "message": "aws.greengrass#SubscribeToIoTCoreConnectionStatus is an unsupported operation."
  }
```
The operation is documented and this PR is merged, but it appears absent from 2.18.3. Is there a planned release containing it, and could the documentation clarify the minimum supported Nucleus version?

rawalexe:
The docs and everything were prepared for a future release, which is a bit delayed for now. Everything was supposed to go together few things fell through. Please use main if you need the feature immediately. We apologize for the confusion.

philbegg:
> The docs and everything were prepared for a future release, which is a bit delayed for now. Everything was supposed to go together few things fell through. Please use main if you need the feature immediately. We apologize for the confusion.

Thanks for the update - its not urgent for us but would like to use it in a future release for some connection monitoring, is there a rough eta on the next release?

rawalexe:
I unfortunately do not have one for you now, maybe @junfuchen99 can provide you one?

PR #1818, opened 2026-07-10
fix: recreate TES HTTP client after a connection failure

aws-kevinrickard:
## Problem

`IotConnectionManager` caches a single `SdkHttpClient` and only rebuilds it when the device's certificate, private key, or root CA file path changes. There is no path to recover from a network-level failure: if the underlying connection pool or DNS resolution held by the client goes stale (for example, after a brief network outage causes a failover from IPv4 to IPv6), every subsequent credential fetch from `CredentialRequestHandler` keeps reusing the same broken client and fails the same way indefinitely. The only recovery today is a manual restart of the TokenExchangeService component or nucleus.

Even with a reset in place, caching the failure for the full error-cache TTL (5 minutes by default) means a transient/local network change still causes a multi-minute credentials outage before the next attempt is made.

## Fix

- Add `IotConnectionManager.reset()`, which discards the cached `SdkHttpClient` so the next `getClient()` call builds a fresh one. This reuses the same invalidation logic previously only triggered by cert/key/CA changes.
- `CredentialRequestHandler` now retries a credential fetch once immediately after a connection-level error (`AWSIotException` / `TLSAuthException`), resetting the connection manager first so the retry goes out on a freshly-built client. Only if that immediate retry also fails does the failure get cached for the existing error-cache TTL, exactly as before.
  - This is a single bounded retry, not an open-ended loop, so it does not add risk of overwhelming the cloud or network: if the network is genuinely down, the retry fails fast against the new client too and behavior falls through to today's TTL-based backoff.
  - If the failure was caused by a stale client (e.g. a local network change), the retry succeeds immediately instead of waiting out the full TTL.
- `IotCloudHelper.sendHttpRequest` gained an overload taking an explicit `maxAttempts`. The post-reset retry uses `maxAttempts=1`, skipping the default overload's own internal 3-attempt/200ms-backoff retry loop for that second attempt. A client that was just rebuilt and fails immediately is unlikely to succeed within a couple hundred more milliseconds of internal retries, so skipping them keeps this method's worst-case added latency bounded to roughly one extra connect/read attempt rather than compounding two independent retry loops (up to 6 total attempts) — which matters because some callers of `getCredentialsBypassCache()` (via `getCredentials()` -> `LazyCredentialProvider`, used as the credentials provider for Nucleus's own internal AWS SDK clients) block indefinitely on the result.
- A connection error that self-heals on the immediate retry is now logged at `DEBUG` instead of `WARN`, since it isn't customer-impacting; `WARN` is only emitted if the retry also fails, so log severity tracks actual customer impact.
- A small jitter (0-200ms) is applied before the immediate retry. Motivation: a shared-endpoint failure (e.g. a regional connectivity blip, or a change rolled out to the credentials endpoint) can cause many devices to fail their current request at the same instant, independent of each device's own cache-expiry schedule (each device's cache expiry is anchored to when it individually last fetched credentials, so cache-driven fetches are already staggered fleet-wide). Without jitter, every affected device's immediate retry would land in the same instant as every other device's retry, with no smoothing between the two waves of requests hitting the shared endpoint.

## Testing

- `IotConnectionManagerTest`: unit test asserting `reset()` causes the next `getClient()` call to return a new instance. Verified this fails against the previous behavior (asserts the same client is returned) and passes with the fix.
- `IotCloudHelperTest`: unit test asserting the `maxAttempts=1` overload does not retry internally, even when a retry would have succeeded.
- `CredentialRequestHandlerTest`:
  - Unit test asserting `reset()` is invoked on the connection manager when a credential fetch throws a connection-level exception, and that both the initial attempt (default overload) and the immediate retry (`maxAttempts=1` overload) are made.
  - Unit test asserting that when the immediate retry succeeds, the caller gets the recovered credentials rather than the cached failure.
  - Verified both of the above fail without the corresponding behavior and pass with it.

## Notes

One thing worth calling out explicitly rather than leaving implicit: `IotConnectionManager.getClient()` and `reset()` both synchronize only on the `client` field read/write, not on the actual request I/O — `IotCloudHelper.sendHttpRequest` fetches the client once, then issues the HTTP call (and any internal retries) outside that lock. So a thread mid-request on the old client can have it closed out from under it by a concurrent `reset()` from another thread hitting a connection error.

This gap already existed before this change (cert/key/CA rotation triggered the same close-without-coordination), but this PR increases how often the trigger fires, since transient connection errors are far more common than cert rotation.

Traced through what actually happens when this race is hit: `ApacheHttpClient`'s connection pool shutdown fails fast (`IllegalStateException`/similar wrapped as `IOException`) rather than hanging or corrupting state, and that surfaces as `AWSIotException`, which is exactly the exception type this PR's own retry-once logic already handles. So the unlucky thread just gets a redundant reset + one bounded extra attempt, not a hang or crash.

Given that, I don't think this needs a code change now, but flagging it as a known, accepted tradeoff rather than something overlooked — happy to discuss if there's appetite for tightening it (e.g. reference-counting in-flight client usage before close) as a follow-up.

### DIFF
--- a/src/main/java/com/aws/greengrass/tes/CredentialRequestHandler.java
+++ b/src/main/java/com/aws/greengrass/tes/CredentialRequestHandler.java
@@ -340,6 +340,11 @@ public class CredentialRequestHandler implements HttpHandler {
             tesCache.get(iotCredentialsPath).responseCode = HttpURLConnection.HTTP_INTERNAL_ERROR;
             tesCache.get(iotCredentialsPath).expiry = newExpiry;
             tesCache.get(iotCredentialsPath).credentials = response;
+            // Discard the cached HTTP client so the next retry (after the cache TTL above) builds a fresh one
+            // instead of reusing a client whose connection pool/DNS resolution may be stale for the current
+            // network state (e.g. after a transient outage causes an IPv4->IPv6 failover). Without this, TES
+            // can get stuck indefinitely reusing a poisoned client until the component is manually restarted.
+            iotConnectionManager.reset();
             LOGGER.atWarn().kv(IOT_CRED_PATH_KEY, iotCredentialsPath)
                     .log("Encountered error while fetching credentials", e);
         } finally {
--- a/src/main/java/com/aws/greengrass/tes/CredentialRequestHandler.java
+++ b/src/main/java/com/aws/greengrass/tes/CredentialRequestHandler.java
@@ -43,6 +43,7 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicReference;
@@ -72,6 +73,14 @@ public class CredentialRequestHandler implements HttpHandler {
     public static final int DEFAULT_CLOUD_4XX_ERROR_CACHE_IN_SEC = 120;
     public static final int DEFAULT_CLOUD_5XX_ERROR_CACHE_IN_SEC = 60;
     public static final int DEFAULT_UNKNOWN_ERROR_CACHE_IN_SEC = 300;
+    // Small jitter window applied before the single immediate retry in sendRequestResettingOnceOnConnectionError().
+    // A shared-endpoint failure (e.g. a regional connectivity blip, or a breaking change rolled out to the
+    // credentials endpoint) causes many devices to fail their *current* request at the same instant, independent
+    // of each device's own cache-expiry schedule (each device's cache expiry is anchored to when it individually
+    // last fetched, so cache-driven fetches are already staggered fleet-wide and are not the concern here).
+    // Without jitter, every affected device's immediate retry would also land in the same instant as every other
+    // device's retry, with no smoothing between the two waves of requests hitting the shared endpoint.
+    private static final int RETRY_JITTER_MAX_MILLIS = 200;
 
     private volatile int cloud4xxErrorCacheInSec = DEFAULT_CLOUD_4XX_ERROR_CACHE_IN_SEC;
     private volatile int cloud5xxErrorCacheInSec = DEFAULT_CLOUD_5XX_ERROR_CACHE_IN_SEC;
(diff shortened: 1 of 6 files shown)


src/main/java/com/aws/greengrass/tes/CredentialRequestHandler.java
```diff
@@ -357,6 +369,86 @@ private byte[] getCredentialsBypassCache() {
...
+     * {@link #sendRequestResettingOnceOnConnectionError()}.
+     *
+     * <p>Without this, a shared-endpoint failure affecting many devices at once (e.g. a regional connectivity
+     * blip, or a breaking change rolled out to the credentials endpoint) would cause every affected device's
+     * retry to land at the exact same instant as every other device's retry, with no smoothing between the two
+     * waves of requests hitting the shared endpoint.</p>
+     */
+    private void sleepWithJitterBeforeRetry() {
+        try {
+            TimeUnit.MILLISECONDS.sleep(ThreadLocalRandom.current().nextInt(RETRY_JITTER_MAX_MILLIS + 1));
```
MikeDombo, line 419: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

Using insecure random number generators can lead to predictable random values. This predictability can be exploited by attackers to guess sensitive values or break cryptographic operations. To fix this, use `new SecureRandom()` without providing an external, fixed, or predictable seed, so that `java.security.SecureRandom` is seeded automatically by the platform&#39;s secure randomness source, rather than using non-cryptographic random number generators (like `java.util.Random`, `Math.random()`, or Apache Commons random utilities). For more information about secure random number generation, see: https://owasp.org/www-community/vulnerabilities/Insecure_Randomness

src/main/java/com/aws/greengrass/tes/CredentialRequestHandler.java
```diff
@@ -357,6 +361,67 @@ private byte[] getCredentialsBypassCache() {
...
+     * @throws AWSIotException  if both the initial attempt and the retry fail with an IoT error
+     * @throws TLSAuthException if both the initial attempt and the retry fail with a TLS/connection error
+     */
+    private IotCloudResponse sendRequestResettingOnceOnConnectionError() throws AWSIotException, TLSAuthException {
+        try {
+            return iotCloudHelper.sendHttpRequest(iotConnectionManager, thingName, iotCredentialsPath,
+                    IOT_CREDENTIALS_HTTP_VERB, null);
+        } catch (AWSIotException | TLSAuthException e) {
+            // Note: TLSAuthException can also come from client construction itself (bad cert/key/CA), in which
+            // case the reset+retry below is a no-op that fails the same way again; that's fine, it's bounded.
```
junfuchen99, line 387: any other scenario where we don't want to recreate http client?

aws-kevinrickard: There is one case — the cert/key/CA files are corrupt or otherwise invalid (as opposed to a permission/expiry issue, which comes back as a 4xx and doesn't hit this path at all) — where recreating the client would fail the same way again. But detecting that case ahead of time would be more work than just doing the bounded retry, so it's handled as an accepted cost.

alter-mage (approved):

junfuchen99 (approved):

PR #1820, opened 2026-07-14
fix: Use USE_LONG_FOR_INTS to prevent unnecessary config.tlog writes

alter-mage:
## Problem

LogManager's `ConfigUtil.updateFromMapWhenChanged()` has a guard to prevent writing unchanged values to config.tlog:

```java
if (!Objects.equals(node.getOnce(), value)) {
    node.withValueChecked(timestamp, value);
}
```

This check **always fails** for numeric values after a reboot because of a type mismatch:

- **Left side:** Jackson deserializes JSON integers from config.tlog as `java.lang.Integer` (default for values fitting in 32 bits)
- **Right side:** LogManager's `long startPosition` field autoboxes to `java.lang.Long`
- **Result:** `Objects.equals(Integer(5518), Long(5518))` → `false` in Java

This causes every tracked file position to be re-written to config.tlog on every processing cycle, regardless of whether it changed.

## Impact

On a customer device with 36 tracked log streams over 182 days:
- config.tlog grew to **773 MB** (3.4M entries)
- **98% of entries** (2.2M) were unnecessary duplicate writes
- Boot time was **~180 seconds** just to parse the tlog on an ARM Cortex-A53

## Fix

Enable `DeserializationFeature.USE_LONG_FOR_INTS` on the ObjectMapper in `Coerce.java`. This causes Jackson to deserialize all JSON integers as `Long`, matching the types LogManager produces. The equality check then correctly prevents no-op writes.

## Testing

Reproduced in Docker (Nucleus 2.15.1 + LogManager 2.3.11):
- **Before fix:** After reboot, unchanged position value `174` was re-written to tlog
- **After fix:** 72 new entries written, **0 duplicates**. Unchanged values correctly skipped.

## Related

- Previous partial fix: LogManager v2.3.11 PR #230 (cleaned up uploaded file entries but did not address the type mismatch for active files)

### DIFF
--- a/src/main/java/com/aws/greengrass/util/Coerce.java
+++ b/src/main/java/com/aws/greengrass/util/Coerce.java
@@ -8,6 +8,7 @@ package com.aws.greengrass.util;
 import com.aws.greengrass.config.Topic;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 
 import java.io.IOException;
@@ -23,7 +24,8 @@ import static com.aws.greengrass.util.Utils.isEmpty;
 public final class Coerce {
     private static final Pattern SEPARATORS = Pattern.compile(" *, *");
     private static final Pattern unwrap = Pattern.compile(" *\\[ *(.*) *\\] *");
-    private static final ObjectMapper MAPPER = new ObjectMapper();
+    private static final ObjectMapper MAPPER = new ObjectMapper()
+            .enable(DeserializationFeature.USE_LONG_FOR_INTS);
 
     private Coerce() {
     }
--- a/src/main/java/com/aws/greengrass/lifecyclemanager/GenericExternalService.java
+++ b/src/main/java/com/aws/greengrass/lifecyclemanager/GenericExternalService.java
@@ -540,7 +540,7 @@ public class GenericExternalService extends GreengrassService {
 
             Topic timeoutTopic = config.find(SERVICE_LIFECYCLE_NAMESPACE_TOPIC, LIFECYCLE_RUN_NAMESPACE_TOPIC,
                     Lifecycle.TIMEOUT_NAMESPACE_TOPIC);
-            Integer timeout = timeoutTopic == null ? null : (Integer) timeoutTopic.getOnce();
+            Integer timeout = timeoutTopic == null ? null : Coerce.toInt(timeoutTopic);
             if (timeout != null) {
                 Exec processToClose = runResult.getExec();
                 context.get(ScheduledExecutorService.class).schedule(() -> {
--- a/src/test/java/com/aws/greengrass/config/ConfigurationTest.java
+++ b/src/test/java/com/aws/greengrass/config/ConfigurationTest.java
@@ -130,7 +130,7 @@ class ConfigurationTest {
         ConfigurationWriter.dump(config, p);
         assertEquals(config.getRoot(), config.getRoot());
         Configuration c2 = ConfigurationReader.createFromTLog(config.context, p);
-        assertEquals(44, c2.lookup("x", "z").getOnce());
+        assertEquals(44L, c2.lookup("x", "z").getOnce());
         assertEquals(config, c2);
         Topic nv = config.lookup("number");
     }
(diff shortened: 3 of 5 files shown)


src/main/java/com/aws/greengrass/util/Coerce.java
```diff
@@ -23,7 +24,8 @@
 public final class Coerce {
     private static final Pattern SEPARATORS = Pattern.compile(" *, *");
     private static final Pattern unwrap = Pattern.compile(" *\\[ *(.*) *\\] *");
-    private static final ObjectMapper MAPPER = new ObjectMapper();
+    private static final ObjectMapper MAPPER = new ObjectMapper()
```
saranyailla, line 27: Is this impacted with this change? https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/6d0b7ae28fb5e1fd8b0272ecc02b35dfad3ac6cd/src/main/java/com/aws/greengrass/lifecyclemanager/GenericExternalService.java#L543

saranyailla: Could you confirm if other plugin components are not impacted by this change?

alter-mage: Good catch — this would throw `ClassCastException` after our change since `getOnce()` now returns `Long` instead of `Integer` for tlog-deserialized values. Fixed in the latest push: replaced `(Integer) timeoutTopic.getOnce()` with `Coerce.toInt(timeoutTopic)`.

alter-mage: Confirmed no other components are impacted. [GitHub org-wide search for `(Integer)` casts in Java](https://github.com/search?q=org%3Aaws-greengrass+%22%28Integer%29%22+language%3AJava&type=code) returns only 2 results: the `GenericExternalService.java` line (now fixed) and a test file in `aws-greengrass-logging-java`. [Searching specifically for `(Integer)` + `getOnce`](https://github.com/search?q=org%3Aaws-greengrass+%22%28Integer%29%22+getOnce+language%3AJava&type=code) confirms only the one we fixed. All plugin components use `Coerce.toInt`/`toLong`/`toDouble` which handle both `Integer` and `Long` via the `Number` interface.

saranyailla (approved):

yitingb (approved):

src/main/java/com/aws/greengrass/util/Coerce.java
```diff
@@ -23,7 +24,8 @@
...
+            .enable(DeserializationFeature.USE_LONG_FOR_INTS);
```
aws-kevinrickard, line 28: The fix works for the reported symptom, but I think the comparison sites are a better place for this fix, and this may be worth a follow-up. Since JSON numbers are typeless, the root issue isn't which type we deserialize to — it's that the config store decides "did this value change?" using `Objects.equals`, which is type-sensitive for boxed numerics.

`Topic.withNewerValue()` has the same check that LogManager's `ConfigUtil` guard duplicates:

```java
if ((Objects.equals(proposed, currentValue) || ...)) {
    return this;   // skip write
}
```

Normalizing `Number` comparison there (compare by `longValue()` for integral types, `doubleValue()` when either side is floating) would fix the whole bug class at the single choke point where the store decides "changed", for all writers and all ingestion paths.

Two concerns with the `USE_LONG_FOR_INTS` approach:

1. **Observable type change for all in-process components.** Every integer config value now comes back from the tlog as `Long`. This PR itself had to fix a `(Integer)` cast in `GenericExternalService` — third-party in-process plugins (e.g. custom provisioning) with similar casts can't be audited and will get `ClassCastException` after upgrading.

2. **The mismatch is relocated, not eliminated.** Other ingestion paths still produce `Integer`: `SerializerFactory.FAIL_SAFE_JSON_OBJECT_MAPPER` (deployment documents) and `Configuration.yamlMapper` (initial config) don't have `USE_LONG_FOR_INTS`. A numeric value merged from a deployment doc enters the store as `Integer`, round-trips through the tlog as `Long`, and the next merge of the same document sees a false diff — spurious changed event, tlog write, and subscriber notifications. Before this change, that path was self-consistent (`Integer` in, `Integer` back). So the fix aligns types with `long`-producing writers like LogManager while misaligning the JSON/YAML-ingestion paths.

Would a follow-up that moves the equality normalization into `Topic.withNewerValue` make sense?

aws-kevinrickard:
Note on the description:

Does the mismatch self-heal after the first write (the stored value becomes Long, so subsequent cycles compare equal)? It's one duplicate write per topic per reboot rather than every processing cycle, which matches the test result of a single duplicate of 174 after reboot. If so, this is still a problem, but not always failing

src/main/java/com/aws/greengrass/util/Coerce.java
aws-kevinrickard: Outcomes: 

We will revert this PR, and the fix is here instead: https://github.com/aws-greengrass/aws-greengrass-log-manager/pull/258

We also discussed the possibility of having nucleus do the deduplication of config writes instead of log manager, that will be looked at later

PR #1822, opened 2026-07-15
feat(mqtt): receive AWS IoT Core direct messages via RECEIVE_ONLY

tushar-aws:
Adds an on-device path for AWS IoT Core direct messages (point-to-point, no cloud subscription). A component subscribes over IPC with subscriptionMode=RECEIVE_ONLY; the Nucleus routes matching inbound messages locally without ever issuing a cloud SUBSCRIBE.
  
  - Subscribe (v5): add skipCloudSubscribe, a routing flag (not an MQTT wire
    option). Unsubscribe is unchanged — local-only removal matches by
    (topic, callback) alone.
  - MqttClient: keep local-only registrations in a separate set; the router
    runs an independent second delivery pass over them, matching each filter
    with MqttTopic.topicIsSupersetOf and invoking each callback in its own
    try/catch. The "no subscriber" drop is logged only when both the cloud
    and local-only passes miss.
  - MqttProxyIPCAgent: SubscribeToIoTCore branches on subscriptionMode.
    RECEIVE_ONLY builds a local-only Subscribe and skips qos validation (qos
    is meaningless with no cloud subscription); SUBSCRIBE/absent/unknown keep
    the existing cloud path. Authorization runs unchanged for both modes.
  - Tests: MqttClientTest (router + local-only delivery), MqttProxyIPCAgentTest
    (handler modes + auth), IPCMqttProxyTest (real IPC), and
    DirectMessageRoutingIntegTest (real IPC handler + real router end to end,
    asserting no cloud SUBSCRIBE).

Note: requires an aws-iot-device-sdk with subscriptionMode; the pom bump to a published SDK is intentionally not committed (kept as a local dev-only change).

**Issue #, if available:**

**Description of changes:**

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/ipc/IPCMqttProxyTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/ipc/IPCMqttProxyTest.java
@@ -30,11 +30,14 @@ import org.junit.jupiter.api.io.TempDir;
 import org.mockito.ArgumentCaptor;
 import software.amazon.awssdk.aws.greengrass.GreengrassCoreIPCClient;
 import software.amazon.awssdk.aws.greengrass.SubscribeToIoTCoreResponseHandler;
+import software.amazon.awssdk.aws.greengrass.model.InvalidArgumentsError;
 import software.amazon.awssdk.aws.greengrass.model.IoTCoreMessage;
 import software.amazon.awssdk.aws.greengrass.model.PublishToIoTCoreRequest;
 import software.amazon.awssdk.aws.greengrass.model.QOS;
 import software.amazon.awssdk.aws.greengrass.model.ServiceError;
 import software.amazon.awssdk.aws.greengrass.model.SubscribeToIoTCoreRequest;
+import software.amazon.awssdk.aws.greengrass.model.SubscriptionMode;
+import software.amazon.awssdk.aws.greengrass.model.UnauthorizedError;
 import software.amazon.awssdk.crt.io.SocketOptions;
 import software.amazon.awssdk.eventstreamrpc.EventStreamRPCConnection;
 import software.amazon.awssdk.eventstreamrpc.StreamResponseHandler;
@@ -57,11 +60,13 @@ import static com.aws.greengrass.lifecyclemanager.GreengrassService.SERVICES_NAM
 import static com.aws.greengrass.testcommons.testutilities.ExceptionLogProtector.ignoreExceptionOfType;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.instanceOf;
 import static org.hamcrest.Matchers.is;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.lenient;
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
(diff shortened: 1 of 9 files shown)


src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -140,6 +140,8 @@ public class MqttClient implements Closeable {
     private final Map<Subscribe, IndividualMqttClient> subscriptions = new ConcurrentHashMap<>();
     private final Map<MqttTopic, IndividualMqttClient> subscriptionTopics = new ConcurrentHashMap<>();
     private final Map<Subscribe, AtomicBoolean> subscribeCancellations = new ConcurrentHashMap<>();
+    // Local-only (direct-message) subscriptions: routed on-device but never subscribed in the cloud.
+    private final Set<Subscribe> localOnlySubscriptions = ConcurrentHashMap.newKeySet();
```
muboheng: nit: we should keep the same naming convention, use recieveOnly here

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -454,6 +456,32 @@ public CompletableFuture<SubscribeResponse> subscribe(Subscribe request)
...
 
+            if (request.isSkipCloudSubscribe()) {
+                // A (topic, callback) may live in only ONE mode -- cloud OR local-only, never both -- otherwise
+                // the router's two delivery passes both fire the same callback. Reject this local-only subscribe
+                // when the same (topic, callback) is already a cloud subscription. The cloud path enforces the
+                // mirror case.
+                boolean alreadyCloudSubscribed = subscriptions.keySet().stream()
+                        .anyMatch(s -> isSameSubscription(s, request.getTopic(), request.getCallback()));
+                if (alreadyCloudSubscribed) {
+                    throw new MqttRequestException(
```
muboheng: should we really be throwing an exception here? In the previous implementation, a subscription request will still get put on the map, but the nature of the map de-duplicate it, while still allowing the operation to go through

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -588,11 +616,17 @@ public CompletableFuture<Void> unsubscribe(Unsubscribe request) throws MqttReque
             if (isClosed.get()) {
                 throw new MqttRequestException("MQTT client is shut down");
             }
+            // Local-only registrations opened no cloud subscription. Match by (callback identity, topic) -- the
+            // same key the cloud loop uses below. subscribe() rejects registering the same (topic, callback) in
+            // both modes, so a hit here means it was local-only; a miss falls through to the cloud path.
+            if (localOnlySubscriptions.removeIf(
```
muboheng: for the unsubscribe path are we differentiating between receive_only and recieve_and_subscribe?

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -972,19 +1010,41 @@ Consumer<Publish> getMessageHandlerForClient(IndividualMqttClient client) {
...
+            // connection, so to match the cloud path's single-delivery guarantee it behaves exactly like a cloud
+            // sub owned by one designated connection: deliver here only if this client owns a covering cloud
+            // subscription AND is the designated owner, OR this client is the sole/odd receiver. This dedups the
+            // multi-connection fan-out case while still delivering messages that arrive on a client with no
+            // covering cloud subscription (a direct message, or the IoT Jobs case).
+            Set<Subscribe> matchedLocalSubs = new HashSet<>();
+            if (!localOnlySubscriptions.isEmpty()) {
+                // Deliver locally if this client is the sole/odd receiver (no matching cloud sub of its own),
+                // or it is the designated owner among the connections that do cover the topic.
+                boolean deliverLocalHere = !clientOwnsMatchingCloudSub
```
muboheng: why this check? a direct message could land out of the subscribe order, there could be a subscription on Nucleus#1 and 2(with 1 higher on the ordering list) but the DM is sent to Nucleus#2 client. which would cause this DM to not get delivered.

src/integrationtests/java/com/aws/greengrass/integrationtests/ipc/IPCMqttProxyTest.java
```diff
@@ -217,6 +220,156 @@ public void onStreamClosed() {
...
+        };
+
+        SubscribeToIoTCoreResponseHandler responseHandler = greengrassCoreIPCClient.subscribeToIoTCore(
+                subscribeToIoTCoreRequest, Optional.of(streamResponseHandler));
+        responseHandler.getResponse().get(TIMEOUT_FOR_MQTTPROXY_SECONDS, TimeUnit.SECONDS);
+
+        ArgumentCaptor<Subscribe> subscribeRequestArgumentCaptor = ArgumentCaptor.forClass(Subscribe.class);
+        verify(mqttClient).subscribe(subscribeRequestArgumentCaptor.capture());
+        Subscribe capturedSubscribeRequest = subscribeRequestArgumentCaptor.getValue();
+        assertThat(capturedSubscribeRequest.getTopic(), is(subscribeTopic));
```
muboheng, line 264: can we somehow check if no subscribe to iot core call is made?

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
tushar-aws: Yes this is a edge case where direct message would get dropped. The better handling would be to deliver the message on receiveOnly subscriptions irrespective of the connection on which the message arrives. This would stop the direct message to get dropped in any case.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
tushar-aws: That is not needed since we unsubscribe using the (topic+callback) matching which is unique across the modes.

src/integrationtests/java/com/aws/greengrass/integrationtests/ipc/IPCMqttProxyTest.java
tushar-aws: The test is present here `DirectMessageRoutingIntegTest.java:202`. 
In the verify(mockMqtt5Client, never()).subscribe(any()), it wires a real MqttClient with a mocked CRT Mqtt5Client, so the real skipCloudSubscribe branch executes and the CRT layer was never asked to subscribe.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
tushar-aws: Removing this block, since subscribe method is consumed from MqttProxyIPCAgent `SubscribeToIoTCoreHandler` which anyway creates a new reference of the callback by wrapping it with `forwardToSubscriber` method.

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
muboheng, line 451: it doesn't make sense for receive only to throw cause it doesn't actually make a subscribe request to the cloud.

muboheng (dismissed):

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
tushar-aws: I think keeping this check is reasonable coz it checks for the device configs(cert, policy, private key, etc) with the IoT core so if the device isn't configured for the cloud, that message can never arrive, so better to fail the subscribe right away than register something that silently never fires.

src/integrationtests/java/com/aws/greengrass/mqttclient/DirectMessageRoutingIntegTest.java
```diff
@@ -0,0 +1,203 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.aws.greengrass.mqttclient;
```
aws-kevinrickard: DirectMessageRoutingIntegTest will run in the wrong test phase. The pom's surefire executions split unit vs integration tests by class-path pattern: the unit execution excludes **/integrationtests/** and the integration execution includes only that pattern. Since this test's package is com.aws.greengrass.mqttclient, its compiled class path doesn't match — so it runs during mvn test (unit phase) and is skipped in the integration phase.

Suggested fix: move the test to com.aws.greengrass.integrationtests.mqttclient, and add a small non-test helper class in package com.aws.greengrass.mqttclient under src/integrationtests/java that exposes the package-private getMessageHandlerForClient hook.

(AI-assisted comment)

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
aws-kevinrickard: nit: naming this var just `subs` is an existing problem, but your additions amplify that problem.

e.g. it's confusing when reading `if (subs.isEmpty() && matchedReceiveOnlySubs.isEmpty())` down below. It's not clear at that point if subs is matched or all. I would suggest making the var name better by having it match your style introduced: `matchedCloudSubs`

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -972,19 +986,32 @@ Consumer<Publish> getMessageHandlerForClient(IndividualMqttClient client) {
...
+            // may therefore deliver more than once.
+            Set<Subscribe> matchedReceiveOnlySubs = new HashSet<>();
+            receiveOnlySubscriptions.stream()
+                    .filter(s -> MqttTopic.topicIsSupersetOf(s.getTopic(), message.getTopic()))
+                    .forEach(matchedReceiveOnlySubs::add);
+
+            if (subs.isEmpty() && matchedReceiveOnlySubs.isEmpty()) {
+                // No cloud subscriber and no receive-only registration matched, so we'll log an error and exit.
+                logger.atError().kv(TOPIC_KEY, message.getTopic()).kv(CLIENT_ID_KEY, client.getClientId())
+                        .log("Somehow got message from topic that no one subscribed to");
```
aws-kevinrickard: As discussed in the demo, I think we should change this. Here's a suggested rewrite: 

> Warn: "Received a message on a topic with no matching subscription or receive-only registration. Dropping the message"


Because:
1. With direct messages this can now happen normally. "Somehow" implies we thought it was impossible.
2. Warn, not Error: the Nucleus is working as designed, but the user still wants to know a message sent to the device went unhandled.
3. We should not only say what the input was but also what the result was (in this case, dropped).

src/main/java/com/aws/greengrass/builtin/services/mqttproxy/MqttProxyIPCAgent.java
```diff
@@ -199,9 +201,17 @@ public CompletableFuture<SubscribeToIoTCoreResponse> handleRequestAsync(Subscrib
...
-                Subscribe subscribeRequest = Subscribe.builder().callback(callback).topic(topic)
-                        .qos(qos).build();
+                // RECEIVE_ONLY only registers the subscription on device, for direct messages. It sends no cloud
+                // subscribe and ignores qos. An absent mode, or any other mode this Nucleus recognizes, subscribes
+                // in the cloud and requires a valid qos. A mode this Nucleus does not recognize is rejected.
+                Subscribe.SubscribeBuilder subscribeBuilder = Subscribe.builder().callback(callback).topic(topic);
+                if (SubscriptionMode.RECEIVE_ONLY.getValue().equals(request.getSubscriptionModeAsString())) {
+                    subscribeBuilder.skipCloudSubscribe(true);
+                } else {
+                    validateSubscriptionMode(request.getSubscriptionModeAsString(), serviceName);
```
aws-kevinrickard: `validateSubscriptionMode` checks "does the SDK model know this string?" (`SubscriptionMode.get(s) != null`), but what matters here is "does this handler implement this mode?". Those are the same set today, but if a future SDK bump adds a third mode, this validation would let it silently fall through to the cloud-subscribe path — with no handler change. An explicit allowlist fails loud until the mode is actually implemented, and drops the helper:

```java
String mode = request.getSubscriptionModeAsString();
if (SubscriptionMode.RECEIVE_ONLY.getValue().equals(mode)) {
    subscribeBuilder.skipCloudSubscribe(true);
} else if (mode == null || SubscriptionMode.SUBSCRIBE_AND_RECEIVE.getValue().equals(mode)) {
    subscribeBuilder.qos(validateQoS(request.getQosAsString(), serviceName));
} else {
    LOGGER.atError().kv(COMPONENT_NAME, serviceName).kv("subscriptionMode", mode)
            .log(INVALID_SUBSCRIPTION_MODE_ERROR);
    throw new InvalidArgumentsError(INVALID_SUBSCRIPTION_MODE_ERROR + ": " + mode);
}
```

(AI-assisted comment)

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -959,43 +973,70 @@ Consumer<Publish> getMessageHandlerForClient(IndividualMqttClient client) {
...
+            // AND any normal message reaching the device on a matching topic. A message reaching covering connections
+            // may therefore deliver more than once.
+            Set<Subscribe> matchedReceiveOnlySubs = new HashSet<>();
+            receiveOnlySubscriptions.stream()
+                    .filter(s -> MqttTopic.topicIsSupersetOf(s.getTopic(), message.getTopic()))
+                    .forEach(matchedReceiveOnlySubs::add);
+
+            if (matchedCloudSubs.isEmpty() && matchedReceiveOnlySubs.isEmpty()) {
+                // No cloud subscriber and no receive-only registration matched, so we'll log an error and exit.
+                logger.atError().kv(TOPIC_KEY, message.getTopic()).kv(CLIENT_ID_KEY, client.getClientId())
```
aws-kevinrickard, line 1010: Any reason to keep this as Error? I think should make it warn as mentioned in last comment:

> Warn, not Error: the Nucleus is working as designed, but the user still wants to know a message sent to the device went unhandled.

aws-kevinrickard (approved):

muboheng (approved):

PR #1824, opened 2026-07-16
fix: prevent builtin service config loss on dependency graph damage

aws-kevinrickard:
**Issue #, if available:** n/a

**Description of changes:**

Defense-in-depth fix for a silent config-loss failure chain, as five small commits:

1. **Prevent** (`GreengrassService.setupDependencies`): preserve the existing `isDefault` flag when re-adding a dependency listed in the `dependencies` topic, instead of downgrading it to `false`. Default (builtin) dependencies can no longer be evicted by a topic update.
2. **Tolerate** (`DeploymentActivator`): key builtin merge protection on static builtin identity (new `KernelLifecycle.AUTOSTART_BUILTIN_SERVICE_NAMES`) unioned with the live dependency graph, and add the same protection to the rollback merge behavior, which previously had none. Even with a damaged dependency graph, deployments and rollbacks can no longer delete builtin services' config.
3. **Heal** (`KernelConfigResolver.getMainConfig`): union the static autostart builtin names (`name:HARD`) into main's resolved dependencies, so a deployment on an already-damaged device repairs the dependency graph instead of perpetuating the damage until reboot.
4. **Heal fully** (`GreengrassService.setupDependencies`): when main re-adds a statically known autostart builtin that is not currently a default dependency, restore launch-equivalent default protection (`isDefault=true`). Applies only to main — other services declaring builtin names keep normal removable-dependency behavior.
5. Refactor: shared helper for the builtin protection name set, null-annotation guard, test import order.

The autostart builtin services (DeploymentService, FleetStatusService, TelemetryAgent, UpdateSystemPolicyService — the builtins with `autostart=true`) are injected into main's in-memory dependency map at every launch as *default* dependencies, protected from eviction by topic updates. That protection was fragile, and everything downstream keyed off the same mutable runtime state:

1. On a fresh device the `services/main/dependencies` topic contains only the nucleus; the builtins exist only in the in-memory map.
2. The first deployment writes the topic with the full resolved list, and the subscriber re-added every entry with `isDefault=false` — silently erasing the launch protection.
3. If that first deployment fails after activation and rolls back, the rollback force-replays the pre-deployment snapshot, restoring the topic to nucleus-only — and the subscriber **evicts every builtin from the dependency graph** while the services keep running.
4. Builtin merge protection was keyed on `kernel.orderedDependencies()`, so the **next deployment deleted the evicted builtins' entire config subtrees** (thing-group membership state, fleet-status sequence number). The running services kept writing to the detached config nodes, so nothing looked wrong.
5. Every tlog truncation rewrites the file from the attached tree, omitting the detached subtrees. At the next restart the device boots from a valid config missing its thing-group state; a subsequent deployment then treats installed components as orphaned group data and removes them.

Observed in the field as: fleet-status sequence number resetting to 0 and `fleetConfigArns` becoming empty after a routine restart, followed by a startup-time local deployment removing the customer's components.

TokenExchangeService and DockerApplicationManager are builtins but not autostart: they are not launch-injected and not in the static set; they remain covered by the graph-derived protection as before.

Each behavioral commit adds a regression test verified in both polarities (fails without the fix, passes with it):

- `SetupDependencyTest`: a default dependency survives a topic update omitting it (real config + subscriber path); a builtin re-added to main gets launch-equivalent protection, while a non-main service declaring the same name keeps removable behavior.
- `DefaultActivatorTest`: with no builtins in `orderedDependencies()`, a deployment merge retains builtin config (and still removes absent non-builtins); a rollback whose snapshot lacks builtin config retains it (and still discards absent non-builtins).
- `KernelConfigResolverTest`: `resolve()` with an empty main dependency map still emits every autostart builtin as `name:HARD`.

Healthy-graph behavior is unchanged: all pre-existing tests in the touched areas pass unmodified (`KernelConfigResolverTest`, `DefaultActivatorTest`, `KernelUpdateActivatorTest`, `DeploymentActivatorFactoryTest`, `SetupDependencyTest`, `GreengrassServiceTest`, `KernelTest`, `KernelLifecycleTest`, `LifecycleTest`, `MergeTest`), with checkstyle/PMD active. The full failure chain was additionally reproduced end-to-end against unmodified v2.10.2 config/lifecycle classes; re-running that reproduction against this patched build confirms the chain breaks at step 2 (no flag erasure, no eviction, no config deletion, all state survives restart).

- Static set = `@ImplementsService(autostart=true)` builtins — the same criterion `launch()` uses — so on a healthy device the unions add nothing. Protecting a name absent from a device's graph is harmless (MERGE only retains existing config), and statically added names always resolve since autostart builtins are compiled into the nucleus.
- One deliberate rollback behavior change: nodes newly created under a builtin during the failed deployment are retained rather than discarded; snapshot values are still restored exactly as before.
- `createRollbackMergeBehavior` widened from private to protected for testability, matching `createDeploymentMergeBehavior`.
- Related PRs from the same investigation: #1819 (regression tests for the `cleanupGroupData` one-way ratchet that amplifies this loss into component removal) and #1820 (fix for the LogManager-driven config churn that drove near-constant tlog truncation, shrinking the loss-exposure window). This PR supersedes #1823 (characterization tests demonstrating the pre-fix behavior, by @saranyailla): its two bug demonstrations are covered here as inverted regression tests, and its two baseline invariant tests are incorporated (with attribution) in the final test commit.

**Documentation Checklist:**

**Compatibility Checklist:**
- [x] For external library version updates, I have reviewed its change logs and Nucleus does not consume

---

# "Simple" Explanation of the Situation:

### Background: the builtin services

The nucleus runs four internal "builtin" services compiled into it —
**DeploymentService** (processes deployments), **FleetStatusService**
(reports device status to the cloud), **TelemetryAgent**, and
**UpdateSystemPolicyService**. They are not deployed as components;
instead, at every boot the nucleus adds them to its in-memory service
dependency graph with a **"shield flag" that exists only in memory**
(🛡️ below) — nothing about them is persisted to disk.

They matter here for two reasons:

1. **Their config holds critical device state.** DeploymentService's config
   records which thing groups this device belongs to and which components
   each group wants running. If that record is lost, the next deployment
   treats those components as orphaned and **removes them**.
2. **Their protection is fragile.** Two things guard them — the **shield
   flag** (🛡️), a memory-only marker guarding their place in the graph,
   and the **protection list**, built fresh at each deployment to guard
   their config during the merge. Both are computed from *current runtime
   state* rather than from the fixed builtin list, and breaking the first
   silently disables the second:

   ```
   shield flag erased → builtin evicted from graph
                              │
     protection list ("who's in the graph right now?") ── empty
                              ▼
            next deployment deletes the builtin's config
   ```

### Background: deployments — the merge, and cloud vs local

Each deployment applies its configuration by **merging** it over the
existing config tree ("the merge" below). Two deployment types matter:
**cloud** (thing-group) deployments carry the target group's full
component list from the cloud; **local** deployments (submitted on the
device itself — some fleets run a component that auto-submits one at
every startup) carry only their own changes, and preserve every *other*
group's components solely from the device's **local records**. (Both
types fetch group *membership* from the cloud — lab-verified — but
membership alone cannot rebuild a missing local record of a group's
components.) That difference decides 💥 vs ✅ in the scenarios below.

### Background: state - kept in three places

- **GRAPH** — main's in-memory dependency list (🛡️ = protected/default entry)
- **MEM** — in-memory config tree (DeploymentService's group state);
  *attached/DETACHED* = whether the subtree is still reachable from the
  config root (see primer below)
- **DISK** — config.tlog (what survives a restart); "deps topic" = the
  persisted copy of main's dependency list stored inside it (a "topic" is
  a node in the nucleus config tree — Greengrass jargon, not MQTT). The
  file = last periodic full rewrite (a root walk, every ~21 min on this
  device) plus individual appends since — writes to detached nodes still
  append, but each rewrite discards them and omits detached data

### Background: attached vs detached config

```
        ATTACHED (normal)                DETACHED (after a deployment
                                          merge deletes the subtree)

             root                             root
              │                                │
           services                         services ──── LogManager ...
        ┌─────┴────────┐
 DeploymentService   LogManager     DeploymentService ◄── still held by the
        │                                  │              running service's
   GroupToRootComponents              GroupToRootComponents   own pointer
        │                                  │
   thinggroup/G → components          thinggroup/G → components

 reachable from root ✓               unreachable from root ✗
 → saved by every disk rewrite       → alive and fully usable in memory,
                                       invisible to every disk rewrite
```

Like a reference to a linked list item which is no longer in the list:
the running service keeps reading and writing its subtree through its own
pointer — writes even still append lines to config.tlog — but anything that
persists config by walking the tree from the root (truncation rewrites,
effectiveConfig.yaml) no longer sees it. The subtree evaporates for good
when the process exits.

---

## Scenario A — the failure path (the incident)

```
① FIRST BOOT (fresh device)
   GRAPH: builtins 🛡️  (injected at launch, memory-only)
   MEM:   DeploymentService config attached, no group state yet
   DISK:  deps topic = [Nucleus]           ← builtins never written to disk

② FIRST DEPLOYMENT FAILS after activation → ROLLBACK          [enters W1]
   forward merge wrote deps topic incl. builtins → shield flags erased
   (🛡️ → ✗): whenever the deps topic changes, the graph is rebuilt from
   it, and every listed service is re-registered as an ordinary
   UNSHIELDED entry — re-registration overwrites the boot-granted flag
   rollback restores deps topic = [Nucleus] → builtins EVICTED from graph
   (not listed, and no longer shielded)
   GRAPH: builtins GONE (services still running)
   MEM:   config still attached & complete
   DISK:  still complete

③ NEXT DEPLOYMENT SUCCEEDS                                    [enters W2]
   the protection list is built from the damaged graph → comes up
   empty → the merge deletes the builtins' config
   GRAPH: builtins still gone
   MEM:   group state alive but DETACHED (services hold old refs)
   DISK:  group state alive immediately following the deployment (via
          appends), but purged at the next tlog truncation (~21 min) —
          truncation rewrites the file from ATTACHED nodes only — and
          it stays gone, since only deployments write group state again
   ...device runs normally like this for days/weeks — even processing
   local deployments safely, since the group data is still alive in
   detached memory (cleanup finds the membership and keeps everything);
   nothing surfaces until a restart...

④ RESTART (e.g. operator systemctl restart)
   GRAPH: builtins 🛡️ again (boot re-injects — graph heals)
   MEM:   rebuilt from disk → group state MISSING
   DISK:  unchanged (missing group state)

⑤ LOCAL DEPLOYMENT at startup (from a component that auto-submits
   one at every boot — common in managed fleets, incl. this one)
   no local record of the group's components exists → resolution has
   nothing to preserve them from (membership, though fetched, cannot
   rebuild the record)
   💥 COMPONENTS REMOVED
```

## The two windows

Scenario A passed through two named degraded states — they differ in
whether the damage has reached the disk, which decides what a restart does:

| | **Window 1** (A②) | **Window 2** (A③) |
|---|---|---|
| Damage | graph eviction only (memory) | + builtin config deleted; disk state deleted after config.tlog truncation |
| Device looks | normal | normal |
| A restart... | **heals** (disk still complete) | **triggers the loss** (memory held the last good copy) |
| Ends when | the next deployment runs (→ W2) | the restart happens (→ A④) |

W1 is transient (hours, typically); W2 can persist for weeks while the
device looks perfectly healthy — one restart away from impact.

## Scenario B — first deployment succeeds (the normal fleet path, safe)

```
① FIRST BOOT            GRAPH: builtins 🛡️   DISK: deps=[Nucleus]
② FIRST DEPLOYMENT OK   shield flags erased (🛡️ → ✗) as in A②, BUT
                        DISK: deps topic now CONTAINS builtins
                        (carried forward from the healthy graph —
                        deployments always persist the graph's builtins)
③ any LATER deployment fails → rollback restores a deps list
                        WITH builtins → nothing evicted → SAFE forever
```

The whole vulnerability lives in one corner: **only a rollback to the
pre-first-deployment snapshot restores a builtin-less list.** After one
successful deployment, the persisted list includes the builtins and
rollbacks are harmless.

Related safe fork: a first deployment that fails **without rollback**
(failure handling policy DO_NOTHING) is also harmless — no snapshot
replay means no builtin-less list is ever restored, so nothing is
evicted. Scenario A specifically requires *fail + rollback*.

## Scenario C — restart during W1 (safe, heals)

Same start as Scenario A, but the device restarts before any further
deployment runs:

```
① FIRST BOOT (fresh device)
   GRAPH: builtins 🛡️  (injected at launch, memory-only)
   MEM:   DeploymentService config attached, no group state yet
   DISK:  deps topic = [Nucleus]

② FIRST DEPLOYMENT FAILS after activation → ROLLBACK          [enters W1]
   shield flags erased by the forward merge, then rollback evicts the
   builtins (as in A②)
   GRAPH: builtins GONE (services still running)
   MEM:   config still attached & complete
   DISK:  still complete                   ← this is what makes the
                                             coming restart safe

③ RESTART (before any further deployment)
   GRAPH: builtins 🛡️ re-injected at launch → graph healed
   MEM:   rebuilt from disk → complete (disk was never damaged)
   DISK:  unchanged, complete
   → FULLY HEALED — device is now equivalent to Scenario B after ②:
     the next deployment writes the deps topic including the builtins,
     and later rollbacks are harmless

④ (any further deployments / restarts)
   → normal operation, no failure path remains
```

The heal works because in W1 the disk still holds everything: the restart
discards only the damaged in-memory graph (rebuilt at boot) and loses no
data. W1 is transient though — it ends at the next deployment (→ A③), and
fleets with auto-submitting components deploy often.

## Scenario D — restart during W2, but no deployment afterward (latent)

```
① FIRST BOOT (fresh device)
   GRAPH: builtins 🛡️        MEM: attached, empty       DISK: deps=[Nucleus]

② FIRST DEPLOYMENT FAILS after activation → ROLLBACK          [enters W1]
   GRAPH: builtins EVICTED   MEM: attached, complete    DISK: complete

③ NEXT DEPLOYMENT SUCCEEDS                                    [enters W2]
   GRAPH: builtins gone
   MEM:   group state alive but DETACHED
   DISK:  group state purged at the next truncation, stays gone

④ RESTART
   GRAPH: builtins 🛡️ (boot heals the graph)
   MEM:   rebuilt from disk → group state MISSING
          (FSS sequence number resets to 0, membership empty)
   DISK:  unchanged (missing group state)

⑤ ...NO deployment processes...
   components keep running normally — the damage is invisible except
   via the cloud-visible symptoms: the FleetStatusService sequence
   number resets to 0 and components report empty fleetConfigArns
   (their thing-group associations)
   💤 IMPACT STAYS LATENT until ANY deployment runs (then → A⑤ 💥
      or → E⑤ ✅ depending on its type)
```

On the incident device, this window lasted 18 seconds — the gap between
the restart and the component's auto-submitted local deployment.

## Scenario E — cloud deployment after the restart (recovers)

```
① FIRST BOOT (fresh device)
   GRAPH: builtins 🛡️        MEM: attached, empty       DISK: deps=[Nucleus]

② FIRST DEPLOYMENT FAILS after activation → ROLLBACK          [enters W1]
   GRAPH: builtins EVICTED   MEM: attached, complete    DISK: complete

③ NEXT DEPLOYMENT SUCCEEDS                                    [enters W2]
   GRAPH: builtins gone
   MEM:   group state alive but DETACHED
   DISK:  group state purged at the next truncation, stays gone

④ RESTART
   GRAPH: builtins 🛡️ (boot heals the graph)
   MEM:   rebuilt from disk → group state MISSING
   DISK:  unchanged (missing group state)

⑤ CLOUD DEPLOYMENT to the device's thing group G
   carries G's full component list in its deployment doc — the part a
   local deployment lacks — and fetches membership from the cloud
   resolution = G's components from the doc
   GRAPH: builtins 🛡️ (shield flags intact since ④)
   MEM:   GroupMembership + GroupToRootComponents(G) rebuilt, attached
   DISK:  group state persisted again at completion
   ✅ DEVICE FULLY RECOVERED (single-group device)
```

Yes — this succeeds and recovers the device, because the cloud deployment
carries the group's component list with it: it doesn't need the (missing)
local record. This is why the remediation guidance is "deploy to the
degraded device" and it is essentially how the incident device was
recovered. Lab-verified addendum: the safety of this path is inherent,
not timing luck — the resumed/processed cloud deployment persists the
group record before any startup local deployment can be processed
(deployments are handled serially, in order).

Variants: multi-group devices → Scenario F; the recovery deployment itself
failing → Scenario G.

## Scenario F — multi-group device, cloud deployment after the restart (partial recovery, collateral removal)

Device belongs to TWO thing groups, Alpha and Beta, each deploying its own
components:

```
① FIRST BOOT (fresh device)
   GRAPH: builtins 🛡️        MEM: attached, empty       DISK: deps=[Nucleus]

② FIRST DEPLOYMENT (Alpha) FAILS after activation → ROLLBACK   [enters W1]
   GRAPH: builtins EVICTED   MEM: attached, complete    DISK: complete

③ DEPLOYMENTS for Alpha and Beta SUCCEED                       [enters W2]
   both groups' components installed and running
   MEM:   group state for Alpha AND Beta alive but DETACHED
   DISK:  group state for both GONE

④ RESTART
   GRAPH: builtins 🛡️ (boot heals the graph)
   MEM:   rebuilt from disk → BOTH groups' state missing
   DISK:  missing both

⑤ CLOUD DEPLOYMENT to Alpha
   membership fetched from cloud → shows Alpha AND Beta ✓
   resolution: Alpha's components come from the deployment doc ✓
               Beta's components are preserved ONLY if Beta's LOCAL
               GroupToRootComponents record exists — it doesn't
   💥 BETA'S COMPONENTS REMOVED, Alpha recovered
   MEM/DISK: group state for Alpha rebuilt; Beta absent

⑥ CLOUD DEPLOYMENT to Beta (a second, separate deployment)
   ✅ Beta recovered — full recovery requires one deployment PER group
```

The membership lookup protects a group's *record* from cleanup, but a
missing record is never *rebuilt* from the cloud — component preservation
still reads only local state. This one-way property is what we call the
**"ratchet"**: group records are destroyed automatically, but only ever
rebuilt by a deployment targeting that specific group. So recovery is
per-group, and each group's deployment removes the still-unrecovered
groups' components in the meantime.

## Scenario G — recovery deployment fails and rolls back (re-arms the bug)

Single-group device; the recovery attempt itself fails after activation:

```
① FIRST BOOT (fresh device)
   GRAPH: builtins 🛡️        MEM: attached, empty       DISK: deps=[Nucleus]

② FIRST DEPLOYMENT FAILS after activation → ROLLBACK          [enters W1]
   GRAPH: builtins EVICTED   MEM: attached, complete    DISK: complete

③ NEXT DEPLOYMENT SUCCEEDS                                    [enters W2]
   GRAPH: builtins gone
   MEM:   group state alive but DETACHED
   DISK:  group state purged at the next truncation, stays gone
          note: deps topic was written by this damaged-graph
          resolution, i.e. WITHOUT the builtins

④ RESTART
   GRAPH: builtins 🛡️ (boot heals the graph)
   MEM:   rebuilt from disk → group state MISSING
   DISK:  unchanged (missing group state; deps topic still
          builtin-less)

⑤ CLOUD DEPLOYMENT arrives (the recovery attempt)
   snapshot taken for rollback = current config, whose deps topic
   is WITHOUT builtins
   forward merge writes deps topic WITH builtins → boot-granted
   shield flags erased (🛡️ → ✗, as in A②)
   ...deployment FAILS after activation (e.g. a component breaks)...

⑥ ROLLBACK replays the snapshot
   deps topic restored to the builtin-less list → builtins EVICTED again
   GRAPH: builtins GONE      [degraded again: W1's structure — evicted
                              graph, builtin config still attached —
                              but group state is already lost, so it
                              lacks W1's "restart heals" property]
   MEM:   group state still missing (rollback restored pre-⑤ config,
          which had none)
   DISK:  still missing group state

⑦ from here: a later successful cloud deployment recovers it (→ E),
   another deployment while degraded re-deletes builtin config (→ A③),
   a restart re-heals the graph but restores no data (→ A④)
```

The pre-first-deployment corner generalizes: on an unfixed nucleus, ANY
rollback whose snapshot holds a builtin-less deps topic re-evicts the
builtins. A degraded device's snapshots hold exactly that — so every
failed deployment on a degraded device re-arms the cycle.

## Scenario H — nucleus UPGRADE fails on a W2 device

A nucleus upgrade is a *bootstrap* deployment: it reboots the device
mid-deployment, and a failure rolls back by REBOOTING AGAIN into the old
nucleus from a rollback snapshot. On a W2 device that snapshot is
poisoned:

```
① – ③ as Scenario A: device is W2-degraded (builtin config detached,
   disk missing group state), still running normally — this is exactly
   the state a fleet-remediation upgrade targets

④ UPGRADE DEPLOYMENT (new nucleus version) arrives
   rollback snapshot taken = a DUMP of the config tree — walked from
   the root, so it OMITS the detached group state
   merge + bootstrap: device REBOOTS into the new nucleus...

⑤ ...UPGRADE FAILS during activation → KERNEL ROLLBACK
   device reboots AGAIN, into the OLD nucleus, loading the snapshot
   GRAPH: builtins 🛡️ (boot re-injects)
   MEM:   rebuilt from the poisoned snapshot → group state MISSING
   DISK:  missing group state
   → the device lands directly in the State-B condition (as after A④)

⑥ LOCAL DEPLOYMENT at startup (auto-submitting component)
   💥 COMPONENTS REMOVED — the failed remediation attempt itself
      triggered the incident it was meant to prevent
```

Unlike Scenario G (ordinary deployment rollback — no reboot, memory
survives, nothing new lost), the upgrade's rollback path REBOOTS, so the
detached in-memory group state evaporates and the loss lands immediately.

Operational implication (confirmed): upgrades of degraded devices should
be monitored, and a failed upgrade followed immediately by a fresh
thing-group deployment — lab-verified to fully restore the device
(with a component-down window in between).

Contrast: the same failed upgrade on a **W1** device is harmless — its
snapshot is complete (everything still attached), the rollback reboot
heals it like any W1 restart (→ C), and a W1 device has no deployed
components to lose anyway (its first deployment never succeeded).

Status: Reproduced. A W2-degraded device upgraded
toward a fixed build (deployment also carried a broken component) hit
FAILED_ROLLBACK_COMPLETE; post-rollback, DeploymentService booted with
empty group state, and the first local deployment removed the running
component. Notably, the cloud membership fetch SUCCEEDED and did not
help — the decisive variable is the missing local record of the group's
components, which membership alone cannot rebuild. The fix status of the
upgrade target is irrelevant: the rollback executes old-nucleus code.

## Scenario I — reboot(s) before the first deployment, which then fails (same as A)

```
① FIRST BOOT (fresh device)
   GRAPH: builtins 🛡️        MEM: attached, empty       DISK: deps=[Nucleus]

② REBOOT (no deployment has arrived yet — one or many, doesn't matter)
   GRAPH: builtins 🛡️ (re-injected, as at every boot)
   MEM:   rebuilt from disk → identical to ①
   DISK:  unchanged — deps topic still [Nucleus]

③ FIRST DEPLOYMENT FAILS after activation → ROLLBACK
   → proceeds exactly as A② (shield flags erased by the forward merge,
     rollback restores [Nucleus], builtins evicted) and onward
```

Pre-deployment reboots change nothing: each boot recreates the same
state. The vulnerable corner — a builtin-less deps topic on disk — exists
from provisioning until the first deployment *writes builtins into the
topic*, no matter how many reboots happen in between. Only a deployment
can close the corner (→ B) or spring it (→ A).

## Scenario J — the FIRST deployment is a bootstrap deployment that fails (safe!)

A bootstrap deployment (e.g. one changing the nucleus version) REBOOTS
the device mid-deployment, before activation completes. Assume the
rollback itself does not bootstrap/reboot (no bootstrap-on-rollback):

```
① FIRST BOOT (fresh device)
   GRAPH: builtins 🛡️        MEM: attached, empty       DISK: deps=[Nucleus]

② BOOTSTRAP DEPLOYMENT arrives (first deployment ever)
   rollback snapshot taken (deps=[Nucleus], tree complete)
   forward merge writes deps topic WITH builtins → shield flags
   erased (🛡️ → ✗), as in A②...
   ...then the device REBOOTS to continue the deployment

③ BOOT (mid-deployment, activation stage)
   GRAPH: builtins 🛡️ — the reboot RE-INJECTS the shield flags!
   the erasure from ② did not survive the restart (flags are re-granted
   at every launch)

④ DEPLOYMENT FAILS during activation → ROLLBACK (no reboot)
   snapshot replayed: deps topic restored to [Nucleus]
   the builtins are absent from the list BUT carry fresh shield
   flags → NOT evicted 🛡️
   GRAPH: builtins 🛡️        MEM: rolled back, fresh    DISK: complete
   ✅ NO DAMAGE — device is a healthy fresh device again
```

The eviction in Scenario A requires the flag-erasure (②) and the
snapshot replay (rollback) to happen **within one uninterrupted boot
session**. A bootstrap deployment inserts a reboot between them, which
re-arms the shields — so it is immune. This is also why the incident's
trigger had to be an ordinary component/config deployment: those never
reboot, so nothing restores the flags before the rollback replays the
poisoned list.


saranyailla (approved):

junfuchen99 (approved):

aws-kevinrickard:
Nit (non-blocking): `getMainConfig` uses both accessors for the same value — `greengrassService.getName()` for the dependency entry and `greengrassService.getServiceName()` for the dedup set. In production they're identical (`getName()` is a one-line delegate to `getServiceName()`), but on Mockito mocks the delegation doesn't exist, so they're separately stubbable: an unstubbed `getServiceName()` in the older mock-based resolver tests silently adds `null` to the dedup set. Harmless today since those tests assert containment, but using one accessor in both places would remove the trap.

Long term: `getName()`/`getServiceName()` is historical duplication (the alias pair predates the 2020 EvergreenService → GreengrassService rename). It would be worth deprecating `getName()` and migrating call sites to `getServiceName()` in a separate cleanup, so new code stops picking accessors at random and mock-based tests can't diverge on which one is stubbed. Fine as a follow-up; not something this PR needs to take on.

PR #1809, opened 2026-06-25
fix: support glob wildcards in noProxyAddresses and prevent PatternSyntaxException crash

xuchuwan:
## Problem

`noProxyAddresses` entries with standard NO_PROXY glob patterns (e.g. `*.sts.amazon.com`) crash Nucleus with `PatternSyntaxException: Dangling meta character '*'`. The crash kills all MQTT connectivity (Shadow, Jobs, Spooler, IPC) and the device becomes permanently unreachable since the config persists across restart.

**Root cause:** [`MqttClient.java:255`](https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/main/src/main/java/com/aws/greengrass/mqttclient/MqttClient.java#L255) and [`StandaloneMqttConnector.java:129`](https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/main/src/main/java/com/aws/greengrass/mqttclient/StandaloneMqttConnector.java#L129) use `endpoint::matches` which is Java's `String.matches(regex)`. `*.domain.com` is invalid regex — the `*` quantifier has no preceding character.

Note: The AWS SDK HTTP path (`ProxyConfiguration.nonProxyHosts()`) already handles wildcards natively. Only the MQTT/CRT paths crash.

## Fix

Added `ProxyUtils.noProxyMatches(endpoint, pattern)` that converts standard NO_PROXY glob patterns to proper regex before matching:

- `*.domain.com` → `.*\Q.domain.com\E` (wildcard subdomain)
- `.domain.com` → `.*\Q.domain.com\E` (leading-dot convention, same as `*. `)
- `domain.com` → `\Qdomain.com\E` (exact match, dots are literal)
- Invalid patterns → logged as warning, returns false (no crash)

Replaced `endpoint::matches` at both call sites with the new helper.

## Validation

**1. Reproduced the crash (before fix):**
```
$ greengrass-cli deployment create --merge "aws.greengrass.Nucleus=2.17.0" \
  --update-config '{"aws.greengrass.Nucleus":{"MERGE":{"networkProxy":{"proxy":{"url":"http://127.0.0.1:3128"},"noProxyAddresses":"*.sts.amazon.com,*.credentials.iot.us-west-2.amazonaws.com"}}}}'

# Result: 6x PatternSyntaxException within 1 second:
java.util.regex.PatternSyntaxException: Dangling meta character '*' near index 0
*.sts.amazon.com
^
  at java.lang.String.matches(String.java:2024)
[...]

PR #1825, opened 2026-07-17
fix: never crash on invalid noProxyAddresses entries

xuchuwan:
**Issue #, if available:** Related: #1809 (this PR carves out the crash-safety portion; glob/wildcard support will follow separately)

**Description of changes:**

An invalid `noProxyAddresses` entry — e.g. the glob-style `*.example.com`, which is not valid Java regex — threw an uncaught `PatternSyntaxException` ("Dangling meta character '*'") from `String.matches()` at both MQTT proxy-bypass call sites, killing MQTT connectivity for the device.

This change makes noProxy handling crash-safe without changing the documented regex semantics:

- **`ProxyUtils.noProxyMatches(endpoint, pattern)`** (new): regex semantics preserved, but
  - entries are trimmed before matching (note: previously a space-padded entry from a comma-space separated list could never match — small intentional behavior change),
  - a null/empty endpoint or blank entry never matches,
  - an uncompilable entry is logged (`WARN`) and treated as a non-match instead of throwing.
- **`MqttClient`** and **`StandaloneMqttConnector`** bypass checks now route through it.
- **`ProxyUtils.validateNoProxyAddresses()`** (new) + wiring in `BootstrapManager.networkProxyHasChanged`: deployments carrying a `networkProxy` section are rejected early with a clear `ComponentConfigurationValidationException` naming the offending entries when an entry contains whitespace or does not compile — instead of a runtime crash (uncompilable) or silent no-match. Empty segments from stray commas (`a.com,,b.com`) are skipped, matching the runtime matcher's tolerance.

Users naturally write glob-style entries like `*.sts.amazon.com` (the convention every other NO_PROXY implementation uses). Today that config crashes MQTT connection setup entirely — a config mistake should degrade to "entry doesn't match" plus a clear deployment-time error, never loss of connectivity. Crash safety is split out from wildcard support so this fix can land independently; glob semantics need a wider change (the SDK HTTP path in `getProxyConfiguration()` passes raw patterns to `nonProxyHosts()`, which also does raw regex matching) and will come as a follow-up.

Recovery note: a device already carrying an invalid entry is not deadlocked by the new validation — a deployment with a corrected value passes validation and applies.

11 new unit tests: 9 in `ProxyUtilsTest` (regex behavior preserved, invalid pattern no-throw, null/empty endpoint, blank entries, trim, validation accept/skip/reject paths) and 2 in `BootstrapManagerTest` (invalid entry → deployment rejected with the entry named in the message; valid entries with stray commas pass validation). Full runs of `ProxyUtilsTest` (24), `BootstrapManagerTest` (24), `MqttClientTest` (101), `StandaloneMqttConnectorTest` (16) — all pass. Checkstyle and PMD (main + tests) clean locally.

Validation strictness deliberately mirrors runtime tolerance: whatever the matcher silently skips (blank/stray-comma segments) is also accepted by validation; whatever the matcher cannot ever match (uncompilable, embedded whitespace) is rejected at deployment time. Validation only runs when a deployment carries a `networkProxy` section; pre-existing bad config in the tlog is covered by the runtime no-throw guard.

**Documentation Checklist:**
 - [ ] Updated the README if applicable. (N/A — no user-facing interface change; regex semantics unchanged.)

**Compatibility Checklist:**
- [x] Any modification or deletion of public interfaces does not impact other plugin components. (Additive only — two new public static methods; no signature changes.)
- [x] For external library version updates, I have reviewed its change logs and Nucleus does not consume any deprecated method or type. (No dependency changes.)

### DIFF
--- a/src/main/java/com/aws/greengrass/util/ProxyUtils.java
+++ b/src/main/java/com/aws/greengrass/util/ProxyUtils.java
@@ -35,6 +35,8 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.regex.Pattern;
+import java.util.regex.PatternSyntaxException;
 import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 import javax.net.ssl.TrustManager;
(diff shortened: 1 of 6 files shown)


src/main/java/com/aws/greengrass/util/ProxyUtils.java
```diff
@@ -442,4 +444,68 @@ public static String getNoProxyEnvVarValue(DeviceConfiguration deviceConfigurati
...
+            if (trimmed.isEmpty()) {
+                // A stray comma is benign -- skip it rather than fail the deployment
+                continue;
+            }
+            if (trimmed.contains(" ") || trimmed.contains("\t")) {
+                invalid.add(trimmed);
+                continue;
+            }
+            try {
+                Pattern.compile(trimmed);
```
MikeDombo, line 503: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

When using regular expressions (regex) in Java, it&#39;s best practice to first compile your regex pattern separately instead of using pattern matching methods directly. By compiling first, you can reuse the same `Pattern` instance repeatedly without having to recompile.

sahith (approved):

src/main/java/com/aws/greengrass/util/ProxyUtils.java
xuchuwan: Agreed in principle, but leaving as-is for this PR: this path only runs during MQTT connection setup against a small, comma-separated config list, so recompilation cost is negligible, and caching compiled Patterns would add static state for little gain. A follow-up PR (glob wildcard support for noProxyAddresses) restructures this exact code around a shared entry-translation helper — that's the right place to revisit pattern reuse if it matters.

PR #1826, opened 2026-07-20
fix: do not treat unknown thing group membership as empty membership

aws-kevinrickard:
## Summary

Fixes a defect where an **unknown** thing-group membership response was treated as an authoritative **"member of no thing groups"** — causing a single local deployment to remove every recorded thing group's components from the device and delete the groups' bookkeeping records.

## The defect

During deployment processing, `DefaultDeploymentTask.getNonTargetGroupToRootPackagesMap()` fetches the device's current thing-group membership via `ThingGroupHelper.listThingGroupsForDevice()` to decide which recorded thing groups' components to preserve. That lookup can end three different ways, and it is important to distinguish them:

1. **Authoritative answer** — the device is configured for the cloud, the `ListThingGroupsForCoreDevice` call succeeds, and the response (possibly an empty list) is the truth about membership.
2. **Attempted and failed** — the device is configured for the cloud, the call is made, but it throws (network unreachable/offline, timeout, `FORBIDDEN`, retries exhausted). This surfaces as an exception and is handled by existing fallback paths that use the persisted group records instead. *This path was already correct and is not changed by this PR.*
3. **Unknown, never attempted (`Optional.empty()`)** — `listThingGroupsForDevice()` has a guard at the top: if `isDeviceConfiguredToTalkToCloud()` is false, it returns `Optional.empty()` **without making any cloud call**. This means the device's cloud provisioning configuration (thing name, IoT data/credential endpoints, certificate path, private key path, root CA path, AWS region) is missing or invalid, so the question "which groups am I in?" cannot even be asked.

**When does case 3 actually come up?** A device can be in that state because:
- it was provisioned for **local-only / offline operation** and never had cloud config (harmless — such a device has no thing-group records to lose); or
- it **had valid cloud config that is now missing or corrupted** — e.g. configuration state loss, a bad reconfiguration, or a partially-applied provisioning change. This is the dangerous case: the device still has thing-group records and thing-group components running, but its provisioning config no longer validates.

Note that a device that is merely **offline** (network down, cloud unreachable) does *not* produce `Optional.empty()` — its config still validates, so the call is attempted, throws, and takes the already-safe fallback path (case 2). `Optional.empty()` occurs specifically when the configuration itself is invalid.

Because case 3 returns an empty `Optional` rather than throwing, none of the case-2 fallbacks fire. The buggy code coerced the unknown into an empty membership set — an authoritative "member of no thing groups". Consequences of one local deployment processed in that state:

1. Every recorded thing group fails the membership filter → its root components are dropped from the merge's preserved root set → **removed from the device**.
2. The `GroupMembership` snapshot is rebuilt empty → the post-deployment cleanup **deletes every thing group's `GroupToRootComponents` / `GroupToLastDeployment` records**, so the components stay gone even after the configuration is repaired, until each group is redeployed from the cloud.

No prior loss of deployment state is required. A plausible end-to-end trigger: a device's cloud config becomes invalid, and a component auto-submits a local deployment at startup (a common fleet pattern) — that single local deployment uninstalls every thing-group component on the device.

## The fix

Unknown membership (`Optional.empty()`) now falls back to the membership implied by the persisted group records — exactly the behavior of the existing attempted-and-failed fallback paths. "We could not determine membership" is no longer treated as "we determined the device has no memberships". One code path, nine lines.

| Scenario | What it means | Before | After |
|---|---|---|---|
| Lookup returns unknown (`Optional.empty()`) | Cloud provisioning config missing/invalid — no lookup attempted | All recorded thing groups' components removed; all group records deleted by cleanup | Recorded groups preserved; records kept (same as fetch-failure handling) |
| Authoritative lookup confirming membership | Call succeeded; groups listed | Preserved | Preserved (unchanged) |
| Authoritative lookup returning an **empty set** | Call succeeded; device genuinely in no thing groups | Components removed, records cleaned | Removed and cleaned (unchanged) — pinned by a test |
| Lookup attempted and failed | Config valid; call made but threw (offline, timeout, `FORBIDDEN`) | Falls back to persisted membership | Unchanged |

## Tests

`ThingGroupMembershipPreservationTest`:
- Regression test (verified **failing with the fix disabled**): unknown membership preserves recorded groups' components and rebuilds `GroupMembership` from persisted info.
- Control test: confirmed membership preserves (passes with and without the fix).
- Legitimate-flow test: authoritative *empty* membership still removes — pins the `Optional.empty()` vs `Optional.of(emptySet)` distinction, guarding against over-preservation (passes with and without the fix).
- Scope-pinning test (verified **failing without the thing-group filter**): the fallback implies membership only for `thinggroup/`-prefixed records; a record whose name is not a thing group is not resurrected as a membership.

## Fallback scope

The fallback returns only `thinggroup/`-prefixed record names — the only names a real membership lookup can return. Without that restriction, a record whose group name is not a thing group (e.g. a raw configuration ARN that failed to parse) would be treated as a membership, and its pinned root components would enter the merge's preserved root set, where they can impose stale version constraints on later deployments and make dependency resolution unsatisfiable. For devices deployed through the cloud APIs this is behavior-neutral: configuration ARNs always parse, so persisted records are `thinggroup/`-prefixed, `thing/`-prefixed, or `LOCAL_DEPLOYMENT`.

Full `mvn clean test` with checkstyle/PMD/spotbugs active: 1196 tests, the only failure is pre-existing and unrelated (identical on the unmodified branch).

## Scope history

This PR was originally broader (running-component preservation for local deployments and a post-restart cleanup deferral). After further lab testing showed those changes address no remaining demonstrated failure once #1824 is deployed, they were narrowed out of this PR and parked on the `local-deployment-preservation-parked` branch for a possible future defense-in-depth PR. This defect is the one remaining demonstrated bug not covered by #1824.

## Related

#1824 fixes the loss of persisted configuration (builtin service config protection) observed in the field. This defect is independent of it — it needs no prior state loss.

### DIFF
--- a/src/main/java/com/aws/greengrass/deployment/DefaultDeploymentTask.java
+++ b/src/main/java/com/aws/greengrass/deployment/DefaultDeploymentTask.java
@@ -349,8 +349,7 @@ public class DefaultDeploymentTask implements DeploymentTask {
         Topics groupsToRootPackages =
                 deploymentServiceConfig.lookupTopics(DeploymentService.GROUP_TO_ROOT_COMPONENTS_TOPICS);
         return Optional.of(groupsToRootPackages.children.values().stream().map(Node::getName)
-                .filter(g -> !LOCAL_DEPLOYMENT_GROUP_NAME.equals(g))
-                .filter(g -> !g.startsWith(DEVICE_DEPLOYMENT_GROUP_NAME_PREFIX))
+                .filter(g -> g.startsWith(ThingGroupHelper.THING_GROUP_RESOURCE_TYPE_PREFIX))
                 .collect(Collectors.toSet()));
     }
 
--- a/src/main/java/com/aws/greengrass/deployment/DefaultDeploymentTask.java
+++ b/src/main/java/com/aws/greengrass/deployment/DefaultDeploymentTask.java
@@ -277,6 +277,15 @@ public class DefaultDeploymentTask implements DeploymentTask {
         Optional<Set<String>> groupsForDeviceOpt;
         try {
             groupsForDeviceOpt = thingGroupHelper.listThingGroupsForDevice(maxAttemptCount);
+            if (!groupsForDeviceOpt.isPresent()) {
+                // Membership is unknown (the device is not configured to talk to the cloud, so no lookup was
+                // attempted). Unknown must not be treated as an authoritative "member of no thing groups" -
+                // that would drop every recorded group's components from the merge and delete the groups'
+                // records in the post-deployment cleanup. Fall back to the memberships implied by the
+                // persisted group records, exactly as the fetch-failure paths below do.
+                logger.atDebug().log("Thing group membership is unknown, falling back to persisted membership");
+                groupsForDeviceOpt = getPersistedMembershipInfo();
+            }
         } catch (GreengrassV2DataException e) {
             if (e.statusCode() == HttpStatusCode.FORBIDDEN) {
                 // Getting group hierarchy requires permission to call the ListThingGroupsForCoreDevice API which
--- a/src/test/java/com/aws/greengrass/deployment/ThingGroupMembershipPreservationTest.java
+++ b/src/test/java/com/aws/greengrass/deployment/ThingGroupMembershipPreservationTest.java
@@ -219,4 +219,33 @@ class ThingGroupMembershipPreservationTest {
                 "An authoritative empty membership response must keep removing: the device genuinely belongs "
                         + "to no thing groups");
     }
+
+    /**
+     * SCOPE-PINNING TEST (must pass with and without the fix): the unknown-membership fallback implies
+     * membership only for thing-group records ({@code thinggroup/} prefix — the only names a real membership
+     * lookup can return). A record whose name is not a thing group (for example a raw configuration ARN that
+     * failed to parse) must not be resurrected as a membership: its components are not added to the merge's
+     * preserved root set, so it cannot impose stale version constraints on later deployments.
+     */
+    @Test
+    void GIVEN_membership_unknown_WHEN_non_thing_group_record_exists_THEN_it_is_not_treated_as_membership()
+            throws Exception {
+        String nonThingGroupName = "SomeRawConfigurationArn";
+        String nonThingGroupComponent = "component3";
+        groupToRootConfig.lookupTopics(nonThingGroupName, nonThingGroupComponent)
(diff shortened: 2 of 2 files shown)


src/main/java/com/aws/greengrass/deployment/DefaultDeploymentTask.java
```diff
@@ -333,6 +374,137 @@ private Map<String, Set<ComponentRequirementIdentifier>> getNonTargetGroupToRoot
...
+     * Components a local deployment request explicitly asks to remove. Explicit removal is always honored,
+     * even when the local bookkeeping that would normally reflect it is missing.
+     */
+    private Set<String> getComponentsExplicitlyRemovedByLocalRequest() {
+        if (!Deployment.DeploymentType.LOCAL.equals(deployment.getDeploymentType())
+                || deployment.getDeploymentDocument() == null) {
+            return Collections.emptySet();
+        }
+        try {
+            LocalOverrideRequest request = SerializerFactory.getFailSafeJsonObjectMapper()
```
MikeDombo: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

src/test/java/com/aws/greengrass/deployment/ThingGroupMembershipPreservationTest.java
```diff
@@ -0,0 +1,251 @@
...
+
+        List<String> rootPackages = rootPackagesCaptor.getValue();
+        assertNotNull(rootPackages);
+        assertTrue(rootPackages.contains(ROOT_COMPONENT_NAME),
+                "The thing-group record's component must still be preserved");
+        assertFalse(rootPackages.contains(nonThingGroupComponent),
+                "A non-thing-group record must not be treated as a thing group membership by the "
+                        + "unknown-membership fallback");
+    }
+}
```
alter-mage, line 251: **Suggestion (non-blocking):** Would you consider adding a test in this class for the throw-path (case 2 — `listThingGroupsForDevice()` throws an exception, e.g. network unreachable)? The PR description notes that path is already correct, but pinning it here would complete the four-corner contract and guard against future regressions that accidentally weaken the catch-block fallback. Something like:

```java
@Test
void GIVEN_membership_lookup_throws_WHEN_local_deployment_THEN_recorded_group_components_are_preserved()
        throws Exception {
    when(mockThingGroupHelper.listThingGroupsForDevice(anyInt()))
            .thenThrow(new IOException("network unreachable"));
    // ... same assertion pattern as the unknown-membership test
}
```

Not a blocker — the existing exception handling is untouched by this change — but it would make this test class a single source of truth for all membership-resolution outcomes.

PR #1830, opened 2026-07-28
feat: compact config.tlog at boot when it exceeds a size threshold

muboheng:
On a normal reboot config.tlog is only appended to, never rewritten, so a device that reboots frequently (before the runtime 15,000-entry auto-truncate can fire, since its counter resets each boot) accumulates entries without bound, slowing every subsequent boot's tlog replay.

Compact config.tlog (and its config.tlog~ backup) at startup when either exceeds a configurable size threshold, rewriting from the in-memory effective config via the existing atomic CommitableFile path, before the append writer is attached:
- Gate 1 compacts a bloated active tlog on a normal reboot (the !readFromTlog deployment/rollback/recovery paths already rewrite it).
- Gate 2 reclaims a bloated config.tlog~ backup left by the demote, and self-heals a crash between the two dumps on a later boot.

Threshold: Nucleus bootConfigTlogCompactionThresholdBytes (default 20 MB, 0 disables). Long-running devices remain bounded by the existing 15,000-entry runtime truncate.

**Issue #, if available:**

**Description of changes:**

**Documentation Checklist:**

**Compatibility Checklist:**

### DIFF
--- a/src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
+++ b/src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
@@ -9,6 +9,7 @@ import com.amazon.aws.iot.greengrass.component.common.DependencyType;
 import com.aws.greengrass.componentmanager.plugins.docker.DockerApplicationManagerService;
 import com.aws.greengrass.config.ConfigurationReader;
 import com.aws.greengrass.config.ConfigurationWriter;
+import com.aws.greengrass.config.Topic;
 import com.aws.greengrass.config.Topics;
 import com.aws.greengrass.config.UpdateBehaviorTree;
 import com.aws.greengrass.dependency.EZPlugins;
@@ -39,6 +40,7 @@ import com.aws.greengrass.status.FleetStatusService;
 import com.aws.greengrass.telemetry.TelemetryAgent;
 import com.aws.greengrass.telemetry.impl.config.TelemetryConfig;
 import com.aws.greengrass.tes.TokenExchangeService;
+import com.aws.greengrass.util.Coerce;
 import com.aws.greengrass.util.CommitableFile;
 import com.aws.greengrass.util.NucleusPaths;
 import com.aws.greengrass.util.RetryUtils;
@@ -90,6 +92,7 @@ public class KernelLifecycle {
     // TODO:  Use the enum from common library when available
     private static final String DEFAULT_PROVISIONING_POLICY = "PROVISION_IF_NOT_PROVISIONED";
     private static final String SYSTEM_SHUTDOWN_EVENT = "system-shutdown";
+    private static final String PATH_LOG_KEY = "path";
     private static final int MAX_PROVISIONING_PLUGIN_RETRY_ATTEMPTS = 3;
 
     public static final String MULTIPLE_PROVISIONING_PLUGINS_FOUND_EXCEPTION = "Multiple provisioning plugins found "
@@ -345,6 +348,20 @@ public class KernelLifecycle {
             if (!readFromTlog) {
                 kernel.writeEffectiveConfigAsTransactionLog(transactionLogPath);
             }
+
+            // On a normal reboot config.tlog is only appended to, never rewritten, so a device that reboots
+            // frequently (never crossing the runtime 15k-entry auto-truncate, whose counter resets each boot)
+            // accumulates duplicate no-op entries without bound, slowing every subsequent boot's tlog replay.
+            // Compact here when over threshold, reusing the same crash-safe CommitableFile dump path.
+            try {
+                compactConfigTlogIfNeeded(transactionLogPath, readFromTlog);
+            } catch (IOException e) {
+                // Compaction is an optimization; a bloated-but-valid tlog still boots. Never let a failed
+                // dump (e.g. disk full) abort launch — log and continue; the next boot retries.
+                logger.atError().setCause(e).kv(PATH_LOG_KEY, transactionLogPath)
+                        .log("Failed to compact config.tlog at boot; continuing with uncompacted tlog");
+            }
+
             kernel.writeEffectiveConfig();
 
             // hook tlog to config so that changes over time are persisted to the tlog
@@ -353,6 +370,84 @@ public class KernelLifecycle {
         } catch (IOException ioe) {
             logger.atError().setEventType("nucleus-read-config-error").setCause(ioe).log();
             throw new RuntimeException(ioe);
+        }
+    }
+
+    /**
+     * Compact config.tlog and its backup at boot when either exceeds the configured size threshold.
+     *
+     * <p>Both gates reuse the existing atomic {@link Kernel#writeEffectiveConfigAsTransactionLog}
+     * (CommitableFile) path, and run before the append writer is attached, so there is no open-writer
+     * contention and no concurrent service writes:
+     * <ol>
+     *     <li>ACTIVE: on a normal reboot ({@code readFromTlog == true}) a bloated config.tlog is rewritten
+     *     from the in-memory effective config, collapsing duplicate no-op entries. The {@code !readFromTlog}
+     *     paths (deployment/rollback/recovery) already rewrote config.tlog above, so they skip this gate.</li>
+     *     <li>BACKUP: the dump above demotes the old (possibly bloated) config.tlog to config.tlog~. If that
+     *     backup is still over threshold, dump once more so the now-small config.tlog cascades into
+     *     config.tlog~, atomically evicting the bloated backup while keeping a valid, current first-tier
+     *     recovery copy. This gate also self-heals a crash that occurred between the two dumps on a
+     *     subsequent boot (active is small but the backup is still large).</li>
+     * </ol>
+     *
+     * @param transactionLogPath path to config.tlog
+     * @param readFromTlog       whether config was loaded from the existing tlog (a normal reboot)
+     * @throws IOException on failure to stat or rewrite the tlog
+     */
+    private void compactConfigTlogIfNeeded(Path transactionLogPath, boolean readFromTlog) throws IOException {
+        // Read the threshold without mutating the config tree; DeviceConfiguration seeds the default later.
+        Topic thresholdTopic = kernel.getConfig().find(SERVICES_NAMESPACE_TOPIC,
+                DeviceConfiguration.DEFAULT_NUCLEUS_COMPONENT_NAME, CONFIGURATION_CONFIG_KEY,
+                DeviceConfiguration.BOOT_CONFIG_TLOG_COMPACTION_THRESHOLD_BYTES);
+        long threshold = DeviceConfiguration.DEFAULT_BOOT_CONFIG_TLOG_COMPACTION_THRESHOLD_BYTES;
+        Object rawValue = thresholdTopic == null ? null : thresholdTopic.getOnce();
+        if (rawValue instanceof Number) {
+            // Preserve the lenient numeric handling (e.g. a value deserialized as a float).
+            threshold = Coerce.toLong(rawValue);
+        } else if (rawValue != null) {
+            // Parse the raw string ourselves: Coerce.toLong collapses unparseable input to 0, which would
+            // be indistinguishable from the intentional disable value, so a typo would silently disable.
+            try {
+                threshold = Long.parseLong(Coerce.toString(rawValue).trim());
+            } catch (NumberFormatException e) {
+                logger.atWarn().kv("value", Coerce.toString(rawValue))
+                        .kv("defaultBytes", threshold)
+                        .log("Invalid bootConfigTlogCompactionThresholdBytes; using default");
+            }
+        }
+        if (threshold < 0) {
+            logger.atWarn().kv("value", threshold)
+                    .log("Negative bootConfigTlogCompactionThresholdBytes; treating as disabled");
+        }
+        if (threshold <= 0) {
+            return;
+        }
+
+        // Gate 1: compact a bloated active tlog on a normal reboot (the !readFromTlog paths already dumped).
+        if (readFromTlog && Files.exists(transactionLogPath)
+                && Files.size(transactionLogPath) > threshold) {
+            logger.atInfo().kv(PATH_LOG_KEY, transactionLogPath).kv("thresholdBytes", threshold)
+                    .kv("sizeBytes", Files.size(transactionLogPath))
+                    .log("config.tlog exceeds boot compaction threshold; compacting");
+            kernel.writeEffectiveConfigAsTransactionLog(transactionLogPath);
+        }
+
+        // Gate 2: reclaim a bloated backup (config.tlog~) left by the demote above or by a prior boot.
+        Path backupTlogPath = CommitableFile.getBackupFile(transactionLogPath);
+        if (Files.exists(backupTlogPath) && Files.size(backupTlogPath) > threshold) {
+            logger.atInfo().kv(PATH_LOG_KEY, backupTlogPath).kv("thresholdBytes", threshold)
+                    .kv("sizeBytes", Files.size(backupTlogPath))
+                    .log("config.tlog~ backup exceeds threshold; compacting to reclaim space");
+            kernel.writeEffectiveConfigAsTransactionLog(transactionLogPath);
+        }
+
+        // If the effective config itself exceeds the threshold, compaction can't get below it and would
+        // re-fire every boot. Warn so the operator can raise the threshold.
+        if (Files.exists(transactionLogPath) && Files.size(transactionLogPath) > threshold) {
+            logger.atWarn().kv(PATH_LOG_KEY, transactionLogPath).kv("thresholdBytes", threshold)
+                    .kv("sizeBytes", Files.size(transactionLogPath))
+                    .log("config.tlog still exceeds threshold after compaction; effective config may be larger "
+                            + "than the threshold — consider raising bootConfigTlogCompactionThresholdBytes");
         }
     }
 
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/lifecyclemanager/KernelRestartTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/lifecyclemanager/KernelRestartTest.java
@@ -10,15 +10,20 @@ import com.aws.greengrass.integrationtests.BaseITCase;
 import com.aws.greengrass.integrationtests.util.ConfigPlatformResolver;
 import com.aws.greengrass.lifecyclemanager.Kernel;
 import com.aws.greengrass.lifecyclemanager.exceptions.ServiceLoadException;
+import com.aws.greengrass.util.Coerce;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Test;
 
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.time.Duration;
 
 import static com.github.grantwest.eventually.EventuallyLambdaMatcher.eventuallyEval;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.greaterThan;
 import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.lessThan;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 
 class KernelRestartTest extends BaseITCase {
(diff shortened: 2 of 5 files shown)


aws-kevinrickard:
I was thinking about the 20MB default threshold, if it's too high. The close to worst case for line count for 20MB might be ~200k lines. I ran a benchmark on a datacenter x86 machine and it took ~1 second to process that many lines, so a multiple of that on a weaker edge device is possible. I think this is acceptable, and the 20MB threshold should give headroom for reducing the chances of compacting on boot. 

No problems, just posting the results of the benchmark.

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
```diff
@@ -356,6 +365,58 @@ void initConfigAndTlog() {
...
+        Topic thresholdTopic = kernel.getConfig().find(SERVICES_NAMESPACE_TOPIC,
+                DeviceConfiguration.DEFAULT_NUCLEUS_COMPONENT_NAME, CONFIGURATION_CONFIG_KEY,
+                DeviceConfiguration.BOOT_CONFIG_TLOG_COMPACTION_THRESHOLD_BYTES);
+        long threshold = thresholdTopic == null
+                ? DeviceConfiguration.DEFAULT_BOOT_CONFIG_TLOG_COMPACTION_THRESHOLD_BYTES
+                : Coerce.toLong(thresholdTopic);
+        if (threshold <= 0) {
+            return;
+        }
+
```
saranyailla, line 425: Tradeoff discussion: A config.tlog file with very few lines but large payloads (secrets, big config blobs) could hit a byte threshold quickly even when there's no real "bloat" problem and no repeated/superseded entries to collapse. This can trigger compaction where it doesn't actually help, wasting the write cost for little gain.

On the other hand, a topic that changes many times (many small writes) but never crosses a size threshold might sit there accumulating history indefinitely under a byte-based scheme
even though it's exactly the case line-based compaction is designed to fix. Under a size threshold, this device might compact few times despite having real bloat (in count/replay terms) to reclaim.

That said, checking byte size is cheaper, so let's keep it. But apply it consistently across all truncation paths, rather than bytes at boot and a hardcoded 15,000-entry count at runtime.

Also, some customers want quick boot up times, so can we consider making it optional?
```
configCompaction:
        maxConfigTlogSizeInBytes: 5 MB #default
        compactOnBoot: true # default
```

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
```diff
@@ -356,6 +365,58 @@ void initConfigAndTlog() {
...
+     *     recovery copy. This gate also self-heals a crash that occurred between the two dumps on a
+     *     subsequent boot (active is small but the backup is still large).</li>
+     * </ol>
+     *
+     * @param transactionLogPath path to config.tlog
+     * @param readFromTlog       whether config was loaded from the existing tlog (a normal reboot)
+     * @throws IOException on failure to stat or rewrite the tlog
+     */
+    private void compactConfigTlogIfNeeded(Path transactionLogPath, boolean readFromTlog) throws IOException {
+        // Read the threshold without mutating the config tree; DeviceConfiguration seeds the default later.
```
aws-kevinrickard, line 398: > "DeviceConfiguration seeds the default later," 

As written, `bootConfigTlogCompactionThresholdBytes` exists only as a Java constant: the default value lives in code (`KernelLifecycle`'s fallback when `find()` returns null), and no topic for it is ever created in the config tree. That means a device running with the default shows no trace of this setting in its effective config — users can't discover that the setting exists or what it's set to by inspecting the device; they'd have to already know the key name from docs or source.

The code comment in `compactConfigTlogIfNeeded` ("DeviceConfiguration seeds the default later") suggests seeding was intended, but nothing in the PR does it.

Other Nucleus settings make their defaults visible through a `DeviceConfiguration` getter with a lazy default, e.g. `deploymentConfigurationTimeSource` (#1706):

```java
public Topic getDeploymentConfigurationTimeSource() {
    return getTopic(DEVICE_PARAM_DEPLOYMENT_CONFIGURATION_TIME_SOURCE)
            .dflt(DEPLOYMENT_CONFIGURATION_TIME_SOURCE_DEPLOYMENT_CREATION_TIME);
}
```

That works there because the consumer reads the value through the getter at runtime, so `.dflt()` creates the topic on first access. This key can't do that on its consuming path — compaction runs before `DeviceConfiguration` is constructed, so the `find()`-with-fallback in `KernelLifecycle` is correct and should stay. But that also means nothing ever creates the topic.

Suggestion: add the getter per the usual pattern and invoke it once in the constructor, which is exactly what `componentStoreMaxSizeBytes` does:

```java
// constructor
getBootConfigTlogCompactionThresholdBytes().dflt(DEFAULT_BOOT_CONFIG_TLOG_COMPACTION_THRESHOLD_BYTES);
```

No behavior change (the seeded default and the code fallback are the same value); it just makes the setting and its default visible in the effective config from first boot onward — and makes the existing code comment in `compactConfigTlogIfNeeded` true. If leaving it unseeded is intentional, please correct that comment instead.

(AI-assisted comment)

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
```diff
@@ -345,6 +347,13 @@ void initConfigAndTlog() {
             if (!readFromTlog) {
                 kernel.writeEffectiveConfigAsTransactionLog(transactionLogPath);
             }
+
+            // On a normal reboot config.tlog is only appended to, never rewritten, so a device that reboots
+            // frequently (never crossing the runtime 15k-entry auto-truncate, whose counter resets each boot)
+            // accumulates duplicate no-op entries without bound, slowing every subsequent boot's tlog replay.
+            // Compact here when over threshold, reusing the same crash-safe CommitableFile dump path.
+            compactConfigTlogIfNeeded(transactionLogPath, readFromTlog);
```
aws-kevinrickard: This call introduces a new way for boot to fail on the normal-reboot path. `compactConfigTlogIfNeeded` throws `IOException`, and the outer catch in `initConfigAndTlog` converts that into a `RuntimeException`, so a compaction failure aborts the Nucleus launch.

Before this change, the `readFromTlog` path only read `config.tlog` — a device with a valid tlog booted regardless of whether it could write a new copy of it. After this change, a device whose tlog is over the threshold must successfully write a full effective-config dump (open `config.tlog+`, write, rename) to get through boot. The failure case isn't hypothetical: a device that has accumulated a 20 MB tlog is disproportionately likely to also be low on disk, and disk-full is precisely the condition under which the dump fails — turning a device that booted yesterday into one that doesn't boot today. It's also self-defeating in that scenario: the compaction that would have freed the space is what's preventing startup.

Since compaction is an optimization (a bloated-but-valid tlog is still fully functional — this is exactly how every device behaves before this feature), failure to compact shouldn't be fatal. Suggest catching `IOException` inside `compactConfigTlogIfNeeded` (or around the call), logging at error, and continuing boot with the uncompacted tlog:

```java
try {
    compactConfigTlogIfNeeded(transactionLogPath, readFromTlog);
} catch (IOException e) {
    logger.atError().setCause(e).kv("path", transactionLogPath)
            .log("Failed to compact config.tlog at boot; continuing with uncompacted tlog");
}
```

The device then retries automatically on every subsequent boot, so a transient failure self-heals.

Worth noting the asymmetry with the runtime truncation, which already treats this failure as non-fatal: `ConfigurationWriter.truncateTlog` catches its own IO errors, undoes the rename, and schedules a retry rather than taking the process down. Boot compaction failing softly would be consistent with that precedent.

(For the sub-steps inside the dump itself, `CommitableFile.commit()` already swallows move errors, so partial-commit states are handled by the existing backup-recovery path on the next boot — the only genuinely throwing step is creating/writing the new file, which is exactly the disk-full case above.)

(AI-assisted comment)

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
```diff
@@ -356,6 +365,58 @@ void initConfigAndTlog() {
...
+     * @throws IOException on failure to stat or rewrite the tlog
+     */
+    private void compactConfigTlogIfNeeded(Path transactionLogPath, boolean readFromTlog) throws IOException {
+        // Read the threshold without mutating the config tree; DeviceConfiguration seeds the default later.
+        Topic thresholdTopic = kernel.getConfig().find(SERVICES_NAMESPACE_TOPIC,
+                DeviceConfiguration.DEFAULT_NUCLEUS_COMPONENT_NAME, CONFIGURATION_CONFIG_KEY,
+                DeviceConfiguration.BOOT_CONFIG_TLOG_COMPACTION_THRESHOLD_BYTES);
+        long threshold = thresholdTopic == null
+                ? DeviceConfiguration.DEFAULT_BOOT_CONFIG_TLOG_COMPACTION_THRESHOLD_BYTES
+                : Coerce.toLong(thresholdTopic);
```
aws-kevinrickard: If the configured value doesn't parse as a number (say someone sets `bootConfigTlogCompactionThresholdBytes: "20MB"`), `Coerce.toLong` returns 0, which is indistinguishable from the documented disable value — so a typo silently turns the feature off, with no log line hinting at why a device's tlog keeps growing. That's the least helpful interpretation of the user's intent: they were clearly trying to configure compaction, not disable it.

Suggest validating the raw value and falling back to the default when it's invalid:

- **Unparseable** → warn and use `DEFAULT_BOOT_CONFIG_TLOG_COMPACTION_THRESHOLD_BYTES`. Note this can't be detected from `Coerce.toLong`'s output (it collapses bad input to 0); the check needs the raw topic value, e.g. `Long.parseLong(Coerce.toString(thresholdTopic))` in a try/catch.
- **0** → keep as the documented disable, no log needed.
- **Negative** → keep treating as disabled (the current `<= 0` behavior), but log a warn. Deliberately *not* suggesting falling back to the default here: `-1` is a common "disable" idiom, and replacing it with an enabled 20 MB threshold would invert the user's likely intent. The warn covers the case where the negative value was a mistake.

Something like:

```java
long threshold = DeviceConfiguration.DEFAULT_BOOT_CONFIG_TLOG_COMPACTION_THRESHOLD_BYTES;
if (thresholdTopic != null) {
    try {
        threshold = Long.parseLong(Coerce.toString(thresholdTopic));
        if (threshold < 0) {
            logger.atWarn().kv("value", threshold)
                    .log("Negative boot tlog compaction threshold; treating as disabled");
        }
    } catch (NumberFormatException e) {
        logger.atWarn().kv("value", Coerce.toString(thresholdTopic))
                .kv("default", threshold)
                .log("Invalid boot tlog compaction threshold; using default");
    }
}
if (threshold <= 0) {
    return;
}
```

On the design itself: 0-as-disable is what makes the typo case sting — the parser's failure value (`Coerce.toLong` → 0) collides exactly with the disable sentinel, and read literally, threshold 0 would mean "compact whenever the tlog exceeds 0 bytes" (always), the opposite of what it does here. A separate `bootConfigTlogCompactionEnabled` boolean plus the threshold would avoid both issues and is worth considering since config keys are hard to change once released. That said, 0-to-disable is a common enough idiom, and with the validation above the collision is neutralized — so keeping the single key as-is with the validation fix seems like the right tradeoff. Flagging the alternative mainly in case you have your own reasons to prefer the explicit flag.

(AI-assisted comment)

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
aws-kevinrickard: On your first point, I was thinking something similar, but didn't find any preferred alternatives.

Byte checking is simpler at boot. Switching the runtime check to bytes is possible (track bytes appended plus file size at attach), but the existing line-count trigger is battle-tested, and there's no known customer benefit from changing its metric — so keeping the runtime behavior as-is reduces risk. Switching to lines for both checks has the opposite problem: plumbing the line count out of the reader back to this point in the code is a much bigger change.

On the large-payload case: this is a good point, and it's a bit worse than stated — if the effective config itself is over the threshold, the compacted file stays over it, so boot compaction would re-fire every boot for no gain. Might be worth a warn log when the post-compaction size is still above the threshold, advising raising it.

On making it optional: setting the threshold to 0 already disables boot compaction, so I don't think we need a separate compactOnBoot flag. Lowering the default from 20 MB is reasonable though — from some quick measurements, replaying a 20 MB / ~200k-line tlog costs about a second on x86 (and the file is parsed twice at boot), so something in the 2–5 MB range keeps the worst-case boot penalty small while still rarely triggering.

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
```diff
@@ -356,6 +365,58 @@ void initConfigAndTlog() {
...
+        if (threshold <= 0) {
+            return;
```
alter-mage, line 423: **Nit: unparseable threshold value silently disables compaction**

`Coerce.toLong` returns `0` for non-numeric strings (e.g., `"20MB"`), and the code treats `0` as the documented disable value. This means a configuration typo silently turns off compaction with no log output — the user intended to configure it, not disable it.

Consider distinguishing between an explicit `0` (intentional disable) and a parse failure (likely a typo). A warning log with a fallback to the default would make this more debuggable:

```java
long threshold;
if (thresholdTopic == null) {
    threshold = DeviceConfiguration.DEFAULT_BOOT_CONFIG_TLOG_COMPACTION_THRESHOLD_BYTES;
} else {
    threshold = Coerce.toLong(thresholdTopic);
    if (threshold == 0 && thresholdTopic.getOnce() != null
            && !Integer.valueOf(0).equals(thresholdTopic.getOnce())) {
        logger.atWarn().kv("configuredValue", thresholdTopic.getOnce())
            .log("bootConfigTlogCompactionThresholdBytes is not a valid number; "
                 + "using default of {} bytes",
                 DeviceConfiguration.DEFAULT_BOOT_CONFIG_TLOG_COMPACTION_THRESHOLD_BYTES);
        threshold = DeviceConfiguration.DEFAULT_BOOT_CONFIG_TLOG_COMPACTION_THRESHOLD_BYTES;
    }
}
```

aws-kevinrickard:
Finished review, summary of comments:

### Blocking:

Config parsing improvements: 
https://github.com/aws-greengrass/aws-greengrass-nucleus/pull/1830#discussion_r3669932752 

Don't abort for compaction failure: 
https://github.com/aws-greengrass/aws-greengrass-nucleus/pull/1830#discussion_r3669899184 

 

### Not blocking:

Add default value to config seeding:
https://github.com/aws-greengrass/aws-greengrass-nucleus/pull/1830#discussion_r3669869304

warn log when the post-compaction size is still above the threshold, advising raising it: 
https://github.com/aws-greengrass/aws-greengrass-nucleus/pull/1830#discussion_r3670029486

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
muboheng: agree with Kevin's comment here. Byte based option is best of the boot case and while a byte based solution during run time will help with consistency, it may introduce changes to an already tested behavior.

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
muboheng: ack

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
muboheng: ack

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
muboheng: going to reduce the default to 10 as a conservative default.

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
```diff
@@ -356,6 +373,80 @@ void initConfigAndTlog() {
...
+     *     config.tlog~, atomically evicting the bloated backup while keeping a valid, current first-tier
+     *     recovery copy. This gate also self-heals a crash that occurred between the two dumps on a
+     *     subsequent boot (active is small but the backup is still large).</li>
+     * </ol>
+     *
+     * @param transactionLogPath path to config.tlog
+     * @param readFromTlog       whether config was loaded from the existing tlog (a normal reboot)
+     * @throws IOException on failure to stat or rewrite the tlog
+     */
+    private void compactConfigTlogIfNeeded(Path transactionLogPath, boolean readFromTlog) throws IOException {
```
MikeDombo, line 397: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

The cyclomatic complexity of this method is 12. By comparison, 98% of the methods in the CodeGuru reference dataset have a lower cyclomatic complexity. This indicates the method has a high number of decisions and it can make the logic difficult to understand and test.

We recommend that you simplify this method or break it into multiple methods. For example, consider extracting the code block on lines 409-447 into a separate method.

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
```diff
@@ -356,6 +373,80 @@ void initConfigAndTlog() {
...
+        // Read the threshold without mutating the config tree; DeviceConfiguration seeds the default later.
+        Topic thresholdTopic = kernel.getConfig().find(SERVICES_NAMESPACE_TOPIC,
+                DeviceConfiguration.DEFAULT_NUCLEUS_COMPONENT_NAME, CONFIGURATION_CONFIG_KEY,
+                DeviceConfiguration.BOOT_CONFIG_TLOG_COMPACTION_THRESHOLD_BYTES);
+        long threshold = DeviceConfiguration.DEFAULT_BOOT_CONFIG_TLOG_COMPACTION_THRESHOLD_BYTES;
+        if (thresholdTopic != null && thresholdTopic.getOnce() != null) {
+            try {
+                // Parse the raw value: Coerce.toLong collapses unparseable input to 0, which would be
+                // indistinguishable from the intentional disable value, so a typo would silently disable.
+                threshold = Long.parseLong(Coerce.toString(thresholdTopic.getOnce()).trim());
```
aws-kevinrickard: nit: this rejects values that `Coerce.toLong` used to accept and would make sense to accept (although are a bit weird) — anything deserialized as a floating-point number, e.g. YAML `1e7`, stringifies to `"1.0E7"` and lands in the "Invalid ... using default" path. A byte count arguably shouldn't be a float, so warning and falling back to the default is a reasonable response; the only mildly unfortunate case is `1e7`, which is just integer shorthand.

If you want to preserve the old leniency, `Coerce.toLong` for anything already numeric and string-parse only the rest:

```java
Object rawValue = thresholdTopic.getOnce();
if (rawValue instanceof Number) {
    threshold = Coerce.toLong(rawValue);
} else {
    // string parsing, as today
}
```

aws-kevinrickard (dismissed):

alter-mage (dismissed):

alter-mage (approved):

src/main/java/com/aws/greengrass/lifecyclemanager/KernelLifecycle.java
```diff
@@ -356,6 +373,84 @@ void initConfigAndTlog() {
...
+     *     config.tlog~, atomically evicting the bloated backup while keeping a valid, current first-tier
+     *     recovery copy. This gate also self-heals a crash that occurred between the two dumps on a
+     *     subsequent boot (active is small but the backup is still large).</li>
+     * </ol>
+     *
+     * @param transactionLogPath path to config.tlog
+     * @param readFromTlog       whether config was loaded from the existing tlog (a normal reboot)
+     * @throws IOException on failure to stat or rewrite the tlog
+     */
+    private void compactConfigTlogIfNeeded(Path transactionLogPath, boolean readFromTlog) throws IOException {
```
MikeDombo, line 397: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

The cyclomatic complexity of this method is 13. By comparison, 98% of the methods in the CodeGuru reference dataset have a lower cyclomatic complexity. This indicates the method has a high number of decisions and it can make the logic difficult to understand and test.

We recommend that you simplify this method or break it into multiple methods. For example, consider extracting the code block on lines 413-451 into a separate method.

aws-kevinrickard (approved):

Issue #1832, opened 2026-08-06
(com.aws.greengrass.mqttclient.spool): DiskSpooler persistentQueueSync() blocks all component startup on main thread


gregbreen:
**Describe the bug**
When using the Disk Spooler plugin (`aws.greengrass.DiskSpooler`) with a large number of persisted messages, all component startup is blocked for the duration of the spooler queue synchronization. `Spool.java#persistentQueueSync()` iterates every persisted message synchronously
on the main Kernel initialization thread, reading each message payload from SQLite individually. No component lifecycle transitions (NEW → INSTALLED → STARTING → RUNNING) can begin until this completes.

On constrained hardware with ~26,500 spooled messages (accumulated during a network fault), this results in a ~5 minute startup delay before any user component begins.

**To Reproduce**

1. Deploy `aws.greengrass.DiskSpooler` (v1.0.2) with Nucleus configured for disk storage:
 ```yaml
 mqtt:
   spooler:
	 maxSizeInBytes: "262144000"
	 storageType: "Disk"
```

2. Allow messages to accumulate in the spooler while the device is offline (e.g., ~26,000 messages)
3. Reboot the device

**Expected behavior**

Component startup should not be blocked by the spooler loading persisted messages. User components that do not depend on spooled message state should be able to start while the spooler drains in the background.

**Actual behavior**

All component startup is delayed by the synchronous persistentQueueSync() call. The delay scales linearly with message count and is unbounded:

```
02:21:12 [INFO] KernelLifecycle: system-start
02:21:14 [INFO] DiskSpool: Finished setting up Database
														← 5 min 19 sec gap (no log output)
02:26:33 [INFO] Spool: Messages added to spool runtime queue. {numMessagesAdded=26568}
02:26:33 [INFO] Kernel: Attempt to load service from plugins. {serviceName=UserComponent}
02:26:42 [INFO] GenericExternalService: service-set-state. {serviceName=UserComponent, newState=STARTING}
```

The same device with an empty spooler starts all components in under 3 seconds.

The blocking code path is in Spool.java (https://github.com/aws-greengrass/aws-greengrass-nucleus/blob/main/src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java):

```
public void persistentQueueSync(Iterable diskQueueOfIds, CloudMessageSpool persistenceSpool)
	  throws InterruptedException, SpoolerStoreException {
  for (long currentId : diskQueueOfIds) {
[...]

PR #1835, opened 2026-08-13
fix(spool): move persistent queue sync off main thread to unblock sta…

saranyailla:
…rtup

**Issue #, if available:**
https://github.com/aws-greengrass/aws-greengrass-nucleus/issues/1832

**Description of changes:**

####   Problem

  DiskSpooler's `persistentQueueSync()` ran synchronously in the Spool constructor on the kernel's main init thread, blocking all component lifecycle transitions until every persisted message was read from SQLite individually (~5 min for 26,500 messages on constrained hardware).

####   Changes
  - `persistentQueueSync` now runs on a background thread via CompletableFuture.runAsync() instead of blocking the `Spool` constructor
  - `popId()` and `popOutMessagesWithQosZero()` are gated on sync completion via `awaitDiskQueueLoaded()` so drain ordering is preserved
  - `addMessage()` waits for the disk load to complete.
  - If using in-memory storage, `diskQueueLoaded` is set to an already-completed future (no-op path)

####   Testing

  - Updated unit tests use a same-thread ExecutorService implementation to keep tests deterministic
  - Existing tests for disk spooler setup, sync failure fallback, and add-on-disk-failure all pass with the new constructor signature
  - Verified popId() blocks until background sync completes
  - Verified addMessage() does not block during background sync

**Documentation Checklist:**

**Compatibility Checklist:**


src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
muboheng: There is a collision here by making the setup asyn when adding to the spooler. Any messages that arrive during this window will start incrementing from 0 until the background sync is done

muboheng: the message id needs some redesign for this to work, other alternative is to have the write operation to block on the queue sync, but I don't think this will be the right decision.

muboheng: This operation might also by out of order if new messages arrive during the sync window. not sure how important ordering is to diskspooler.

muboheng, line 222: addMessage does a lock here but in the asyn persistentQueueSync it does not.

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -151,13 +171,21 @@ private CloudMessageSpool getPersistenceSpoolGGService()
...
+
+    /**
+     * Blocks until all persisted message IDs have been loaded from disk into the in-memory deque.
+     * Returns immediately if already complete or if using Memory storage.
+     *
+     * <p>Must be called by any method that consumes from or iterates the deque to guarantee
+     * all persisted messages are present and ordered correctly.</p>
+     */
+    private void awaitDiskQueueLoaded() {
+        diskQueueLoaded.join();
```
muboheng: since diskQueueLoaded is a CompletableFuture, the .join operation ignores InterruptedException, consider using .get here with a timeout

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -110,35 +119,46 @@ private void setSpoolerConfigFromDeviceConfig(Topics topics) {
...
+    private CloudMessageSpool setupSpooler(ExecutorService executorService) {
         if (config.getStorageType() == SpoolerStorageType.Disk) {
             try {
-                return getPersistenceSpoolGGService();
+                return setupDiskSpooler(executorService);
             } catch (ServiceLoadException | IOException e) {
-                //log and use InMemorySpool
                 logger.atWarn()
                         .kv(PERSISTENCE_SPOOL_SERVICE_NAME_KEY, config.getPersistenceSpoolServiceName())
                         .cause(e).log("Persistence spool set up failed, defaulting to InMemory Spooler");
```
muboheng, line 130: since this is a sync process we will likely not catch these exceptions in time to default back the the inMemorySpooler

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
muboheng, line 424: went down the rabit hole to trace the disk spookler code for a bit, but this is an expensive operation with a DB I/O call every time no wonder it takes so long. The right fix would be to add an interface to load or have batch calls, but for the purpose of this PR this might be overkill

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -110,35 +119,46 @@ private void setSpoolerConfigFromDeviceConfig(Topics topics) {
...
-            persistenceSpool.initializeSpooler();
+        if (!(locatedService instanceof CloudMessageSpool)) {
+            throw new ServiceLoadException(
+                    "The Greengrass service located was not an instance of CloudMessageSpool");
+        }
+        CloudMessageSpool persistenceSpool = (CloudMessageSpool) locatedService;
+        persistenceSpool.initializeSpooler();
+        // Run the per-message sync on a background thread to avoid blocking kernel startup
+        // (issue #1832). popId() will wait for this to complete before draining messages.
+        diskQueueLoaded = CompletableFuture.runAsync(() -> {
```
alter-mage, line 155: **Concurrency: `persistentQueueSync` now races `addMessage` on `nextId`/`queueOfMessageId` (no shared lock).**

Before this change, `persistentQueueSync` ran inside the `Spool` constructor on the main thread, before the instance was visible to any other thread — so it always completed (and advanced `nextId` to `highestId + 1`) before any `addMessage` could run. Moving it onto `CompletableFuture.runAsync(...)` makes it run concurrently with `addMessage` for the first time, and the two don't share synchronization:

- `addMessage` holds `lock` around `nextId.getAndIncrement()` and `queueOfMessageId.putLast(id)`.
- `persistentQueueSync` takes no lock around its `putLast` loop or the final `nextId.set(Math.max(nextId.get(), highestId + 1))` (a non-atomic get-then-set).

`nextId` starts at 0 on a fresh restart and is only advanced past the persisted range on the last line of the sync. If enough messages are added during the sync window to push `nextId` up into the persisted-id range before that `set` runs, an added message and a persisted message can claim the same id, colliding in `queueOfMessageId` and in the id-keyed store. Timing-dependent, but the unsynchronized access is the defect and it's new to this PR.

**Suggested fix:** don't hold `lock` across the whole sync (that would move the large-backlog stall onto the `addMessage` path and defeat this PR's goal). Instead advance `nextId` past the highest persisted id **synchronously, before spawning the background task**, so `addMessage` always allocates above the persisted range; the background loop then only appends. If reading the max id up front isn't cheap via the DiskSpooler API, a narrower alternative is to take `lock` only around each individual `putLast`/`nextId` update inside the loop (releasing between iterations).

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -205,11 +233,14 @@ private void addMessageToSpooler(long id, SpoolMessage message) {
...
     /**
      * Pop the id of the oldest PublishRequest.
+     * Waits for disk-to-memory sync to complete before draining to ensure
+     * all persisted messages are available and ordered correctly.
      *
      * @return message id
      * @throws InterruptedException the thread is interrupted while popping the first id from the queue
      */
     public long popId() throws InterruptedException {
+        awaitDiskQueueLoaded();
```
alter-mage, line 252: **Ordering: a mid-load `addMessage` can be published before older persisted messages.**

`popId` uses `queueOfMessageId.takeFirst()`, so publish order = deque position, not id value. Because `addMessage` is (intentionally) not gated on the sync, a message added while the sync is mid-load is appended between not-yet-loaded persisted ids:

```
sync: putLast(5), putLast(6)
app:  putLast(9)          // lands here
sync: putLast(7), putLast(8)
deque: [5, 6, 9, 7, 8] -> popId drains 5,6,9,7,8
```

So a newer message (9) can go out before older persisted messages (7, 8). No collision — purely a positional reordering, and only within the sync window. Is strict oldest-first a guarantee the spooler makes, or is delivery order best-effort? If best-effort this is fine; if it's a guarantee, the id-collision fix above doesn't address it (that only fixes ids) and appends would need to be sequenced after the persisted load completes.

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
alter-mage: **`awaitDiskQueueLoaded()` uses an unbounded, uninterruptible `join()`.**

`diskQueueLoaded.join()` waits forever and ignores interruption. If the background sync stalls (e.g. a slow/stuck read on a large `spooler.db`), the first `popId()` blocks indefinitely — and unlike the pre-PR behavior, Nucleus reports itself started while the publish path is silently wedged. `popId` already declares `throws InterruptedException` and the spooler loop shuts down on interrupt, but `join()` won't observe it. Consider `diskQueueLoaded.get()` (interruptible, fits the existing `InterruptedException` contract) or a bounded `get(timeout)` that logs a warning so a stuck sync surfaces instead of hanging.

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -151,12 +172,25 @@ private CloudMessageSpool getPersistenceSpoolGGService()
...
+    /**
+     * Blocks until all persisted message IDs have been loaded from disk into the in-memory deque.
+     * Returns immediately if already complete or if using Memory storage.
+     *
+     * <p>Must be called by any method that consumes from or iterates the deque to guarantee
+     * all persisted messages are present and ordered correctly.</p>
+     */
+    private void awaitDiskQueueLoaded() throws InterruptedException {
+        try {
+            diskQueueLoaded.get();
```
MikeDombo, line 190: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

It looks like you are using `CompletableFuture.get()` or `Future.get()` without setting its timeout parameter. Be sure to set timeout values or handle timeout exceptions to prevent threads from waiting for a process to end. This can lead to downtime for your customers. For more information, see [CompletableFuture](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#get-long-java.util.concurrent.TimeUnit-) and [Future](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/Future.html#get-long-java.util.concurrent.TimeUnit-) documentation.

Suggested remediation:
Set timeout parameters for `CompletableFuture.get()` or `Future.get()`.Exceptions, such as `ExecutionException`, `InterruptedException` and `TimeoutException` are caught and errors are logged.
```
@@ -189,3 +189,3 @@
         try {
-            diskQueueLoaded.get();
+            diskQueueLoaded.get(100, TimeUnit.MILLISECONDS);
         } catch (ExecutionException e) {
@@ -193,2 +193,8 @@
             logger.atDebug().cause(e).log(&#34;Disk queue load completed with an error&#34;);
+        } catch (InterruptedException ie) {
+            log.error(&#34;Thread interrupted.&#34;, ie);
+            throw new InterruptedException(ie.getMessage());
+        } catch (TimeoutException te) {
+            log.error(&#34;Future is not done when timeout.&#34;, te);
+            throw new TimeoutException(te.getMessage());
         }

```

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
saranyailla: Yeah, I am also working on the second approach to not break the customer's Nucleus when they don't update their spooler component. 

See: https://github.com/aws-greengrass/aws-greengrass-nucleus/pull/1834 and https://github.com/aws-greengrass/aws-greengrass-disk-spooler/pull/45

saranyailla: Resolved by waiting for initial sync to finish before adding new messages to spooler

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
saranyailla: The code path that gives `ServiceLoadException` runs in the same thread. All of `setupDiskSpooler` is not async.

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -325,29 +405,42 @@ public void persistentQueueSync(Iterable<Long> diskQueueOfIds, CloudMessageSpool
...
+        nextId.set(highestId + 1);
+
+        // load message bodies into the runtime queue up to capacity.
         int numMessages = 0;
+        SpoolerStoreException e = null;
         int queueOfMessageIdInitSize = queueOfMessageId.size();
         for (long currentId : diskQueueOfIds) {
             numMessages++;
-            //Check for queue space and remove if necessary
             SpoolMessage message = persistenceSpool.getMessageById(currentId);
```
alter-mage, line 424: Nit: `persistentQueueSync` rebuilds `queueOfMessageId` in the order `getAllMessageIds()` returns ids, and `popId()` publishes in deque order — so oldest-first delivery depends on that returned order. In the DiskSpooler plugin, `getAllMessageIds()` runs `SELECT message_id FROM spooler;` with no `ORDER BY`. Since `message_id` is `INTEGER PRIMARY KEY`, SQLite usually returns rows in ascending rowid order, but that's incidental, not contractual — so FIFO restoration here is "works by luck." (The DAO method's javadoc already claims "ordered iterable," so the code doesn't match its documented contract either.)

Adding `ORDER BY message_id ASC` to that query would make oldest-first a guarantee. It's performance-neutral — `message_id` is the primary key, so it's just an ordered PK-btree walk with no extra sort step. That change lives in the aws-greengrass-disk-spooler repo, not here — flagging it so this PR's ordering assumption rests on a real guarantee rather than incidental behavior.

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
alter-mage, line 190: `awaitDiskQueueLoaded()` calls `diskQueueLoaded.get()` with no timeout, so `popId()` and `addMessage()` block for the entire duration of the background load. On a device with a large `spooler.db` (or slow/constrained disk), that can be many minutes; if the DiskSpooler plugin stalls (e.g. a held SQLite lock), it's unbounded. The `.get()` is interruptible so shutdown is fine, but during normal operation nothing caps the wait — and unlike the old synchronous behavior, the device reports itself started while the publish path is silently parked.

Suggest bounding the wait with `diskQueueLoaded.get(timeout, unit)`, and making the timeout customer-configurable (e.g. an `mqtt.spooler` config key, with a sane default). On timeout, log a clear warning and proceed with whatever IDs have been loaded so far, so the MQTT path stays responsive instead of hanging indefinitely. Worth deciding on-timeout semantics explicitly: proceed-with-partial (responsive, some persisted messages not yet queued) vs keep-waiting.

saranyailla: I looked into this earlier and don't think a timeout is the right fix. There's no correct value for it as load time depends on db size and disk speed. And proceeding with a partial queue on timeout means `addMessage` can reuse an `id` that's still on disk, overwriting the existing messages. Looking at the code more closely, the thing we actually want to guard against is a single stuck DB call, which isn't common and belongs in the spooler plugin component. Making this configurable would need a separate design of its own.

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
saranyailla: Yes, this is noted.

alter-mage (dismissed):

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -110,35 +120,46 @@ private void setSpoolerConfigFromDeviceConfig(Topics topics) {
...
+            throw new ServiceLoadException(
+                    "The Greengrass service located was not an instance of CloudMessageSpool");
+        }
+        CloudMessageSpool persistenceSpool = (CloudMessageSpool) locatedService;
+        persistenceSpool.initializeSpooler();
+        // Run the per-message sync on a background thread to avoid blocking kernel startup
+        // (issue #1832). popId() will wait for this to complete before draining messages.
+        diskQueueLoaded = CompletableFuture.runAsync(() -> {
             try {
                 persistentQueueSync(persistenceSpool.getAllMessageIds(), persistenceSpool);
```
yitingb: Before this PR, `getAllMessageIds()` ran synchronously inside `getPersistenceSpoolGGService()`, so an `IOException` propagated to the catch in `setupSpooler()` and the spool fell back to `InMemorySpool`. Degraded but coherent: the disk store was never touched again for that boot.

With this change, `getAllMessageIds()` runs inside the `runAsync` lambda, after `setupDiskSpooler()` has already returned the disk spooler to the constructor. The new `catch (IOException)` logs a warning and completes `diskQueueLoaded` normally. On that path:

1. No persisted id ever enters `queueOfMessageId`, so previously spooled messages are stranded on disk and never drained.
2. Pass 1 of `persistentQueueSync` never runs, so `nextId` stays at its initial value. `addMessage()` then hands out ids 0, 1, 2... which can collide with rows still in the SQLite DB. Depending on how the plugin's `add()` handles a duplicate key, a new message either overwrites an unsent persisted one, or the later `removeMessageById()` after publish deletes the wrong row. This id-collision hazard could not occur before, and it is the same class of bug pass 1 was added to prevent on the `SpoolerStoreException` path.

The existing test `GIVEN_spooler_config_disk_WHEN_setup_spooler_failed_THEN_use_in_memory_spool` still passes but no longer tests what its name says: with the change, the spool keeps the (mocked) disk spooler.

Is this an expected behavior change?

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -151,12 +186,25 @@ private CloudMessageSpool getPersistenceSpoolGGService()
...
+    /**
+     * Blocks until all persisted message IDs have been loaded from disk into the in-memory deque.
+     * Returns immediately if already complete or if using Memory storage.
+     *
+     * <p>Must be called by any method that consumes from or iterates the deque to guarantee
+     * all persisted messages are present and ordered correctly.</p>
+     */
+    private void awaitDiskQueueLoaded() throws InterruptedException {
+        try {
+            diskQueueLoaded.get();
```
MikeDombo, line 206: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

It looks like you are using `CompletableFuture.get()` or `Future.get()` without setting its timeout parameter. Be sure to set timeout values or handle timeout exceptions to prevent threads from waiting for a process to end. This can lead to downtime for your customers. For more information, see [CompletableFuture](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#get-long-java.util.concurrent.TimeUnit-) and [Future](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/Future.html#get-long-java.util.concurrent.TimeUnit-) documentation.

Suggested remediation:
Set timeout parameters for `CompletableFuture.get()` or `Future.get()`.Exceptions, such as `ExecutionException`, `InterruptedException` and `TimeoutException` are caught and errors are logged.
```
@@ -203,3 +203,3 @@
         try {
-            diskQueueLoaded.get();
+            diskQueueLoaded.get(100, TimeUnit.MILLISECONDS);
         } catch (ExecutionException e) {
@@ -207,2 +207,8 @@
             logger.atDebug().cause(e).log(&#34;Disk queue load completed with an error&#34;);
+        } catch (InterruptedException ie) {
+            log.error(&#34;Thread interrupted.&#34;, ie);
+            throw new InterruptedException(ie.getMessage());
+        } catch (TimeoutException te) {
+            log.error(&#34;Future is not done when timeout.&#34;, te);
+            throw new TimeoutException(te.getMessage());
         }

```

alter-mage (dismissed):

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -261,18 +301,41 @@ public void removeMessageById(long messageId) {
...
+        }
         removeMessagesWithQosZero(false);
     }
 
     private void removeMessagesWithQosZero(boolean needToCheckCurSpoolerSize) {
         if (!qos0MessageCheckRequired.get()) {
             return;
         }
+        try {
+            awaitDiskQueueLoaded();
```
yitingb, line 358: **`popOutMessagesWithQosZero()` now blocks the CRT event-loop thread until the background load finishes, which prevents reconnection for that entire time.**

On the default MQTT5 path the chain is synchronous end to end: aws-c-mqtt5 emits the disconnect lifecycle event on the connection's event-loop thread (`AWS_FATAL_ASSERT(aws_event_loop_thread_is_callers_thread(client->loop))` immediately before the emit in `mqtt5_client.c`), the JNI binding invokes `AwsIotMqtt5Client.onDisconnection` inline, and from there `CallbackEventManager` → `MqttClient` callback → `spool.popOutMessagesWithQosZero()` → untimed `diskQueueLoaded.get()`, all direct calls with `keepQos0WhenOffline=false` (the default). Reconnect is driven by the client's service task scheduled on that same event loop, so while the thread is parked, no reconnect attempt starts even if the network recovers seconds later. Pre-PR this couldn't happen: the load finished in the constructor, so no connection existed while a load was pending.

Impact scales with the backlog: using #1832's numbers (26,568 messages in 5 min 19 s, ~12 ms/message), a 1-second blip during the load window costs up to ~5 minutes offline, longer for larger spools. The devices most exposed are exactly the ones this PR targets — a big backlog implies flaky connectivity, so a disconnect during the load is likely. The wait also runs before the `qos0MessageCheckRequired` check, so it's paid even when there is nothing to purge.

The gate itself is correct (the purge must not iterate a half-loaded queue) — the problem is which thread waits. Suggest decoupling the purge from the CRT callback thread so `onDisconnection` returns promptly, while preserving the invariant that the purge only runs over the fully loaded queue. A purge that consequently completes after reconnection is fine under QoS 0 at-most-once semantics. Decoupling would also fix a pre-existing issue this PR widens: `removeMessagesWithQosZero` does a disk read per queued id, so even before this change a disconnect with a large disk queue could run minutes of SQLite I/O on the event-loop thread.

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -261,33 +317,67 @@ public void removeMessageById(long messageId) {
...
+     * done, the prune runs synchronously on the calling thread.</p>
+     */
     public void popOutMessagesWithQosZero() {
-        removeMessagesWithQosZero(false);
+        if (diskQueueLoaded.isDone()) {
+            removeMessagesWithQosZero(false);
+        } else {
+            // Do not block the caller (potentially the CRT event-loop thread) on the disk load.
+            // Defer pruning until the load finishes, on the managed executor.
+            diskQueueLoaded.thenRunAsync(() -> removeMessagesWithQosZero(false), executorService)
```
MikeDombo, line 342: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

This code is using non-async method `whenComplete` on `CompletableFuture` (or `CompletionStage`). Please be aware that task subscribed to `CompletableFuture`(or `CompletionStage`) through non-async method may be executed by the thread that completes the current `CompletableFuture` (or `CompletionStage`), or the thread that calls this non-async method to add subscription. In other words, the thread that executes the subscription is non-deterministic. If you prefer having fully control over threads, you may consider using the async variants instead. In addition, please make sure tasks subscribed to `CompletableFuture` (or `CompletionStage`) are short lived and non-blocking to avoid deadlock. Learn more: [https://docs.oracle.com/en/java/javase/18/docs/api/java.base/java/util/concurrent/CompletableFuture.html](https://docs.oracle.com/en/java/javase/18/docs/api/java.base/java/util/concurrent/CompletableFuture.html)

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -151,12 +188,25 @@ private CloudMessageSpool getPersistenceSpoolGGService()
...
+    /**
+     * Blocks until all persisted message IDs have been loaded from disk into the in-memory deque.
+     * Returns immediately if already complete or if using Memory storage.
+     *
+     * <p>Must be called by any method that consumes from or iterates the deque to guarantee
+     * all persisted messages are present and ordered correctly.</p>
+     */
+    private void awaitDiskQueueLoaded() throws InterruptedException {
+        try {
+            diskQueueLoaded.get();
```
MikeDombo, line 206: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

It looks like you are using `CompletableFuture.get()` or `Future.get()` without setting its timeout parameter. Be sure to set timeout values or handle timeout exceptions to prevent threads from waiting for a process to end. This can lead to downtime for your customers. For more information, see [CompletableFuture](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#get-long-java.util.concurrent.TimeUnit-) and [Future](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/Future.html#get-long-java.util.concurrent.TimeUnit-) documentation.

Suggested remediation:
Set timeout parameters for `CompletableFuture.get()` or `Future.get()`.Exceptions, such as `ExecutionException`, `InterruptedException` and `TimeoutException` are caught and errors are logged.
```
@@ -205,3 +205,3 @@
         try {
-            diskQueueLoaded.get();
+            diskQueueLoaded.get(100, TimeUnit.MILLISECONDS);
         } catch (ExecutionException e) {
@@ -209,2 +209,8 @@
             logger.atDebug().cause(e).log(&#34;Disk queue load completed with an error&#34;);
+        } catch (InterruptedException ie) {
+            log.error(&#34;Thread interrupted.&#34;, ie);
+            throw new InterruptedException(ie.getMessage());
+        } catch (TimeoutException te) {
+            log.error(&#34;Future is not done when timeout.&#34;, te);
+            throw new TimeoutException(te.getMessage());
         }

```

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
saranyailla: Addressed this by making the qos0 message draining asynchronous after the disk queue is fully loaded as this depends on that.

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -151,12 +188,25 @@ private CloudMessageSpool getPersistenceSpoolGGService()
...
+        } catch (ExecutionException e) {
+            // Background load already logged its own failure; proceed with whatever was loaded.
+            logger.atDebug().cause(e).log("Disk queue load completed with an error");
```
alter-mage, line 209: `awaitDiskQueueLoaded()` catches `ExecutionException` from a failed background `persistentQueueSync` and logs it at `atDebug()`. If the background load fails partway through, this is currently the only trace of it — below the level most fleets monitor by default. Could this go to at least WARN instead of DEBUG?

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -205,11 +258,14 @@ private void addMessageToSpooler(long id, SpoolMessage message) {
...
     /**
      * Pop the id of the oldest PublishRequest.
+     * Waits for disk-to-memory sync to complete before draining to ensure
+     * all persisted messages are available and ordered correctly.
      *
      * @return message id
      * @throws InterruptedException the thread is interrupted while popping the first id from the queue
      */
     public long popId() throws InterruptedException {
+        awaitDiskQueueLoaded();
```
alter-mage, line 268: `DiskSpoolerStartupBlockingTest` is a solid addition and proves kernel/component startup is not blocked by a slow disk load, but it never calls `popId()` or `addMessage()` concurrently with an in-progress background sync to verify they actually block until the load completes and return correctly-ordered results afterward. Thats the one part of the new concurrency model introduced by this PR with no direct test. Could you extend this test (or add a sibling) that calls `popId()`/`addMessage()` from another thread while `SlowMockDiskSpoolerService` is still running, and asserts it blocks until `syncCompleted` and then returns correctly?

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -182,6 +232,9 @@ public void addId(long id) {
      */
     public SpoolMessage addMessage(Publish request) throws InterruptedException,
             SpoolerStoreException {
+        // Wait for the disk-to-memory sync to complete before adding to guarantee that the nextId picked is higher
+        // than all the persisted ids on disk.
+        awaitDiskQueueLoaded();
```
alter-mage, line 237: The PR description says `addMessage()` "proceeds immediately without waiting." In the current diff, `addMessage()`s first line is `awaitDiskQueueLoaded()` — it does wait for the background sync before proceeding (needed so `nextId` is settled relative to everything still loading). Minor, but worth updating the description so it matches the actual behavior for future readers.

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -151,12 +188,25 @@ private CloudMessageSpool getPersistenceSpoolGGService()
...
             }
-            logger.atInfo().log("Persistent Spooler has been set up");
-            return persistenceSpool;
-        } else {
-            throw new ServiceLoadException(
-                    "The Greengrass service located was not an instance of CloudMessageSpool"
-            );
+        }, executorService);
+
+        logger.atInfo().log("Persistent Spooler has been set up");
```
alter-mage, line 193: This log fires immediately after scheduling the background load, not after it finishes. In the old code this same log ran after `persistentQueueSync` had already completed synchronously, so it was accurate then. Now it fires before any messages are necessarily loaded — an operator reading logs during a slow boot could reasonably conclude the spooler is fully populated when its actually just started loading in the background. Could this be reworded (e.g. "Persistent Spooler background load scheduled") or moved to fire on `diskQueueLoaded`s completion instead?

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -261,33 +317,67 @@ public void removeMessageById(long messageId) {
...
+    /**
+     * Remove all QoS 0 messages from the spooler queue. Called when the device goes offline and the
+     * spooler is configured to not keep QoS 0 messages while disconnected.
+     *
+     * <p>This may be invoked on the MQTT connection-event (CRT event-loop) thread, so it must not
+     * block. If the disk-to-memory load is still in progress, the prune is deferred to run on the
+     * spool executor once the load completes, rather than waiting inline. If the load is already
+     * done, the prune runs synchronously on the calling thread.</p>
+     */
     public void popOutMessagesWithQosZero() {
```
alter-mage, line 336: This defers the purge via `diskQueueLoaded.thenRunAsync(() -> removeMessagesWithQosZero(false), executorService)`, which runs unconditionally once the disk load completes — it never checks whether the device is still offline at that point. If the connection resumes before the load finishes, the purge still fires anyway and drops QoS0 messages that are now perfectly deliverable, just because the device happened to be offline at some earlier moment. This kind of gap existed before this change too, but its still bad UX and worth fixing here: could the deferred continuation re-check current connectivity before dropping messages, instead of assuming the device is still offline by the time it runs?

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -61,20 +67,28 @@ public class Spool {
...
+
     /**
      * Constructor.
      *
      * @param deviceConfiguration the device configuration
      * @param kernel              a kernel instance
+     * @param executorService     executor for running background tasks
      */
-    public Spool(DeviceConfiguration deviceConfiguration, Kernel kernel) {
+    public Spool(DeviceConfiguration deviceConfiguration, Kernel kernel, ExecutorService executorService) {
```
alter-mage, line 84: Out of curiosity, not a blocker: `Spool` is constructed with `new` inside `MqttClient`s constructor rather than being resolved via `context.get(Spool.class)` like `MqttClient` itself is elsewhere in Nucleus. Was that a deliberate choice, or just how its always been?

src/main/java/com/aws/greengrass/mqttclient/MqttClient.java
```diff
@@ -281,7 +281,7 @@ protected MqttClient(DeviceConfiguration deviceConfiguration,
         eventLoopGroup = new EventLoopGroup(Coerce.toInt(mqttTopics.findOrDefault(1, MQTT_THREAD_POOL_SIZE_KEY)));
         hostResolver = new HostResolver(eventLoopGroup);
         clientBootstrap = new ClientBootstrap(eventLoopGroup, hostResolver);
-        spool = new Spool(deviceConfiguration, kernel);
+        spool = new Spool(deviceConfiguration, kernel, executorService);
```
alter-mage, line 284: `onConnectionResumed()` calls `triggerSpooler()`, which submits `runSpooler()` to the executor. The first thing that worker does when it tries to publish a spooled message is call `spool.popId()`, whose first line is `awaitDiskQueueLoaded()` — it blocks until the background disk-to-memory sync finishes. So if the connection comes back up while that load is still running, the spooler thread starts, calls `popId()`, and sits blocked for however long the remaining load takes before publishing anything. Not a correctness bug, and a similar wait already existed before this PR too (just hidden behind the kernel not starting up at all) — but this is still a bad UX for anyone watching the device: it reports "connected" and appears ready, while nothing is actually flowing. Worth raising so its tracked, even without a concrete fix proposed here yet.

saranyailla: That is correct by design. We want to deliver messages in order and for that all the messages existing on the disk must be loaded to determine the order. The only difference here is that we won't block starting up services due to this loading process.

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
saranyailla: Device connectivity should be the caller's concern. Dropping QoS-0 on disconnect is the configured behavior (keepQos0WhenOffline=false), and this method is just the mechanism that does it, so keeping the connectivity decision out of here maps to single-responsibility. If we ever want to narrow the timing window, that belongs in the caller (skip or cancel the scheduled purge on connection resume), not in this method. Adding a connectivity check here would also couple  the spooler implementation to the MQTT client, which isn't the case today.

It also can't be fully avoided even with a gate: the device can come back online in the middle of removing QoS-0 messages, so the check would just move the race, not close it. Not fixing this time, since it's neither a regression nor incorrect behavior.

src/main/java/com/aws/greengrass/mqttclient/spool/Spool.java
```diff
@@ -151,12 +190,28 @@ private CloudMessageSpool getPersistenceSpoolGGService()
...
+    /**
+     * Blocks until all persisted message IDs have been loaded from disk into the in-memory deque.
+     * Returns immediately if already complete or if using Memory storage.
+     *
+     * <p>Must be called by any method that consumes from or iterates the deque to guarantee
+     * all persisted messages are present and ordered correctly.</p>
+     */
+    private void awaitDiskQueueLoaded() throws InterruptedException {
+        try {
+            diskQueueLoaded.get();
```
MikeDombo, line 210: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

It looks like you are using `CompletableFuture.get()` or `Future.get()` without setting its timeout parameter. Be sure to set timeout values or handle timeout exceptions to prevent threads from waiting for a process to end. This can lead to downtime for your customers. For more information, see [CompletableFuture](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#get-long-java.util.concurrent.TimeUnit-) and [Future](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/Future.html#get-long-java.util.concurrent.TimeUnit-) documentation.

Suggested remediation:
Set timeout parameters for `CompletableFuture.get()` or `Future.get()`.Exceptions, such as `ExecutionException`, `InterruptedException` and `TimeoutException` are caught and errors are logged.
```
@@ -209,3 +209,3 @@
         try {
-            diskQueueLoaded.get();
+            diskQueueLoaded.get(100, TimeUnit.MILLISECONDS);
         } catch (ExecutionException e) {
@@ -214,2 +214,8 @@
             logger.atWarn().cause(e).log(&#34;Disk queue load completed with an error&#34;);
+        } catch (InterruptedException ie) {
+            log.error(&#34;Thread interrupted.&#34;, ie);
+            throw new InterruptedException(ie.getMessage());
+        } catch (TimeoutException te) {
+            log.error(&#34;Future is not done when timeout.&#34;, te);
+            throw new TimeoutException(te.getMessage());
         }

```

PR #1369, opened 2022-11-28
fix(ecr): retry download on network errors

leoSRM:
**Issue #1323:**

**Description of changes:**
Add retry logic when ECR download on network errors. When network errors, throws a **ConnectionException**. Than **RetryUtils** catches this exception, do download again. If other exceptions, do not try.
Now, ConnectionException includes these conditions：
_read: connection timed out
net/http: TLS handshake timeout
Temporary failure in name resolution
Client.Timeout exceeded while awaiting headers_

**Why is this change necessary:**
This enhancement ensures thar docker images can be download.

**How was this change tested:**
- [x] Updated or added new unit tests.
- [x] Manual tests.
[network-error-download-retry.zip](https://github.com/aws-greengrass/aws-greengrass-nucleus/files/10215243/network-error-download-retry.zip)
[greengrass-nameserver-error.log](https://github.com/aws-greengrass/aws-greengrass-nucleus/files/10215640/greengrass-nameserver-error.log)

**Any additional information or context required to review the change:**

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1836, opened 2026-08-17
fix: retry docker pull network errors instead of failing deployments

aws-kevinrickard:
## Problem

`DefaultDockerClient.pullImage` classifies a failed `docker pull` by substring-matching docker's stderr against seven network-error strings:

```java
"read: connection timed out"  "net/http"  "Temporary failure in name resolution"
"request canceled"  "timeout"  "no such host"  "dial tcp"
```

A match becomes `ConnectionException`, which retries indefinitely. Everything else becomes `DockerPullException`, which no retry config lists as retryable, and `RetryUtils` rethrows when nothing matches. So an unmatched error fails the deployment on the first attempt, with no backoff and no second try.

A device whose connection fails over mid-pull reports:

```
DOCKER_PULL_ERROR: Failed to download docker image. Unexpected error while trying to perform
docker pull - Error response from daemon: Get "https://<acct>.dkr.ecr.<region>.amazonaws.com/v2/
<repo>/manifests/sha256:...": read tcp 172.28.1.230:52044->34.204.60.241:443:
read: connection reset by peer
```

That matches none of the seven: it says `read tcp` rather than `dial tcp`, and `read: connection reset by peer` rather than `read: connection timed out`.

`dial tcp` does match, which is why the failure looks intermittent. A failover landing before the connection opens retries indefinitely and stays invisible; one landing mid-transfer produces `read tcp` and fails outright. Pulls run long enough to make mid-transfer failover the likely case.

The gap is structural. Each incident has extended the allowlist by one string (#1369, then #1507 for `dial tcp`), so any transport error nobody anticipated fails a deployment with zero retries. Every other downloader classifies by exception *type* instead — `S3Downloader` and `GreengrassRepositoryDownloader` retry `IOException`, so a connection reset during an S3 download already retries today.

### One more string will not fix this

docker exposes no machine-readable signal. On docker 25.0.14, a DNS failure, a `manifest unknown` and an invalid reference all exit `1`, and `docker pull` offers no `--format`. The Engine API reports failures as `{"errorDetail":{"message":"..."}}` inside a 200 response — the same string wearing JSON.

Four layers wrap that text: the docker CLI, dockerd, containerd, and the Go standard library. Go once exposed the predicate this code needs and has since deprecated it:

```go
// Deprecated: Temporary errors are not well-defined.
// Most "temporary" errors are timeouts, and the few exceptions are surprising.
// Do not use this method.
Temporary() bool
```

The layer producing these errors gave up classifying them. This change therefore layers imperfect signals rather than chasing an exhaustive list.

## Fix

Six separately reviewable commits described below in logical rather than commit order, plus one test-only commit.

**1. Classify by the Go `net.OpError` shape.** Go renders an `OpError` as `"<op> <net> <source>-><addr>: <cause>"`. Matching the op prefix — `read tcp` and `write tcp`, joining `dial tcp` — catches any TCP-layer failure whatever cause follows, including causes docker has never emitted. Doing so moves the dependency onto the stable half of the string: the standard library generates the prefix and has not changed it since Go 1.0, whereas the cause is a syscall errno string that varies by platform and kernel. The old allowlist enumerated the varying half.

Transport causes still match on their own, because docker and containerd sometimes drop the op prefix when wrapping an error raised while copying a layer: `connection reset by peer`, `connection refused`, `connection aborted`, `broken pipe`, `network is unreachable`, `network is down`, `host is unreachable`, `no route to host`, `i/o timeout`, `EOF`, `unexpected EOF`, `context deadline exceeded`, `server misbehaving`.

Six of the seven original strings remain, so nothing that already retried indefinitely stops — with one deliberate exception. Bare `timeout` is dropped: the specific forms it stood in for are matched directly (`read: connection timed out`, `i/o timeout`, `net/http`), and as a bare substring it could fire on an image name, a tag, or registry prose and retry a permanent failure forever. A timeout form matching none of the specific entries now takes the bounded retry of item 4 rather than failing outright, which is what makes dropping it safe. Two of the remaining strings now match case-insensitively, which only widens them.

The HTTP/2 entries in item 2 are the two specific transport forms rather than an `http2:` package prefix, since the prefix would also match framing and flow-control faults — not connectivity, and capable of persisting through a broken intermediary, which under the prefix would retry forever.

**2. Match transport failures raised above `net`.** Some layers never build an `OpError`. `crypto/tls` reports `remote error: tls: bad record MAC`, `tls: use of closed connection` and `remote error: tls: internal error`; HTTP/2 reports `http2: server sent GOAWAY`, `http2: client connection lost` and `stream error: stream ID N; INTERNAL_ERROR`; containerd discards the `OpError` while copying a layer, leaving `read |0: file already closed`. These now match, along with `context canceled`.

Only wire-level TLS failures qualify. When the peer rejects our certificate or we distrust theirs, no retry helps, so `x509:`, `bad certificate` and `handshake failure` stay out and the code says why. Registry 5xx stays out too: retryable, but not transport, and a registry returning one persistently is a fault that should surface.

**3. Fall back on device connectivity.** `runWithConnectionErrorCheck` already infers connectivity for `DockerServiceUnavailableException`; the same inference now covers `UnknownDockerPullException`. When MQTT is down, an unrecognized pull failure more likely reflects that outage than anything the registry said, so it retries until connectivity returns. An online device still gets the bound, which preserves the no-hang property.

Two limits. MQTT reachability does not prove registry reachability, since the endpoints differ and can diverge under a proxy. And detection lags: keep-alive defaults to 60s with a 30s ping timeout, so `getMqttOnline` can report online for up to roughly 90s after connectivity drops. Item 4's retry budget is sized to outlast that lag, so an unrecognized failure during a real outage reaches the inference once the flag flips; the string matching still carries the reported case directly. This signal backs up the strings rather than replacing them.

**4. Retry the remainder instead of failing immediately.** Six unrecoverable errors — `manifest unknown`, `no matching manifest for`, `invalid reference format`, `no space left on device`, `authentication required`, `requested access to the resource is denied` — now map explicitly to `DockerPullException` and fail fast. All six already failed fast, so they become non-retryable by intent rather than by omission. Anything matching neither list becomes the new `UnknownDockerPullException` and retries 7 times at 10s→2m, roughly 195–390 seconds with jitter — sized so the budget outlasts the ~90s MQTT detection lag from item 3 on any jitter draw, since a shorter budget could expire before the device notices it is offline and the fallback ever fires. Only the currently broken bucket changes behavior.

The budget bounds this retry tier, not the deployment: an ECR credential expiring mid-pull re-enters the pull sequence with a fresh budget. That can happen at most once in practice, since the refreshed token is valid for hours, so the effective worst case is about twice the figure above.

This tier stays bounded because it catches errors nothing is known about. Nucleus imposes no deployment timeout by design — `DefaultDeploymentTask` and `DeploymentService` both block without one and delegate detection downstream — so routing unknowns to the indefinite tier would hang a deployment forever on a permanent error the list does not yet name, recoverable only by cancelling it. Guessing wrong should cost minutes, not a stuck device. That also rules out reusing `finiteAttemptsRetryConfig`, whose 30 attempts at 10s→32m take hours.

`UnknownDockerPullException extends DockerPullException`, so exhausting the retries still reports `DOCKER_PULL_ERROR` with the same text. Because `RetryUtils` matches with `isInstance` and only the subclass appears in a retry config, a plain `DockerPullException` still fails fast.

**5. Back off each differentiated retry config independently.** `runWithRetry(DifferentiatedRetryConfig)` tracked one `retryInterval` shared by every config in the list, and slept on it before clamping to the matched config's ceiling — so a config with a high ceiling could ramp the interval and leave a config with a lower ceiling sleeping far past its own. Both existing callers hid this: `DeploymentDocumentDownloader` pairs two configs both flat at 1 minute, and the single-config overload cannot interfere with itself. Pairing 32-minute and 2-minute ceilings, as item 4 does, makes it reachable — several `DockerServiceUnavailableException` failures ramp the interval toward 32 minutes, and an `UnknownDockerPullException` arriving afterwards sleeps that long despite its own ceiling. Tracking the interval per config fixes it and leaves both existing callers unaffected. Without this, item 4's bounded tier does not hold to the 195–390s it claims.

**6. Lower the indefinite tier's ceiling from 64 to 5 minutes.** Droppable without affecting the rest. At 64 minutes, a device seven failures in waits an hour between attempts, so backoff gates recovery even after the network returns. The other downloaders retry at a flat 1 minute. Five minutes keeps exponential backoff — a 225s average sleep against flat-1-minute's 45s, so roughly a fifth of the requests during a sustained outage — while bounding recovery latency.

Note the scope: `infiniteAttemptsRetryConfig` also covers the ECR auth-token fetch and docker login, so this raises their steady-state retry rate too. That looks desirable for the same reason, but it is wider than the pull path.

## Testing

`DefaultDockerClientTest` (new) covers the two extracted predicates with 35 parameterized cases across network, known non-retryable and unrecognized errors. Each case also asserts the two classifications agree, since `isConnectionError` runs first. The network group quotes the reported string verbatim and fails against the previous code, and includes a `read tcp` error whose cause is deliberately a string docker does not emit, pinning the op prefix as the mechanism rather than the cause list. Two further tests pin the matching mechanics themselves: one runs classification under `tr-TR` — whose case rules break a locale-sensitive fold, since docker writes `stream ID` and Turkish lowercases `I` to a dotless ı — and fails if the `Locale.ROOT` argument is dropped (verified by mutation); the other asserts every entry in both lists is lower case, since matching folds only the error text and an upper-case entry could never match.

`DockerImageDownloaderTest` gains four cases: an unknown error retried to the bound and then surfacing as `PackageDownloadException`, asserting the attempt count so it fails against the old single-attempt behavior; one clearing on the second attempt; one on an **offline** device succeeding on the fifth attempt, which exceeds the bounded limit and so passes only via the offline inference — reverting that inference makes it fail, confirming it discriminates; and a plain `DockerPullException` still failing in one attempt.

`RetryUtilsTest` gains a case pairing a config that ramps a 2s interval to 4s with one fixed at 10ms, asserting total elapsed time. It fails at 5641ms against the shared interval and passes at 1932ms with the per-config fix. A single large interval rather than a ramp over several small ones puts roughly 490ms either side of the 2500ms threshold, so no plausible `Thread.sleep` overshoot can flip the result.

The pre-existing `GIVEN_network_error_WHEN_download_docker_image_THEN_retry_download_image_until_succeed` passes unchanged and guards fail-fast against regression. All 65 tests across the two docker classes pass, as do `RetryUtilsTest` and `DeploymentDocumentDownloaderTest` (the other `DifferentiatedRetryConfig` caller), and `checkstyle:check`, `pmd:check` and `spotbugs:check` run clean.

## Notes

Left for separate changes:

- Bare `"timeout"` (the old `DOCKER_PULL_TIMEOUT`) is dropped rather than kept, as described in item 1 — the one deliberate exception to the rule that every previously matched string still matches.
- Classification still reads docker stderr. The `OpError` shape and the connectivity fallback narrow the exposure without removing it.
- `DockerImageDownloader.run()` nests the indefinite tier inside the finite one, so a `ConnectionException` never reaches the outer tier.

### DIFF
--- a/src/main/java/com/aws/greengrass/componentmanager/plugins/docker/DefaultDockerClient.java
+++ b/src/main/java/com/aws/greengrass/componentmanager/plugins/docker/DefaultDockerClient.java
@@ -27,6 +27,8 @@ import java.io.IOException;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Optional;
 
@@ -38,25 +40,56 @@ public class DefaultDockerClient {
     public static final Logger logger = LogManager.getLogger(DefaultDockerClient.class);
 
     /**
-     * connect error messages.
+     * Errors indicating that a docker pull failed at the network transport layer rather than being rejected by the
+     * registry. Such a failure is recoverable once connectivity is restored, so it is retried indefinitely.
+     *
+     * <p>Two shapes are matched. Go renders a {@code net.OpError} as
+     * {@code "<op> <net> <source>-><addr>: <cause>"}, so matching on the op prefix classifies a failure at the TCP
+     * layer regardless of which cause follows, including causes not seen before. Only {@code dial tcp} was matched
+     * previously, which covers a connection that fails to be established; a connection that dies mid-transfer
+     * instead reports {@code read tcp} or {@code write tcp}. The remaining entries are transport-level causes,
+     * matched on their own because docker and containerd also surface them without an op prefix, for example when
+     * wrapping an error raised while copying an image layer.
      */
-    private static final String READ_CONNECTION_TIME_OUT = "read: connection timed out";
-    private static final String NET_HTTP_TIMEOUT = "net/http";
-    private static final String TEMPORARY_FAILURE_IN_NAME_RESOLUTION = "Temporary failure in name resolution";
-    private static final String REQUEST_CANCELED = "request canceled";
-    private static final String DOCKER_PULL_TIMEOUT = "timeout";
-    private static final String NO_SUCH_HOST = "no such host";
-    private static final String DIAL_TCP = "dial tcp";
+    private static final List<String> CONNECTION_ERRORS = Collections.unmodifiableList(Arrays.asList(
+            // Go net.OpError op prefixes for the transport operations
+            "dial tcp",
+            "read tcp",
+            "write tcp",
+            // Transport-level causes
+            "connection reset by peer",
+            "connection refused",
+            "connection aborted",
+            "broken pipe",
+            "network is unreachable",
+            "network is down",
+            "host is unreachable",
+            "no route to host",
+            "read: connection timed out",
+            "i/o timeout",
+            "unexpected eof",
+            "\": eof",
+            // Timeouts and cancellations, which all mean the request did not complete rather than that the registry
+            // rejected it
+            "net/http",
+            "timeout",
+            "request canceled",
+            "context deadline exceeded",
+            // Name resolution failures
+            "no such host",
+            "temporary failure in name resolution",
+            "server misbehaving"));
 
     /**
      * Errors that a retry cannot recover from, so a docker pull failing with one of these should fail fast.
      */
-    private static final String MANIFEST_UNKNOWN = "manifest unknown";
-    private static final String NO_MATCHING_MANIFEST = "no matching manifest for";
-    private static final String INVALID_REFERENCE_FORMAT = "invalid reference format";
-    private static final String NO_SPACE_LEFT_ON_DEVICE = "no space left on device";
-    private static final String AUTHENTICATION_REQUIRED = "authentication required";
-    private static final String ACCESS_DENIED = "requested access to the resource is denied";
+    private static final List<String> NON_RETRYABLE_ERRORS = Collections.unmodifiableList(Arrays.asList(
+            "manifest unknown",
+            "no matching manifest for",
+            "invalid reference format",
+            "no space left on device",
+            "authentication required",
+            "requested access to the resource is denied"));
 
     /**
      * Sanity check for installation.
@@ -177,14 +210,7 @@ public class DefaultDockerClient {
      * @return true if the error is a known network error
      */
     static boolean isConnectionError(String err) {
-        String lowerCaseErr = err.toLowerCase();
-        return err.contains(READ_CONNECTION_TIME_OUT)
-                || err.contains(TEMPORARY_FAILURE_IN_NAME_RESOLUTION)
-                || lowerCaseErr.contains(NET_HTTP_TIMEOUT)
-                || lowerCaseErr.contains(REQUEST_CANCELED)
-                || lowerCaseErr.contains(DOCKER_PULL_TIMEOUT)
-                || lowerCaseErr.contains(NO_SUCH_HOST)
-                || lowerCaseErr.contains(DIAL_TCP);
+        return containsAny(err, CONNECTION_ERRORS);
     }
 
     /**
@@ -194,13 +220,12 @@ public class DefaultDockerClient {
      * @return true if the error is known to be non-retryable
      */
     static boolean isNonRetryableError(String err) {
-        String lowerCaseErr = err.toLowerCase();
-        return lowerCaseErr.contains(MANIFEST_UNKNOWN)
-                || lowerCaseErr.contains(NO_MATCHING_MANIFEST)
-                || lowerCaseErr.contains(INVALID_REFERENCE_FORMAT)
-                || lowerCaseErr.contains(NO_SPACE_LEFT_ON_DEVICE)
-                || lowerCaseErr.contains(AUTHENTICATION_REQUIRED)
-                || lowerCaseErr.contains(ACCESS_DENIED);
+        return containsAny(err, NON_RETRYABLE_ERRORS);
+    }
+
+    private static boolean containsAny(String err, List<String> lowerCaseNeedles) {
+        String lowerCaseErr = err.toLowerCase(Locale.ROOT);
+        return lowerCaseNeedles.stream().anyMatch(lowerCaseErr::contains);
     }
 
     /**
--- a/src/main/java/com/aws/greengrass/componentmanager/plugins/docker/DefaultDockerClient.java
+++ b/src/main/java/com/aws/greengrass/componentmanager/plugins/docker/DefaultDockerClient.java
@@ -50,8 +50,11 @@ public class DefaultDockerClient {
      * instead reports {@code read tcp} or {@code write tcp}. The remaining entries are transport-level causes,
      * matched on their own because docker and containerd also surface them without an op prefix, for example when
      * wrapping an error raised while copying an image layer.
+     *
+     * <p>Every entry must be lower case, since {@link #containsAny} only folds the case of the error text it is
+     * matched against. {@code DefaultDockerClientTest} enforces this.
      */
-    private static final List<String> CONNECTION_ERRORS = Collections.unmodifiableList(Arrays.asList(
+    static final List<String> CONNECTION_ERRORS = Collections.unmodifiableList(Arrays.asList(
             // Go net.OpError op prefixes for the transport operations
             "dial tcp",
             "read tcp",
@@ -75,17 +78,22 @@ public class DefaultDockerClient {
             "tls: use of closed connection",
             "bad record mac",
             "tls: internal error",
-            // HTTP/2 transport failures, matched on the package prefix for the same reason "net/http" is. A stream
-            // error carries no such prefix, so it is matched separately.
-            "http2:",
+            // HTTP/2 transport failures. Matched as specific forms rather than on the "http2:" package prefix,
+            // because the prefix would also match framing and flow-control faults that are not connectivity
+            // problems and could persist, for example through a broken proxy. A stream error carries no package
+            // prefix, so it is matched separately.
+            "http2: server sent goaway",
+            "http2: client connection lost",
             "stream error: stream id",
             // Transport failures that containerd surfaces while copying an image layer, having discarded the
             // net.OpError that produced them
             "file already closed",
             // Timeouts and cancellations, which all mean the request did not complete rather than that the registry
-            // rejected it
+            // rejected it. The bare "timeout" that was previously matched is gone: "read: connection timed out",
+            // "i/o timeout" and "net/http" cover the forms it stood in for, and as a bare substring it could match
+            // an image name or registry prose and retry a permanent failure indefinitely. A timeout form matching
+            // none of these now gets the bounded unknown-error retry instead of failing outright.
             "net/http",
-            "timeout",
             "request canceled",
             "context canceled",
             "context deadline exceeded",
@@ -96,8 +104,10 @@ public class DefaultDockerClient {
 
     /**
      * Errors that a retry cannot recover from, so a docker pull failing with one of these should fail fast.
+     *
+     * <p>Lower case for the same reason as {@link #CONNECTION_ERRORS}.
      */
-    private static final List<String> NON_RETRYABLE_ERRORS = Collections.unmodifiableList(Arrays.asList(
+    static final List<String> NON_RETRYABLE_ERRORS = Collections.unmodifiableList(Arrays.asList(
             "manifest unknown",
             "no matching manifest for",
             "invalid reference format",
--- a/src/main/java/com/aws/greengrass/componentmanager/plugins/docker/DefaultDockerClient.java
+++ b/src/main/java/com/aws/greengrass/componentmanager/plugins/docker/DefaultDockerClient.java
@@ -69,11 +69,25 @@ public class DefaultDockerClient {
             "i/o timeout",
             "unexpected eof",
             "\": eof",
+            // TLS transport failures. Only wire-level failures are listed. An error saying the peer rejected our
+            // certificate, or that we do not trust theirs, is a configuration problem a retry cannot fix, so
+            // "x509:", "bad certificate" and "handshake failure" are deliberately absent.
+            "tls: use of closed connection",
+            "bad record mac",
+            "tls: internal error",
+            // HTTP/2 transport failures, matched on the package prefix for the same reason "net/http" is. A stream
+            // error carries no such prefix, so it is matched separately.
+            "http2:",
+            "stream error: stream id",
+            // Transport failures that containerd surfaces while copying an image layer, having discarded the
+            // net.OpError that produced them
+            "file already closed",
             // Timeouts and cancellations, which all mean the request did not complete rather than that the registry
             // rejected it
             "net/http",
             "timeout",
             "request canceled",
+            "context canceled",
             "context deadline exceeded",
             // Name resolution failures
             "no such host",
--- a/src/main/java/com/aws/greengrass/componentmanager/plugins/docker/DefaultDockerClient.java
+++ b/src/main/java/com/aws/greengrass/componentmanager/plugins/docker/DefaultDockerClient.java
@@ -12,6 +12,7 @@ import com.aws.greengrass.componentmanager.plugins.docker.exceptions.DockerLogin
 import com.aws.greengrass.componentmanager.plugins.docker.exceptions.DockerPullException;
 import com.aws.greengrass.componentmanager.plugins.docker.exceptions.DockerServiceUnavailableException;
 import com.aws.greengrass.componentmanager.plugins.docker.exceptions.InvalidImageOrAccessDeniedException;
+import com.aws.greengrass.componentmanager.plugins.docker.exceptions.UnknownDockerPullException;
 import com.aws.greengrass.componentmanager.plugins.docker.exceptions.UserNotAuthorizedForDockerException;
 import com.aws.greengrass.logging.api.Logger;
 import com.aws.greengrass.logging.impl.LogManager;
@@ -46,6 +47,16 @@ public class DefaultDockerClient {
     private static final String DOCKER_PULL_TIMEOUT = "timeout";
     private static final String NO_SUCH_HOST = "no such host";
     private static final String DIAL_TCP = "dial tcp";
+
+    /**
+     * Errors that a retry cannot recover from, so a docker pull failing with one of these should fail fast.
+     */
+    private static final String MANIFEST_UNKNOWN = "manifest unknown";
+    private static final String NO_MATCHING_MANIFEST = "no matching manifest for";
+    private static final String INVALID_REFERENCE_FORMAT = "invalid reference format";
+    private static final String NO_SPACE_LEFT_ON_DEVICE = "no space left on device";
+    private static final String AUTHENTICATION_REQUIRED = "authentication required";
+    private static final String ACCESS_DENIED = "requested access to the resource is denied";
 
     /**
      * Sanity check for installation.
@@ -110,7 +121,9 @@ public class DefaultDockerClient {
      *                                             the registry
      * @throws UserNotAuthorizedForDockerException when current user is not authorized to use docker
      * @throws ConnectionException                 network error
-     * @throws DockerPullException                 unexpected error
+     * @throws DockerPullException                 an error that a retry cannot recover from, or, as
+     *                                             {@link UnknownDockerPullException}, an unrecognized error that may
+     *                                             be transient
      */
     public void pullImage(Image image) throws DockerServiceUnavailableException, InvalidImageOrAccessDeniedException,
             UserNotAuthorizedForDockerException, DockerPullException, ConnectionException {
@@ -135,16 +148,18 @@ public class DefaultDockerClient {
                     throw new InvalidImageOrAccessDeniedException(
                             String.format("Invalid image or login - %s", response.err));
                 }
-                if (response.err.contains(READ_CONNECTION_TIME_OUT)
-                        || response.err.contains(TEMPORARY_FAILURE_IN_NAME_RESOLUTION)
-                        || response.err.toLowerCase().contains(NET_HTTP_TIMEOUT)
-                        || response.err.toLowerCase().contains(REQUEST_CANCELED)
-                        || response.err.toLowerCase().contains(DOCKER_PULL_TIMEOUT)
-                        || response.err.toLowerCase().contains(NO_SUCH_HOST)
-                        || response.err.toLowerCase().contains(DIAL_TCP)) {
+                if (isConnectionError(response.err)) {
                     throw new ConnectionException(String.format("Network issue when docker pull - %s", response.err));
                 }
-                throw new DockerPullException(
+                if (isNonRetryableError(response.err)) {
+                    throw new DockerPullException(
+                            String.format("Unexpected error while trying to perform docker pull - %s", response.err),
+                            response.failureCause);
+                }
+                // The error is not recognized as either a network error or a non-retryable one. Assume it may be
+                // transient and let the caller apply a small, bounded number of retries rather than failing the
+                // deployment on the first attempt.
+                throw new UnknownDockerPullException(
                         String.format("Unexpected error while trying to perform docker pull - %s", response.err),
                         response.failureCause);
             }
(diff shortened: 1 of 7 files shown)


src/main/java/com/aws/greengrass/componentmanager/plugins/docker/DefaultDockerClient.java
```diff
@@ -154,6 +169,40 @@ public void pullImage(Image image) throws DockerServiceUnavailableException, Inv
...
 
+    /**
+     * Check if a docker CLI error indicates a network-level failure, which is recoverable once connectivity
+     * is restored and so is retried indefinitely.
+     *
+     * @param err stderr emitted by the docker CLI
+     * @return true if the error is a known network error
+     */
+    static boolean isConnectionError(String err) {
+        String lowerCaseErr = err.toLowerCase();
```
MikeDombo: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

#### toLowerCase [](|2045d6a2-68fa-4cde-87a6-566ffddb2b76|)

The parameterless versions of `String.toLowerCase()` and `String.toUpperCase()` use the default locale of the JVM when transforming strings. This can have unintended results, including difficult-to-debug transient failures because case mapping differs based on locale. 

**Even if you aren’t writing code for a user experience that’s displayed to a customer**, system internal code often uses case folding to normalize strings for comparison.
Not doing so can missing locale lead to errors when running on systems that are configured to use a different system locale. This can include developer desktops or build servers (not just your fleet hosts): many hours have been spent debugging errors that only occur on a specific machine due to its locale configuration.

#### Recommended solutions:

Always pass in a `Locale`.

**Language or locale-dependent processing:**

Get the locale from the platform, use the language of the content, or configure it in your code and pass that to the method.
```
String firstName = &#34;Ichabod&#34;; 
Locale locale = // Where you obtain the locale from is platform dependent
String lowerCaseFirstName = firstName.toLowerCase(locale);
``` 

**Internal processing that isn’t locale-dependent:**

*Example: S3 bucket name*

Always specify the language/country-neutral locale `Locale.ROOT`. 

```
String imgTag = &#34;IMG&#34;; 
if (&#34;img&#34;.equals(imgTag.toLowerCase(Locale.ROOT)) {    
// do something 
}
```

src/main/java/com/aws/greengrass/componentmanager/plugins/docker/DefaultDockerClient.java
```diff
@@ -154,6 +169,40 @@ public void pullImage(Image image) throws DockerServiceUnavailableException, Inv
...
+    }
+
+    /**
+     * Check if a docker CLI error is one that a retry cannot recover from, such as a missing image or a full disk.
+     *
+     * @param err stderr emitted by the docker CLI
+     * @return true if the error is known to be non-retryable
+     */
+    static boolean isNonRetryableError(String err) {
+        String lowerCaseErr = err.toLowerCase();
```
MikeDombo: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

#### toLowerCase [](|2045d6a2-68fa-4cde-87a6-566ffddb2b76|)

The parameterless versions of `String.toLowerCase()` and `String.toUpperCase()` use the default locale of the JVM when transforming strings. This can have unintended results, including difficult-to-debug transient failures because case mapping differs based on locale. 

**Even if you aren’t writing code for a user experience that’s displayed to a customer**, system internal code often uses case folding to normalize strings for comparison.
Not doing so can missing locale lead to errors when running on systems that are configured to use a different system locale. This can include developer desktops or build servers (not just your fleet hosts): many hours have been spent debugging errors that only occur on a specific machine due to its locale configuration.

#### Recommended solutions:

Always pass in a `Locale`.

**Language or locale-dependent processing:**

Get the locale from the platform, use the language of the content, or configure it in your code and pass that to the method.
```
String firstName = &#34;Ichabod&#34;; 
Locale locale = // Where you obtain the locale from is platform dependent
String lowerCaseFirstName = firstName.toLowerCase(locale);
``` 

**Internal processing that isn’t locale-dependent:**

*Example: S3 bucket name*

Always specify the language/country-neutral locale `Locale.ROOT`. 

```
String imgTag = &#34;IMG&#34;; 
if (&#34;img&#34;.equals(imgTag.toLowerCase(Locale.ROOT)) {    
// do something 
}
```

src/main/java/com/aws/greengrass/componentmanager/plugins/docker/DockerImageDownloader.java
```diff
@@ -66,6 +68,16 @@ public class DockerImageDownloader extends ArtifactDownloader {
...
                             SdkClientException.class, ServerException.class)).build();
 
+    // A docker pull error that is recognized as neither a network error nor a non-retryable error may still be
+    // transient, so allow a small number of retries. Deliberately much shorter than finiteAttemptsRetryConfig: the
+    // error is not known to be recoverable, so this only needs to be long enough to ride out a brief blip before
+    // reporting the failure.
+    @Setter(AccessLevel.PACKAGE)
+    private RetryUtils.RetryConfig unknownErrorRetryConfig =
+            RetryUtils.RetryConfig.builder().initialRetryInterval(Duration.ofSeconds(10L))
+                    .maxRetryInterval(Duration.ofMinutes(2L)).maxAttempt(5).retryableExceptions(
```
alter-mage: ### `unknownErrorRetryConfig`'s budget can expire before MQTT notices the device is offline

The connectivity fallback added in this PR is the right idea, but with `maxAttempt(5)` the bounded budget is short enough that it can run out before `getMqttOnline()` has flipped — so for the first ~90s of a real outage the fallback is blind, and the deployment fails on exactly the class of transient failure the fallback exists to catch.

The two timelines:

- `unknownErrorRetryConfig` is 5 attempts, 10s → 2m ceiling (#L76-L79). With sleeps jittered at 50–100% of the interval (`RetryUtils.java#L108`), total wall clock works out to roughly 75–150s.
- `mqttOnline` is a cached `AtomicBoolean` (`MqttClient.java#L146`) written only by `setMqttOnline` (#L1096), which is called only from `onConnectionInterrupted` / `onConnectionResumed` (#L172-L183). There is no polling and no synchronous liveness check on the public surface. So until the CRT client's keep-alive cycle expires, `getMqttOnline()` returns the stale last-known value. Defaults are 60s keep-alive + 30s ping timeout (#L101, #L103) — the ~90s already noted in the `runWithConnectionErrorCheck` comment.

The minimum-jitter case (~75s) is below the 90s detection lag, so the budget can be exhausted before the flag ever flips.

Since `getMqttOnline()` is polled fresh on every attempt (confirmed by `verify(mqttClient, times(4)).getMqttOnline()` in the offline test), this is a race rather than a guaranteed failure: a jittery-high run (~150s) will likely survive long enough for the flag to flip, a jittery-low run (~75s) won't. That still leaves the outcome decided by jitter.

Worth noting both timeouts are device-configurable (`keepAliveTimeoutMs`, `pingTimeoutMs`, #L100-L103), so ~90s is the default, not an upper bound — a fleet that has tuned keep-alive upward has a proportionally wider blind window.

**Suggested fix:** `maxAttempt(5)` → `maxAttempt(7)` at the same 2m ceiling. That puts the range at roughly 195–390s, which clears the 90s default even on the unluckiest jitter draw, while staying far from a hung deployment. It preserves the "cost minutes, not a stuck device" framing exactly — it just picks a bound that accounts for the detection lag the comment already documents.

I looked at whether the final attempt could instead force a liveness check rather than read the cached flag, but `MqttClient` exposes no synchronous ping — that would mean new API, which is clearly outside this PR.

src/main/java/com/aws/greengrass/componentmanager/plugins/docker/DockerImageDownloader.java
```diff
@@ -56,7 +58,7 @@ public class DockerImageDownloader extends ArtifactDownloader {
     @Setter(AccessLevel.PACKAGE)
     private RetryUtils.RetryConfig infiniteAttemptsRetryConfig =
             RetryUtils.RetryConfig.builder().initialRetryInterval(Duration.ofMinutes(1L))
-                    .maxRetryInterval(Duration.ofMinutes(64L)).maxAttempt(Integer.MAX_VALUE).retryableExceptions(
+                    .maxRetryInterval(Duration.ofMinutes(5L)).maxAttempt(Integer.MAX_VALUE).retryableExceptions(
```
alter-mage, line 61: ### Was the ECR request-rate change intended, and is this worth splitting?

Agreed that 64 minutes is bad — a device that has backed off fully waiting another hour after connectivity returns is poor behaviour, and 5 minutes is a reasonable ceiling.

Flagging the reach rather than the value: `infiniteAttemptsRetryConfig` also wraps `get-ecr-auth-token` (#L204), and `EcrAccessor` deliberately rethrows `ServerException | SdkClientException` as retryable (`EcrAccessor.java#L84-L86`) — both of which are in this config's list (#L62). So tail retry rate for ECR `GetAuthorizationToken` goes from one attempt per 64 min to one per 5 min, and with `maxAttempt(Integer.MAX_VALUE)` it persists for the whole duration of an outage. Those are exactly the exception types a regional ECR event produces fleet-wide.

That may well be intended and desirable — it's the same argument as for the pull, and 12.8× more attempts is also 12.8× more recovery opportunities once the service is back. I have no data suggesting the extra rate is a problem, so this is a question about intent rather than an objection: was the change to ECR auth-token behaviour deliberate, or a side effect of sharing the config with the pull path?

**Suggestion regardless of the answer:** consider splitting this into its own PR. `56b80ac8` already isolates it, so the cost is near zero, and it buys an independent revert boundary — the pull-classification fix is well-scoped and low-risk, whereas this one touches behaviour for every device talking to ECR. Bundled, a production issue with the ceiling change means reverting the classification fix too, which is the part that fixes the reported bug. Worth noting the PR title is entirely about pull-error classification, so a reviewer skimming it wouldn't expect ECR auth timing to change.

src/main/java/com/aws/greengrass/componentmanager/plugins/docker/DockerImageDownloader.java
```diff
@@ -285,8 +297,11 @@ private <T> void run(CrashableSupplier<T, Exception> task, String description, S
             throws PackageDownloadException, InterruptedException {
         try {
             // Finite retry attempts for errors that are not due to connectivity issues and
-            // might need explicit intervention to recover from
-            RetryUtils.runWithRetry(finiteAttemptsRetryConfig, () -> RetryUtils
+            // might need explicit intervention to recover from, and a shorter finite retry for errors that could
+            // not be classified at all
+            RetryUtils.runWithRetry(RetryUtils.DifferentiatedRetryConfig.builder()
```
alter-mage, line 312: ### Question: is the bounded-retry guarantee local to the tier, or end-to-end for the deployment?

Not asking for a change here — everything below is pre-existing on `main` and outside this PR's diff. Recording it because it bears on the rationale for `unknownErrorRetryConfig`, and I'd rather it be written down than rediscovered later.

The PR's justification for routing unknown errors to a bounded tier rather than the indefinite one is that Nucleus imposes no deployment timeout, so an unbounded tier could hang a deployment forever on a permanent unnamed error. That reasoning is sound for the tier itself. But `performDownloadSteps` wraps the whole login+pull sequence in a credential-refresh `do/while` whose only exit condition is `credentialRefreshNeeded == false` (#L196-L262), and each iteration calls `run()` — this line — which constructs a fresh `DifferentiatedRetryConfig`. Since `attemptMap` and the new `retryIntervalMap` are locals inside `runWithRetry` (`RetryUtils.java#L66-L70`), every loop iteration starts the unknown-error budget over at 5 attempts.

The reachable path is via the pull: `credentialsUsable()` is checked at #L250, inside the `run()` lambda but outside the inner retry at #L251. If an ECR token expires inside the ~75–150s bounded window, the loop re-fetches and re-enters, restarting the budget. Self-limiting in practice — the refreshed token is good for another 12 hours, so the reset can happen at most once, giving roughly 2× the intended bound rather than an unbounded loop. (The same reset is reachable via a long `docker-login` retry sequence through `finiteAttemptsRetryConfig`, though the timing there is much tighter.)

So the ~75–150s figure is a bound on the retry tier, not on the deployment. Is that the intended reading? If so it might be worth saying so explicitly in the `unknownErrorRetryConfig` comment, since "bounded so a deployment can't hang" invites the end-to-end interpretation.

Two adjacent pre-existing observations from the same trace, purely for the record:

- `Credentials.isValid()` is `Instant.now().isBefore(expiresAt)` (`Registry.java#L117-L119`) against ECR's raw `expiresAt` with no buffer (`EcrAccessor.java#L82`). A token can be judged valid immediately before expiry and then rejected by the registry mid-pull. A small buffer (say 60s) would make the check mean "valid long enough to be useful."
- Non-ECR registries get `Instant.MAX` (`Registry.java#L96`), so `isValid()` is permanently true and this loop never re-runs. Everything above is ECR-private-registry only.

src/main/java/com/aws/greengrass/componentmanager/plugins/docker/DefaultDockerClient.java
```diff
@@ -37,15 +40,70 @@ public class DefaultDockerClient {
...
+            // error carries no such prefix, so it is matched separately.
+            "http2:",
+            "stream error: stream id",
+            // Transport failures that containerd surfaces while copying an image layer, having discarded the
+            // net.OpError that produced them
+            "file already closed",
+            // Timeouts and cancellations, which all mean the request did not complete rather than that the registry
+            // rejected it
+            "net/http",
+            "timeout",
```
alter-mage: ### `"timeout"` is now redundant, and it sits on the branch where false positives are costliest

Ordering first, since it sets the stakes for the whole list. `isConnectionError` is checked before `isNonRetryableError` (#L198 vs #L201), so on any overlap the network list wins. That makes the consequences asymmetric:

- false positive on the **network** list → indefinite tier → deployment retries forever
- false positive on the **non-retryable** list → fails fast, customer sees a retryable error
- match on neither → bounded ~2min → benign

So the network list is the one where breadth is expensive, and it's the list this PR grows from 7 entries to 29. Worth stating explicitly somewhere, because it's the lens for judging any future addition — and a false positive here produces the hung deployment that `unknownErrorRetryConfig` was specifically designed to avoid.

Credit where it's due: the anti-collision property is already tested. The non-retryable case asserts both `isNonRetryableError(err)` and `assertFalse(isConnectionError(err))` (`DefaultDockerClientTest.java#L86-L89`), so the six fail-fast strings are proven not to leak into the indefinite tier. I checked all six against all 29 connection needles and found zero collisions.

**The suggestion:** consider dropping bare `"timeout"` (this line). It's pre-existing, not added here — but this PR is what makes it redundant. The list now carries `"read: connection timed out"` (#L66) and `"i/o timeout"` (#L69), which are the specific forms that bare `"timeout"` was standing in for. It now adds no coverage while keeping the widest match surface on the highest-consequence branch: it fires on the substring appearing anywhere, including inside an image name, a tag, or a registry's own error prose. You already flag this tension for `DOCKER_PULL_TIMEOUT` in the PR description; the specific entries added here look like they resolve it.

I can't demonstrate a real collision — that's an empirical question about docker/containerd output I can't answer from the Nucleus source. So: redundant and structurally risky rather than a proven bug. Your call on whether that's worth the change.

**Smaller question:** `"http2:"` (#L80) is matched as a package prefix, justified by analogy to `"net/http"`. Reasonable, but `net/http` errors are almost always transport, whereas `http2:` also covers framing and flow-control faults that aren't connectivity problems. Is the broader prefix intentional, or would the specific forms (`GOAWAY`, `client connection lost`) be enough?

For the record, the rest of the list looks well-chosen to me — the Go op prefixes are anchored, the multi-word causes are distinctive, and `"\": eof"` (#L71) is anchored to quote-colon specifically to avoid matching a stray "eof", which is exactly the care this kind of list needs.

src/main/java/com/aws/greengrass/componentmanager/plugins/docker/DefaultDockerClient.java
```diff
@@ -154,6 +216,32 @@ public void pullImage(Image image) throws DockerServiceUnavailableException, Inv
...
+     *
+     * @param err stderr emitted by the docker CLI
+     * @return true if the error is known to be non-retryable
+     */
+    static boolean isNonRetryableError(String err) {
+        return containsAny(err, NON_RETRYABLE_ERRORS);
+    }
+
+    private static boolean containsAny(String err, List<String> lowerCaseNeedles) {
+        String lowerCaseErr = err.toLowerCase(Locale.ROOT);
```
alter-mage, line 251: ### `Locale.ROOT` fix looks good — one suggestion to lock it in

Confirming the CodeGuru threads are genuinely resolved at head, not just stale: there is now exactly one `toLowerCase` call in the file (this line) and it carries `Locale.ROOT`. Rather than patching both call sites, both predicates were collapsed into a single `containsAny` helper (#L226, #L236, #L240-L242), which makes the hazard structurally unreachable instead of fixed twice. That's the better fix and worth noting.

**Suggestion: add a test that pins it.** `DefaultDockerClientTest` has no locale coverage, so nothing in the suite fails if the `Locale.ROOT` argument is dropped in a future edit.

That's more than hygiene here, because one of the needles this PR adds depends on the case-folding to work at all. `"stream error: stream id"` (#L81) matches docker's `stream ID 15; INTERNAL_ERROR` only because the haystack is lowercased first — docker emits a capital I. Under `tr-TR`, `"I".toLowerCase()` yields dotless ı (U+0131), so with a parameterless `toLowerCase()` that needle silently stops matching, and a transport error gets reclassified as unknown or non-retryable. No exception, no log — just a misrouted retry on a device in a locale nobody tests in. Same exposure for the uppercase-bearing entries at #L71 and #L75-L78.

Something like:

```java
@Test
void GIVEN_turkish_locale_WHEN_classified_THEN_still_matches() {
    Locale original = Locale.getDefault();
    try {
        Locale.setDefault(new Locale("tr", "TR"));
        assertTrue(DefaultDockerClient.isConnectionError(
                "http2: stream error: stream ID 15; INTERNAL_ERROR"));
    } finally {
        Locale.setDefault(original);
    }
}
```

Worth noting there's no precedent for this in the repo — `grep -rn "Locale.setDefault" src/test/` comes back empty, so this would be the first test mutating the default locale. The try/finally is essential to avoid leaking it into sibling tests. If mutating global state in a unit test is against convention here, an alternative is to have `containsAny` take the locale as a parameter (package-private) so the test can pass `tr-TR` directly without touching the JVM default — happy to defer to whichever you prefer.

src/main/java/com/aws/greengrass/componentmanager/plugins/docker/DockerImageDownloader.java
aws-kevinrickard: Good catch — a low-jitter run really could exhaust the budget inside the detection lag. Bumped to `maxAttempt(7)` in a5b5a3e7 and documented the sizing rationale on the config.

(AI-assisted comment)

src/main/java/com/aws/greengrass/componentmanager/plugins/docker/DockerImageDownloader.java
aws-kevinrickard: Deliberate — the Notes section of the description flags that this config also covers the ECR auth-token fetch and docker login, and the same recovery-latency argument applies there. Keeping it in this PR, but it stays isolated in 56b80ac8 if an independent revert is ever needed.

(AI-assisted comment)

src/main/java/com/aws/greengrass/componentmanager/plugins/docker/DockerImageDownloader.java
aws-kevinrickard: Per-tier is the intended reading. Documented on the config in a5b5a3e7, including the at-most-2× practical worst case from a credential refresh. Filed #1839 for the `isValid()` expiry-buffer observation so it is not lost in this thread.

(AI-assisted comment)

src/main/java/com/aws/greengrass/componentmanager/plugins/docker/DefaultDockerClient.java
aws-kevinrickard: Took both suggestions in 56dc6cc0: dropped bare `"timeout"` and narrowed `"http2:"` to the two specific transport forms. The description now records the `"timeout"` drop as the one deliberate exception to the no-string-stops-matching claim — safe now because an unmatched timeout form gets the bounded retry rather than failing outright.

(AI-assisted comment)

src/main/java/com/aws/greengrass/componentmanager/plugins/docker/DefaultDockerClient.java
aws-kevinrickard: Added in b61ee851: a `tr-TR` test in the try/finally `setDefault` shape (mutation-checked — it fails with a parameterless `toLowerCase()`), plus a test asserting every needle is lower case, since an upper-case entry could never match and nothing else would catch it.

(AI-assisted comment)

src/main/java/com/aws/greengrass/componentmanager/plugins/docker/DefaultDockerClient.java
aws-kevinrickard: Resolved — a single `toLowerCase(Locale.ROOT)` call remains in `containsAny`, and `DefaultDockerClientTest` now pins locale-independent classification under `tr-TR`.

(AI-assisted comment)

src/main/java/com/aws/greengrass/componentmanager/plugins/docker/DefaultDockerClient.java
aws-kevinrickard: Resolved — same as above: one `toLowerCase(Locale.ROOT)` call in `containsAny`, now pinned by a `tr-TR` locale test.

(AI-assisted comment)

Issue #1837, opened 2026-08-28
(deployment): cancelling a Nucleus-update deployment orphans its launch directory under alts/


aws-kevinrickard:
**Describe the bug**

Cancelling a deployment that requires a Nucleus restart leaves its launch directory behind at `<root>/alts/<deployment-id>`, and nothing ever deletes it.

`KernelAlternatives.prepareBootstrap` creates that directory and points `alts/current` and `alts/old` at it. Only two methods delete it, and both run solely when the deployment reaches a terminal status: `activationSucceeds()` deletes the directory `old` points at, and `rollbackCompletes()` deletes the one `broken` points at. A cancelled deployment reaches neither.

The next deployment calls `cleanupLaunchDirectoryLinks()`, which deletes the `old`, `broken`, and `new` **links** but not the directories they point at. So cancelling severs the last reference to the directory.

**To Reproduce**

1. Deploy a component at version 1.0.0 whose recipe declares a `bootstrap` lifecycle step, so its deployments take the Nucleus-restart path.
2. Deploy version 2.0.0 of that component, with a startup slow enough to leave a window — a `run` script that sleeps 60 seconds before signalling readiness, for example.
3. Once the Nucleus has restarted and is waiting for services, issue a third deployment. It cancels the second.
4. Inspect `<root>/alts/`.

**Expected behavior**

`alts/` contains only `current`, `init`, and the directory `current` points at.

**Actual behavior**

`alts/` also holds a directory named after the cancelled deployment's ID, with no link referring to it. It survives subsequent deployments and restarts.

**Environment**

- Nucleus version: reproduced on 2.18.x; the cleanup paths are unchanged on `main`. The cause is not OS-specific.

**Additional context**

Each leak is small: `prepareBootstrap` copies the launch directory with `NOFOLLOW_LINKS`, so the directory holds link entries rather than a copy of the Nucleus distribution. (Established from the copy flags, not measured on a device.)

The growth is unbounded, though. A device that regularly has deployments superseded mid-rollout gains a directory and an inode per cancellation, with nothing to prune them. It also makes `alts/` misleading during diagnosis: a directory named after a deployment ID normally indicates launch state for a deployment in progress, and here it indicates nothing.

`determineDeploymentStage` keys on the `old` and `broken` links rather than on directory names, so an orphaned directory does not cause the Nucleus to try to resume a cancelled deployment.

(AI-assisted post)

PR #1838, opened 2026-08-28
fix(deployment): drive an interrupted deployment to a terminal state

aws-kevinrickard:
**Issue #, if available:** None. (#1837 was filed while investigating this, but is a separate defect that this change does not address.)

**Description of changes:**

A deployment that requires no Nucleus restart takes the `DefaultActivator` path, which merges the new configuration into the live config tree with the transaction-log writer attached and flushing immediately. The merged version therefore reaches durable storage before anything knows whether the component starts, and the config tree is the only record of current state — nothing distinguishes a running, verified version from one written a moment ago.

Interrupt the Nucleus between that merge and a terminal deployment status, and three things follow. The unverified version survives as the device's current state, because `initConfigAndTlog` prefers `config.tlog` and nothing marks the value provisional. Nothing resumes or reverts it, because `determineDeploymentStage` returns `DEFAULT` unless kernel-alts directories exist and the `DEFAULT` launch path never reads `deployment_metadata.json`. Processing the deployment again then rewrites its rollback snapshot from that state, so a later rollback restores the version being deployed rather than the one that was running.

Superseding a deployment mid-activation reaches the same state with no interruption at all: `waitForServicesToStart` returns early on cancellation instead of throwing, so `DefaultActivator` never rolls back, and the superseding deployment inherits the cancelled deployment's merged version as its rollback baseline.

Seven changes:

- **Snapshot regardless of `FailureHandlingPolicy`.** `DefaultActivator` previously recorded a rollback snapshot only when the deployment requested rollback or switched the IoT data endpoint, so a `DO_NOTHING` deployment — the default for local deployments — left nothing to restore from. The policy governs whether a *failed* deployment rolls back; it cannot govern whether an *interrupted* one is recoverable.
- **Carry the rollback snapshot forward.** A terminal status is what removes the `ongoing` link, so finding that link when a new deployment starts means live config may hold unverified changes. The unfinished deployment's `rollback_snapshot.tlog` is then the last verified configuration, so preserve it into the new deployment's directory — which `createNewDeploymentDirectory` otherwise deletes — and keep it rather than overwriting it from live config. This covers re-processing the same deployment and a different deployment superseding it.
- **Re-process an abandoned local deployment at launch.** When the launch stage is `DEFAULT` and an unfinished *local* deployment's metadata is still on disk, offer it back onto the deployment queue, as the kernel-activation stage already does. Only local deployments: a cloud deployment that never reported a terminal status is re-delivered by the cloud with the document the cloud holds, so re-processing it from disk would race a better-informed mechanism — and resolution from disk can fail where the cloud succeeds, turning a deployment that would have completed into a failed one. Local deployments get no re-delivery, so they are the case that needs this.
- **Count an attempt where the deployment is applied, not where it is received.** The same deployment can be delivered several times — by the device re-processing it at launch and by the cloud re-delivering it after a reconnect — while only some deliveries reach the point of changing anything.
- **Cap the attempts, and restore the device when they run out.** A deployment that reliably takes the device down mid-activation would otherwise retry forever. The launch then boots from the deployment's rollback snapshot rather than `config.tlog`, which is the same config override the kernel-rollback stage uses; doing it before any service starts is what makes it safe. `DeploymentService` reports `FAILED`, which stops cloud re-delivery and releases the device to accept a newer deployment. `maxInterruptedDeploymentAttempts` sets the limit, defaulting to three; below 1 there is no limit, for a fleet whose devices may legitimately need several attempts because the environment is not ready yet. The value is parsed strictly, because coercing it would turn anything non-numeric into 0 and silently disable the protection; a malformed value falls back to the default and one too large for an int is clamped rather than rejected. The limit is recorded alongside the count when the deployment is applied, because the launch consults the count from `parseArgs`, before `config.tlog` has been read — so a deployment is judged by the limit in force when it was applied.
- **Remember a refused deployment until something else finishes.** Reporting the failure clears the `ongoing` link, so a re-delivery already in flight would otherwise get a fresh allowance and re-apply the version just backed out. Such a re-delivery is now rejected, identified by the last-failed link and the attempt count inside it. Rejecting rather than failing it leaves that record intact; any other deployment reaching a terminal state clears it, so a new revision is always accepted.
- **Stop a cancelled deployment before `removeObsoleteServices`,** rather than mutating services alongside the deployment that superseded it.

The re-queue and the restore both swallow any failure. They are recovery steps, so a corrupt metadata file must never keep the device from starting.

If the unverified version is broken the device does not recover, and getting there needs no operator action: a device that never reported a terminal status re-delivers the deployment to itself after reconnecting, and that re-delivery is what poisons the rollback snapshot.

This affects any deployment that changes only components, which is most routine application deployments. Deployments that update the Nucleus or include a component with a `bootstrap` step take the restart-based path and are unaffected, because `KernelUpdateActivator` closes the transaction-log writer before merging and records its target separately.

A second, independent defect is fixed in the first commit: the same-version branch of `handleComponentVersionConfigs` read `previousVersion` from the version topic, so any deployment silently rotated `previousVersion` onto the current version for every component whose version was not changing. `ComponentManager.getVersionsToKeep` reads that field to decide which versions are non-stale, so the replaced version's artifacts became eligible for cleanup.

- [ ] If my code makes a remote network call, it was tested with a proxy. (No network calls added.)

Unit tests cover each part: the snapshot carry-forward and its boundaries, the attempt counter, the cap and its unlimited setting, recognising a re-delivery of a refused deployment, the launch-time re-queue and restore, a failing re-queue still letting the device start, and a cancelled deployment stopping before it removes services. Each test that pins a defect was checked in both directions — it fails with the corresponding production change reverted.

End-to-end coverage is an 18-scenario suite run on devices; it is not held in this repository. Twelve reproduce the defects and fail on the current release line; six are controls guarding behaviour that already works, including the restart-based path surviving the same interruption and an interruption during a rollback. All 18 pass with this change, with the interrupted-deployment scenarios stable both alone and in the full suite. A full sweep of the wider suite — 787 scenarios — found one regression, an interrupted custom-nucleus deployment that this change re-processed locally instead of leaving for the cloud to re-deliver; that is what narrowed re-processing to local deployments.

Three deliberate behaviour changes worth your attention:

- A failed snapshot write now fails any deployment rather than only rollback-enabled ones, matching what `KernelUpdateActivator` already does. Note the distinction from a *missing* deployment directory, which is not a failure: with no directory there is nowhere to put a snapshot and nothing tracking the deployment that could use one, so the snapshot is skipped and the deployment proceeds. `DeploymentService` creates the directory before submitting, so every deployment that reaches a device is still snapshotted.
- A deployment interrupted on every allowed attempt is failed and reverted, where previously another attempt could succeed. This trades eventual success for a predictable failure the operator can retry, and is what keeps a deployment that repeatedly takes the device down from blocking newer ones. It defaults to three attempts and can be turned off with `maxInterruptedDeploymentAttempts`, which restores the previous behaviour.
- A re-delivery of a deployment that used up its attempts is rejected, so it reports `REJECTED` where it previously reported `IN_PROGRESS` and then a result.

Two properties I chose not to change:

- Constructing a `Kernel` performs the boot-time restore even in a process that never launches, because the restore works by choosing which file `initConfigAndTlog` reads. That is what makes it safe — no services are running, so nothing has to be reconciled — and moving it into `launch()` would mean merging a snapshot into an already-loaded configuration. `KERNEL_ACTIVATION` and `KERNEL_ROLLBACK` already override the config file the same way; this change widens when that happens, not what it does. Untested consequence: a second process constructing a `Kernel` while the Nucleus is running with attempts exhausted would rewrite `config.tlog` from the snapshot.
- A re-delivery arriving after a *different* deployment has reached a terminal state gets a fresh allowance, because that deployment's terminal status clears the record. A cloud deployment older than the group's last one is rejected as stale, and a local deployment is never re-delivered, so this is bounded elsewhere. A unit test documents the boundary.

Cancellation *semantics* are unchanged: a cancelled deployment still does not revert, and the log still tells the operator to deploy again. What changes is that the next deployment no longer inherits unverified state as its rollback baseline.

**Documentation Checklist:**
 - [ ] Updated the README if applicable. (Not applicable — no user-facing configuration or interface changes.)

**Compatibility Checklist:**
- [ ] For external library version updates, I have reviewed its change logs and Nucleus does not consume any deprecated method or type. (No dependency changes.)

Nothing is removed or renamed and no new requirement is added, so no customer has to change anything. The behaviour changes above alter recovery outcomes only; if you read any of them as requiring a corresponding customer change, that is the part to push back on.

### DIFF
--- a/src/integrationtests/java/com/aws/greengrass/integrationtests/deployment/DeploymentTaskIntegrationTest.java
+++ b/src/integrationtests/java/com/aws/greengrass/integrationtests/deployment/DeploymentTaskIntegrationTest.java
@@ -1466,8 +1466,17 @@ class DeploymentTaskIntegrationTest extends BaseITCase {
 
     @SuppressWarnings("PMD.AvoidCatchingGenericException")
     private Future<DeploymentResult> submitSampleJobDocument(URI uri, Long timestamp) throws Exception {
-        kernel.getContext().get(DeploymentDirectoryManager.class)
-                .createNewDeploymentDirectory("testFleetConfigArn" + deploymentCount.getAndIncrement());
+        DeploymentDirectoryManager deploymentDirectoryManager =
+                kernel.getContext().get(DeploymentDirectoryManager.class);
+        // DeploymentService removes the ongoing link when a deployment reports a terminal status. This
+        // helper bypasses DeploymentService, so do it here: an ongoing link left behind means the next
+        // deployment treats the previous one as interrupted and inherits its rollback snapshot as the
+        // baseline, which would roll back past a deployment that actually completed.
+        if (deploymentDirectoryManager.hasUnfinishedDeployment()) {
+            deploymentDirectoryManager.persistLastSuccessfulDeployment();
+        }
+        deploymentDirectoryManager.createNewDeploymentDirectory(
+                "testFleetConfigArn" + deploymentCount.getAndIncrement());
 
         sampleJobDocument = OBJECT_MAPPER.readValue(new File(uri), DeploymentDocument.class);
         sampleJobDocument.setTimestamp(timestamp);
--- a/src/main/java/com/aws/greengrass/deployment/DeploymentDirectoryManager.java
+++ b/src/main/java/com/aws/greengrass/deployment/DeploymentDirectoryManager.java
@@ -342,7 +342,7 @@ public class DeploymentDirectoryManager {
                 return null;
             }
             Path unfinishedDir = Files.readSymbolicLink(ongoingDir).toAbsolutePath();
-            if (requiredDirName != null && !requiredDirName.equals(unfinishedDir.getFileName().toString())) {
+            if (requiredDirName != null && !isDirectoryNamed(unfinishedDir, requiredDirName)) {
                 return null;
             }
             Path file = unfinishedDir.resolve(fileName);
@@ -456,13 +456,22 @@ public class DeploymentDirectoryManager {
                 return false;
             }
             Path failedDir = Files.readSymbolicLink(previousFailureDir).toAbsolutePath();
-            return getSafeFileName(deploymentId).equals(failedDir.getFileName().toString())
+            return isDirectoryNamed(failedDir, getSafeFileName(deploymentId))
                     && hasExhaustedAttempts(failedDir);
         } catch (IOException e) {
             logger.atError().kv(DEPLOYMENT_ID_LOG_KEY, deploymentId)
                     .log("Unable to tell whether this deployment already used up its attempts", e);
             return false;
         }
+    }
+
+    /**
+     * Whether a directory has the given name. A path has no file name when it is a root, so the name has
+     * to be null-checked before it is compared.
+     */
+    private static boolean isDirectoryNamed(Path directory, String name) {
+        Path fileName = directory.getFileName();
+        return fileName != null && name.equals(fileName.toString());
     }
 
     public static String getSafeFileName(String fleetConfigArn) {
--- a/src/main/java/com/aws/greengrass/deployment/DeploymentDirectoryManager.java
+++ b/src/main/java/com/aws/greengrass/deployment/DeploymentDirectoryManager.java
@@ -19,8 +19,10 @@ import lombok.AccessLevel;
 import lombok.Getter;
 
 import java.io.IOException;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.StandardCopyOption;
 import java.util.concurrent.atomic.AtomicReference;
 import javax.inject.Inject;
 
@@ -36,6 +38,17 @@ public class DeploymentDirectoryManager {
     static final String ROLLBACK_BOOTSTRAP_TASK_FILE = "rollback_bootstrap_task.json";
     static final String DEPLOYMENT_METADATA_FILE = "deployment_metadata.json";
     static final String CONFIG_SNAPSHOT_ERROR = "config_snapshot_error";
+    static final String PROCESSING_ATTEMPTS_FILE = "processing_attempts";
+
+    /**
+     * A deployment that never reaches a terminal status gets processed again, either by this device on
+     * the next launch or by the cloud re-delivering it after a reconnect. Cap those attempts so a
+     * deployment that reliably takes the device down cannot keep the device from accepting a newer one.
+     */
+    static final int MAX_PROCESSING_ATTEMPTS = 3;
+
+    // Files preserved from an unfinished deployment are parked here while its directory is cleaned up.
+    private static final String PRESERVED_FILE_PREFIX = ".preserved-";
 
     private static final String LINK_LOG_KEY = "link";
     private static final String FILE_LOG_KEY = "file";
@@ -187,6 +200,33 @@ public class DeploymentDirectoryManager {
     }
 
     /**
+     * Take the rollback snapshot for the ongoing deployment, unless one was carried forward from an
+     * unfinished deployment: that snapshot holds the last verified configuration and live config may
+     * not, so overwriting it would make unverified state the rollback target.
+     *
+     * @throws IOException if write fails
+     */
+    public void takeRollbackSnapshot() throws IOException {
+        Path filepath = getSnapshotFilePath();
+        if (Files.exists(filepath)) {
+            logger.atInfo().kv(FILE_LOG_KEY, filepath)
+                    .log("Keeping the rollback snapshot carried forward from an unfinished deployment");
+            return;
+        }
+        takeConfigSnapshot(filepath);
+    }
+
+    /**
+     * Whether a deployment is recorded as still in progress. The ongoing link is created when a
+     * deployment starts and removed once its result is reported.
+     *
+     * @return true if an unfinished deployment's directory is still linked
+     */
+    public boolean hasUnfinishedDeployment() {
+        return Files.isSymbolicLink(ongoingDir);
+    }
+
+    /**
      * Resolve snapshot file path.
      *
      * @return Path to snapshot file
@@ -243,11 +283,22 @@ public class DeploymentDirectoryManager {
     /**
      * Create or return the directory for a given deployment.
      *
+     * <p>An ongoing directory left behind by an unfinished deployment means the live configuration may
+     * hold that deployment's merged but never verified changes, so its rollback snapshot — not live
+     * config — is the last verified configuration. Preserve that snapshot into the new deployment's
+     * directory, whether the same deployment is re-processed or a different one supersedes it. Preserve
+     * the attempt count too when the same deployment is re-processed, so repeated interruptions can be
+     * capped.
+     *
      * @param deploymentId Deployment id
      * @return Path to the deployment directory
      * @throws IOException on I/O errors
      */
     public Path createNewDeploymentDirectory(String deploymentId) throws IOException {
+        final Path preservedSnapshot = stashFileFromUnfinishedDeployment(ROLLBACK_SNAPSHOT_FILE, null);
+        final Path preservedAttempts =
+                stashFileFromUnfinishedDeployment(PROCESSING_ATTEMPTS_FILE, getSafeFileName(deploymentId));
+
         cleanupPreviousDeployments(ongoingDir);
         Path path = deploymentsDir.resolve(getSafeFileName(deploymentId));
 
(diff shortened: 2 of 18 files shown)


src/main/java/com/aws/greengrass/deployment/DeploymentDirectoryManager.java
```diff
@@ -265,11 +333,172 @@ public Path createNewDeploymentDirectory(String deploymentId) throws IOException
...
+            return false;
+        }
+    }
+
+    private boolean hasExhaustedAttempts(Path deploymentDir) throws IOException {
+        int limit = readNumber(deploymentDir.resolve(ATTEMPT_LIMIT_FILE), 0);
+        return limit >= 1 && readNumber(deploymentDir.resolve(PROCESSING_ATTEMPTS_FILE), 0) >= limit;
+    }
+
+    private int readNumber(Path file, int fallback) throws IOException {
```
alter-mage, line 456: **Should an unreadable attempt limit fall back to the default rather than to "no limit"?**

`hasExhaustedAttempts` reads the limit with a fallback of `0`:

```java
int limit = readNumber(deploymentDir.resolve(ATTEMPT_LIMIT_FILE), 0);
return limit >= 1 && readNumber(deploymentDir.resolve(PROCESSING_ATTEMPTS_FILE), 0) >= limit;
```

and `readNumber` returns that fallback when the file cannot be parsed. Since `limit >= 1` is required, a `0` means no limit. So if `processing_attempt_limit` is empty, truncated, or garbled, does the cap silently switch off and the deployment retry indefinitely?

If that reading is right, it seems reachable rather than hypothetical, because of how the file is written in `recordProcessingAttempt`:

```java
Files.write(deploymentDir.resolve(ATTEMPT_LIMIT_FILE),
        String.valueOf(attemptLimit).getBytes(StandardCharsets.UTF_8));
Path attemptsFile = deploymentDir.resolve(PROCESSING_ATTEMPTS_FILE);
int attempts = readNumber(attemptsFile, 0) + 1;
Files.write(attemptsFile, String.valueOf(attempts).getBytes(StandardCharsets.UTF_8));
```

These are the only durable writes in `DeploymentDirectoryManager` that don't go through `CommitableFile`/`CommitableWriter`, and an interrupt during them is the same class of event the feature exists to handle — so the trigger condition for the feature is also what could corrupt the file that governs it. There's also a window between the two writes where the limit is updated and the count isn't.

This looks like the same inversion `9b569f6f` addressed one layer up, where the commit message calls it "the one direction a safety setting must not fail in" and `getEffectiveMaxInterruptedDeploymentAttempts()` falls back to `DEFAULT_MAX_INTERRUPTED_DEPLOYMENT_ATTEMPTS` for exactly that reason. Was the on-disk path intentionally different, or is the config-side reasoning meant to apply here too?

Two possible directions:

1. Use `DEFAULT_MAX_INTERRUPTED_DEPLOYMENT_ATTEMPTS` as the fallback for the limit specifically (keeping `0` for the count, where absent genuinely means zero), and route both writes through `CommitableFile`.
2. Store both numbers in `deployment_metadata.json`, which is already written via `CommitableWriter` and read via `CommitableReader` — that would make the two-write window and the partial-file case go away rather than being handled.

src/main/java/com/aws/greengrass/deployment/DeploymentDirectoryManager.java
```diff
@@ -265,11 +333,172 @@ public Path createNewDeploymentDirectory(String deploymentId) throws IOException
...
+     * @return the attempt number, starting at 1
+     * @throws IOException on I/O errors
+     */
+    public int recordProcessingAttempt(int attemptLimit) throws IOException {
+        if (!hasUnfinishedDeployment()) {
+            // Nothing is tracking this deployment, so there is nowhere to keep a count.
+            return 0;
+        }
+        Path deploymentDir = getDeploymentDirectoryPath();
+        Files.write(deploymentDir.resolve(ATTEMPT_LIMIT_FILE),
```
alter-mage, line 407: **Why separate plain-text files for the attempt count rather than `deployment_metadata.json`?**

The count and limit are stored as two bare files written with `Files.write`. Two concerns, and I think they have a common fix:

**1. Neither write is atomic, and a corrupt file disables the protection.** These are the only durable writes in `DeploymentDirectoryManager` that don't go through `CommitableWriter`/`CommitableFile`. An interrupt between the two writes — which is precisely the scenario this feature exists to handle — leaves the limit updated and the count stale, or a partially written file. And `readNumber` maps an unparseable value to its fallback of `0`, which `hasExhaustedAttempts` treats as "no limit", so a garbled limit file silently turns the cap off. (Detailed separately on `readNumber`.)

**2. Would `deployment_metadata.json` work instead?** It's in the same deployment directory, so it's carried forward by the same mechanism, and it already has the properties these files need:

- written atomically via `CommitableWriter.commitOnClose`
- read via `CommitableReader.of(filePath).read(...)`, which tolerates a partial write
- already read from `parseArgs` in the `KERNEL_ACTIVATION`/`KERNEL_ROLLBACK` branches, so it's demonstrably available before `config.tlog` is loaded — which is the constraint given in `a64c0786` for not reading the limit from config
- `Deployment` is `@JsonInclude(NON_NULL)` with a fail-safe mapper, so adding `processingAttempts`/`processingAttemptLimit` fields should deserialize older files without migration

Folding both numbers into that one atomic write would also remove the two-write window, and would let `readNumber`, both constants, and two of the three stash/restore pairs go away.

Is there a reason this was avoided — e.g. wanting to keep the `parseArgs` read cheap rather than deserializing the whole deployment document, or not wanting `readDeploymentMetadata`'s `IOException`-on-missing-symlink behaviour on this path? If it's the cost, that seems worth trading for atomicity, but you'd know the boot-time budget better than I do.

If keeping separate files is preferred, the minimum would be routing both through `CommitableFile` and defaulting a malformed limit to the configured default rather than `0`.

src/main/java/com/aws/greengrass/deployment/DeploymentDirectoryManager.java
```diff
@@ -265,11 +333,172 @@ public Path createNewDeploymentDirectory(String deploymentId) throws IOException
...
+                    .log("Preserving file from unfinished deployment for the next deployment");
+            return stash;
+        } catch (IOException e) {
+            logger.atError().kv(FILE_LOG_KEY, fileName)
+                    .log("Unable to preserve file from unfinished deployment", e);
+            return null;
+        }
+    }
+
+    private void restorePreservedFile(Path stash, Path destination) {
```
alter-mage, line 379: **Can a failed restore here silently reintroduce the contaminated-snapshot bug?**

In `createNewDeploymentDirectory`, the snapshot is moved out to a stash, the old directory is deleted, and then the stash is moved back. `restorePreservedFile` catches and logs rather than propagating:

```java
} catch (IOException e) {
    logger.atError().kv(FILE_LOG_KEY, destination)
            .log("Unable to restore preserved file into new deployment directory", e);
}
```

If that move fails, the original directory is already gone (`cleanupPreviousDeployments` ran first), so the deployment proceeds with no snapshot in place. `takeRollbackSnapshot()` then takes the `!Files.exists` branch and snapshots live config — which at that point holds the previous deployment's merged, unverified changes. That looks like exactly the inverted-rollback state this PR exists to prevent, reached through a failed move rather than through the original code path.

A few related observations:

1. **The stash is never cleaned up.** `PRESERVED_FILE_PREFIX` appears only twice in the PR — the constant and the `Files.move` that creates the stash. Nothing deletes it, and nothing enumerates `deployments/`, so a stranded `.preserved-rollback_snapshot.tlog` would persist indefinitely.

2. **The stash path is shared across deployments.** `deploymentsDir.resolve(PRESERVED_FILE_PREFIX + fileName)` is a single fixed path with `REPLACE_EXISTING`, so a stranded file from one deployment is silently replaced by the next. Would a path inside the deployment directory, or one qualified by deployment ID, be safer?

3. **No test covers a failed preserve/restore.** The new `DeploymentDirectoryManagerTest` cases cover carried-forward, superseded, completed, exhausted, and re-delivered — all success paths. A case where the restore move fails, asserting the deployment doesn't then overwrite the snapshot from live config, would pin the behaviour.

Would it be safer to fail the deployment when the snapshot cannot be restored, rather than continue without one? Given the PR's own reasoning — that a deployment applied without a verified rollback baseline is the root problem — continuing seems like the riskier default, and `takeConfigSnapshot` failing already fails the deployment via `FAILED_NO_STATE_CHANGE`, so there's precedent for treating a missing snapshot as fatal.

Alternatively, `Files.copy` to the stash and delete the source only after the restore succeeds, so a failure can't lose the only good copy.

src/main/java/com/aws/greengrass/deployment/DeploymentDirectoryManager.java
```diff
@@ -243,11 +298,24 @@ private Path getDeploymentDirectoryPath() throws IOException {
...
+     * directory, whether the same deployment is re-processed or a different one supersedes it. Preserve
+     * the attempt count too when the same deployment is re-processed, so repeated interruptions can be
+     * capped.
+     *
      * @param deploymentId Deployment id
      * @return Path to the deployment directory
      * @throws IOException on I/O errors
      */
     public Path createNewDeploymentDirectory(String deploymentId) throws IOException {
+        final Path preservedSnapshot = stashFileFromUnfinishedDeployment(ROLLBACK_SNAPSHOT_FILE, null);
```
alter-mage, line 313: **This method resolves the `ongoing` symlink four times, and pairs stash/restore by hand**

Two related things.

**Four independent resolutions of the same symlink.** `Files.isSymbolicLink(ongoingDir)` appears four times in this class. Three of those are reached from one `createNewDeploymentDirectory` call, because each of the three `stashFileFromUnfinishedDeployment` invocations independently does:

```java
if (!Files.isSymbolicLink(ongoingDir)) { return null; }
Path unfinishedDir = Files.readSymbolicLink(ongoingDir).toAbsolutePath();
```

…for three files in the same directory, and then `cleanupPreviousDeployments(ongoingDir)` resolves it again. Resolving once up front and passing the directory down would remove the repetition, and would also remove a fragility: as written, each call re-reads the link independently, so the three stashes are not guaranteed by construction to have come from the same directory. Not reachable today given this runs single-threaded, but it's an invariant held by accident.

**The stash/restore pairing is maintained manually across ~20 lines.** Six lines varying only in the filename, with `getSafeFileName(deploymentId)` computed twice. The correctness requirement is that every stashed file is restored — currently enforced only by remembering to add both halves. A `Map<String, Path>` (filename → stash path) built in one pass and drained in another would make it structurally impossible to stash without restoring, and would collapse this to two loops.

That matters a little more than usual because a stash that isn't restored is not benign: the source file has already been `Files.move`d out and the original directory is then deleted, so a missed restore loses the file outright.

Both of these are refactors rather than defects — happy to be told they're not worth churn here. If the `deployment_metadata.json` suggestion is taken up, two of the three stashed files disappear anyway and this shrinks on its own.

src/main/java/com/aws/greengrass/lifecyclemanager/Kernel.java
```diff
@@ -302,12 +303,92 @@ public Kernel launch() {
...
+     * is durable but was never verified, so the device launches from its rollback snapshot instead — the
+     * same config override the kernel-rollback stage uses. Doing it before any service starts is what
+     * makes it safe: nothing has to be reconciled afterwards. DeploymentService then reports the
+     * deployment failed.
+     */
+    // Catching RuntimeException is deliberate: see the catch block.
+    @SuppressWarnings("PMD.AvoidCatchingGenericException")
+    private String configToRestoreAbandonedDeploymentFrom(DeploymentDirectoryManager deploymentDirectoryManager,
+                                                          DeploymentStage stage) {
+        if (!deploymentDirectoryManager.hasUnfinishedDeployment()
```
alter-mage, line 327: **`hasUnfinishedDeployment()` is checked three times on the same launch path**

`hasExhaustedProcessingAttempts()` already opens with the same check:

```java
public boolean hasExhaustedProcessingAttempts() {
    if (!hasUnfinishedDeployment()) {
        return false;
    }
```

So the first clause of this guard looks dead — `hasExhaustedProcessingAttempts()` cannot return true without an unfinished deployment. And in `parseArgs` the `default` branch checks it a third time after the restore declines:

```java
String restoreFrom = configToRestoreAbandonedDeploymentFrom(...);   // checks twice inside
if (!Utils.isEmpty(restoreFrom)) { break; }
if (deploymentDirectoryManager.hasUnfinishedDeployment()) {          // third call
```

Three `Files.isSymbolicLink` calls where one result would serve. Not a performance concern at boot, but the duplicated guard means a future change to one of them can silently diverge from the others.

Would it be cleaner to resolve the state once in `parseArgs` and pass it down — or drop the redundant first clause here and rely on the callee's own guard? The latter is a one-line change and keeps the "is there an unfinished deployment" question owned in a single place.

src/main/java/com/aws/greengrass/deployment/DeploymentService.java
```diff
@@ -612,6 +632,22 @@ private void createNewDeployment(Deployment deployment) {
                 return;
             }
 
+            if (attemptsExhausted) {
+                // The launch that re-processed this deployment already restored the configuration from its
+                // rollback snapshot, so the device is back on the last verified configuration. Reporting a
+                // terminal status stops the cloud re-delivering the deployment and releases the device to
+                // accept a newer one.
+                logger.atError().kv(GG_DEPLOYMENT_ID_LOG_KEY_NAME, deployment.getGreengrassDeploymentId())
+                        .kv("attempts", deploymentDirectoryManager.getProcessingAttempts())
```
alter-mage, line 641: **`getProcessingAttempts()` is public API for a single log field**

This `kv` is its only production caller. The method has its own javadoc, its own `hasUnfinishedDeployment()` guard, its own symlink resolution and its own `readNumber` call — all to populate one log field, on a path where `hasExhaustedProcessingAttempts()` has just read the same two files a few lines earlier.

`recordProcessingAttempt` already returns the count, so one option is to thread that value through rather than re-read it. Failing that, could this be package-private, since nothing outside the package needs it? Small thing, but it's a public method on a class that's already large enough to have picked up `@SuppressWarnings("PMD.ExcessiveClassLength")` in this PR.

src/test/java/com/aws/greengrass/lifecyclemanager/KernelTest.java
```diff
@@ -409,6 +414,112 @@ void GIVEN_kernel_launch_WHEN_deployment_activation_happy_path_THEN_inject_deplo
...
+        doReturn(DEFAULT).when(kernelAlternatives).determineDeploymentStage(any(), any());
+        doReturn(tempFile).when(kernelAlternatives).getLaunchParamsPath();
+        kernel.getContext().put(KernelAlternatives.class, kernelAlternatives);
+
+        KernelCommandLine kernelCommandLine = mock(KernelCommandLine.class);
+        doReturn(deploymentDirectoryManager).when(kernelCommandLine).getDeploymentDirectoryManager();
+        doReturn(mock(BootstrapManager.class)).when(kernelCommandLine).getBootstrapManager();
+        kernel.setKernelCommandLine(kernelCommandLine);
+
+        kernel.parseArgs().launch();
```
alter-mage, line 519: **No test exercises `parseArgs()` without `launch()`, which is the shape `GreengrassSetup` uses**

The exhaustion restore and the re-queue are deliberately split across two methods — `configToRestoreAbandonedDeploymentFrom` is called from `parseArgs()`'s `default` branch, `requeueAbandonedDeployment` from `launch()`. The reason given in `a64c0786` makes sense: the count has to be read before `config.tlog` is loaded.

All five new `KernelTest` cases go through this helper, which ends in `kernel.parseArgs().launch()`. So `parseArgs()` is only ever tested with `launch()` immediately following.

`GreengrassSetup.performSetup()` is a production caller with a different shape — it calls `kernel.parseArgs(...)` and then has two `kernel.shutdown(); return;` exits (the `--setup-system-service` path and the `kernelStart == false` path) before it reaches `kernel.launch()`. In those cases `parseArgs()` runs and the process leaves without launching.

I'm not claiming a corruption bug here: the `!readFromTlog` rewrite in `initConfigAndTlog` is pre-existing, goes through `CommitableFile`, and demotes the old file to `config.tlog~`, so the write itself is atomic and recoverable. And the restore is gated on `hasUnfinishedDeployment() && hasExhaustedProcessingAttempts()`, which is narrow.

But it is an untested shape for a newly behaviour-bearing method, and it raises a design question:

1. **Should the restore live in `launch()` instead**, alongside `requeueAbandonedDeployment()`, so recovery only happens in a process that is actually booting the device? If the ordering constraint rules that out, would a launch-intent flag that `GreengrassSetup` doesn't set be worth it?
2. **Could a `KernelTest` case call `parseArgs()` alone** — with an exhausted deployment staged — and assert on whether the config override is applied? That pins the intended behaviour either way, and documents which one is intended.

Related and smaller: this PR aligns `DefaultActivator`'s snapshot behaviour with `KernelUpdateActivator`'s unconditional snapshot, and the CI bots flag `KernelUpdateActivator` below threshold on both unit (60/65/55 vs 65) and integration (25/29/22 vs 58) coverage. I checked and its production code isn't touched by this PR, so that's pre-existing rather than a regression introduced here — but it does mean the reference implementation this change aligns to is the least-covered class nearby, so the shared invariant is asserted rather than tested. Not a blocker, just worth naming.

src/integrationtests/java/com/aws/greengrass/integrationtests/deployment/DeploymentTaskIntegrationTest.java
```diff
@@ -1466,8 +1466,17 @@ void GIVEN_successful_deployment_with_multiple_unsuccessful_deployments_with_fin
...
     private Future<DeploymentResult> submitSampleJobDocument(URI uri, Long timestamp) throws Exception {
-        kernel.getContext().get(DeploymentDirectoryManager.class)
-                .createNewDeploymentDirectory("testFleetConfigArn" + deploymentCount.getAndIncrement());
+        DeploymentDirectoryManager deploymentDirectoryManager =
+                kernel.getContext().get(DeploymentDirectoryManager.class);
+        // DeploymentService removes the ongoing link when a deployment reports a terminal status. This
+        // helper bypasses DeploymentService, so do it here: an ongoing link left behind means the next
+        // deployment treats the previous one as interrupted and inherits its rollback snapshot as the
+        // baseline, which would roll back past a deployment that actually completed.
+        if (deploymentDirectoryManager.hasUnfinishedDeployment()) {
```
alter-mage, line 1475: **This harness change looks correct — worth confirming the implication**

This is the one harness adjustment in the PR that changes test semantics rather than just stub strictness, so I traced it.

`submitSampleJobDocument` drives `DeploymentConfigMerger` directly and never goes through `DeploymentService`, so nothing called `persistLastSuccessfulDeployment`/`persistLastFailedDeployment` — and it's `persistPointerToLastFinishedDeployment` that does the `Files.delete(ongoingDir)`. So the `ongoing` link survived from one deployment to the next for the entire test. This PR then treats a surviving `ongoing` link as "the previous deployment was interrupted," which makes each new deployment inherit the previous one's rollback snapshot as its baseline.

That reading matches the `f53959e7` commit message, and the helper now calls the same production method `DeploymentService` uses, so I read this as an honest harness fix rather than a workaround — the test was silently wrong before and the new semantics surfaced it.

Two things I'd like to confirm:

1. **Is it worth an in-repo test that pins the behaviour this exposed?** The harness fix makes the existing tests pass again, but the underlying scenario — an `ongoing` link surviving into the next deployment, and the next deployment inheriting the previous snapshot — is now load-bearing production behaviour. Right now it's covered at the unit level in `DeploymentDirectoryManagerTest`, but the integration path that surfaced it is only fixed, not asserted.

2. **Was anything hidden by always recording success?** The helper unconditionally calls `persistLastSuccessfulDeployment()`, even for deployments the test expects to fail or roll back. Since `wasRefusedForExhaustedAttempts` reads the `previous-failure` link specifically, recording every prior deployment as a success means these tests can never exercise that path. Would `persistLastFailedDeployment()` be more faithful for the rollback cases, or does it not matter here because those tests don't reach the attempt cap?

On the `ServiceDependencyLifecycleTest` side, the four `when(...)` → `lenient().when(...)` changes on `getFailureHandlingPolicy()` look straightforwardly correct — the activator no longer reads the policy on the success path now that the snapshot is unconditional, so strict stubbing would flag them. No concern there.

